PR #5, opened 2026-04-30
fix: if a provider is failing to set up, do not panic, add to retry list

Tomelia1999:
What the branch fixes                                                                                                                     
   
  The branch hardens protocol/rpcsmartrouter/rpcsmartrouter.go against three startup/runtime failure modes that previously caused panics,   
  silent data loss, or unnecessary outages:
                                                                                                                                            
  1. A single broken static provider crashes the whole smart router — startup validation called LavaFormatError("[PANIC] …") on the first   
  failure, taking the process down even when other providers were healthy.
  2. Two providers sharing the same Name could mass-exclude each other — the failed-set was keyed by provider.Name, so deduplication by name
   caused collateral exclusions.                                                                                                            
  3. Race between updateEpoch and the new retry goroutine — both write to rpsr.providerSessions and push to the consumer session manager
  (CSM); without coordination they could push snapshots in the opposite order they wrote the map, silently dropping providers until the next
   epoch.                 
                                                                                                                                            
  How it fixes them — by commit

  2ffb84f — don't panic on setup failure; retry in background                                                                               
   
  - Replaces [PANIC] errors in Phase 1 static-provider validation with LavaFormatWarning and adds the failed provider to a per-endpoint     
  failedStaticSet / failedStaticEndpoints list. Validation continues for the remaining providers instead of aborting.
  - Only fails the endpoint hard (errCh <- err) when all static providers for that api-interface fail (healthyCount == 0).                  
  - Restructures the flow so session registration happens after validation using only healthyStaticProviders / healthyBackupProviders.      
  Previously the unfiltered list was registered first and then re-registered after backup validation — the new path is "one call, correct   
  from the start" (rpcsmartrouter.go:462).                                                                                                  
  - Adds a new field failedStaticProviders map[string][]*lavasession.RPCStaticProviderEndpoint on RPCSmartRouter (rpcsmartrouter.go:155) and
   a new background goroutine retryFailedStaticProviders (rpcsmartrouter.go:~1820) that:                                                    
    - Wakes every 3 minutes (matches SpecValidator's disabled-chain interval).
    - Re-runs GetChainRouter + ChainFetcher.Validate against each failed endpoint inside a scoped 30s context.WithTimeout, with defer       
  cancel() — fixes goroutine/connection leaks from connector goroutines that only exit on ctx cancellation.                                 
    - On recovery, copy-on-write merges recovered sessions into rpsr.providerSessions (the old map may still be iterated by probeProviders /
   cleanupStaleTrackers without the lock) and calls sessionManager.UpdateAllProviders.                                                      
    - Exits when the failed list drains.
  - Also wraps the Phase 1 verification calls in a scoped context.WithTimeout(ctx, 30s) so a hung/blackholed provider can't stall validation
   of the others, and so temporary chain routers don't leak goroutines for the app's lifetime (rpcsmartrouter.go:~770,                      
  rpcsmartrouter.go:~800).                                                                                                                  
  - ChainTracker now uses healthyStaticProviders[0] instead of relevantStaticProviderList[0] so it doesn't poll a known-dead node.          
  - Adds 451 lines of tests in rpcsmartrouter_test.go.                                                                                      
                                                                                                                                            
  6763481 — duplicate provider names cause collateral exclusion                                                                             
                                                                                                                                            
  - Re-keys the failed sets from map[string]struct{} (keyed by provider.Name) to map[*lavasession.RPCStaticProviderEndpoint]struct{} (keyed 
  by pointer identity). Same fix applied to the backup-provider path (failedBackupNames → failedBackupSet).
  - Two providers configured with the same Name are now treated as distinct entities; only the actually-failing one is excluded.            
                                                                                                                                            
  9473684 — mutex around UpdateAllProviders to fix epoch-update race                                                                        
                                                                                                                                            
  - Pulls the existing rpsr.mu lock around the read → build-fresh-sessions → write-back → sessionManager.UpdateAllProviders block in        
  updateEpoch (rpcsmartrouter.go:~1690–1750), and the matching block in retryFailedStaticProviders.
  - The invariant being protected: the pair (write to rpsr.providerSessions, push snapshot to CSM) must be atomic. Without the lock,        
  updateEpoch and the retry goroutine can interleave so that CSM ends up with the older of the two snapshots while rpsr.providerSessions    
  holds the newer one — silently dropping providers until the next epoch tick repairs it.
  - The locked region is bounded CPU work (map rebuild + UpdateAllProviders's synchronous body, which dispatches probing to a goroutine).   
  The genuinely heavy cleanupStaleTrackers call stays outside the lock.                                                                     
  - Also changes updateEpoch to snapshot the session-manager keys under the lock first, then iterate, so map iteration isn't concurrent with
   the retry goroutine's writes.                                                                                                            
                          
  Net effect                                                                                                                                
                          
  A misconfigured or temporarily-down static provider no longer takes the smart router down at boot; it is excluded, logged loudly, and     
  re-validated every 3 minutes until it recovers — at which point it gets re-registered without an epoch boundary. Concurrent epoch ticks
  and recovery merges can no longer corrupt the CSM view of the provider set.

### DIFF
--- a/protocol/rpcsmartrouter/rpcsmartrouter.go
+++ b/protocol/rpcsmartrouter/rpcsmartrouter.go
@@ -1792,11 +1792,16 @@ func (rpsr *RPCSmartRouter) updateEpoch(epoch uint64) {
 			rpsr.backupProviderSessions[chainKey] = freshBackupSessions
 		}
 		server := rpsr.rpcServers[chainKey]
+
+		// UpdateAllProviders stays under rpsr.mu so the (rpsr.providerSessions write
+		// → csm push) pair is atomic with retryFailedStaticProviders' matching pair.
+		// Otherwise the two callers can push snapshots to csm in the opposite order
+		// they wrote rpsr.providerSessions, silently dropping providers until the
+		// next epoch. The synchronous body of UpdateAllProviders is a bounded map
+		// rebuild; probing is dispatched to a goroutine.
+		err := sessionManager.UpdateAllProviders(epoch, freshProviderSessions, freshBackupSessions)
 		rpsr.mu.Unlock()
 
-		// Heavy calls stay outside the lock (UpdateAllProviders triggers async probing,
-		// cleanupStaleTrackers does tracker removal)
-		err := sessionManager.UpdateAllProviders(epoch, freshProviderSessions, freshBackupSessions)
 		if err != nil {
 			utils.LavaFormatError("Failed to update providers on epoch change", err,
 				utils.LogAttr("epoch", epoch),
@@ -1804,8 +1809,9 @@ func (rpsr *RPCSmartRouter) updateEpoch(epoch uint64) {
 			)
 		}
 
-		// Cleanup stale ChainTrackers for endpoints no longer in the provider sessions
-		// This must happen AFTER UpdateAllProviders so connections are closed first
+		// cleanupStaleTrackers is the genuinely heavy work (tracker teardown +
+		// connection close) and stays outside the lock. Must run AFTER
+		// UpdateAllProviders so connections are closed first.
 		if server != nil {
 			rpsr.cleanupStaleTrackers(chainKey, server, freshProviderSessions, freshBackupSessions)
 		}
@@ -1954,10 +1960,16 @@ func (rpsr *RPCSmartRouter) retryFailedStaticProviders(
 
 			sessionManager := rpsr.sessionManagers[sessionManagerKey]
 			backupSessions := rpsr.backupProviderSessions[sessionManagerKey]
+
+			// UpdateAllProviders stays under rpsr.mu so this (rpsr.providerSessions
+			// write → csm push) pair is atomic with updateEpoch's matching pair.
+			// Otherwise a concurrent epoch tick can push to csm in the opposite
+			// order it wrote rpsr.providerSessions, silently dropping recovered
+			// providers until the next epoch.
+			err := sessionManager.UpdateAllProviders(currentEpoch, mergedSessions, backupSessions)
 			rpsr.mu.Unlock()
 
-			// Re-register all providers with session manager
-			if err := sessionManager.UpdateAllProviders(currentEpoch, mergedSessions, backupSessions); err != nil {
+			if err != nil {
 				utils.LavaFormatWarning("retry: failed to re-register recovered providers", err,
 					utils.LogAttr("chain", rpcEndpoint.ChainID),
 				)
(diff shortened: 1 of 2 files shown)


protocol/rpcsmartrouter/rpcsmartrouter.go
avitenzer, line 1: (a) TOCTOU race in retry → UpdateAllProviders (rpcsmartrouter.go:1923–1971). The retry path is:
                                                                                                                                                                                                              
  1. Take mu, capture currentEpoch, write mergedSessions to rpsr.providerSessions[key], capture sessionManager and backupSessions, release mu.
  2. Outside the lock, call sessionManager.UpdateAllProviders(currentEpoch, mergedSessions, backupSessions).                                                                                                  
                                              
  Between steps 1 and 2, updateEpoch(currentEpoch+1) can run, see mergedSessions, build fresh objects from it, and call UpdateAllProviders(currentEpoch+1, fresh, …). The retry's subsequent call is then     
  UpdateAllProviders(currentEpoch /* stale */, …). consumer_session_manager.go:176–178 rejects older-epoch updates outright:
  if epoch < previousEpoch {                                                                                                                                                                                  
      return utils.LavaFormatError("trying to update provider list for older epoch", …)                                                                                                                       
  }                                                                                                                                                                                                           
  Outcome: recovered providers are persisted in rpsr.providerSessions (so they'll be registered at the next epoch tick), but the retry's "[+] static provider re-registered successfully" log won't fire —    
  instead an error log appears. So this is a "delayed registration up to one epoch", not a permanent loss, but the warning at line 1961 will fire regularly under load and the immediate-availability claim of
   the retry isn't always true.                                                                                                                                                                               
                                                                                                                                                                                                              
  Fix options: (a) call UpdateAllProviders inside the lock (it's already called inside the lock at startup), or (b) re-read rpsr.epochTimer.GetCurrentEpoch() after unlock, bail if it changed, and let the   
  next epoch tick pick up the merged map.                                                                                                                                                                     
                                                                    
  (b) Same-key TOCTOU on updateEpoch: the symmetric inverse race also exists — updateEpoch holds the lock through the read/build/write section but releases it before its own UpdateAllProviders call (lines  
  1795–1799). If retry + epoch fire near-simultaneously, the order in which their two UpdateAllProviders calls reach the manager isn't determined by the order in which they wrote to rpsr.providerSessions.  
  Same fix as (a) — keep UpdateAllProviders under the lock or sequence with epoch read.

avitenzer, line 1: Duplicate provider names cause collateral exclusion (static_provider_endpoints.go:18–23 only validates non-empty Name; no uniqueness check). failedStaticNames is keyed by provider.Name, and the filter
   (rpcsmartrouter.go:992–998) drops every provider matching a failed name. If an operator names two providers identically with different URLs and one fails, both are excluded. The new test
  TestGracefulFailure_DuplicateNameCollision documents this rather than fixing it. Either reject duplicates in static_provider_endpoints.go at config-load time, or key the failed set on something more
  unique (e.g., Name + first URL).

avitenzer (approved):

PR #6, opened 2026-05-04
feat: Added graceful shutdown

sotskov-do:
# Graceful shutdown for `smartrouter`

## Summary

Implements end-to-end graceful shutdown for the smart router so that SIGINT/SIGTERM (and Kubernetes pod termination) drain in-flight HTTP requests and notify active WebSocket clients with a proper `1001 Going Away` close frame, instead of being killed mid-flight.

Adds a new `--shutdown-grace-period` flag (default **25s**, leaving a 5s margin under Kubernetes' default `terminationGracePeriodSeconds: 30s`).

## Behavior change

Before:

- `Start` blocked on a hard-coded `os.Interrupt` channel.
- On signal, the process exited immediately. In-flight HTTP relays were cut. Active WS clients saw an abrupt TCP close — indistinguishable from a crash. Provider WS pools and gRPC streaming pools leaked.

After:

- Cobra wires `signal.NotifyContext(ctx, SIGINT, SIGTERM)`. `Start` returns once endpoints are up; the cobra `RunE` waits on `<-ctx.Done()` and then calls `RPCSmartRouter.Stop(gracePeriod)`.
- `Stop` runs in two phases:
    1. **Phase 1 — drain client-facing layer in parallel.** Every chain listener's `Shutdown(ctx)` runs concurrently. Listeners stop accepting new connections, send `1001 Going Away` close frames on active WS connections, and wait for in-flight HTTP handlers + WS goroutines to unwind.
    2. **Phase 2 — close upstream pools.** `DirectWSSubscriptionManager.Close()` and `grpcSubscriptionManager.Stop()` run _after_ Phase 1 so in-flight relays don't lose their pools mid-call.
- A second SIGINT/SIGTERM during the drain phase force-terminates (we explicitly `cancel()` to restore default signal handling).

## Design choices

- **`ChainListener.Shutdown(ctx) error` is a new interface method** ([[protocol/chainlib/chainlib.go](vscode-webview://0h4nbkg85b592rtil09r5srijg806333fe45r2phameu9fne61cv/protocol/chainlib/chainlib.go)](vscode-webview://0h4nbkg85b592rtil09r5srijg806333fe45r2phameu9fne61cv/protocol/chainlib/chainlib.go)). Implemented on JSON-RPC, Tendermint, REST, gRPC, and `EmptyChainListener`. Each implementation knows how to drain its own runtime: fiber `ShutdownWithContext` for HTTP-backed listeners, `http.Server.Shutdown` (HTTP/2 GOAWAY) for gRPC.
    
- **Per-listener `wsWG sync.WaitGroup`** tracks active WS goroutines so Shutdown can wait for them after HTTP drains. The `Add(1)` is done **synchronously inside the upgrade middleware** (per-route on `/ws` and `/websocket`), _not_ inside the hijacked websocket goroutine — otherwise `wsWG.Wait` would race with a late `Add(1)` (fasthttp stops tracking hijacked connections, so `app.ShutdownWithContext` could return before the goroutine ran). The middleware also balances `Add` with `Done` if the upgrade itself fails (status ≠ 101), so a rejected upgrade doesn't deadlock `Wait`.
    
- **Single-writer discipline in `ConsumerWebsocketManager`** ([[protocol/chainlib/consumer_websocket_manager.go](vscode-webview://0h4nbkg85b592rtil09r5srijg806333fe45r2phameu9fne61cv/protocol/chainlib/consumer_websocket_manager.go)](vscode-webview://0h4nbkg85b592rtil09r5srijg806333fe45r2phameu9fne61cv/protocol/chainlib/consumer_websocket_manager.go)). The underlying gorilla/fasthttp websocket library is **not safe for concurrent writes**, so a single goroutine now serves both regular outbound frames and the server-shutdown close frame. All call-sites that previously did `websocketConnWriteChan <- msg` go through a `sendWS` helper that selects on `ctx`/`webSocketCtx` so they unblock when the writer exits — without this, shutting down would deadlock the per-conn goroutine and prevent `wsWG.Done()` from firing.
    
- **`ListenWithRetry` takes a `ctx`** ([[protocol/chainlib/common.go](vscode-webview://0h4nbkg85b592rtil09r5srijg806333fe45r2phameu9fne61cv/protocol/chainlib/common.go)](vscode-webview://0h4nbkg85b592rtil09r5srijg806333fe45r2phameu9fne61cv/protocol/chainlib/common.go)). Without this, the retry-sleep loop would re-bind the port after `Shutdown` returned and undo the shutdown.
    
- **Drain helper deduplicated** ([[protocol/chainlib/common.go](vscode-webview://0h4nbkg85b592rtil09r5srijg806333fe45r2phameu9fne61cv/protocol/chainlib/common.go)](vscode-webview://0h4nbkg85b592rtil09r5srijg806333fe45r2phameu9fne61cv/protocol/chainlib/common.go) — `drainHTTPThenWS`). Both fiber-based listeners (JSON-RPC, Tendermint) had byte-identical drain bodies; they now share one helper that documents the HTTP-then-WS ordering rationale in one place.

### DIFF
--- a/protocol/chainlib/jsonRPC.go
+++ b/protocol/chainlib/jsonRPC.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/goccy/go-json"
@@ -332,6 +333,8 @@ type JsonRPCChainListener struct {
 	wsSubscriptionManager      WSSubscriptionManager
 	listeningAddress           string
 	websocketConnectionLimiter *WebsocketConnectionLimiter
+	app                        *fiber.App     // captured during Serve so Shutdown can drain HTTP
+	wsWG                       sync.WaitGroup // tracks active downstream WS goroutines for Shutdown drain
 }
 
 // NewJrpcChainListener creates a new instance of JsonRPCChainListener
@@ -364,6 +367,7 @@ func (apil *JsonRPCChainListener) Serve(ctx context.Context, cmdFlags common.Con
 	test_mode := common.IsTestMode(ctx)
 	// Setup HTTP Server
 	app := createAndSetupBaseAppListener(cmdFlags, apil.endpoint.HealthCheckPath, apil.healthReporter)
+	apil.app = app
 
 	app.Use("/ws", func(c *fiber.Ctx) error {
 		apil.websocketConnectionLimiter.HandleFiberRateLimitFlags(c)
@@ -381,6 +385,8 @@ func (apil *JsonRPCChainListener) Serve(ctx context.Context, cmdFlags common.Con
 	apiInterface := apil.endpoint.ApiInterface
 
 	webSocketCallback := websocket.New(func(websocketConn *websocket.Conn) {
+		apil.wsWG.Add(1)
+		defer apil.wsWG.Done()
 		canOpenConnection, decreaseIpConnection := apil.websocketConnectionLimiter.CanOpenConnection(websocketConn)
 		defer decreaseIpConnection()
 		if !canOpenConnection {
@@ -411,7 +417,7 @@ func (apil *JsonRPCChainListener) Serve(ctx context.Context, cmdFlags common.Con
 			headerRateLimit:        uint64(rateLimit),
 		})
 
-		consumerWebsocketManager.ListenToMessages()
+		consumerWebsocketManager.ListenToMessages(ctx)
 	})
 	websocketCallbackWithDappID := constructFiberCallbackWithHeaderAndParameterExtraction(webSocketCallback, apil.logger.StoreMetricData)
 	app.Get("/ws", websocketCallbackWithDappID)
@@ -538,7 +544,7 @@ func (apil *JsonRPCChainListener) Serve(ctx context.Context, cmdFlags common.Con
 		apil.listeningAddress = addr
 	}()
 
-	ListenWithRetry(app, apil.endpoint.NetworkAddress, addrChannelSafe)
+	ListenWithRetry(ctx, app, apil.endpoint.NetworkAddress, addrChannelSafe)
 }
 
 func (apil *JsonRPCChainListener) GetListeningAddress() string {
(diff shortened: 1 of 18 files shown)


protocol/chainlib/jsonRPC.go
avitenzer, line 1: ▎ protocol/chainlib/jsonRPC.go:387-389 (and tendermintRPC.go:407-409) — wsWG.Add(1) race                                                                                                                                               
  ▎ Calling Add(1) inside the WS goroutine doesn't satisfy the WaitGroup happens-before rule. After fasthttp hijacks the connection it stops tracking it, so app.ShutdownWithContext can return before this goroutine is scheduled — 
  ▎ meaning wsWG.Wait may see 0 and return, then this Add(1) runs and the eventual Done is unmatched. Move the Add to the synchronous upgrade middleware:                                                                                
  ▎ app.Use("/ws", func(c *fiber.Ctx) error {
  ▎     apil.websocketConnectionLimiter.HandleFiberRateLimitFlags(c)                                                                                                                                                                     
  ▎     if websocket.IsWebSocketUpgrade(c) {                                                                                                                                                                                             
  ▎         apil.wsWG.Add(1)        // before any goroutine is spawned                                                                                                                                                                   
  ▎         c.Locals("allowed", true)                                                                                                                                                                                                    
  ▎         return c.Next()                                                                                                                                                                                                              
  ▎     }                                                                                                                                                                                                                                
  ▎     return fiber.ErrUpgradeRequired
  ▎ })                                                                                                                                                                                                                                   
  ▎ // …then in the WS callback:
  ▎ defer apil.wsWG.Done()                                                                                                                                                                                                               
  ▎ If the upgrade fails after Add, decrement via a deferred path / Locals flag.

protocol/chainlib/jsonRPC_test.go
avitenzer, line 1: ▎ protocol/chainlib/jsonRPC_test.go:16-17 — duplicate import                                                                                                                                                                           
  ▎ "github.com/gorilla/websocket"                                                                                                                                                                                                       
  ▎ gws "github.com/gorilla/websocket"                                                                                                                                                                                                   
  ▎ Same package twice. Pick one (websocket) and update the new tests to use it.

protocol/chainlib/jsonRPC.go
avitenzer, line 1: ▎ protocol/chainlib/jsonRPC.go:975-994 and tendermintRPC.go:875-899 — duplicate Shutdown bodies                                                                                                                                        
  ▎ The two Shutdown implementations are byte-identical except for the warning string. Extract a helper:
  ▎ // in common.go                                                                                                                                                                                                                      
  ▎ func drainHTTPThenWS(ctx context.Context, app *fiber.App, wg *sync.WaitGroup, name string) error {
  ▎     if app == nil { return nil }                                                                                                                                                                                                     
  ▎     httpErr := app.ShutdownWithContext(ctx)                                                                                                                                                                                          
  ▎     done := make(chan struct{})                                                                                                                                                                                                      
  ▎     go func() { wg.Wait(); close(done) }()                                                                                                                                                                                           
  ▎     select {                                                                                                                                                                                                                         
  ▎     case <-done:
  ▎     case <-ctx.Done():                                                                                                                                                                                                               
  ▎         utils.LavaFormatWarning(name+": WS goroutines did not finish within shutdown grace period", nil)
  ▎     }                                                                                                                                                                                                                                
  ▎     return httpErr
  ▎ }

avitenzer (commented):
Fill in the PR description

protocol/chainlib/jsonRPC.go
sotskov-do: Done

protocol/chainlib/jsonRPC_test.go
sotskov-do: Done

protocol/chainlib/jsonRPC.go
sotskov-do: Done

avitenzer (approved):

PR #7, opened 2026-05-05
fix(CV consistency): fix in CV in the case of consistency mismatch in  chosen providers

AnnaR-prog:
# Description

CV stall: fail fast when consistency filter drops sessions below quorum                                                        
                                                                                                                                 
  Bug                                                                                                                            
   
  CrossValidation requests stall ~30 s and return generic context deadline exceeded whenever the pre-request consistency filter  
  drops some — but not all — of the sessions returned by GetSessions. UsedProviders.sessionsLatestBatch stays at the original
  count, so RelayProcessor.checkEndProcessing waits for responses that will never arrive; the precise upstream cause is masked.  
                  
  Fix

  Three load-bearing pieces; only deliver the fast-fail in combination.                                                          
   
  - lavasession / rpcsmartrouter — counter sync + post-filter guard. New UsedProviders.ReleaseFromLatestBatch decrements the     
  counter for sessions dropped pre-dispatch. Wired into the failed-sessions release loop. Adds a post-filter CV guard that
  returns PairingListEmptyError with a precise message (sessionsAcquired / sessionsAfterFilter / sessionsFiltered) when survivors
   fall below AgreementThreshold.
  - relaypolicy — CV short-circuit on batch errors. Without it, batch errors route through SendRetry with NumOfProviders=1,
  silently violating quorum and re-masking the error. OnSendRelayResult gains a Selection parameter and short-circuits to        
  SendStop in CV mode.
  - rpcsmartrouter — release surviving sessions in the early-return path. The post-filter guard must release the surviving valid 
  sessions before returning, not just the failed ones. The state machine's validateReturnCondition requires CurrentlyUsed() == 0;
   without this release the request still stalls the full timeout. Discovered during the integration repro.
                                                                                                                                 
  Plus tests + comment cleanup: integration test that exercises the full path (3 sessions, AT=2, filter drops 2) and pins all    
  four invariants — < 1 s, error wraps PairingListEmptyError, CurrentlyUsed() == 0, SessionsLatestBatch() == 0. Verified to fail
  when the surviving-session release is removed. Updates a stale comment that no longer matches the post-short-circuit behavior. 
                  
  Reproduction / measurement                                                                                                     
   
  Three direct-rpc REST endpoints — one healthy, two via testutils/mock_node_proxy against the same upstream. Bootstrap,         
  set-block?height=5, advance seenBlock past EndpointLagThreshold, send a CV request with max-participants=3,
  agreement-threshold=2.                                                                                                         
                  
  End-to-end wall clock: 30.067 s → 0.085 s. Response now carries insufficient sessions for cross-validation consensus after     
  consistency filter with the diagnostic fields, instead of context deadline exceeded.
                                                                                                                                 
  Test plan       

  - lavasession, relaypolicy, relaycore, rpcsmartrouter all green.
  - Unit tests for the new primitives (TestReleaseFromLatestBatch,
  TestOnSendRelayResult/CrossValidation_stops_immediately_on_any_error).                                                         
  - Integration test verified to fail without the surviving-session release and pass with it.
  - Manual end-to-end repro on the proxy setup confirms 30 s → ~85 ms.                                                           
                                                                                                                                 
  Compatibility                                                                                                                  
                                                                                                                                 
  RelayPolicyInf.OnSendRelayResult signature gains a Selection parameter. Single production call site updated; test call sites   
  pass Stateless where appropriate.

### DIFF
--- a/protocol/rpcsmartrouter/rpcsmartrouter_server.go
+++ b/protocol/rpcsmartrouter/rpcsmartrouter_server.go
@@ -752,10 +752,15 @@ func (rpcss *RPCSmartRouterServer) sendRelayToDirectEndpoints(
 	// Pre-request consistency validation: filter out endpoints that are too far behind
 	validSessions, failedSessions, filterErr := rpcss.filterEndpointsByConsistency(ctx, sessions, protocolMessage)
 
-	// Release failed sessions via OnSessionFailure:
-	// - Provides QoS punishment (AppendRelayFailure)
-	// - Properly unlocks the session
-	// - Adds provider to unwantedProviders for retry exclusion
+	// Release failed sessions:
+	// - ReleaseFromLatestBatch decrements UsedProviders.sessionsLatestBatch so
+	//   RelayProcessor.checkEndProcessing matches the goroutines we will
+	//   actually launch. Without this the CV path can wait the full
+	//   processingTimeout (~30s) for responses that never arrive.
+	// - OnSessionFailure provides QoS punishment, unlocks the session, and
+	//   marks the provider unwanted for retry exclusion.
+	usedProviders := relayProcessor.GetUsedProviders()
+	releaseRouterKey := lavasession.NewRouterKeyFromExtensions(protocolMessage.GetExtensions())
 	for endpointAddress, sessionInfo := range failedSessions {
 		if sessionInfo != nil && sessionInfo.Session != nil {
 			utils.LavaFormatDebug("releasing stale session via OnSessionFailure",
@@ -763,6 +768,7 @@ func (rpcss *RPCSmartRouterServer) sendRelayToDirectEndpoints(
 				utils.LogAttr("error", lavasession.ConsistencyPreValidationError),
 				utils.LogAttr("GUID", ctx),
 			)
+			usedProviders.ReleaseFromLatestBatch(endpointAddress, releaseRouterKey, lavasession.ConsistencyPreValidationError)
 			rpcss.sessionManager.OnSessionFailure(sessionInfo.Session, lavasession.ConsistencyPreValidationError)
 		}
 	}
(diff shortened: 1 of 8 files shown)


protocol/rpcsmartrouter/rpcsmartrouter_server.go
avitenzer, line 1: 781-785 — Comment is misleading
  ▎                                                                                                                                                                                                      
  ▎ The comment says "Returning PairingListEmptyError lets the state machine retry with a fresh batch", but in this PR the CV branch in Policy.OnSendRelayResult (relaypolicy/policy.go:121) makes that 
  ▎ error result in SendStop, not retry. Future readers will look at this and assume retry is the intended behavior; the actual goal is fail-fast with the precise upstream cause. Suggest:              
  ▎
  ▎ // Post-filter CV guard: fail fast with a precise error message when the                                                                                                                             
  ▎ // filter dropped the surviving session count below the agreement threshold.
  ▎ // Returning PairingListEmptyError surfaces the precise cause; the CV                                                                                                                                
  ▎ // short-circuit in Policy.OnSendRelayResult ensures the state machine                                                                                                                               
  ▎ // stops immediately rather than retrying with NumOfProviders=1 (which
  ▎ // would silently violate quorum and mask this error).

avitenzer (commented):
▎ protocol/rpcsmartrouter/rpcsmartrouter_server_test.go — Missing integration test                                                                                                                     
  ▎                                                                                                                                                                                                      
  ▎ The 30s→0.085s benchmark in commit 97266e4 was manual. Please add a test (using the existing MockRelayProcessorForHeaders patterns or a small new test harness) that:
  ▎                                                                                                                                                                                                      
  ▎ 1. Sets selection = CrossValidation, AgreementThreshold = 2, gives 3 sessions to sendRelayToDirectEndpoints.
  ▎ 2. Stubs filterEndpointsByConsistency (or its dependencies) to return only 1 valid session.                                                                                                          
  ▎ 3. Asserts:                                                     
  ▎   - sendRelayToDirectEndpoints returns within < 1s (well under processingTimeout).                                                                                                                   
  ▎   - The returned error wraps lavasession.PairingListEmptyError. 
  ▎   - usedProviders.CurrentlyUsed() == 0 and usedProviders.SessionsLatestBatch() == 0 after return (this is the regression catch from commit 97266e4).

protocol/rpcsmartrouter/rpcsmartrouter_server.go
AnnaR-prog: Done

avitenzer (approved):

PR #9, opened 2026-05-05
feat: Added spec re-verification

sotskov-do:
## Summary

Adds **periodic re-verification of statically-configured upstreams** to the smart router. On every epoch tick, the router now re-validates each provider in the `direct-rpc` and `backup-direct-rpc` pools, demotes the ones that fail, and promotes recovered ones back into the active set — so a flapping upstream no longer stays unusable until the process restarts.

## What's in the change

- **New re-verifier** — runs synchronously from the existing epoch callback, bounded by `SpecReVerifyConcurrency = 5` and `SpecReVerifyAttemptTimeout = 30s`. Logs a warning when a per-chain cycle exceeds the attempt timeout so operators can spot ticks approaching epoch length.
- **Demote / promote reconciliation** — `updateEpoch` now diffs re-verified results against the freshly built session maps for both static and backup tiers. Demoted sessions' `DirectRPCConnections` are closed _after_ `UpdateAllProviders` to avoid racing in-flight relays.
- **Safe parser cloning for gRPC validation** — `NewGrpcChainProxy → setupForProvider` mutates the parser's `registry`/`codec` to the verification connection's lifetime; reusing the live parser would panic gRPC relays on a nil connector when the per-attempt context cancels. `CloneChainParserForValidation` isolates that mutation; non-gRPC parsers pass through unchanged.

## Tests
- demote, promote, mixed-tier, and timeout paths.
- verifies `cloneForValidation` produces an isolated parser.

### DIFF
--- a/protocol/rpcsmartrouter/spec_reverifier.go
+++ b/protocol/rpcsmartrouter/spec_reverifier.go
@@ -49,6 +49,11 @@ type chainReverifyInputs struct {
 	convertProvidersToSessions func([]*lavasession.RPCStaticProviderEndpoint) map[uint64]*lavasession.ConsumerSessionsWithProvider
 	configuredStatic           []*lavasession.RPCStaticProviderEndpoint
 	configuredBackup           []*lavasession.RPCStaticProviderEndpoint
+	// validateFn is the per-provider validation callback applyReverification
+	// dispatches to. Production leaves it nil and applyReverification falls back
+	// to validateProvider (real network probe). Tests inject a fake to exercise
+	// updateEpoch's orchestration without standing up upstreams.
+	validateFn func(context.Context, *lavasession.RPCStaticProviderEndpoint) error
 }
 
 // applyReverification revalidates configured providers for one tier and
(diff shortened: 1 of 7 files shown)


protocol/rpcsmartrouter/spec_reverifier.go
avitenzer, line 1: **Issue:** closeDemotedDirectConnections runs in a goroutine without panic recovery. A panic in dc.Close() kills the goroutine silently.                                                                   
  **Suggestion:**                                                                                                                                                                                            
  func closeDemotedDirectConnections(demoted []*lavasession.ConsumerSessionsWithProvider) {                                                                                                              
      defer func() {                                                                       
          if r := recover(); r != nil {                                                                                                                                                                  
              utils.LavaFormatError("re-verify: panic closing demoted connections", nil,                                                                                                                 
                  utils.LogAttr("panic", r))                                            
          }                                                                                                                                                                                              
      }()                                                           
      // ... existing body                                                                                                                                                                               
  }

avitenzer, line 1: **Issue:** All TestUpdateEpoch_* tests construct RPCSmartRouter directly, leaving reverifyInputs nil; no integration test exercises the new code path through updateEpoch.
  **Suggestion:** Add a TestUpdateEpoch_AppliesReverification that populates rpsr.reverifyInputs[chainKey] with a convertProvidersToSessions fake and an injected validate (the latter requires either       
  parameterizing the production-path validate or a test-only seam — a small refactor: store validateFn on chainReverifyInputs, default-initialize to the production callback). Assert: demoted entries
  dropped from rpsr.providerSessions[chainKey]; promoted entries appear with PairingEpoch == epoch.

sotskov-do: Done

sotskov-do: This is impossible case, nothing in `closeDemotedDirectConnections` could cause panic: `http` and `websocket` connections returning `nil`. `grpc` handle closing under mutex and also checking everything related to `nil`. So I think this check is redundant.

AnnaR-prog (approved):

avitenzer (approved):

PR #11, opened 2026-05-07
fix(rpcsmartrouter): honor force-cache-refresh on retry path

AnnaR-prog:
addArchiveExtension and removeArchiveExtension rebuild metadata from scratch with only the new extension override. After re-parse the new chainMessage's forceCacheRefresh flag defaults to false, so retries silently consult the cache despite the client setting lava-force-cache-refresh: true. Copy the directive from the source protocolMessage onto the new one after each archive transition.

Also keep "Cached" in the Lava-Provider-Address list when retries occurred — the previous code excluded it, leaving Lava-Retries=N and a single provider name that disagreed on how many actors participated.

Provider-selection directives (lava-select-provider, lava-stickiness) are intentionally not preserved: the failover path may need to fall through to a different provider.

MAG-1653

### DIFF
--- a/protocol/relaycore/relay_state.go
+++ b/protocol/relaycore/relay_state.go
@@ -193,6 +193,7 @@ func addArchiveExtension(ctx context.Context, protocolMessage chainlib.ProtocolM
 		utils.LavaFormatError("Failed adding archive extension", err, utils.LogAttr("apiUrl", relayRequestData.ApiUrl))
 		return protocolMessage
 	}
+	preserveForceCacheRefresh(protocolMessage, newProtocolMessage)
 	archiveStatus.isUpgraded.Store(true)
 	archiveStatus.isArchive.Store(true)
 	return newProtocolMessage
@@ -224,8 +225,26 @@ func removeArchiveExtension(ctx context.Context, protocolMessage chainlib.Protoc
 		utils.LavaFormatError("Failed removing archive extension", err, utils.LogAttr("apiUrl", relayRequestData.ApiUrl))
 		return protocolMessage
 	}
+	preserveForceCacheRefresh(protocolMessage, newProtocolMessage)
 	archiveStatus.isArchive.Store(false)
 	return newProtocolMessage
+}
+
+// preserveForceCacheRefresh copies the lava-force-cache-refresh directive from
+// src onto dst. ParseRelay only sees the metadata we hand it (just the new
+// extension override) and so calls SetForceCacheRefresh(false) on dst — without
+// this copy, a client that sent lava-force-cache-refresh: true gets cache-served
+// data on the retry/failover path because the post-archive message no longer
+// carries the directive (MAG-1653). Provider-selection directives like
+// lava-select-provider and lava-stickiness are intentionally NOT copied here:
+// the failover path may need to fall through to a different provider.
+func preserveForceCacheRefresh(src, dst chainlib.ProtocolMessage) {
+	if src == nil || dst == nil {
+		return
+	}
+	if src.GetForceCacheRefresh() {
+		dst.SetForceCacheRefresh(true)
+	}
 }
 
 // cacheBlockHashes marks all requested block hashes as irrelevant for future queries.
--- a/protocol/relaycore/relay_state.go
+++ b/protocol/relaycore/relay_state.go
@@ -193,7 +193,7 @@ func addArchiveExtension(ctx context.Context, protocolMessage chainlib.ProtocolM
 		utils.LavaFormatError("Failed adding archive extension", err, utils.LogAttr("apiUrl", relayRequestData.ApiUrl))
 		return protocolMessage
 	}
-	preserveForceCacheRefresh(protocolMessage, newProtocolMessage)
+	preserveRetrySafeDirectives(protocolMessage, newProtocolMessage)
 	archiveStatus.isUpgraded.Store(true)
 	archiveStatus.isArchive.Store(true)
 	return newProtocolMessage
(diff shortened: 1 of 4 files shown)


protocol/relaycore/relay_state.go
avitenzer, line 1: **Issue:** This helper fixes lava-force-cache-refresh, but the same rebuild path also drops lava-providers-block, lava-relay-timeout, and lava-debug-relay. In particular,                  
  BaseProtocolMessage.GetBlockedProviders() will return nil on the post-upgrade message, so a client that explicitly blocked providers can have them selected on the failover path — same
  shape of bug as the one being fixed.                                                                                                                                                    
  **Suggestion:** Either preserve all retry-safe directives in one place or, at a minimum, file a follow-up under MAG-1653 and document the gap. A general-purpose helper would scale better
  than per-directive copies:                                                                                                                                                              
                                                                                                                                                                                          
  // preserveDirectives carries directives that are tied to consumer intent and
  // must not be reset by the rebuild. It deliberately omits lava-select-provider                                                                                                         
  // and lava-stickiness — the failover path may need to fall through.                                                                                                                    
  func preserveDirectives(src, dst chainlib.ProtocolMessage) {                                                                                                                            
      if src == nil || dst == nil {                                                                                                                                                       
          return                                                                                                                                                                          
      }                                                              
      if src.GetForceCacheRefresh() {                                                                                                                                                     
          dst.SetForceCacheRefresh(true)                             
      }
      // also: copy lava-providers-block, lava-relay-timeout, lava-debug-relay                                                                                                            
      // from src.GetDirectiveHeaders() into dst.                                                                                                                                         
  }

avitenzer, line 1: **Issue:** The chain message's flag is updated but BaseProtocolMessage.directiveHeaders on dst does not get a lava-force-cache-refresh entry, so the directive map is out of sync with the  
  underlying chain message after this call.                                                                                                                                             
  **Suggestion:** If you keep the per-directive helper, also set the entry in the directive headers map for consistency (would require an interface to expose the map setter, or build it in  
  via the broader preserve pattern above).

protocol/rpcsmartrouter/rpcsmartrouter_server.go
avitenzer, line 1: **Issue:** With "Cached" now allowed, the resulting Lava-Provider-Address value can be "Cached,lava@providerX" (or vice versa) depending on map ordering. Consumers parsing this header may 
  assume a stable ordering or that "Cached" is always first/last.                                                                                                                        
  **Suggestion:** Either sort allProvidersList for stability or keep "Cached" pinned to a deterministic position. Even a sort.Strings(allProvidersList) would make integration tests less     
  flaky and downstream parsers happier.

protocol/relaycore/relay_state.go
AnnaR-prog: Done — expanded the helper to cover relay-timeout and debug-relay. One note on
  providers-block first, then what I changed.                                                                            
                                             
  On lava-providers-block specifically: it's actually already preserved, just not via the directive map.                 
  GetBlockedProviders() has exactly one production caller — lavasession.NewUsedProviders at used_providers.go:22. That   
  constructor runs once at the top of ProcessRelaySend (rpcsmartrouter_server.go:575) against the original
  protocolMessage, copies the addresses into unwantedProviders/originalUnwantedProviders, and from then on the block list
   lives in usedProviders, which is reused across every retry. Post-upgrade GetBlockedProviders() returning nil is a
  tree-falls-in-a-forest event — nothing reads it after construction. I left a comment in the helper noting this so the
  next reader doesn't add it back unnecessarily.

  What I added: relay_state.go now has preserveRetrySafeDirectives(src, dst) as the orchestrator, called from both       
  addArchiveExtension and removeArchiveExtension. Underneath it, three single-purpose helpers — one per directive — each
  handling its own storage shape:                                                                                        
                  
  - preserveForceCacheRefresh → chainMessage flag + map entry                                                            
  - preserveRelayTimeout → chainMessage timeoutOverride + map entry
  - preserveDebugRelay → map entry only (no chainMessage field)                                                          
                                                                                                                         
  I went with explicit per-directive helpers rather than a blanket "copy every directive except a denylist" because the  
  denylist would be load-bearing in a non-obvious way: lava-select-provider and lava-stickiness must not be preserved or 
  test_2_1_one_provider_down_retry_to_next regresses (the pin would stick across retries and failover would never reach a
   different provider). An allowlist makes that intent explicit at the call site instead of buried in an exclusion check.

  Tests: added TestArchiveAddPreservesRelayTimeout, TestArchiveAddDoesNotInventRelayTimeout,                             
  TestArchiveAddPreservesDebugRelay, and — as a guard rail — TestArchiveAddDoesNotPreserveSelectProvider. That last one
  fails loudly if anyone ever expands the helper to include the pin, and the failure message points directly at the      
  failover test that would regress.

  So the only directives the rebuild can still drop are lava-select-provider and lava-stickiness, both intentional, both 
  pinned by tests.

AnnaR-prog: Done

protocol/rpcsmartrouter/rpcsmartrouter_server.go
AnnaR-prog: switched to insertion-order build with a seen set instead of for provider := range         
  allProvidersMap. Walks protocolErrorResults → nodeErrorResults → successResults, then appends the resolver             
  (providerAddress, which is "Cached" when the cache served the response). The seen set deduplicates without losing      
  first-seen position.                                                                                                   
                                                                                                                         
  Two properties this gives us:
                                                                                                                         
  1. Deterministic across runs. Map iteration was the only source of randomness — gone now. Same inputs always produce   
  the same header value.                                                                                                 
  2. last entry == response source. Whether the cache or a real provider answered, the resolver is appended last.        
  Examples:                                                                                                              
    - Cache resolves after p1 fails → Lava-Provider-Address: lava@p1,Cached
    - p2 succeeds after p1 fails → Lava-Provider-Address: lava@p1,lava@p2                                                
                                                                                                                         
  That's a usable contract for downstream parsers — they can read the last entry to attribute the response, and walk     
  earlier entries to see the failure trail in chronological-ish order.                                                   
                                                                                                                         
  Honest caveat: it's chronological within the iteration order of each result slice, but we don't track per-result       
  attempt timestamps, so in concurrent/batched scenarios the ordering could still drift from strict wall-clock time.
  Failures-before-resolver and stable-across-runs both hold either way.

avitenzer (approved):

PR #14, opened 2026-05-10
feat: Added support of backup providers to WS/gRPC

sotskov-do:
# Description

- Adds a two-tier (primary → backup) endpoint cascade to `DirectWSSubscriptionManager` and `DirectGRPCSubscriptionManager`, matching the HTTP backup-provider model already used by `ConsumerSessionManager`.
- Backup endpoints are consulted only when the primary tier is exhausted; sticky-session lookups span both tiers via a unified URL map.
- `CreateSmartRouterEndpoint` now harvests WS/gRPC URLs from `healthyBackupProviders` in addition to `healthyStaticProviders` and wires them into the new manager constructors. Endpoint collection is factored into shared `collectWSEndpoints` / `collectGRPCEndpoints` helpers that log the tier each endpoint came from.

### DIFF
--- a/protocol/rpcsmartrouter/direct_grpc_subscription_manager.go
+++ b/protocol/rpcsmartrouter/direct_grpc_subscription_manager.go
@@ -250,12 +250,14 @@ func (dgm *DirectGRPCSubscriptionManager) cleanupStaleSubscriptions() {
 // This enables tools like grpcurl to discover services through the smart router.
 // The cleanup function should be called when the connection is no longer needed.
 func (dgm *DirectGRPCSubscriptionManager) GetReflectionConnection(ctx context.Context) (*grpc.ClientConn, func(), error) {
-	if len(dgm.grpcEndpoints) == 0 {
-		return nil, nil, fmt.Errorf("no gRPC endpoints available for reflection")
+	// Select an endpoint via the primary→backup cascade. Reflection is read-only
+	// metadata so it isn't client-pinned; clientKey is empty.
+	endpoint, err := dgm.selectEndpoint(ctx, "", nil)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


protocol/rpcsmartrouter/direct_grpc_subscription_manager.go
avitenzer, line 1: IsStreamingMethod panics when only backup endpoints are configured

  protocol/rpcsmartrouter/direct_grpc_subscription_manager.go:283
  
  pool, err := dgm.getOrCreatePool(ctx, dgm.grpcEndpoints[0])

  There is no bounds check. The PR's own constructor docstring now states "Either slice may be nil or empty, but at least one must be non-empty for selection to succeed," and
  CreateSmartRouterEndpoint instantiates the manager when either tier is non-empty. So a config with WS in static-providers: and gRPC only in backup-providers: is now a legitimate
  configuration — and it will crash on the first call to IsStreamingMethod with runtime error: index out of range [0] with length 0. Pre-existing code, but the PR is what makes the
  failure path reachable.

avitenzer, line 1: **Issue:** Same problem as above in a milder form — guard only inspects grpcEndpoints, ignoring grpcBackupEndpoints. Error message becomes misleading ("no gRPC endpoints available" when
   there are backup ones).
  **Suggestion:** Same fix — replace the direct grpcEndpoints[0] with selectEndpoint(ctx, "", nil).

sotskov-do: Done.

sotskov-do: Done.

AnnaR-prog (approved):

PR #16, opened 2026-05-11
fix(rpcsmartrouter): propagate HTTP status into classifier message (MAG-1666)

avitenzer:
## Summary

- The classifier's `HTTPStatusContains(N)` rules for JSON-RPC transport were never firing for HTTP 4xx + JSON-RPC error body because the consumer-side message synthesis only carried the body's `error.message`, dropping the HTTP status digits.
- Result: HTTP 404, 405, 413 (declared `Retryable: false` in `error_codes.go`) were silently misclassified as UNKNOWN → Priority 5 fail-open retryable.
- Fix: prepend `"HTTP <status>: "` to the classifier message at the two sites that drive retry/health decisions (`direct_rpc_relay.go::sendJSONRPCRelay` and `node_error_handler.go::ExtractNodeErrorDetails`). `CodeEquals` matchers still run before `HTTPStatusContains`, so meaningful JSON-RPC codes classify identically — the prefix only changes the verdict for HTTP 4xx/5xx + generic body code where the HTTP status is the authoritative signal.

Closes: [MAG-1666](https://magmadevs.atlassian.net/browse/MAG-1666). The same change closes two follow-up drafts filed against 404 and 413 in [smart-router-automation BUGS/](https://github.com/Magma-Devs/smart-router-automation/tree/main/BUGS) — same defect, same fix.

## What changed

| File | Change |
|---|---|
| `protocol/rpcsmartrouter/direct_rpc_relay.go` | Prepend `"HTTP <status>: "` to `classifierMessage` before `IsNonRetryableNodeErrorWithContext` when status is non-2xx. |
| `protocol/chainlib/node_error_handler.go` | Same prefix in `ExtractNodeErrorDetails` step 1 for the telemetry/health path through `classifyDirectRPCError`. |
| `protocol/common/error_registry_test.go` | New `TestClassifyError_GenericJsonRPCHTTPStatusMappings` — JSON-RPC analogue of the existing REST classifier test; locks the contract for all 8 status codes covered by `httpStatusMessageMappings`. |
| `protocol/chainlib/handle_and_classify_test.go` | Updated `TestExtractNodeErrorDetails_JSONRPCBodyBeatsHTTPStatus` to assert the composed `"HTTP 400: opaque"` message while keeping the code-priority invariant (`-32700` still wins via `CodeEquals`). |

## Test plan

- [x] `go test ./protocol/common/...` — all green, includes new `TestClassifyError_GenericJsonRPCHTTPStatusMappings` (8/8 status codes).
- [x] `go test ./protocol/chainlib/...` — all green, includes updated `TestExtractNodeErrorDetails_JSONRPCBodyBeatsHTTPStatus`.
- [x] `go test ./protocol/rpcsmartrouter/...` — all green (78s).
- [x] `go test ./protocol/relaypolicy/... ./protocol/relaycore/...` — all green.
- [x] Integration: `tests/simulator/production_tests/test_phase1_2_http_status_codes.py` (smart-router-automation) — all 13 status codes (400-504) pass against a k3d cluster running this branch. The MAG-1666 xfail on HTTP 405 was removed in the same change (XPASS'd).
- [x] Integration: adjacent `test_router_retry_group2_failover.py` and `test_router_retry_group13_error_classifier_tiers.py` — green (the 405 xfail in group13 was also removed in the same change).

## Notes for review

- `CodeEquals` matchers run before `HTTPStatusContains` in the classifier, so a real JSON-RPC standard code (-32601, -32603, -32700, …) is unaffected by the prefix — its `CodeEquals` rule still wins. This is verified by `TestExtractNodeErrorDetails_JSONRPCBodyBeatsHTTPStatus` (`-32700` continues to classify as `USER_PARSE_ERROR` even with the new prefix).
- HTTP 429 and 504 take a separate sentinel path through `ValidateStatusCodes` whose error string already carries the digits — those continue to classify correctly on the existing path, untouched by this change.
- The REST + gRPC transports already include `httpStatusCodeMappings` (`CodeEquals` for HTTP statuses) so they were not affected by this defect.

[MAG-1666]: https://magmadevs.atlassian.net/browse/MAG-1666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/smartrouter_examples/smartrouter_eth.yml
+++ b/config/smartrouter_examples/smartrouter_eth.yml
@@ -14,11 +14,11 @@ direct-rpc:
     chain-id: "ETH1"
     api-interface: "jsonrpc"
     node-urls:
-      - url: "https://eth.llamarpc.com"
+      - url: "https://lb.drpc.live/ethereum/AgR9ugK3ak9roEnJY1YOg8O1VFMFBNUR8bRUehXRfUMv"
         skip-verifications:
           - chain-id
           - pruning
-      - url: "wss://eth.llamarpc.com"
+      - url: "wss://lb.drpc.live/ethereum/AgR9ugK3ak9roEnJY1YOg8O1VFMFBNUR8bRUehXRfUMv"
 
   # HTTP+WS Endpoint 2
   - name: "eth-rpc-2"
@@ -29,7 +29,7 @@ direct-rpc:
         skip-verifications:
           - chain-id
           - pruning
-      - url: "wss://json-rpc.8zfcse2amst1lajmh299uq4jn.blockchainnodeengine.com/?key=AIzaSyDyUtm6b-e-xKDQgVWzlroHdVTytiXEDik"
+      - url: "wss://g.w.lavanet.xyz:443/gateway/eth/rpc/4926f3fd246058892909cdda0c88f8c7"
 
   # HTTP+WS Endpoint 3
   - name: "eth-rpc-3"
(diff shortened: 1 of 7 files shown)


AnnaR-prog (commented):
Medium: protocol/common/error_registry_test.go claims it would catch removing the prefix at the direct relay call site, but it only calls ClassifyError with an already-prefixed message. A regression in protocol/rpcsmartrouter/direct_rpc_relay.go would still pass this test. I’d add a direct-relay test for HTTP 404 / 413 with a JSON-RPC body and generic code.

It's done

AnnaR-prog (approved):

PR #25, opened 2026-05-15
feat(compose): local docker-compose for smart-router

ssejzer:
Closes: #MAG-1755

- Adds `docker compose` stack under `compose/` for testing local smart-router binary ( routers + cache + Traefik on a single network)
- `./scripts/compose_up.sh` rebuilds the binary statically
  (`CGO_ENABLED=0 -trimpath -tags netgo -ldflags '-w -s'`) into `build/smartrouter`,
  Top-level `Dockerfile` (CI multi-stage build) is untouched.
- `config/values.yml` is the source of truth; `scripts/render_compose.py`
  generates `compose/docker-compose.yml`, `compose/configs/router-*.yml`,
  `compose/traefik/dynamic.yml`, and `compose/usage.txt` (all gitignored).
- Bundles `specs/{cosmoshub,cosmossdkv50,cosmoswasm}.json` so the `lava-grpc`
  router can resolve the COSMOSHUB spec graph locally.
- Ships 4 routers covering every smart-router interface: `eth` (jsonrpc),
  `lava-rest` (rest), `lava-tm` (tendermintrpc), `lava-grpc` (grpc).
- 22 unit tests for the renderer (`scripts/test_render_compose.py`).
- Uses same Dockerfile
- Makefile can trigger compose scripts and now displays a usage screen when invoked without target

## Known caveat

`lava-grpc-router` crash-loops against the public Lava gateway at startup —
the smart-router gRPC connector opens up to 100 simultaneous TLS handshakes
(`protocol/chainlib/chainproxy/connector.go:194`,
`protocol/lavasession/direct_rpc_connection.go:648`) which the gateway
rate-limits. This is a binary-level issue independent of compose; documented
in `compose/README.md`. The other 3 interfaces work end-to-end.


compose/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# syntax=docker/dockerfile:1
```
nimrod-teich: why are we using a different Dockerfile for the docker-compose than the one that is used anywhere else?

nimrod-teich (changes requested):
Check my comments + make the scripts part of the make file

compose/Dockerfile
ssejzer: the one in the root dir takes longer ti build. It is used by the `release_docker` job to tag (example v6.2.0)  multi-arch images and push them to ghcr.io/magma-Devs/smart-router.
I'll try to improve it

ssejzer: we are now using the same dockerfile

PR #28, opened 2026-05-17
fix(chainlib): detect malformed upstream responses for retry

nimrod-teich:
## Summary
Upstream JSON-RPC and REST responses that were truncated mid-write or missing required envelope fields (`result` / `error`) were forwarded to clients as HTTP 200 successes with no retry — the router was the protection layer that wasn't protecting. This PR adds two complementary defenses so those bodies are caught and routed through the existing retry pipeline against another provider.

## What changed

**1. Transport-level detection — `protocol/rpcsmartrouter/direct_rpc_relay.go`**
- `sendJSONRPCRelay`: every 2xx response with a non-empty body is passed through `json.Valid`. Failures are wrapped via `classifyAndWrap(..., common.TransportJsonRPC)` so the relay pipeline treats them as protocol errors (same bucket as a connection RST), not node errors — flaky upstreams stay attributable in per-provider dashboards.
- `sendRESTRelay`: same check, but gated on `looksLikeJSONOpening` (first non-whitespace byte is `{` or `[`) so REST endpoints serving legitimate plain-text or binary content aren't false-flagged. Wrapped as `common.TransportREST`.
- gRPC path is unchanged — binary wire format means truncation surfaces at the HTTP/2 transport layer below this code.

**2. Application-level envelope shape check — `protocol/chainlib/chainproxy/rpcInterfaceMessages/`**
- New `jsonRPCEnvelopeScanner.go`: a single-pass, depth-aware, string-aware byte scanner that records the presence and byte range of the top-level `result` / `error` fields. Zero-allocation on the value bytes (sub-slicing, not copying), preserving the optimization the previous implementation captured by omitting `Result` from its decode struct.
- `jsonRPCMessage.go`: shared `checkJsonrpcEnvelope` helper distinguishes three outcomes — malformed bytes, schema-incomplete (missing both fields, including `error: null` with no result), and healthy. `JsonrpcMessage.CheckResponseError` and `CheckResponseErrorForJsonRpcBatch` both call into it so the rules stay in lockstep.
- `tendermintRPCMessage.go`: `TendermintrpcMessage.CheckResponseError` reuses `checkJsonrpcEnvelope` for the envelope check and adds Tendermint's `{"response":{"code":N,"log":M}}` inner-error inspection on top.

Signatures of `CheckResponseError` and `CheckResponseErrorForJsonRpcBatch` are unchanged — `(hasError bool, errorMessage string)`. The verdict refactor is internal to `checkJsonrpcEnvelope`.

## Behavior change worth flagging

`{"error": null}` with no `result` previously decoded as `result.Error == nil` and was silently returned to the client as `(false, "")`. It is now classified as a schema violation and routed through the retry pipeline. Any upstream that emits this shape (rare; some custom RPCs) will start triggering retries. Symmetric case `{"result": X, "error": null}` is preserved as healthy.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./protocol/...` clean
- [x] New scanner unit tests (37 cases) cover truncation mid-value/string/escape, escape handling, depth tracking with braces in strings, top-level array/scalar rejection, whitespace tolerance.
- [x] New `CheckResponseError` unit tests cover the two ticket shapes (truncated bytes, missing both fields) for JSON-RPC single, JSON-RPC batch, and Tendermint, plus the `error: null` edge cases.
- [x] New integration test `TestDirectRPCRelaySender_MalformedJSONResponseRoutesAsTransportError` asserts that truncated/garbage 2xx bodies surface as errors wrapped in `*common.LavaWrappedError` (so a future refactor bypassing `classifyAndWrap` would be caught), and that non-JSON REST bodies still pass through.
- [x] Two pre-existing tests flipped from documenting the bug (`Invalid JSON should not return error (fail-open)`) to asserting the fixed behavior, with comments explaining the inversion.

## Unblocks
- MAG-1718 Phase 1.6 wrong-data cases (truncated, missing-fields) in the smart-router-automation simulator — those xfails can flip once this lands.

## Follow-ups (not in this PR)
- **`RelayResult.IsMalformed` field + Prometheus counter**: would let dashboards slice retries by `malformed` vs `node_error`. The transport-level detection already attributes correctly to `ProtocolError`; this would add per-provider granularity.
- **WebSocket bypass**: `results_manager.go` still skips `CheckResponseError` for `Subscribe`-tagged responses. WS retry semantics differ (re-subscribe vs re-request) — separate ticket.
- **REST `IsNodeError` escalation**: this PR keeps REST malformed detection at the transport layer (returns an error from `sendRESTRelay`). The pre-existing pattern where Cosmos tx errors (HTTP 200, non-zero code) are detected but don't drive `IsNodeError` is preserved — separate ticket if anyone wants retry on those.

### DIFF
--- a/protocol/chainlib/chainproxy/rpcInterfaceMessages/jsonRPCEnvelopeScanner.go
+++ b/protocol/chainlib/chainproxy/rpcInterfaceMessages/jsonRPCEnvelopeScanner.go
@@ -1,0 +1,283 @@
+package rpcInterfaceMessages
+
+import (
+	"errors"
+	"fmt"
+)
+
+// jsonrpcScanResult holds presence flags and byte ranges for the top-level
+// "error" and "result" fields of a JSON-RPC envelope. Byte slices are
+// sub-slices of the input data — no copying — and are valid only as long
+// as the caller retains a reference to that data.
+type jsonrpcScanResult struct {
+	hasError    bool
+	errorBytes  []byte
+	hasResult   bool
+	resultBytes []byte
+}
+
+// scanJsonrpcEnvelope walks data as a single JSON object and records the
+// presence and byte range of the top-level "error" and "result" fields.
+// Other keys are skipped without allocation. The walk is depth-aware and
+// string-aware: braces inside string values do not affect nesting.
+//
+// Returns a non-nil error when the input is not a syntactically well-formed
+// JSON object (truncated bytes, top-level array/scalar, structural break).
+// Keys are matched by literal byte comparison; spec-legal but exotic
+// spellings like "result" for "result"
+// are not decoded and will not be recognized.
+func scanJsonrpcEnvelope(data []byte) (jsonrpcScanResult, error) {
+	var res jsonrpcScanResult
+	pos := skipWhitespace(data, 0)
+	if pos >= len(data) {
+		return res, errors.New("empty input")
+	}
+	if data[pos] != '{' {
+		return res, fmt.Errorf("expected '{' at position %d, got %q", pos, data[pos])
+	}
+	pos++
+	pos = skipWhitespace(data, pos)
+	if pos < len(data) && data[pos] == '}' {
+		return res, nil
+	}
+	for {
+		if pos >= len(data) {
+			return res, errors.New("truncated: expected key")
+		}
+		if data[pos] != '"' {
+			return res, fmt.Errorf("expected string key at position %d, got %q", pos, data[pos])
+		}
+		keyStart := pos + 1
+		keyEnd, next, err := scanString(data, pos)
+		if err != nil {
+			return res, fmt.Errorf("key: %w", err)
+		}
+		pos = next
+		pos = skipWhitespace(data, pos)
+		if pos >= len(data) || data[pos] != ':' {
+			return res, fmt.Errorf("expected ':' at position %d", pos)
+		}
+		pos++
+		pos = skipWhitespace(data, pos)
+		valueStart := pos
+		valueEnd, err := skipValue(data, pos)
+		if err != nil {
+			return res, fmt.Errorf("value: %w", err)
+		}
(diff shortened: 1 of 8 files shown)


avitenzer (approved):
Approving. The core fix is solid: layering transport-level `json.Valid` with an application-level envelope scanner closes the silent-forwarding gap, splits attribution between protocol-error (truncated bytes) and node-error (schema-incomplete) buckets, and preserves the zero-allocation-on-the-value-bytes optimization the previous implementation captured.

Pushed three small follow-ups as `fe1663b`:

1. **Docstring repair** in `scanJsonrpcEnvelope` — the comment that tried to contrast `"result"` with its unicode-escape spelling rendered identically in source, leaving the sentence as a tautology. Rewritten textually.
2. **Log-body cap** at 2 KB in the two malformed-response `LavaFormatWarning` sites — a persistently flaky upstream returning megabyte-scale garbage would otherwise saturate the log pipeline. New helper `truncateForLog`.
3. **`defer`-restore on `BatchNodeErrorOnAny`** in `partial_success_with_malformed_sibling_default_mode` — matches the strict-mode test pattern and prevents global-state leakage if a future test is added after this one.

PR description also rewritten to reflect what actually ships (the prior description claimed a `RelayResult.IsMalformed` field, a `common.ResponseCheck` struct refactor, and changes in `results_manager.go` / `chain_fetcher.go` that aren't in the diff — those moved to the follow-ups section).

Behavior change worth a release-note line: `{"error": null}` with no `result` previously returned `(false, "")` silently; now it's classified as a schema violation and triggers retry. Symmetric `{"result": X, "error": null}` is preserved as healthy.

Tests pass: `protocol/chainlib/chainproxy/rpcInterfaceMessages` (0.89s) and integration tests for the malformed-detection branches in `protocol/rpcsmartrouter` (3.07s) green after the fixups.

PR #29, opened 2026-05-17
chore(release): drive version from git tag, prune Lava residue

nimrod-teich:
## Summary

Collapses several pieces of upstream Lava residue around versioning and the release pipeline so the binary customers run prints the version you expect, the release process is reproducible, and the GitHub Release page produces a clean, focused artifact set for v1.

> **Note for reviewers:** [#30](https://github.com/Magma-Devs/smart-router/pull/30) is stacked on top of this PR. Merge this one first; GitHub will auto-retarget #30 to `main` after.

### Version handling

- Replaces `types/protocol.Version.ConsumerTarget` (Cosmos `Params` residue from upstream Lava — hand-edited per release, with CI regex-parsing the Go source for the value) with a top-level `version/` package whose vars are injected at build time via `-X .../version.Version=$(git describe --tags --always --dirty)`. Release contract becomes "tag a commit, merge to main" — no source-file bump.
- Fixes a silent no-op: CI was passing `-ldflags "-X main.version=… -X main.commit=…"` but `package main` declared neither var, so the runtime version was *always* whatever was last committed to `params.go`, regardless of what CI injected.
- Wires the same `-X .../version.Version` injection into GoReleaser (`.goreleaser.yaml`) and the release `Dockerfile`, so the standalone binary and the Docker image both print the tag verbatim. Previously each was broken in its own way.
- Sources `GIT_VERSION` in `release.yml` from the same ref-resolution expression the existing checkout step already uses (`workflow_dispatch` input → `github.ref_name`) so the leading `v` survives. Previously, `steps.meta.outputs.version` stripped it — standalone binary printed `v6.2.2`, Docker image printed `6.2.2`.
- Normalizes `SetVersionInner` so `git describe` output (e.g. `v6.2.2-3-gabc1234-dirty`) still parses for the Prometheus gauge instead of zeroing it and logging an error on every non-clean-release startup. **Behavior pinned by a 13-case table-driven test** (`version_metric_test.go`).
- `smartrouter version` now prints the commit SHA alongside the version string (first line scriptable, second line operator-friendly).

### Wire-format change

**Breaking for any consumer that reads `Lavap-Version` from a relay reply:** the HTTP metadata header is renamed to `Smart-Router-Version`. Single-emit, no deprecation window — agreed pre-v1 while we still control all consumers. Per the smart-router semver convention added to the README in this PR, this would be a MAJOR bump in a customer-facing world; pre-v1 it's a free move.

### lavap / lavad cleanup

- Renames user-facing `lavap`/`lavad` strings (startup logs, gRPC banner) — mis-named since the fork.
- Deletes `protocol/statetracker/updaters/version_updater.go` and the `ProtocolVersionResponse` struct in `consumer_state_query.go`: dead Lava-residue with zero external callers, and the only remaining consumers of `protocoltypes.Version` after this change.
- Drops the dead `lavap` build entry from `.goreleaser.yaml` (`cmd/lavap` was retired in `b1bc1e3`; GoReleaser would have failed at the build step the first time the release workflow fired).
- Removes stale `cmd/lavap` references from the README.

### Release pipeline consistency

- `.goreleaser.yaml`: drops `CGO_ENABLED=1` and the o64-clang / oa64-clang / aarch64-linux-gnu-gcc cross-compiler chain. Smart-router has no CGO requirement (no first-party CGO code; Kafka client is `segmentio/kafka-go`, pure Go). `CGO_ENABLED=0 go build ./cmd/smartrouter` produces a clean, statically-linked binary on all four target os/arch combinations. With CGO off, the heavy `goreleaser-cross:v1.22` runner image is no longer required; `release.yml` switches to plain `goreleaser:v1.22`.
- Aligns release microarchitecture with dev: adds `GOAMD64=v3` (Haswell+: AVX2, BMI2, FMA) and `GOARM64=v8.2` (FP16, dotprod) to the global env. `smartrouter.yml` already builds dev images at these levels; release was falling back to baseline (`v1` / `v8.0`). Dev binaries had modern instruction sets while the binaries customers actually deployed were baseline.

### Documentation

- Adds a "Releases" section to the README covering the tag-driven workflow, what artifacts get published, two gotchas a first-time release-cutter hits (draft + pre-release flags, strict `v[0-9]+.[0-9]+.[0-9]+*` tag pattern), and **project-specific semver semantics** (what counts as MAJOR vs MINOR vs PATCH for smart-router).

## Heads-up for the first release

First post-merge image will carry a `git describe` tag like `v0.0.0-alpha-45-g<sha>` (functional, but not what you want long-term). Push a proper release tag (`git tag vX.Y.Z && git push --tags`) on the merge commit immediately after merging to produce a clean release image and to give downstream consumers a stable pin.

## Out of scope (covered by PR #30)

After this PR, two compile passes still produce customer-facing artifacts: GoReleaser compiles the standalone binaries; `release.yml`'s `release_docker` job compiles a different binary inside `Dockerfile` to build the Docker image. Flag-aligned but still distinct compilations. **[PR #30](https://github.com/Magma-Devs/smart-router/pull/30) consolidates** by moving the Docker image build into GoReleaser's `dockers:` + `docker_manifests:` config and deleting the source-build `Dockerfile`, so the standalone binary and the binary inside the Docker image are byte-identical.

## Test plan

- [x] `go test ./protocol/metrics/ -run TestSetVersionInner` — passes locally, 13 cases green.
- [ ] CI green on this branch.
- [ ] `Resolve version from git` job output matches local `git describe --tags --always --dirty` for HEAD.
- [ ] Pushed dev image tag matches the resolved version string.
- [ ] Pull image, run `smart-router version`, confirm output matches the image tag.
- [ ] Inspect a relay reply on a dev deployment: contains `Smart-Router-Version: <version>` header, does **not** contain `Lavap-Version`.
- [ ] `lava_rpcsmartrouter_protocol_version` Prometheus gauge reports a non-zero numeric value on a tag-clean build; zero (and a startup error log) only on parse failure as before.
- [ ] After landing, push a throwaway tag (e.g. `v0.0.1-test`) to validate the full release workflow before tagging the real v1.


Dockerfile
ssejzer: Without this, go defaults to GOAMD64=v1 and GOARM64=v8.0

```suggestion
ARG GIT_COMMIT="unknown"
ENV GOAMD64=v3 GOARM64=v8.2
```

ssejzer: not relevant, this will be fixed in PR30

ssejzer (approved):

PR #32, opened 2026-05-18
fix(relaycore): serialize Consistency writes against ResetState

avitenzer:
## Summary
- Replaces MAG-1893's prescribed 2-pass `Clear` (Option A — "or equivalent") with `sync.RWMutex` serialization on `ConsistencyImpl`. Writes take `RLock`; `ResetState` takes the exclusive `Lock` around `cache.Clear()`. This satisfies ristretto's documented "Set/Get won't be occurring during Clear" precondition rather than racing against it — strictly stronger than a 2-pass clear, and deterministic instead of empirical.
- New unit test `TestResetState_FlushesCorruptionUnderConcurrentWrites` mirrors MAG-1878's production-bug shape: a continuous corruption-write injector spams `SetSeenBlockFromKey` up to the moment `ResetState` fires. Existing `TestSetGet`, `TestBasic`, and `TestDebugMoveClock_ClearsCorruptedSeenBlock` all still pass under `-race`.
- Refs MAG-1878 — code-side leg only. MAG-1878 closes on deploy + 3× XPASS of PR #117 Scenario 2 against the cluster, then xfail-marker flip in `smart-router-automation`.

## Test plan
- [x] `go test ./protocol/relaycore/... -count=5 -race -timeout 120s` — clean (5/5 pass, 20.7 s)
- [x] `go test ./protocol/rpcsmartrouter/... -run "TestDebugMoveClock|TestDebugResetScores|TestDebugResetAll|TestDebugConsistencyReset" -count=2 -race` — clean, includes existing seen-block regression test
- [ ] Build smart-router binary from this commit and redeploy to the eth-sim-router deployment
- [ ] Run `test_debug_reset_scores_does_not_recover_corrupted_seenblock` 3× against the redeployed cluster — expect XPASS each time
- [ ] Flip `xfail(strict=False)` to expected-pass in `smart-router-automation` PR #117
- [ ] Close MAG-1878 once 3× XPASS is confirmed

### DIFF
--- a/protocol/relaycore/consistency.go
+++ b/protocol/relaycore/consistency.go
@@ -1,6 +1,7 @@
 package relaycore
 
 import (
+	"sync"
 	"time"
 
 	"github.com/dgraph-io/ristretto/v2"
@@ -22,10 +23,19 @@ type Consistency interface {
 	Key(userData common.UserData) string
 }
 
-// ConsistencyImpl is the default implementation of Consistency
+// ConsistencyImpl is the default implementation of Consistency.
+//
+// mu serializes write paths against ResetState. ristretto's Clear() docs warn
+// it "is not an atomic operation (but that shouldn't be a problem as it's
+// assumed that Set/Get calls won't be occurring until after this)" — so under
+// concurrent writes, a Set enqueued into setBuf during Clear can commit after
+// Clear returns and survive what was supposed to be a reset. Writes take a
+// read-lock; ResetState takes the exclusive write-lock, which makes ristretto's
+// precondition hold rather than racing against it.
 type ConsistencyImpl struct {
 	cache  *ristretto.Cache[string, any]
 	specId string
+	mu     sync.RWMutex
 }
 
 func (cc *ConsistencyImpl) SetLatestBlock(key string, block int64) {
@@ -54,10 +64,17 @@ func (cc *ConsistencyImpl) Key(userData common.UserData) string {
 }
 
 // used on subscription, where we already have the dapp key stored, but we don't keep the dappId and ip separately
+//
+// The Get-then-Set sequence is held under the read-lock so that ResetState's
+// exclusive lock blocks every in-flight write — including the SetLatestBlock
+// call below — until Clear has run, which is what makes ristretto's
+// "no concurrent Set/Get during Clear" precondition hold.
 func (cc *ConsistencyImpl) SetSeenBlockFromKey(blockSeen int64, key string) {
 	if cc == nil {
 		return
 	}
+	cc.mu.RLock()
+	defer cc.mu.RUnlock()
 	// seen block is only increasing
 	if block, found := cc.GetLatestBlock(key); found && block >= blockSeen {
 		return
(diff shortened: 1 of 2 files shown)


protocol/relaycore/consistency_test.go
```diff
@@ -58,3 +59,75 @@ func TestBasic(t *testing.T) {
...
+// returns, leaving the cache holding the corruption value despite the reset
+// completing successfully. ristretto's own Clear doc warns it "is not an
+// atomic operation (but that shouldn't be a problem as it's assumed that
+// Set/Get calls won't be occurring until after this)" — the lock makes that
+// precondition hold.
+//
+// Iteration amplifies the probability of catching the race across goroutine-
+// scheduling variance. With the fix the test is deterministic (0 leaks);
+// without it, some iterations leak corruption.
+func TestResetState_FlushesCorruptionUnderConcurrentWrites(t *testing.T) {
```
sotskov-do, line 85: Test isn't stable. Try to run 2-3 times:
```
go test ./protocol/relaycore -run TestResetState_FlushesCorruptionUnderConcurrentWrites -count=10
```

protocol/relaycore/consistency.go
```diff
@@ -54,10 +64,17 @@ func (cc *ConsistencyImpl) Key(userData common.UserData) string {
...
+//
+// The Get-then-Set sequence is held under the read-lock so that ResetState's
+// exclusive lock blocks every in-flight write — including the SetLatestBlock
+// call below — until Clear has run, which is what makes ristretto's
+// "no concurrent Set/Get during Clear" precondition hold.
 func (cc *ConsistencyImpl) SetSeenBlockFromKey(blockSeen int64, key string) {
 	if cc == nil {
 		return
 	}
+	cc.mu.RLock()
```
sotskov-do, line 108: There is a case where this function is waiting on `RLock()` while `ResetState()` holds `Lock()`. When `Lock()` is released, the cache has just been cleared, so `GetLatestBlock` returns `found=false`, monotonic guard is bypassed, and `SetLatestBlock` writes value into  just cleared cache.

protocol/relaycore/consistency_test.go
avitenzer: Reproduced with `GOMAXPROCS=2 go test ./protocol/relaycore -run TestResetState_FlushesCorruptionUnderConcurrentWrites -count=200` — 1/50 iterations leaked under the previous RWMutex-only fix on my Mac. Confirmed your read: this is the queued-writer race from comment 2 surfacing as test flakiness, not an unrelated test-design issue.

Closed in 5559cca with a gen counter + tombstone (details in the reply to comment 2). Same reproducer passes 200/200 now.

The test itself was also reshaped to use a fresh `ConsistencyImpl` per iteration — the new tombstone window means the previous iteration's `ResetState` would otherwise suppress the next iteration's pre-reset writes, making the test pass trivially.

protocol/relaycore/consistency.go
avitenzer: Confirmed. Closed in 5559cca with two layered mechanisms in `consistency.go`:

1. **Gen counter** (`atomic.Int64`, advanced in `ResetState` under the exclusive `Lock`). `SetSeenBlockFromKey` snapshots `gen` before `RLock` and drops the write if it advanced while we were waiting. This addresses exactly the scenario you described.

2. **Tombstone window** (`lastResetAtNano atomic.Int64`, stamped in `ResetState`; 100 ms `ResetTombstoneWindow` constant). A subtler variant of the same hole that the gen counter alone doesn't catch: a `Set` whose `gen.Load()` happens *after* `ResetState`'s `gen.Add` captures the post-reset gen, so the gen check passes — but the call's `blockSeen` was determined pre-reset by a relay response still in flight through the relay-processor pipeline. The tombstone drops any write whose `RLock` acquires within 100 ms of the most recent reset, absorbing that pipeline window.

100 ms is empirical — chosen to absorb a typical relay-pipeline's in-flight responses while being short enough that operators triggering `/debug/reset-*` for incident response won't notice the brief write-blocked period. Reads are untouched (`rpcsmartrouter_server.go:1149-1151` already treats `found=false` as "skip validation," so transient post-reset emptiness is benign).

The `GOMAXPROCS=2 -count=200` reproducer that previously flaked now passes 200/200.

sotskov-do (approved):

PR #34, opened 2026-05-18
fix(sync_binary): stage to .new then atomic mv to avoid ETXTBSY on running binary

VicSheCodes:
## Summary

`my_scripts/sync_binary.sh` previously did `scp ... :/root/smart-router-binary`, which fails with `ETXTBSY` whenever the running smart-router pods hold that binary's inode open for execution — Linux refuses to overwrite the file in place.

The fix stages the new binary at `/root/smart-router-binary.new` and then does an `ssh ... 'mv /root/smart-router-binary.new /root/smart-router-binary'`. `mv` on the same filesystem is an atomic rename: it swaps only the directory entry, leaving the running inode untouched, and the pods pick up the new file on the rolling restart that already follows.

Verified end-to-end on the deploy cluster: the redeployed binary contains commit `4147966`, all three `*-router` deployments restarted cleanly to `1/1 Running`, and a smoke `eth_blockNumber` returned a valid block (`0x1312D00`).

## Test plan

- [x] `bash -n my_scripts/sync_binary.sh` — syntax OK
- [x] `bash my_scripts/sync_binary.sh` — build + scp + mv + rolling restart all succeeded
- [x] `strings /root/smart-router-binary | grep 4147966` — commit hash present
- [x] `kubectl get pods -n lava-infra` — `base-router`, `eth-router`, `eth-sim-router` all `1/1 Running`
- [x] `curl -X POST https://eth-sim-jsonrpc.<cluster>/ -d '{"jsonrpc":"2.0","method":"eth_blockNumber","id":1}'` — HTTP 200, valid `result`

### DIFF
--- a/my_scripts/sync_binary.sh
+++ b/my_scripts/sync_binary.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+# Build smart-router binary on Mac, scp to server, rolling-restart all *-router deployments.
+# Run from anywhere: bash /Users/victoria/smart-router/my_scripts/sync_binary.sh
+
+set -e
+
+REMOTE="root@64.176.170.39"
+SSH_OPTS=(-o RemoteCommand=none -o RequestTTY=no)
+LOCAL_REPO="/Users/victoria/smart-router"
+
+cd "${LOCAL_REPO}"
+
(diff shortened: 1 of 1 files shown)


nimrod-teich (changes requested):
This PR shouldn't be here.

PR #38, opened 2026-05-18
fix(MAG-1866): JSON-RPC error response must be Object per spec, not stringified envelope

VicSheCodes:
## Why

When `lava-select-provider` pins to a provider that's not in the valid-providers list, the router currently returns `error` as a **string** (containing a serialized JSON envelope). JSON-RPC 2.0 §5.1 requires `error` to be an Object with `code` + `message` keys. Spec-compliant clients (Web3.py, ethers.js, viem) crash trying to read `response.error.code` because there's no `.code` attribute on a string.

## What changed

A new helper `convertToJsonRpcError` in `protocol/chainlib/common.go` builds a proper `{jsonrpc, id, error: {code, message, data}}` envelope:

- `code: -32000` (JSON-RPC 2.0 implementation-defined server-error range, §5.1)
- `message`: the inner human-readable error string (e.g. "Selected provider not available {selectedProvider:..., validProviders:..., GUID:...}")
- `data.guid`: the request GUID extracted from the existing stringified envelope
- `data.error`: the full inner error message (preserves provider context for debugging)
- `id`: extracted from the incoming request body so the client can correlate; falls back to JSON `null` for unparseable requests (spec-valid)

Only the JSON-RPC POST error path in `jsonRPC.go` switches to the new helper. The pre-existing `convertToJsonError` and its other callers (REST POST/GET, Tendermint GET) are untouched — those interfaces have different client expectations and changing the shared helper in-place would break them.

`GetUniqueGuidResponseForError` in `protocol/metrics/rpcconsumer_logs.go` is also untouched. The new helper parses its existing JSON-string output rather than forcing a signature change on a metrics-package helper used elsewhere.

## Test plan

- [x] `go test ./protocol/chainlib/... ./protocol/rpcsmartrouter/... ./protocol/metrics/... -count=1` — all pass.
- [x] `go vet ./...` — clean.
- [x] `make lint` — clean (lint target is `go vet`).
- [x] Two new unit tests added (`TestConvertToJsonRpcError`, `TestConvertToJsonRpcError_FallbackOnUnparseable`) covering happy path and parser-fallback.
- [ ] After merge: deploy via `sync_binary.sh` to victoria cluster, then run `pytest tests/product/router_behavior/test_unsupported_method_force_quicknode_retry.py::test_unsupported_method_force_quicknode_yields_retry` — `isinstance(response.error, dict)` should now pass.

## Backward compatibility

Any client expecting `error` to be a string was itself spec-violating; the fix forward is correct. No callers within smart-router treat `error` as a string. REST and Tendermint chains are unaffected — they keep `convertToJsonError`.

Closes MAG-1866.

### DIFF
--- a/protocol/chainlib/common.go
+++ b/protocol/chainlib/common.go
@@ -257,6 +257,11 @@ func addHeadersAndSendBytes(c *fiber.Ctx, metaData []pairingtypes.Metadata, data
 // convertToJsonError returns a JSON-encoded `{"error": errorMsg}` body as raw bytes.
 // Returning []byte (rather than string) lets every error-path caller hand the
 // result straight to addHeadersAndSendBytes / fiber.Ctx.Send with no conversion.
+//
+// NOTE: This helper produces a non-spec error envelope (`error` is a string).
+// It is retained for non-JSON-RPC chain interfaces (REST, Tendermint GET) whose
+// clients do not expect JSON-RPC 2.0 §5.1 shape. For JSON-RPC interfaces use
+// convertToJsonRpcError below.
 func convertToJsonError(errorMsg string) []byte {
 	jsonResponse, err := json.Marshal(fiber.Map{
 		"error": errorMsg,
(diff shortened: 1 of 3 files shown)


AnnaR-prog:
`convertToJsonRpcError` has a bug in masked-error mode. `GetUniqueGuidResponseForError` omits the `Error` field via `,omitempty` when `ReturnMaskedErrors == "true"`, so the input is just `{"Error_GUID":"..."}`. In that case the helper's `if parsed.Error != ""` branch is skipped and `message` retains its initial value — the raw JSON envelope string. The response is still spec-compliant (the field is a string), but `error.message` becomes `{"Error_GUID":"..."}` — which is the ugly-stringified-envelope the PR is meant to eliminate, just one layer in.

Suggested one-line fix, right after the existing `if parsed.Error != ""` block:

```go
} else if guid != "" {
    message = "Internal server error"
}
```

Default for `ReturnMaskedErrors` is `"false"` so the happy-path tests don't catch it, but production deploys with masking enabled would hit this branch.

Smaller follow-ups, not merge blockers:
- `data.error` is set to the same string as `error.message` — redundant payload. Either drop it when `data.guid` already carries context, or use `data` for something structured (parsed `selectedProvider`/`validProviders`).
- Single `-32000` for every router-synthesized error collapses distinguishable failures (provider exhausted, consistency rejection, retry-budget exhausted) onto one code. Worth a follow-up if clients want to branch on it.
- Batch requests: `json.Unmarshal(arrayBody, &probe)` type-mismatches → `id` falls back to `null`. Strictly the response should be an array of per-id responses; single `{id:null}` is degraded but tolerable. Worth a comment on the helper.

AnnaR-prog:
Pushed a fix in 93e96f2 for the masked-mode issue from [my prior comment](https://github.com/Magma-Devs/smart-router/pull/38#issuecomment-4486355129).

Adds the `else if guid != ""` branch so when `Error` is elided by `,omitempty` under `ReturnMaskedErrors == "true"`, `error.message` becomes `"Internal server error"` instead of leaking the raw `{"Error_GUID":"..."}` envelope. `data.guid` still carries the correlation ID for debugging.

Also added regression test `TestConvertToJsonRpcError_MaskedModeOmitsErrorField` that asserts `error.message` contains neither `"Error_GUID"` nor `"{"` under the masked-mode input shape.

Verified locally with `go test -race -count=3 -run TestConvertToJsonRpcError ./protocol/chainlib/...` and the full `protocol/chainlib/...` package — both green.

The smaller follow-ups (redundant `data.error`, single `-32000` for distinct errors, batch `id=null`) are left as-is; happy to handle in a separate PR if you'd like.

AnnaR-prog (approved):

PR #39, opened 2026-05-18
fix: [MAG-1891] silence cobra --help dump on fatal startup error (noisy logs)

sotskov-do:
## Summary

On fatal startup error (e.g. "all static providers failed verification"), cobra was dumping the full `--help` text after the error line, swamping `kubectl logs` in a CrashLoopBackOff and burying the real failure.

Set `SilenceUsage: true` on the root cobra command at the construction site (`CreateRPCSmartRouterCobraCommand`). `SilenceErrors` is left at default so cobra still prints the authoritative `Error: <msg>` line.

**Scope:** this only affects startup paths that `return` an error from `RunE` (e.g. invalid endpoints, missing `direct-rpc` config, the "all static providers failed verification" trigger). Other paths in the same `RunE` use `utils.LavaFormatFatal` (config load failure, missing `--geolocation` value, log-level / test-mode / pprof / pyroscope flag parse) and `os.Exit(1)` before cobra's error handler runs — those were never dumping the usage block anyway, and they still emit a structured log without the cobra `Error:` line. Converting those to `return utils.LavaFormatError(...)` for parity is a separate change.

## Changes

- `protocol/rpcsmartrouter/rpcsmartrouter.go` — `SilenceUsage: true` on the root command (one line + comment explaining the operator-visibility motivation).
- `protocol/rpcsmartrouter/rpcsmartrouter_test.go` — regression tests:
  - startup error → no `Usage:` / `Available Commands:` / `Examples:` / `Flags:` in output;
  - `--help` still renders the full usage text.

## Test plan

- [ ] `go test ./protocol/rpcsmartrouter/ -run TestRPCSmartRouterCobraCommand -v` — both new tests pass.
- [ ] `go build ./cmd/smartrouter/` — binary still builds.
- [ ] Manual: invoke `smartrouter` with no `--geolocation` (or a deliberately broken config) and confirm only the structured error log + `Error: <msg>` appear — no `Usage:` block.
- [ ] Manual: `smartrouter --help` still prints the full flag catalogue.

Origin: Sebastian's MAG-1879 sim-crashloop investigation on 2026-05-17.

### DIFF
--- a/protocol/rpcsmartrouter/rpcsmartrouter.go
+++ b/protocol/rpcsmartrouter/rpcsmartrouter.go
@@ -1512,6 +1512,11 @@ func CreateRPCSmartRouterCobraCommand() *cobra.Command {
 	cmdRPCSmartRouter := &cobra.Command{
 		Use:   "rpcsmartrouter [config-file] | { {listen-ip:listen-port spec-chain-id api-interface} ... }",
 		Short: `rpcsmartrouter sets up a centralized server with static providers to perform api requests`,
+		// On startup failure (e.g. all static providers failed verification)
+		// cobra otherwise dumps the full --help text after the error line,
+		// swamping kubectl logs in a CrashLoopBackOff. Operators need to see
+		// the error, not the flag catalogue.
+		SilenceUsage: true,
 		Long: `rpcsmartrouter sets up a centralized server with static and backup providers to perform api requests through the lava protocol.
 		This is the smart router mode that uses pre-configured static providers instead of dynamically discovering providers on-chain.
 		all configs should be located in the local running directory /config or ` + lavaDefaultNodeHome + `
--- a/protocol/rpcsmartrouter/rpcsmartrouter_test.go
+++ b/protocol/rpcsmartrouter/rpcsmartrouter_test.go
@@ -1,7 +1,9 @@
 package rpcsmartrouter
 
 import (
+	"bytes"
 	"context"
+	"errors"
 	"sync"
 	"testing"
 	"time"
@@ -12,6 +14,7 @@ import (
 	"github.com/magma-Devs/smart-router/protocol/provideroptimizer"
 	"github.com/magma-Devs/smart-router/utils/rand"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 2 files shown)


AnnaR-prog:
Worth a one-line scope clarification in the PR description: `SilenceUsage` only takes effect when cobra's error handler runs — i.e. when `RunE` **returns** an error.

Several startup failures in the same `RunE` use `utils.LavaFormatFatal`, which `os.Exit(1)`s before cobra ever sees the error:
- `rpcsmartrouter.go:1592` — config load failure
- `rpcsmartrouter.go:1598` — missing `--geolocation`
- `rpcsmartrouter.go:1610` — log-level parse error
- `rpcsmartrouter.go:1616` — test-mode flag parse
- `rpcsmartrouter.go:1625` — pprof address flag parse

Those paths weren't dumping the usage block anyway (because cobra's handler never runs), so they're already "silent" in the help-text sense — but they also don't get the `Error:` line, just the structured log. No regression, just a coverage gap the description doesn't flag.

The cited trigger — "all static providers failed verification" at `rpcsmartrouter.go:1073` — does use `LavaFormatError` and **does** reach cobra, so the fix lands on that case as intended.

If you want the `LavaFormatFatal` paths to also benefit from the cleaner output, that's a separate change (convert them to `return utils.LavaFormatError(...)`). Out of scope here; just worth noting that "fatal startup error" in the PR title/body currently reads broader than the actual coverage.

Bonus observation, not actionable: cobra propagates `SilenceUsage` from root to subcommands via `c.Root().SilenceUsage`, so `cache`, `test/*`, and `version` subcommands also get the cleaner error output for free.

AnnaR-prog:
Concrete wording suggestion for the description tweak from my prior comment, if helpful:

---

**Title:** `fix: [MAG-1891] silence cobra --help dump on RunE startup errors (noisy logs)`

**Summary (add one sentence to the existing first paragraph):**

> On fatal startup error (e.g. "all static providers failed verification"), cobra was dumping the full `--help` text after the error line, swamping `kubectl logs` in a `CrashLoopBackOff` and burying the real failure.
>
> Set `SilenceUsage: true` on the root cobra command at the construction site (`CreateRPCSmartRouterCobraCommand`). `SilenceErrors` is left at default so cobra still prints the authoritative `Error: <msg>` line.
>
> **Scope:** this only affects startup paths that `return` an error from `RunE` (e.g. invalid endpoints, missing `direct-rpc` config, the "all static providers failed verification" trigger). Other paths in the same `RunE` use `utils.LavaFormatFatal` (config load failure, missing `--geolocation` value, log-level / test-mode / pprof / pyroscope flag parse) and `os.Exit(1)` before cobra's error handler runs — those were never dumping the usage block anyway, and they still emit a structured log without the cobra `Error:` line. Converting those to `return utils.LavaFormatError(...)` for parity is a separate change.

That's it from my side — the code change is good as-is.

[MAG-1891]: https://magmadevs.atlassian.net/browse/MAG-1891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

AnnaR-prog:
PR description updated with the **Scope** paragraph clarifying that the fix only affects `RunE` paths that `return` an error — `LavaFormatFatal` paths weren't dumping usage anyway (they `os.Exit(1)` before cobra's error handler), so the silencing has no effect on them. Title/body now matches actual coverage. Converting the `LavaFormatFatal` paths for parity is called out as a separate change.

AnnaR-prog (approved):

PR #43, opened 2026-05-19
fix(rpcsmartrouter): echo WS subscribe id + fix unsubscribe race (MAG-1824)

avitenzer:
## Summary

Two interlocking bugs in the direct WebSocket subscription path, per [MAG-1824](https://magmadevs.atlassian.net/browse/MAG-1824):

- **JSON-RPC id clobbering on `eth_subscribe`.** `createSubscriptionReply` was writing the upstream rpcclient's internal `nextID()` counter (typically `1`) to the response sent back to the client, so string ids like `"client-uuid-1"` were replaced with `1` — violating JSON-RPC 2.0 §4.2. Fixed by threading the caller's `requestID` through to the reply, and by routing the `subscription-not-found` error response through a new `common.MarshalJsonRPCErrorWithRequestID` helper that echoes string / numeric / null ids verbatim.
- **Subscription "forgotten" after a brief upstream blip.** When `upstreamSub.Err()` fired, `listenForUpstreamMessages` handed off to `handleUpstreamDisconnect` *and* ran its own deferred cleanup. If reconnection succeeded, the new listener picked up notifications (it holds `activeSub` directly) but `dwsm.activeSubscriptions[hp]` had been deleted by the old goroutine's defer — so `eth_unsubscribe` lookup failed with "subscription not found" while notifications kept flowing, matching the bug's reported symptom exactly. Skip the cleanup when reconnect is in flight. Also added a defensive upstream-id fallback in the EVM unsubscribe lookup (ownership still enforced by `clientRouterIDs[clientKey]`).

## Files

- `protocol/rpcsmartrouter/direct_ws_subscription_manager.go` — id-echo, reconnect race, upstream-id fallback. `listenForUpstreamMessages` now takes a consumer-defined `upstreamErrSource` interface (one method, `Err()`) so the race fix is testable without exposing test scaffolding from the rpcclient package.
- `protocol/chainlib/consumer_websocket_manager.go` — unsubscribe-failure response routes through the new id-echoing helper.
- `protocol/common/return_errors.go` — `MarshalJsonRPCErrorWithRequestID` helper.
- Tests: `return_errors_test.go` (new) + 5 new regression tests in `direct_ws_subscription_manager_test.go`.

## Test plan

- [x] `make test-short` — `ok protocol/rpcsmartrouter 78s`
- [x] `make lint` — `go vet ./...` clean
- [x] `go test ./protocol/common ./protocol/chainlib ./protocol/rpcsmartrouter` — all three packages `ok`
- [x] Race-fix test verified to FAIL when the `reconnectInFlight` guard is reverted (then restored)
- [x] End-to-end test `TestEndToEnd_SubscribeUnsubscribeRoundTrip` drives subscribe→notify→unsubscribe against an in-process mock upstream, asserting all three ticket acceptance criteria:
  - Subscribe response id echoes `"client-uuid-1"` verbatim
  - Unsubscribe returns `{"result":true,"id":"client-uuid-2"}`
  - Notifications arrive
  - (Bonus) mock upstream receives the upstream-id on outbound `eth_unsubscribe`, not the router id

Expected outcome on the release-smoke side: `tests/release_smoke/test_ws_subscribe_smoke.py::TestWSSubscribeSmoke::test_unsubscribe_returns_true` flips XFAIL→PASS if the production symptom matches the reconnect race or the upstream-id-echo-back pattern. If it doesn't, capture upstream logs from a reproducing run so we can chase the residual cause.

[MAG-1824]: https://magmadevs.atlassian.net/browse/MAG-1824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/protocol/chainlib/consumer_websocket_manager.go
+++ b/protocol/chainlib/consumer_websocket_manager.go
@@ -289,7 +289,9 @@ func (cwm *ConsumerWebsocketManager) ListenToMessages(ctx context.Context) {
 				if err != nil {
 					utils.LavaFormatWarning("error unsubscribing from subscription", err, utils.LogAttr("GUID", webSocketCtx))
 					if err == common.SubscriptionNotFoundError {
-						msgData, err := json.Marshal(common.JsonRpcSubscriptionNotFoundError)
+						// Echo the caller's id (JSON-RPC 2.0 §4.2) — the template's hardcoded
+						// id is just a fallback for malformed requests with no id field.
+						msgData, err := common.MarshalJsonRPCErrorWithRequestID(common.JsonRpcSubscriptionNotFoundError, msg)
 						if err != nil {
 							continue
 						}
--- a/protocol/common/return_errors.go
+++ b/protocol/common/return_errors.go
@@ -1,9 +1,11 @@
 package common
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/gofiber/fiber/v2"
+	"github.com/tidwall/gjson"
 )
 
 // #######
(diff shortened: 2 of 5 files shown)


AnnaR-prog:
Worth verifying before merge.

**1. Reconnect-failure path must still clean up.**

With `reconnectInFlight=true`, the listener's deferred `cleanupSubscription` is skipped. The fix relies on `handleUpstreamDisconnect` calling cleanup itself when reconnect *fails*. The success path is well-described in the doc comment, but the failure path isn't called out — if `handleUpstreamDisconnect` ever takes an error branch that exits without calling `cleanupSubscription(hashedParams)`, the subscription leaks forever. Worth a code read on `handleUpstreamDisconnect` to confirm every exit either restores the subscription OR explicitly cleans it up. If that's already true, a sentence in the new doc-comment ("cleanup ownership has been transferred to handleUpstreamDisconnect") would seal it.

**2. Tendermint reply path is now lossy.**

Pre-PR, `createSubscriptionReply` for `tendermintrpc` returned `json.Marshal(originalMsg)` — the upstream message verbatim. Post-PR, it builds a fresh envelope with only `{jsonrpc, id, result}`. If the upstream `JsonrpcMessage` ever has additional fields a client cares about (custom extensions, error metadata, anything beyond `result`), they're dropped. The existing test only verifies the `query` field round-trips inside `result`. Worth one of:
- Confirming via Tendermint client SDK docs that nothing else is expected, or
- Copying `originalMsg.Error` too if it can be non-nil on a subscription reply, or
- Doing a `map[string]json.RawMessage` round-trip and overwriting just `id` (preserves whatever else was there).

AnnaR-prog:
Pushed 779468a addressing both points from [the prior comment](https://github.com/Magma-Devs/smart-router/pull/43#issuecomment-4486944425).

**1. Cleanup ownership transfer.** Read `handleUpstreamDisconnect` — confirmed it calls `cleanupSubscription(hashedParams)` on every failure exit: reconnect failure at line 1261, GetConnection-after-reconnect failure at 1276, re-subscribe failure at 1286. So the `reconnectInFlight` guard does not leak. Added a paragraph to the `listenForUpstreamMessages` doc-comment making the ownership transfer explicit.

**2. Tendermint reply now preserves upstream fields.** Switched to marshal-then-round-trip-through-`map[string]json.RawMessage`, overwriting only `id`. This preserves whatever else the upstream returned (`Error` is the most realistic extra field — a non-nil error envelope on a failed subscribe would have been silently dropped by the hand-built `{jsonrpc, id, result}` envelope). Regression test `TestCreateSubscriptionReply_Tendermint_PreservesUpstreamFields` constructs an upstream message with a non-nil `Error` and asserts the error envelope still appears in the response alongside the echoed id.

Verified locally: `go test -race -count=3` on `TestCreateSubscriptionReply*` green; full `protocol/rpcsmartrouter ./protocol/chainlib ./protocol/common` packages green; `go vet` clean.

AnnaR-prog (approved):

PR #63, opened 2026-06-04
refactor(metrics)!: remove dead metric families + CLI flags (BREAKING)

nimrod-teich:
## Summary

Metrics-scoped cleanup. Only `SmartRouterMetricsManager` is instantiated at runtime — the consumer, provider, and health metric families were dead code (verified with `golang.org/x/tools` `deadcode`). This removes them and the two CLI flags that only gated dead behaviour. Net **−2,400 lines**.

Commits (each independently reviewable):
1. `refactor(metrics): remove dead consumer & provider metrics managers` — delete `lava_consumer_*` / `lava_provider_*` managers + the always-nil `pmetrics` chaintracker plumbing; relocate the load-bearing survivors (flag consts → `flags.go`, `SetVersionInner` → `version_metric.go`, `ProviderSelectionScores` → `_inf.go`).
2. `refactor(metrics): retire dead --show-provider-address-in-metrics flag` — its only effect was on the now-deleted consumer metrics.
3. `refactor(metrics): always expose optimizer scores, drop --optimizer-qos-listen` — `lava_rpc_optimizer_selection_score` is now collected unconditionally; the flag (and the deleted `/provider_optimizer_metrics` endpoint it advertised) is removed.
4. `refactor(metrics): remove dead lava_health_* metrics server` — `NewHealthMetrics` had no call site and stood up its own orphaned HTTP server.
5. `docs(metrics): add metrics reference, link from README` — new `docs/METRICS.md` + README link.

## ⚠️ BREAKING CHANGES

Two CLI flags are **removed** from the `smartrouter` binary — passing them now errors with "unknown flag":

- `--optimizer-qos-listen` — optimizer selection-score metrics are now always exposed on `/metrics`, so the flag is redundant. (Remote push is still controlled by `--optimizer-qos-server-address`.)
- `--show-provider-address-in-metrics` — only gated the removed `lava_consumer_*` metrics.

Also removed: the `GET /provider_optimizer_metrics` JSON endpoint (its handler lived on the deleted consumer manager; the same data is the `lava_rpc_optimizer_selection_score` gauge on `/metrics`).

**Dashboards/alerts** referencing `lava_consumer_*`, `lava_provider_*`, or `lava_health_*` / `lava_*_entities` will go empty — but those series were never emitted at runtime, so no live data is lost.

**Deployment check:** grepped `ansible-internal` / `devops` / `lava-terraform-infrastructure` — no deployment invokes the `smartrouter` binary with the removed flags. (The `--optimizer-qos-listen` in `ansible-internal`'s `iprpc_service.service.j2` belongs to the separate `lavap rpcconsumer` binary, which is unaffected.)

## Test plan

- `go build ./...` ✅
- `go vet ./protocol/{metrics,chaintracker,rpcsmartrouter,common}/` ✅
- `go test` for `protocol/metrics`, `protocol/chaintracker`, `protocol/rpcsmartrouter`, `protocol/common` ✅ (SmartRouter test coverage preserved when stripping the consumer test cases)
- `deadcode -test ./...` reports zero dead non-test funcs remaining in `protocol/metrics` ✅


docs/METRICS.md
```diff
@@ -0,0 +1,206 @@
...
+meaning. All metrics are defined under [`protocol/metrics/`](../protocol/metrics/).
+
+## Exposition
+
+Metrics are served in Prometheus text format from an HTTP server started by the
+metrics manager.
+
+| Path | Format | Description |
+| --- | --- | --- |
+| `/metrics` | Prometheus | All registered metrics ([`promhttp.Handler()`](../protocol/metrics/smartrouter_metrics_manager.go#L615)) |
```
avitenzer: nit: stale source anchor. `promhttp.Handler()` is now at `smartrouter_metrics_manager.go#L623`, not `#L615`. Line-number anchors rot quickly across refactors — consider linking to the function/symbol instead of `#Lxxx`.

docs/METRICS.md
```diff
@@ -0,0 +1,206 @@
...
+| `--optimizer-qos-server-address` | `""` | Push optimizer QoS reports to this address. Independent of `/metrics` — the `lava_rpc_optimizer_selection_score` gauge is collected and exposed regardless (see below). |
+
+```bash
+# enable, then scrape
+smartrouter ... --metrics-listen-address ":7779"
+curl http://localhost:7779/metrics
+```
+
+The flag is defined at
+[`rpcsmartrouter.go:1975`](../protocol/rpcsmartrouter/rpcsmartrouter.go#L1975); the flag
```
avitenzer: nit: off-by-one anchor — the metrics-listen flag is registered at `rpcsmartrouter.go#L1974` (both the prose `rpcsmartrouter.go:1975` and the markdown `#L1975` link point one line past it).

protocol/metrics/version_metric.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/magma-Devs/smart-router/utils"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// SetVersionInner encodes a semantic version string onto a protocol-version gauge as
+// major*1e6 + minor*1e3 + patch, so dashboards can compare deployed versions numerically.
+// Shared by every metrics manager that exposes a protocol_version gauge.
```
avitenzer: nit: this comment ("Shared by every metrics manager that exposes a protocol_version gauge") is now stale — after this PR `SmartRouterMetricsManager` is the only metrics manager that remains.

protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -233,18 +232,18 @@ func (rpsr *RPCSmartRouter) Start(ctx context.Context, options *rpcSmartRouterSt
...
-	if options.analyticsServerAddresses.OptimizerQoSAddress != "" || options.analyticsServerAddresses.OptimizerQoSListen {
-		smartRouterOptimizerQoSClient = metrics.NewConsumerOptimizerQoSClient(smartRouterIdentifier, options.analyticsServerAddresses.OptimizerQoSAddress, options.geoLocation, metrics.OptimizerQosServerPushInterval) // start up optimizer qos client
-		smartRouterOptimizerQoSClient.StartOptimizersQoSReportsCollecting(ctx, metrics.OptimizerQosServerSamplingInterval)
-	}
+	// Always collect optimizer QoS reports so the lava_rpc_optimizer_selection_score
+	// metric is exposed on /metrics. When an OptimizerQoSAddress is configured the
+	// reports are also pushed there; with no address the client's QueueSender is nil
+	// (a no-op), so collection still runs but nothing is pushed.
+	smartRouterOptimizerQoSClient := metrics.NewConsumerOptimizerQoSClient(smartRouterIdentifier, options.analyticsServerAddresses.OptimizerQoSAddress, options.geoLocation, metrics.OptimizerQosServerPushInterval)
+	smartRouterOptimizerQoSClient.StartOptimizersQoSReportsCollecting(ctx, metrics.OptimizerQosServerSamplingInterval)
```
avitenzer, line 240: Informational, not a change request — flagging so it's a conscious decision. This sampling goroutine now runs unconditionally (every `OptimizerQosServerSamplingInterval`, default 1s). I traced it: it's the **producer** that fills `reportsToSend`, which `StartSelectionStatsUpdater` reads to populate the `lava_rpc_optimizer_selection_score` gauge on `/metrics`. That gauge updater has no address gate, so making this producer unconditional is exactly what makes "always expose optimizer scores" work — removing it would re-break the gauge. ✅

The only edge case is when metrics are fully off (`--metrics-listen-address=disabled` AND no `--optimizer-qos-server-address`): the producer still ticks but nothing reads the result (`QueueSender` nil, gauge updater no-op). It's a read-only RLock pass and honors `ctx.Done()`, so it's negligible wasted work, not a leak. Optional micro-opt: gate on `MetricsListenAddress != "disabled" || OptimizerQoSAddress != ""`. Fine to leave as-is given you deliberately simplified away the conditional.

avitenzer (commented):
Reviewed the full diff across four dimensions (dangling refs, changed wiring, test coverage, docs/flags). Build + vet + `go test` for the touched packages all pass; no dangling references to any removed symbol; SmartRouter test coverage is preserved. Nicely de-risked deletion — extracting `ConsumerMetricsManagerInf` before deleting the 962-line concrete type, plus `deadcode` as objective evidence, is the right pattern.

No blockers. A few inline nits below, all non-blocking. The one on the optimizer QoS goroutine is informational — I traced it and it's intentional/required, not a bug.

avitenzer (approved):

PR #65, opened 2026-06-04
chore: remove verified dead code (relaypolicy, chaintracker, types/relay, common)

nimrod-teich:
## Summary

Removes verified-dead, zero-reference code flagged by `deadcode` (sweep run with `GOTOOLCHAIN=go1.26.2`, no `-test`, so test-only-reachable code also surfaces). Every candidate was confirmed unreferenced with **package-qualified** `git grep` including `_test.go` — this matters because several dead symbols share a name with a *live* function in another package.

Two independently-reviewable commits:

### 1. `refactor: remove standalone dead code`
Self-contained whole-file / whole-feature removals outside the shared util packages:
- **`protocol/relaypolicy/classify.go`** — `ClassifyNodeError` was never called. ⚠️ Note: the live `chainlib.ClassifyNodeError` is a **different** function; a naive grep conflates them. Also drops the `ErrorClassification` struct in `types.go` that this file was the sole user of.
- **`cmd/common/warning_log_command_wrapper.go`** — `CreateWarningLogCommandWrapper`, no cobra wiring.
- **`protocol/chaintracker/utils.go`** — `BuildProofFromBlocks` only (kept `exponentialBackoff` + `FindRequestedBlockHash`, both still referenced).
- **`types/relay/relay.go`** — `RelayExchange` + constructor + `Signable` methods; struct had zero external refs and is not generated.

### 2. `refactor(common): remove zero-reference dead helpers`
Clean removals in `protocol/common` (each with its companion/live sibling kept):
- `ClassifyMessage`, `LogCodedWarning`, `IsSquareBracketed`/`UnSquareBracket`, `ContextOutOfTime`/`IsTimeout` (+ the imports they solely used).

## Deliberately left out (scope discipline)
`protocol/common` is upstream-Lava parity surface, so I avoided cascading deletions there:
- `errors.go` `IsUnsupportedMethod*` / `IsSolanaNonRetryable*` — a documented parity API family whose removal would also orphan two pattern-var blocks.
- `ApiInterfaceToTransport` + `ClassifyNodeErrorForRetry` — newly orphaned by removing `relaypolicy/classify.go`; removing them cleanly also means unwinding `NodeErrorClassification` + its tests. Flagged as a follow-up.

## Relationship to other PRs
Independent of #63 (metrics) and #64 (statetracker) — no file overlap. Closes out the "Bucket A.3a" findings from the dead-code sweep.

## Test plan
- `go build ./...` ✅
- `go vet ./...` ✅
- `go test` for `common`, `relaypolicy`, `chaintracker`, `types/relay`, `chainlib`, `rpcsmartrouter` ✅
- `deadcode ./...` — all removed symbols gone; no unexpected new orphans beyond the two documented above ✅


avitenzer (approved):

nimrod-teich:

```bash
git fetch origin
```

PR #82, opened 2026-06-09
feat!: OTel usage pipeline + project-id rename + reporter-flow removal

nimrod-teich:
> ⚠️ **BREAKING CHANGE — removed CLI flags.** This removes the flags for the deleted Kafka / reports / relay-server push pipeline. Passing any of them now fails at startup with `unknown flag`, so launch scripts must be updated before upgrading. Full mapping in [`protocol/rpcsmartrouter/README.md`](protocol/rpcsmartrouter/README.md) → "Migrating from the old push / Kafka pipeline".
>
> | Removed | Replacement |
> | --- | --- |
> | `--relay-server-address`, `--relay-kafka-*` (7 flags), `--reports-be-address`, `--optimizer-qos-server-address` | `--usage-otel-*` |
> | `--optimizer-qos-push-interval` | `--optimizer-qos-sampling-interval` |

Ports **lavanet/lava#2285** and **lavanet/lava#2286** into smart-router as a single PR (all three scopes). smart-router has no `rpcconsumer/` or `integration/` dirs, so those upstream changes have no target and are skipped.

### 1. OTel usage pipeline (`relay_usage` + `optimizer_qos`)
Replaces the Kafka producer **and** the relay-server push client with a single OTLP/HTTP log sink emitting two event types. **Off by default** (`--usage-otel-enabled=false`) — when off, the relay and QoS paths each pay one inlinable `NoopUsageSink` no-op; no SDK setup, no goroutines.

- New `protocol/metrics/usage_sink.go` (`UsageEventSink` interface + `NoopUsageSink`), `otel_usage_sink.go` (`OTelUsageSink` + `BatchProcessor`), `hostname.go`.
- `RelayUsageEvent` / `NewRelayUsageEvent` added to `analytics.go`.
- `RPCConsumerLogs` and `ConsumerOptimizerQoSClient` now take a `UsageEventSink`; per-relay + per-subscription-delivery emit points wired into the websocket manager.
- **smart-router divergence preserved:** the QoS client keeps its in-memory `reportsToSend` cache so the Prometheus `/metrics` selection-score endpoint stays populated, and is **always constructed** (not gated on a flag). Each sampling tick now *also* fires the report at the sink.
- 9 new `--usage-otel-*` flags; `usageSink.Close()` deferred at construction so the BatchProcessor drains on shutdown.

### 2. `dapp-id` → `project-id` wire rename
- New `chainlib.ProjectIDHeader = "project-id"` (HTTP header, gRPC metadata key, fiber Locals key).
- fasthttp buffer-aliasing fix: `strings.Clone` at both extraction sites + the Origin Locals stash (the value is serialized asynchronously by the OTel BatchProcessor after fasthttp recycles the request buffer).

### 3. `Reporter` / `ReportsAddressFlag` removal
- Deleted the reports client, kafka client, and relay-server push client (+ tests).
- Shrunk signatures: `blockProvider`, `ReportProvider`, `reportProviderLocked`, `NewConsumerSessionManager`, `NewReportedProviders` (all drop the reporter / `errorsForReport` plumbing).

## Collateral
- OTel modules **1.21 → 1.28** (+ logs API `otlploghttp` / `log` / `sdk/log` v0.4.0), grpc **→ 1.65**, `autoexport` **→ 0.53** (required to keep smart-router's own `protocol/tracing` compatible). `segmentio/kafka-go` dropped.
- `upstream_grpc_pool.go` migrated off the now-deprecated `grpc.DialContext` + `WithBlock` to `grpc.NewClient` + a `Ready`-only state-walk loop that preserves the original fast-fail-on-unreachable semantics. Paired `.golangci.yml` SA1019 exclusion for the remaining (out-of-scope) `grpc.Dial` sites, matching upstream.

## Decisions (vs. literal upstream)
- **Dropped** the relay-server push client for full upstream parity — relay usage flows only through the OTel sink.
- **Surgical** QoS-client port rather than wholesale replacement: smart-router's `SmartRouterMetricsManager` depends on the in-memory selection-stats cache for `/metrics`, so that path is kept intact and sink emission is added alongside it.
- `CalculateQoSScoresForMetrics` keeps smart-router's 4-arg signature (upstream simplified to 2-arg); self-consistent across the `OptimizerInf` interface and optimizer.

## Test plan
- [x] `go build ./...` clean
- [x] `go test ./...` — full suite passes (metrics / lavasession / chainlib / rpcsmartrouter incl. new sink tests)
- [x] `go vet ./protocol/...` clean
- [x] All changed files gofmt-clean
- [x] `golangci-lint` (go1.26-compatible v2) — no findings in changed files beyond pre-existing repo debt; lint is advisory on this repo per `.github/workflows/lint.yml`


avitenzer:
**Per-delivery metric goroutine — please confirm intended** (`consumer_websocket_manager.go`, subscription delivery loop in `ListenToMessages`)

The subscription delivery loop fires `go logger.AddMetricForWebSocket(&perMessage, nil, websocketConn)` **once per pushed notification**, plus a Prometheus-manager mutex acquisition per message on the delivery hot path. For high-frequency subscriptions (`newHeads`, `logs`) that's unbounded goroutine churn under load.

It's **race-safe** — the `subscriptionFields := *metricsData` snapshot is taken before both goroutines spawn, and each per-message emit works on its own value copy — so this is a throughput/overhead question, not a correctness one. Two things to confirm:

1. Is the unbounded per-message goroutine fan-out acceptable, or should it be a bounded worker / batched emit before enabling at scale?
2. Is the flat `DefaultSubscriptionDeliveryCU = 10` per delivered notification the intended billing model?

nimrod-teich:
1. Acceptable.
2. Intended.

avitenzer (approved):

nimrod-teich:

```bash
git fetch origin
```

PR #96, opened 2026-06-11
feat(chaintracker): polling-relief flags to slow polling + widen consistency gate

AnnaR-prog:
## What & why

Proactive chain-tracker polling is billed against the customer's own upstream node
quota, and the cadence (`averageBlockTime / 16`, no floor) is pathological on fast
chains — Solana's 400ms blocks drive polls as fast as ~25ms (~40/s per endpoint).
Add two **opt-in, zero-default** process-wide relief flags, modeled on
**lavanet/lava [#2276](https://github.com/lavanet/lava/pull/2276)** (adapted to the
smart-router's architecture).

| Flag | Effect | Default | Range |
|---|---|---|---|
| `--chain-tracker-polling-multiplier` | overrides the polling divisor; **smaller = slower polling = fewer upstream calls** | 16 | [4,16] |
| `--consistency-block-gap-factor` | widens the endpoint-lag consistency gate (`blockLagForQosSync × factor`) | 2 | [2,8] |

These are a **coupled pair**: slower polling makes each endpoint's cached "latest
block" staler → the consistency gate rejects more endpoints ("no pairings") unless
the gate is widened in step.

## Implementation

- **Multiplier** — a process-wide override (`chaintracker.PollingTimeMultiplierOverride`)
  honored in `newCustomChainTracker`, so it applies to the global tracker *and* every
  per-endpoint tracker with no per-call threading. The resolved value is clamped to a
  floor of **4** (the adaptive tiers compute `base/(m/4)`; `m<4` divides by zero) with
  a warn.
- **Gap-factor** — a `blockGapFactor` parameter on `NewConsistencyValidationConfig`
  (`× 2` → `× factor`). Note `EndpointLagThreshold = max(blockLagForQosSync × factor,
  finalizationDistance, 10)`.
- **Validation** — out-of-range values **warn-and-revert** to the default. Zero = no
  relief. Overrides are set in `RunE` before any tracker/config is built.
- **Logging** — startup log reports the *effective* interval; a "polling-relief active"
  line logs when either override is set.
- **`--consistency-probability-gate` is intentionally NOT ported** — the smart-router
  consistency gate is deterministic (`lag > threshold`) with no Poisson term; that
  model lives in the optimizer, not the consistency path.
- **Dead-code reconcile** — removed the never-wired `PollingMultiplier` /
  `PollingMultiplierFlagName` / `LavaPollingMultiplierFrequency`.

## Coupling math — why the gap-factor need not change at multiplier 4 (but ships anyway)

Slower polling makes an endpoint's cached latest block staler, which feeds the gate's
`lag = seenBlock − endpointLatestBlock`, so in principle the gate should widen in
step. But the *magnitude* of 16→4 is far below the gate's budget:

- **Added staleness is < 1 block, and chain-independent.** Poll interval ≈ `base/m`, so
  the extra staleness from 16→4 is `base/4 − base/16 = (3/16)·base ≈ 0.19 block` — the
  `averageBlockTime` cancels, so it's ~0.19 block on *any* chain (≤ ~1 block even in the
  slowest adaptive tier).
- **The budget is ≥ 10 blocks.** `EndpointLagThreshold = max(blockLagForQosSync ×
  factor, finalizationDistance, 10)` — never below the floor of 10, and ~50 slots on
  Solana (`blockLagForQosSync = 25`, factor 2).

So 16→4 adds a fraction of a block against a ≥10-block (Solana ~50) budget — a
**~50–250× margin**. The multiplier change alone cannot push endpoints over the gate,
so **`--consistency-block-gap-factor` can stay at the default 2 when running multiplier
4** — and since 4 is the floor (you can't go more aggressive), this holds across the
whole `[4,16]` range. Confirmed by the runtime A/B below: it ran with the gap-factor
**unset** and saw no rejections.

### So why add the second flag at all

1. **Coupled companion (lava's lesson).** lava #2276 ships them together because
   slowing polling without widening the gate backfires into rejections under more
   aggressive conditions — the principle holds even where 16→4 doesn't trigger it.
2. **Load-bearing on other chains.** The danger case is a chain pinned at the floor-10
   threshold (small `blockLagForQosSync`, small finalization), where even small added
   staleness eats a larger fraction of the budget — there the operator needs the knob.
3. **Non-polling lag sources.** `lag` also grows from a genuinely slow endpoint or
   network jitter, independent of the multiplier; the gap-factor is a general tolerance
   lever for those.
4. **Free + reversible.** Zero-default = no effect; it's there to answer the
   "all endpoints failed consistency pre-validation" / "no pairings" canary without
   another deploy.

In short: **not required for 16→4** (the math has huge margin), but shipped as the
defensive, chain-portable companion and canary-response knob.

## Verification

- `go build ./...` green; `relaycore` + `rpcsmartrouter` suites green.
- New unit tests: clamp floor at 2/3, effective-multiplier override, and the gap-factor
  widening the threshold on QoS-lag-dominated values.
- **Flag wiring confirmed**: both flags appear in `rpcsmartrouter --help`.
- `chaintracker`: 5/6 runs clean; the one failure was a pre-existing **real-timer**
  timing test (`TestChainTrackerFetchSpreadAcrossPollingTime` / `...PollingTimeUpdate`),
  inert on this change's path — timing flakiness, not a regression.

### ✅ Runtime-measured on the deployed server

Deployed to the **eth-sim** router on the `anna` tenant and counted the actual
chain-tracker polls at the simulator (`GET /history?method=eth_blockNumber`,
per-endpoint, 60s windows):

| config | startup log | polls / 60s (per endpoint) | measured interval | expected |
|--------|-------------|----------------------------|-------------------|----------|
| multiplier **16** (default) | `pollingInterval 812.5ms` | 73 | **822 ms** | base/16 = 812 ms |
| multiplier **4** | `pollingInterval 3.25s` + `polling-relief active` | 19 | **3158 ms** | base/4 = 3250 ms |

→ **73 → 19 polls = 3.84× fewer**; the probe fires at the base/4 cadence. Honored
end-to-end: parsed → `polling-relief active chainTrackerPollingMultiplier=4` →
`pollingInterval 3.25s` → actual polls at ~3.16 s. (eth-sim `averageBlockTime ≈ 13 s`.)

The ~74% reduction here is consistent with the handoff doc's fast-node figure (~84%);
the magnitude is **upstream-latency-dependent** — largest on fast/co-located nodes, and
only ~23% on a high-latency gateway where the synchronous fetch already throttles.

### DIFF
--- a/protocol/chaintracker/chain_tracker.go
+++ b/protocol/chaintracker/chain_tracker.go
@@ -56,17 +56,44 @@ type IChainTracker interface {
 }
 
 const (
-	initRetriesCount               = 4
-	BACKOFF_MAX_TIME               = 10 * time.Minute
-	maxFails                       = 10
-	GoodStabilityThreshold         = 0.3
-	PollingUpdateLength            = 10
-	MostFrequentPollingMultiplier  = 16
-	LavaPollingMultiplierFrequency = 4
-	PollingMultiplierFlagName      = "polling-multiplier"
+	initRetriesCount                      = 4
+	BACKOFF_MAX_TIME                      = 10 * time.Minute
+	maxFails                              = 10
+	GoodStabilityThreshold                = 0.3
+	PollingUpdateLength                   = 10
+	MostFrequentPollingMultiplier         = 16
+	MinPollingTimeMultiplier              = 4 // adaptive tiers compute base/(multiplier/4); below this divides by zero
+	ChainTrackerPollingMultiplierFlagName = "chain-tracker-polling-multiplier"
 )
 
-var PollingMultiplier = uint64(1)
+// PollingTimeMultiplierOverride is the polling-relief process-wide override of
+// MostFrequentPollingMultiplier (default 16). 0 = no relief (current behavior).
+// A smaller value = slower polling = fewer upstream calls. Set once at startup from
+// the --chain-tracker-polling-multiplier flag and honored in newCustomChainTracker,
+// so it applies to every tracker (the global one and all per-endpoint ones).
+var PollingTimeMultiplierOverride = 0
+
+// EffectivePollingMultiplier returns the multiplier actually in force: the
+// polling-relief override when set, otherwise the built-in MostFrequentPollingMultiplier.
+func EffectivePollingMultiplier() int {
+	if PollingTimeMultiplierOverride != 0 {
+		return PollingTimeMultiplierOverride
+	}
+	return MostFrequentPollingMultiplier
+}
+
+// clampPollingMultiplier enforces the divide-by-zero floor: updateTimer's adaptive
+// tiers compute base/(multiplier/4), so a multiplier below 4 divides by zero. Warns
+// when it has to clamp — the flag path is already range-validated, so this only fires
+// if a per-tracker config sets PollingTimeMultiplier below the floor.
+func clampPollingMultiplier(m int) int {
+	if m < MinPollingTimeMultiplier {
+		utils.LavaFormatWarning("chain-tracker polling multiplier below safe floor; clamping", nil,
+			utils.LogAttr("provided", m), utils.LogAttr("floor", MinPollingTimeMultiplier))
+		return MinPollingTimeMultiplier
+	}
+	return m
+}
 
 type ChainFetcher interface {
 	FetchLatestBlockNum(ctx context.Context) (int64, error)
@@ -505,9 +532,6 @@ func (cs *ChainTracker) updateTimer(tickerBaseTime time.Duration, fetchFails uin
 		newPollingTime = tickerBaseTime / cs.pollingTimeMultiplier
 	}
 	newTickerDuration := exponentialBackoff(newPollingTime, fetchFails)
-	if PollingMultiplier > 1 {
-		newTickerDuration /= time.Duration(PollingMultiplier)
-	}
 
 	cs.timer = time.NewTimer(newTickerDuration)
 }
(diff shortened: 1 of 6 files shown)


Tomelia1999 (changes requested):
We actually set the flags in a config file, not via CLI.
The implementation set viper to read in the following order:
read from cli flags
override the default "chain-tracker-polling-multiplier" parameter in case it was set via CLI (it wasn't, so it will be 0)
read from flags from config file (too late)
You should change the order so we:
read flags from cli
read flags from yml config file
overide default "chain-tracker-polling-multiplier"

AnnaR-prog:
@Tomelia1999 good catch — fixed in df80fbb.

You were exactly right about the ordering. The override-reading block in `RunE` ran **before** `viper.ReadInConfig()`, so a flag set in `config.yml` (not CLI) was read as `0` and silently dropped — only CLI-passed flags took effect. I moved the block to just **after** the config file is merged into viper (still before any chain tracker / consistency config is built), so the flags now resolve with viper's normal precedence: **CLI-if-passed > config.yml > default**.

Verified by running the binary both ways:
- **config.yml** (the broken case): now logs `polling-relief active chainTrackerPollingMultiplier=4 consistencyBlockGapFactor=4` — logged nothing before the move.
- **CLI** (`--chain-tracker-polling-multiplier=4`): still honored.

Build clean and the `rpcsmartrouter` suite passes. Ready for re-review 🙏

AnnaR-prog:
Verified on a **live deploy** with the flag set in `config.yml` (a top-level key, **not** a CLI arg) — the exact delivery path you flagged.

Startup logs (fixed binary):
- `read config file successfully … /smart-router/config/config.yml`
- `polling-relief active chainTrackerPollingMultiplier=4 consistencyBlockGapFactor=4` ← read **from the config file** (logged nothing before the reorder)
- `Chain tracker started  pollingInterval=3.25s` (= base/4)

Actual polling, counted at the upstream over a 60s steady-state window:

| delivery | polls/60s (per endpoint) | interval |
|---|---|---|
| baseline (multiplier 16) | 73 | ~822 ms (base/16) |
| **multiplier 4 via `config.yml`** | **18** | **~3333 ms (base/4)** |

→ ~4× fewer polls, matching the CLI-delivered numbers — so config-file delivery is honored end-to-end on a real deploy. (Bonus: as a config-file key it's harmlessly ignored by a binary that doesn't recognize it, unlike a CLI flag which would crash an older build.)

avitenzer (approved):

nimrod-teich:

```bash
git fetch origin
```

PR #100, opened 2026-06-11
fix: drop the per-socket isHealthy selection gate

Tomelia1999:
# Description

Removes the per-socket `DirectRPCConnection.healthy` bit (`isHealthy`) and the gates that read it, fixing the `No pairings available` connection-health **deadlock** (Hypernative/monad + gk8solana incidents). Selection is now driven by QoS + the self-healing `endpoint.Enabled` backoff — the way WebSocket connections already behaved.

## The bug

Each direct-RPC socket carried a `healthy atomic.Bool`. **Any** transport blip (dial refused / reset / EOF / read error) latched it `false`, and **both** the relay path *and* the chain-tracker fetcher refused to send over an `unhealthy` socket. Since the only thing that set it back to `true` was a *successful send over that same socket*, once it latched `false` the curing send could never happen → **permanent `No pairings available` until process restart**.

The relay path and the chain tracker **share one `DirectRPCConnection` per endpoint**, so a failed *background tracker poll* poisons the *user relay path* too (this is the gk8solana mechanism).

## What changed

- Remove the three read-gates on the bit: relay path (`consumer_types.go`), chain-tracker fetcher (`endpoint_chain_fetcher.go`), epoch-transition probe (`consumer_session_manager.go`).
- Delete the now-dead plumbing: `IsHealthy`/`MarkHealthy` (interface + HTTP/gRPC/WS impls), `SessionConnection.IsHealthy()`, the `/debug/reset-scores` connection-health reset (`ResetAllConnectionHealth` + the `connections_reset` response field), and the orphaned `EndpointChainFetcher.GetDirectConnection`.
- Add a nil-element guard on `DirectConnections[0]` in the relay path (latent pre-existing panic risk surfaced during review).
- **Raise `MaxConsecutiveConnectionAttempts` 5 → 50** (`common.go`, commit `25dc90f`). With the health gate gone, `endpoint.Enabled` becomes the *sole* automatic disable mechanism — see the two-latch note below for why 50 and exactly what it trades off.
- Update/repoint tests; net code change ≈ **+45 / −497**.

**Two-latch note (why this is safe, and what the threshold change trades off):** there are two automatic disable mechanisms — the `healthy` bool (single-blip latch, **no** auto-recovery → the deadlock, removed here) and `endpoint.Enabled` (a consecutive-failure backoff that **already self-heals** on a successful relay — `ResetHealth` zeroes the refusal counter — and at every epoch). Today the bool gate fires *first* and starves the `Enabled` resets, making them futile; removing the bool is what makes the `Enabled` backoff effective.

This PR **also raises that backoff threshold from 5 → 50** (`MaxConsecutiveConnectionAttempts`, commit `25dc90f`), and that deserves to be called out plainly: with the gate gone `endpoint.Enabled` is the *only* automatic disable, so the net behavioral change here is bigger than "drop the gate" — it is **"drop the gate *and* make the remaining backoff 10× more lenient."** Concretely, a dead endpoint in a multi-endpoint pool now stays in selection rotation for up to **50** dial-and-fail cycles (each bounded by the relay timeout) instead of 5 before it is backed off — which directly **amplifies the outage-time-dialing tradeoff** noted below. Why 50 and not 5: in the single-endpoint shape (the actual incident shape), disabling the *sole* endpoint merely re-creates `No pairings` until the next epoch re-enables it (see the on-call note), so a higher threshold keeps the lone endpoint fast-failing (consistent `500`s) through brief outages instead of flapping into `No pairings`. The threshold is **shared with the provider-relay path and `blockProvider`**, so this leniency applies there too. (If preferred, `25dc90f` can be split out so the gate removal is reviewed against the original 5-strike backoff — but the value was chosen deliberately for the single-endpoint case.)

## QA / verification (live A/B)

Harness: single-endpoint Solana **direct-RPC** router with a toggleable reverse proxy between the router and the real upstream, so we can cut/restore the upstream (a true transport failure — connection refused, *not* an HTTP error) without touching the router process. Same harness, config, and outage sequence run against both binaries.

| Observation | **`main` @ `7957d91b7` (pre-fix)** | **this branch (fixed)** |
|---|---|---|
| Baseline | `200 200 200` | `200 200 200` |
| While upstream down | `500` (fast-fail ~25 ms) | `500` (fast-fail ~25 ms) |
| **After upstream restored** | `500 ×8` — **never recovers** | `200 ×5` — **immediate** |
| `direct connection is not healthy` (fetcher gate) | **24** | **0** |
| `direct RPC connection is unhealthy` (relay gate) | **20** | **0** |
| `Pairing list empty` / `No pairings` after restore | present (repeating) | absent |
| Cure | **process restart only** | none needed* |

\* For the **tested outage duration**. See the on-call note below for the single-endpoint long-outage edge.

**`main` — the deadlock** (upstream is provably healthy the whole time):
```
ERR failed to fetch all previous blocks and was necessary
  error="[SVMChainTracker] ... direct connection is not healthy for endpoint http://127.0.0.1:8899 ..." fetchFails=13
WRN direct RPC connection is unhealthy endpoint=http://127.0.0.1:8899
DBG Pairing list empty IgnoredProviderList=map[sol-rpc-1:{}] Provider list=sol-rpc-1
DBG SOLANA could not get a provider addresses error="No pairings available."
```

**This branch — self-heals** (same trigger, the tracker hits the dead shared socket, but nothing latches):
```
ERR provider node error error="...connect: connection refused" error_name=PROTOCOL_CONNECTION_REFUSED retryable=true
... upstream restored, no router restart ...
DBG direct RPC relay succeeded in goroutine endpoint=sol-rpc-1 latency=780ms
```
Gate strings over the window: **0**. A reproducible harness (toggleable upstream proxy), raw evidence logs, and a QA suite design are maintained in the team's internal QA notes for this incident (not in-repo).

## On-call note (single-endpoint long outages)

The QA table's *"Cure: none needed"* holds for the **tested outage duration**, not unconditionally. In a **single-endpoint** deployment, if the sole endpoint accrues `MaxConsecutiveConnectionAttempts` (50) consecutive failures — a long, uninterrupted outage — `endpoint.Enabled` flips `false` → there are zero usable endpoints → `probeDirectRPCEndpoints` returns *"no enabled direct RPC endpoints"* and selection again yields `No pairings available`, until `checkAndUnblockHealthyReBlockedProviders` re-enables it at the next epoch (`rpcsmartrouter.go`).

This is **strictly milder than the original deadlock** — it self-heals at the epoch boundary instead of requiring a process restart — and the 5→50 bump makes it far less likely to trip. **On-call should treat a between-epoch `No pairings` under a long single-endpoint outage as expected, self-healing behavior — not a regression.** The proper long-term fixes live in the roadmap: the hard `0.8` availability floor (**Step 2**) and active per-endpoint probing/recovery (**Step 3**).

## Tests

- Unit: bool/`MarkHealthy` tests removed; the meaningful behavior re-expressed as `SendRequest`-level tests (`…_SurfacesTransportError`, `…_4xxReturnsResponseAndError`); the probe's `endpoint.Enabled` guard test kept; the chain-fetcher guard test repointed at the surviving nil check.
- Added `TestFetchEndpointConnection_DirectRPC_BackoffAndSelfHeal` — a regression guard that drives the real selection path against an unreachable upstream and locks in the three properties the fix relies on: selection **ignores transport state**, an endpoint is backed off **only after `MaxConsecutiveConnectionAttempts` consecutive failures**, and a single `ResetHealth` (the successful-relay self-heal) returns it to rotation.
- `go build ./...` + the full `lavasession` and `rpcsmartrouter` suites pass locally.
- ⚠️ CI note: the `Build and Test` workflow currently only runs `go build` (no `go test`), so these test changes aren't gated by CI — verified locally + by the live A/B above.

## Out of scope / follow-ups

- **Outage-time dialing** (intended tradeoff, amplified by the 5→50 bump above): with the gate gone the tracker/relay now actually dial a dead endpoint each attempt instead of short-circuiting. Bounded by the relay timeout + the `Enabled` **50**-strike backoff (raised from 5 in this PR — see the two-latch note); the proactive-probe backoff / node-quota work is **Step 3** of the roadmap.
- Hard `0.8` availability floor (**Step 2**) and a permanent operator block-list (**Step 4**) are separate.
- The `endpoint.Enabled` recovery-before-epoch design is tracked as a future investigation.

---

## Author Checklist

* [x] included the correct type prefix in the PR title (`fix:`)
* [x] included the necessary unit tests (note CI `go test` gap above)
* [x] updated relevant documentation / Go comments

### DIFF
--- a/protocol/lavasession/common.go
+++ b/protocol/lavasession/common.go
@@ -33,6 +33,14 @@ func GetMaxAllowedBlockListedSessionPerProvider() int {
 }
 
 const (
+	// MaxConsecutiveConnectionAttempts is the number of consecutive connection
+	// failures after which an endpoint is backed off (endpoint.Enabled = false).
+	// Raised from 5 to 50 alongside removing the per-socket health gate: with that
+	// gate gone the (often sole) direct endpoint must not be disabled too eagerly on
+	// a transient blip, since disabling the only endpoint reproduces the "No pairings"
+	// symptom until an epoch/successful relay re-enables it. A successful relay resets
+	// the counter (see Endpoint.ResetHealth), so this is a consecutive-failure budget.
+	// Shared by the provider-relay path and blockProvider — this threshold gates both.
 	MaxConsecutiveConnectionAttempts                 = 50
 	TimeoutForEstablishingAConnection                = 1500 * time.Millisecond // 1.5 seconds
 	MaximumNumberOfFailuresAllowedPerConsumerSession = 15
--- a/protocol/lavasession/common.go
+++ b/protocol/lavasession/common.go
@@ -33,7 +33,7 @@ func GetMaxAllowedBlockListedSessionPerProvider() int {
 }
 
 const (
-	MaxConsecutiveConnectionAttempts                 = 5
+	MaxConsecutiveConnectionAttempts                 = 50
 	TimeoutForEstablishingAConnection                = 1500 * time.Millisecond // 1.5 seconds
 	MaximumNumberOfFailuresAllowedPerConsumerSession = 15
 	RelayNumberIncrement                             = 1
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -599,9 +599,8 @@ func (csm *ConsumerSessionManager) probeDirectRPCEndpoints(
 	consumerSessionsWithProvider.Lock.RLock()
 	defer consumerSessionsWithProvider.Lock.RUnlock()
 
-	var healthyEndpoints int
+	var usableEndpoints int
 	var totalEndpoints int
-	minLatency := time.Hour // Start with a large value
 
 	for _, endpoint := range consumerSessionsWithProvider.Endpoints {
 		if !endpoint.IsDirectRPC() {
@@ -610,12 +609,13 @@ func (csm *ConsumerSessionManager) probeDirectRPCEndpoints(
 
 		totalEndpoints++
 
-		// Belt-and-suspenders with conn.IsHealthy(): the endpoint struct tracks
-		// cumulative relay-path failures via MarkUnhealthy → ConnectionRefusals,
-		// while IsHealthy() tracks the last transport attempt. A disabled endpoint
-		// must never be considered healthy at probe time — otherwise a backup with
-		// 5+ consecutive refusals could be unblocked at epoch transition despite
-		// the relay path having already confirmed it's down.
+		// The relay path no longer gates on a per-socket health bit (a dead socket
+		// fails the real relay, which feeds QoS via OnSessionFailure and trips the
+		// endpoint.Enabled consecutive-failure backoff — both self-healing). So the
+		// probe no longer reads connection health either. We still honour
+		// endpoint.Enabled: a backup with 5+ consecutive refusals stays backed off
+		// and must not be optimistically unblocked at epoch transition. Active
+		// per-endpoint probing/recovery is handled by the chain tracker (Step 3).
 		if !endpoint.Enabled {
 			utils.LavaFormatDebug("Direct RPC endpoint is disabled, skipping probe",
 				utils.LogAttr("provider", providerAddress),
@@ -629,28 +629,10 @@ func (csm *ConsumerSessionManager) probeDirectRPCEndpoints(
 				continue
 			}
 
-			// Check connection health - this is a cheap operation
-			// that checks the internal health state without making a network call
-			startTime := time.Now()
-			healthy := conn.IsHealthy()
-			checkLatency := time.Since(startTime)
+			usableEndpoints++
 
-			if healthy {
-				healthyEndpoints++
-				// Track minimum latency (for consistent API with provider probe)
-				if checkLatency < minLatency {
-					minLatency = checkLatency
-				}
-
-				if DebugProbes {
-					utils.LavaFormatDebug("Direct RPC endpoint probe succeeded",
-						utils.LogAttr("provider", providerAddress),
-						utils.LogAttr("url", conn.GetURL()),
-						utils.LogAttr("protocol", conn.GetProtocol()),
-					)
-				}
-			} else {
-				utils.LavaFormatDebug("Direct RPC endpoint is unhealthy",
+			if DebugProbes {
+				utils.LavaFormatDebug("Direct RPC endpoint probe (no health gate)",
 					utils.LogAttr("provider", providerAddress),
 					utils.LogAttr("url", conn.GetURL()),
 					utils.LogAttr("protocol", conn.GetProtocol()),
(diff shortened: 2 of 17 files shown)


avitenzer:
The PR body's safety argument no longer matches the shipped constant
  
  This is the one thing I'd want resolved. The PR description's central "why this is safe" argument leans on the endpoint.Enabled backoff being a "5-consecutive-failure backoff that already self-heals", and
  the in-code comment at consumer_types.go:687 says "after MaxConsecutiveConnectionAttempts ... backed off." But commit 25dc90f, bundled into this same PR, raises that constant 5 → 50
  (protocol/lavasession/common.go:36).

  I traced the mechanics to confirm the impact:
  - Endpoint.MarkUnhealthy() (consumer_types.go:221) increments ConnectionRefusals and only sets Enabled = false at >= MaxConsecutiveConnectionAttempts.
  - A successful relay calls ResetHealth() (rpcsmartrouter_server.go:2091) which zeroes ConnectionRefusals.
  
  So the threshold is 50 consecutive failures with no intervening success. With the health gate gone, endpoint.Enabled is now the sole automatic disable mechanism — and this PR makes it fire 10× more slowly
  than the body claims. The net behavioral change shipped here is bigger than "drop the gate": it's "drop the gate and make the remaining backoff 10× more lenient."
  
  That deserves to be called out explicitly because it directly amplifies the one tradeoff the PR already acknowledges (outage-time dialing): a dead endpoint in a multi-endpoint pool now stays in selection
  rotation for up to 50 dial-and-fail cycles instead of 5 before it's backed off. Either:
  - update the PR body + the consumer_types.go comments to say 50, and justify why 50 (it's not obviously wrong — see the next point — but it's currently undocumented and contradicts the stated rationale), or
  - split 25dc90f out of this PR so the gate removal can be reasoned about against the original 5-strike backoff it was validated against.

avitenzer:
**Single-endpoint setups can still surface No pairings between epochs**

  The live A/B in the PR uses a single-endpoint Solana router, which is also the shape most exposed to a residual edge: if the sole endpoint accrues enough consecutive failures, Enabled flips false → there
  are now zero usable endpoints → probeDirectRPCEndpoints returns "no enabled direct RPC endpoints" and selection yields No pairings available again, until checkAndUnblockHealthyReBlockedProviders re-enables
  it at the next epoch (rpcsmartrouter.go:2108/2135).

  This is milder than the deadlock (it self-heals at epoch boundaries rather than requiring a restart), and raising the threshold to 50 makes it much less likely to trip — which is plausibly the real
  motivation for 25dc90f. But it means the "self-heals, cure: none needed" row in the QA table holds for the tested outage duration, not unconditionally. Worth a sentence in the PR so on-call doesn't treat a
  between-epoch No pairings under a long single-endpoint outage as a regression. The Step 2 (0.8 availability floor) / Step 3 (active probe) roadmap items are the right long-term home for this.

Tomelia1999:
> The PR body's safety argument no longer matches the shipped constant
> 
> This is the one thing I'd want resolved. The PR description's central "why this is safe" argument leans on the endpoint.Enabled backoff being a "5-consecutive-failure backoff that already self-heals", and the in-code comment at consumer_types.go:687 says "after MaxConsecutiveConnectionAttempts ... backed off." But commit [25dc90f](https://github.com/Magma-Devs/smart-router/commit/25dc90fe029dd7c674e1175977ab32b0864c5603), bundled into this same PR, raises that constant 5 → 50 (protocol/lavasession/common.go:36).
> 
> I traced the mechanics to confirm the impact:
> 
> * Endpoint.MarkUnhealthy() (consumer_types.go:221) increments ConnectionRefusals and only sets Enabled = false at >= MaxConsecutiveConnectionAttempts.
> * A successful relay calls ResetHealth() (rpcsmartrouter_server.go:2091) which zeroes ConnectionRefusals.
> 
> So the threshold is 50 consecutive failures with no intervening success. With the health gate gone, endpoint.Enabled is now the sole automatic disable mechanism — and this PR makes it fire 10× more slowly than the body claims. The net behavioral change shipped here is bigger than "drop the gate": it's "drop the gate and make the remaining backoff 10× more lenient."
> 
> That deserves to be called out explicitly because it directly amplifies the one tradeoff the PR already acknowledges (outage-time dialing): a dead endpoint in a multi-endpoint pool now stays in selection rotation for up to 50 dial-and-fail cycles instead of 5 before it's backed off. Either:
> 
> * update the PR body + the consumer_types.go comments to say 50, and justify why 50 (it's not obviously wrong — see the next point — but it's currently undocumented and contradicts the stated rationale), or
> * split [25dc90f](https://github.com/Magma-Devs/smart-router/commit/25dc90fe029dd7c674e1175977ab32b0864c5603) out of this PR so the gate removal can be reasoned about against the original 5-strike backoff it was validated against.

Tomelia1999:
change from 5 to 50 is the expected behavior in order to be less strict until we fix the deadlock bug by developing probing. We now wait for 50 concecutive connection errors before moving Endpoint.enabled to false instead of 5.

avitenzer (approved):

nimrod-teich:

```bash
git fetch origin
```

avitenzer (approved):

PR #102, opened 2026-06-11
feat: cross-validation enhancements — per-method policy, provider-group diversity, mismatch/failure observability

avitenzer:
## Summary

Implements the **Cross Validation Enhancements** PRD: cross-validation moves from a purely caller-driven, per-request mechanism to one an operator can mandate per method, with provider-group diversity requirements and richer mismatch/failure observability. Quorum failures surface as structured signals so clients decide whether to retry, fall back, or fail — recovery logic stays on the client, not the router.

## Use cases delivered

| UC | What | Where |
|----|------|-------|
| **UC-1** | Per-method validation policy (enable/disable/strengthen per method; deterministic, logged resolution) | `cross_validation_policy.go` resolver |
| **UC-2** | Provider-group diversity quorum — a quorum must span ≥ N distinct groups; **per-group quorum** stronger variant (each group reaches its own internal quorum) | group-aware selection + termination + gate |
| **UC-4** | Quorum mismatch → bounded `cross_validation_mismatch_total{group,finality}` metric (+ per-provider disagreements) for alerting | metrics manager |
| **UC-5** | Quorum failure → structured client signal (HTTP 500 + `lava-cross-validation-*` headers + error body), distinguishable from a generic upstream error | header/body emission |
| **UC-6** | Outlier provider excluded from the result set; exclusion recorded in log + metric | outlier detection |
| **UC-7** | No policy configured → existing header-driven behavior, fully backwards compatible | resolver passthrough |

UC-3 (high-value-transaction triggers) was **descoped**.

## End-to-end test harnesses

Driven against live Lava mainnet and the `provider_simulator` mock:
- `scripts/pre_setups/test_uc1_smartrouter_lava.sh` — per-method policy
- `scripts/pre_setups/test_uc2_smartrouter_lava.sh` — group diversity + per-group quorum (16 checks)
- `scripts/pre_setups/test_uc4_smartrouter_sim.sh` — UC-4 + UC-5 + UC-6 from one fault-injection primitive (13 checks, deterministic)

## Code review + fixes

A correctness review of the implementation surfaced **7 bugs, all fixed with regression tests** (commits `6b5f145`, `c0c77a8`):

1. **(security)** a caller could collapse an operator-mandated quorum by sending a small `max-participants` — the effective `max-participants` is now floored at the operator's no-caller threshold/min-groups
2. `lava-select-provider`/stickiness override under group diversity is now logged, not silently dropped
3. quorum-failure reason no longer mislabels `no-agreement` as `diversity-unmet`
4. nil-reply consensus no longer flags real responders as mismatch-metric outliers
5. header-path `MinGroups` defaults to `1`, not `0`
6. early-exit no longer commits to the nil fallback before a preferred real quorum can form
7. structural fail-fast reason now takes precedence over the final-eval reason on failure

## Testing

`protocol/relaycore`, `protocol/chainlib`, `protocol/lavasession`, `protocol/rpcsmartrouter` all green; `go vet` and `gofmt` clean. Each fix ships a test that fails without it.

## Follow-ups (not blocking)

Cleanup/altitude items from the review, deferred as non-correctness refactors: a triplicated group-adequacy predicate, the `"default"` empty-group literal duplicated across ~7 sites, and an O(N) provider-ranking cost on the group-diversity hot path.


protocol/rpcsmartrouter/README.md
```diff
@@ -52,6 +52,179 @@ backup-providers:
       - url: https://eth-mainnet.g.alchemy.com/v2/BACKUP_KEY
 ```
 
+## Cross-Validation
+
+Cross-validation fans a single request out to **N providers in parallel**, hashes each
+successful response (`SHA256(reply.data)`), and only returns an answer once a **quorum** of
```
sotskov-do, line 58: Is it reliable way to compare responses? What if fields in responses will be in different orders?

protocol/rpcsmartrouter/cross_validation_policy_test.go
```diff
@@ -0,0 +1,615 @@
...
+	"github.com/magma-Devs/smart-router/protocol/chainlib"
+	"github.com/magma-Devs/smart-router/protocol/common"
+	"github.com/magma-Devs/smart-router/protocol/lavasession"
+	spectypes "github.com/magma-Devs/smart-router/types/spec"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func intPtr(i int) *int { return &i }
```
sotskov-do: Since we use go 1.26 in this project, there is no need in such function:
https://go.dev/blog/go1.26#language-changes

protocol/lavasession/consumer_session_manager.go
```diff
@@ -101,6 +101,85 @@ func (csm *ConsumerSessionManager) GetNumberOfValidProviders() int {
 	return len(csm.validAddresses)
 }
 
+// countDistinctGroups counts distinct cross-validation group labels across the given provider addresses,
+// folding an empty GroupLabel into the implicit "default" group. Assumes csm.lock is held.
+func (csm *ConsumerSessionManager) countDistinctGroups(addresses []string) int {
+	groups := make(map[string]struct{}, len(addresses))
+	for _, addr := range addresses {
+		label := "default"
```
sotskov-do: Better to make as constant, since there are multiple places where this value is used.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -101,6 +101,85 @@ func (csm *ConsumerSessionManager) GetNumberOfValidProviders() int {
...
+	assignments := make(map[string][]string)
+	for _, addr := range csm.validAddresses {
+		label := "default"
+		if cswp, ok := csm.pairing[addr]; ok && cswp.GroupLabel != "" {
+			label = cswp.GroupLabel
+		}
+		assignments[label] = append(assignments[label], addr)
+	}
+	for label := range assignments {
+		sort.Strings(assignments[label])
```
sotskov-do, line 157: What is the purpose of sorting?

protocol/lavasession/consumer_session_manager.go
```diff
@@ -1288,7 +1391,10 @@ func (csm *ConsumerSessionManager) tryGetConsumerSessionWithProviderFromBlockedP
 			utils.LavaFormatDebug("Epoch changed between getValidConsumerSessionsWithProvider to tryGetConsumerSessionWithProviderFromBlockedProviderList getting pairing from new epoch list", utils.LogAttr("GUID", ctx))
 		}
 		csm.lock.RUnlock() // unlock because getValidConsumerSessionsWithProvider is locking.
-		return csm.getValidConsumerSessionsWithProvider(ctx, wantedProviderNumber, ignoredProviders, cuNeededForSession, requestedBlock, addon, extensions, stateful, virtualEpoch, "", "")
+		// Emergency blocked-provider fallback: group-blind selection (minGroups=1, perGroupTarget=1). The
+		// diversity / per-group gate still enforces the requirement, so a non-diverse fallback fails rather
+		// than under-validates.
+		return csm.getValidConsumerSessionsWithProvider(ctx, wantedProviderNumber, ignoredProviders, cuNeededForSession, requestedBlock, addon, extensions, stateful, virtualEpoch, "", "", 1, 1)
```
sotskov-do: Magic numbers. Better to use `minGroups`, `perGroupTarget`

protocol/relaycore/relay_processor.go
```diff
@@ -114,6 +134,86 @@ func (rp *RelayProcessor) getMaxParticipants() int {
 	return 1
 }
 
+// getMinGroups returns the required number of distinct provider groups (1 = no diversity requirement).
+func (rp *RelayProcessor) getMinGroups() int {
+	if rp.crossValidationParams != nil && rp.crossValidationParams.MinGroups > 1 {
+		return rp.crossValidationParams.MinGroups
+	}
+	return 1
```
sotskov-do: Magic number. Better to define default as constant

protocol/relaycore/relay_processor.go
```diff
@@ -114,6 +134,86 @@ func (rp *RelayProcessor) getMaxParticipants() int {
...
+		}
+	}
+	return false
+}
+
+// quorumGroupOf returns the group label used for diversity counting for a result, folding an empty label
+// into the implicit "default" group.
+func quorumGroupOf(result common.RelayResult) string {
+	if result.ProviderInfo.ProviderGroup == "" {
+		return "default"
```
sotskov-do: Another usage of `"default"`. Better to define constant

protocol/relaycore/relay_processor.go
```diff
@@ -521,63 +695,140 @@ func (rp *RelayProcessor) responsesCrossValidation(results []common.RelayResult,
 		}
 	}
 
-	var maxCount int
-	var mostCommonResult common.RelayResult
-	var consensusHash [32]byte
+	minGroups := rp.getMinGroups()
+	perGroup := rp.perGroupQuorum()
+
+	var (
```
sotskov-do: Winner selection block is too big. Better to move to separate function with additional tests

protocol/relaycore/relay_processor.go
```diff
@@ -521,63 +695,140 @@ func (rp *RelayProcessor) responsesCrossValidation(results []common.RelayResult,
...
+				if !haveWinner || qualGroups > winnerQualGroups || (qualGroups == winnerQualGroups && count.count > winnerCount) {
+					haveWinner = true
+					winnerQualGroups = qualGroups
+					winnerCount = count.count
+					mostCommonResult = count.result
+					consensusHash = hash
+					consensusDistinctGroups = len(count.groupCounts)
+				}
+			}
+		} else if count.count >= crossValidationSize && len(count.groupCounts) >= minGroups {
```
sotskov-do: **Diversity over count: a smaller multi-group agreement beats a larger single-group one — worth documenting, and the fleet-sizing assumption worth flagging.**

  This gate (`count >= threshold && len(groupCounts) >= minGroups`) means a response hash that spans only one group can *never* win, no matter how many providers returned it. I
  validated the consequence with `min-groups: 2`, `threshold: 2`, groups `A:3 nodes, B:1, C:1`, `max-participants` large enough to query all:

  - `A`'s 3 identical responses → `count=3`, `groups={A}`, `len=1` → fails `len(groupCounts) >= 2`, **discarded**.
  - `B`+`C`'s 2 identical responses → `count=2`, `groups={B,C}`, `len=2` → **wins**.

  So two single-node groups override a 3-node group, and group `A` is then flagged as the outlier in `cross_validation_mismatch_total`. This is correct under the anti-Sybil intent (one
  group ≈ one operator/failure domain, so cross-group agreement is trusted over raw node count) — but it bakes in an assumption that **groups are independent and comparably sized**. A
  fleet with single-node groups half-defeats the guarantee: two tiny groups colluding (or simply both wrong at a block boundary on a non-deterministic method) outvote a larger honest
  group, and the diversity is carried by single points of failure.

  Two requests:
  1. Add a sentence to this gate's comment (or `hashQuorumReached` at line 169) making the "single-group consensus never wins regardless of count" rule explicit — it's the core
  semantic and currently has to be reverse-engineered.
  2. Should the startup capacity check (`validateCrossValidationStartup`) warn when an enabled `min-groups`/per-group policy is backed by groups smaller than the agreement threshold
  (e.g. single-node groups), since those make the diversity guarantee rest on SPOFs?

protocol/rpcsmartrouter/cross_validation_policy.go
```diff
@@ -0,0 +1,442 @@
...
+			}
+			return Bound{Floor: &n}, nil
+		default:
+			return data, nil // object form: let mapstructure decode the map into Bound
+		}
+	}
+}
+
+// scalarToInt converts a YAML scalar (int / float / numeric string) to an int for the Bound shorthand.
+func scalarToInt(data interface{}) (int, error) {
```
sotskov-do, line 289: What is the reason to allow user to insert any value (int, float, string)? In what case user could set, for example, `threshold` as `2.9`?
Seems like too flexible contract. Moreover according to this, there could be unexpected behavior, like `2.9` will floor to `2`.
What do you think to simplify UX and return error if user set configuration wrong?

protocol/rpcsmartrouter/cross_validation_policy.go
```diff
@@ -0,0 +1,442 @@
...
+		if isStateful(chainID, apiInterface, method) {
+			return fmt.Errorf("cross-validation policy on stateful method %s/%s/%s is not allowed: cross-validating a transaction-submission response is a no-op (see UC-3); strengthen write paths via the stateful fan-out instead", chainID, apiInterface, method)
+		}
+	}
+	return nil
+}
+
+// splitPolicyKey reverses policyKey for diagnostics. chain-id and api-interface come back lowercased.
+func splitPolicyKey(key string) (chainID, apiInterface, method string) {
+	parts := strings.SplitN(key, "\x00", 3)
```
sotskov-do: Same separator `"\x00"` is used in two places: here and `policyKey`. Better to define as constant

protocol/rpcsmartrouter/cross_validation_policy.go
```diff
@@ -0,0 +1,442 @@
...
+
+// PerGroupRequirements returns the no-caller resolved requirement of every ENABLED per-group-quorum policy
+// for the given chain/api. Empty when no per-group policy applies. The threshold has a caller header, so a
+// caller could raise it at runtime (failing group-quorum-unmet), but the no-caller value is what a
+// well-provisioned fleet must support to satisfy the policy by default.
+func (r *CrossValidationPolicyResolver) PerGroupRequirements(chainID, apiInterface string) []PerGroupRequirement {
+	if r == nil {
+		return nil
+	}
+	wantChain, wantAPI := strings.ToLower(chainID), strings.ToLower(apiInterface)
```
sotskov-do: What do you think building key as `chainID` + `"\x00"` + `apiInterface` and compare with policy key as `strings.HasPrefix`? Without building policy key and splitting it on parts?

sotskov-do (commented):
Did you run `smart-router-automation` tests in order to check possible regression?

protocol/rpcsmartrouter/README.md
avitenzer: We create canonical JSON but it general if we use the same node application that will not be an issue.

avitenzer:
> Did you run `smart-router-automation` tests in order to check possible regression?

No I didnt and Victoria also wrote some automation tests to test the various scenarios.

protocol/rpcsmartrouter/cross_validation_policy_test.go
avitenzer: fixed

protocol/lavasession/consumer_session_manager.go
avitenzer: resolved

protocol/lavasession/consumer_session_manager.go
avitenzer: The sort.Strings(assignments[label]) sorts the address slice within each group.

Its single purpose is to make that one-shot startup diagnostic log deterministic and readable

protocol/lavasession/consumer_session_manager.go
avitenzer: Resolved

protocol/relaycore/relay_processor.go
avitenzer: fixed

protocol/relaycore/relay_processor.go
avitenzer: fixed

protocol/relaycore/relay_processor.go
avitenzer: Request 1 — document the "single-group consensus never wins" rule (protocol/relaycore/relay_processor.go)
  - Expanded the doc comment on hashQuorumReached (the one predicate both checkEndProcessing and HasRequiredNodeResults flow through) to state outright: once minGroups > 1, a hash whose matching providers all
  sit in one group can never win regardless of count, with the anti-Sybil rationale and the B+C-outvote-A example.
  - Added a one-line echo at the selectQuorumWinner default-mode gate the comment was anchored to, pointing back to hashQuorumReached as the source of truth.
  
  Request 2 — startup SPOF warning (rpcsmartrouter_server.go + cross_validation_policy.go)
  - New MinGroupsRequirements resolver method (parallel to PerGroupRequirements, disjoint via the PerGroupQuorum flag) surfacing each enabled default-mode policy's (MinGroups, Threshold).
  - New pure helper groupsBelowThreshold(groupSizes, threshold) → sorted labels of under-staffed groups.
  - A log-only LavaFormatWarning in validateCrossValidationStartup, scoped to default min-groups mode, mirroring the existing len(groupSizes) > 0 skip guard. It returns nil — startup never fails on it.
  - Tests: TestGroupsBelowThreshold, TestMinGroupsRequirements_DisjointFromPerGroup, and two new TestValidateCrossValidationStartup subtests proving the under-staffed config stays NoError and the warning
  payload is right (groupsBelowThreshold=B,C).

protocol/rpcsmartrouter/cross_validation_policy.go
avitenzer: We used Float because it was needed by old config so I'll remove it.

protocol/rpcsmartrouter/cross_validation_policy.go
avitenzer: Fixed

nimrod-teich:

```bash
git fetch origin
```

sotskov-do (approved):

PR #104, opened 2026-06-14
fix(relaycore): canonicalize response before cross-validation hashing (MAG-2062)

avitenzer:
## Summary

Cross-validation bucketed providers by `sha256.Sum256` of the **raw** response bytes, so two providers returning the same semantic result with their JSON envelope keys in a different order (e.g. `{jsonrpc,id,result}` from LavaGateway vs `{id,jsonrpc,result}` from Tatum1) landed in separate hash buckets. Agreement then failed with `maxMatchingResults:1` and the request returned **HTTP 500** with `lava-cross-validation-status: failed`, even though both responses were byte-for-byte identical in their `result` field (**MAG-2062**).

## Root cause

A prior change (memory optimization) replaced canonical-form hashing with raw-byte `sha256.Sum256(Reply.Data)`, silently dropping the key-order invariance that the old `json.Unmarshal`→`json.Marshal` round-trip provided (Go's `encoding/json` sorts map keys on marshal). The bug is latent: it only manifests when **different provider implementations** return the same answer with different envelope serialization — common on multi-implementation chains (Solana: LavaGateway vs Tatum; ETH: Infura vs Alchemy vs QuickNode).

## Fix

Hash a **canonical form** instead — decode and re-marshal so object keys are sorted before hashing. Applied at **both** hash sites (the counting-site hash in `handleResponse` and the `responsesCrossValidation` fallback) so the cached and recomputed hashes always agree.

Design choices:
- **`json.Number` via `UseNumber()`** — plain unmarshal coerces numbers through `float64`, where two distinct large integers can collapse to one value and cause a **false agreement** (CV claiming agreement on different data — worse than the bug being fixed). `UseNumber` preserves the literal token, so it can only ever err toward the safe direction.
- **Decode into `interface{}`** (not `map[string]interface{}`) — correctly normalizes top-level arrays (e.g. `[]` from `eth_getLogs`) and recursively sorts nested objects.
- **Non-JSON fallback** — binary gRPC payloads that fail to parse fall back to raw-byte hashing, preserving prior behavior for non-JSON interfaces.

`Reply.Data` is untouched — canonicalization affects only the internal quorum bucket key; the bytes returned to the client stay verbatim.

## Tests

- `TestCrossValidationAgreesOnDifferentKeyOrder` — the ticket's **exact wire payloads**. Proven to catch the regression: reverting the hash to raw bytes reproduces the literal `maxMatchingResults:1,agreementThreshold:2` failure; with the fix it passes.
- `TestCanonicalResponseHash` — key-order invariance, whitespace invariance, the large-integer false-agreement guard, genuine-divergence disagreement, and non-JSON fallback.

## Verification

- `go build ./...` clean
- Full `relaycore` suite green
- `gofmt` clean

### DIFF
--- a/protocol/relaycore/relay_processor.go
+++ b/protocol/relaycore/relay_processor.go
@@ -1,8 +1,10 @@
 package relaycore
 
 import (
+	"bytes"
 	"context"
 	"crypto/sha256"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
@@ -373,6 +375,37 @@ func (rp *RelayProcessor) HasRequiredNodeResults(tries int) (bool, int) {
 	return false, nodeErrors
 }
 
+// canonicalResponseHash returns a hash of a provider response that is invariant
+// to JSON object key ordering and insignificant whitespace. Cross-validation
+// must place two providers that return the same semantic answer in the same
+// quorum bucket even when their JSON envelope keys are serialized in a different
+// order — e.g. {"jsonrpc","id","result"} from one provider vs
+// {"id","jsonrpc","result"} from another. Hashing the raw bytes would split
+// these into separate buckets and falsely fail agreement (see MAG-2062).
+//
+// It decodes into an interface{} and re-marshals; Go's encoding/json sorts map
+// keys alphabetically on marshal, yielding a deterministic canonical byte form.
+// json.Number (via UseNumber) preserves the literal numeric token rather than
+// coercing through float64 — without this, two distinct large integers could
+// round to the same float64 and produce a *false agreement*, which is worse
+// than the false negative this fixes.
+//
+// If the data is not valid JSON (e.g. binary gRPC payloads) it falls back to
+// hashing the raw bytes, preserving prior behavior for non-JSON interfaces.
+func canonicalResponseHash(data []byte) [32]byte {
+	dec := json.NewDecoder(bytes.NewReader(data))
+	dec.UseNumber()
+	var v interface{}
+	if err := dec.Decode(&v); err != nil {
+		return sha256.Sum256(data)
+	}
+	canonical, err := json.Marshal(v)
+	if err != nil {
+		return sha256.Sum256(data)
+	}
+	return sha256.Sum256(canonical)
+}
+
 func (rp *RelayProcessor) handleResponse(response *RelayResponse) {
 	nodeError := rp.ResultsManager.SetResponse(response, rp.RelayStateMachine.GetProtocolMessage())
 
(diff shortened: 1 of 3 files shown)


AnnaR-prog:
### Memory under concurrent cross-validation — worth a follow-up (non-blocking)

Flagging a memory characteristic of `canonicalResponseHash`. It's **not a blocker** for this PR — the `selection == CrossValidation` gate already keeps it off the Stateless/Stateful hot path, which is the right call — but it's worth tracking before CV is enabled on any large-payload method.

`canonicalResponseHash` does a full `json.Decode` into `interface{}` **plus** a re-marshal, vs the old `sha256.Sum256(data)` which allocated nothing. I benchmarked it on `eth_getLogs`-shaped payloads (`-benchmem`, on the branch):

| Payload | Raw `sha256` | Canonical (this PR) |
|---|---|---|
| 773 KB | 0 B, 0 allocs | 7.4 MB, 57K allocs |
| 3.78 MB | 0 B, 0 allocs | ~33 MB, 285K allocs |

So transient allocation ≈ **9× the response size**, scaling linearly — a 10 MB `debug_traceTransaction` ≈ 90 MB transient per hash. The dominant cost is the `interface{}` tree, not the marshal; streaming the canonical bytes straight into the hasher (`json.NewEncoder(sha256.New())`) shaves ~13% but doesn't change the order of magnitude.

**Within a single request this is bounded** — `handleResponse` is serialized through the response channel, so ~1 tree is alive at a time and it's freed right after (only the 32-byte digest is retained). **The concern is across concurrent requests:** each request gets its own `RelayProcessor` and they run in parallel, and I don't see a cap on (a) concurrent in-flight CV requests, (b) `max-participants` (only `positive` + `threshold ≤ participants` are validated), or (c) response-body size on the hashing path (the `maxResponseSizeForBlockExtraction = 1MB` guard in `direct_rpc_relay.go` covers block extraction, not CV hashing). So the peak multiplies:

```
added transient  ≈  N_concurrent_CV_requests × (≈9 × response_size)
```

e.g. 10 concurrent CV requests on 10 MB traces → ~900 MB transient spike, on top of the ~N×K×response_size of raw bytes already held for comparison. On a memory-constrained router that's an OOM, and the blast radius is the whole process (all traffic, not just CV).

Two qualifiers keep it from being urgent: it only fires for **opt-in** CV traffic on large-payload methods, and it's a **transient spike** (freed per-response), not a retained footprint.

Possible follow-ups, if/when CV runs on heavy methods:
- **Concurrency semaphore** bounding simultaneous CV canonicalizations — preserves correctness and caps the peak regardless of N. The codebase already uses the buffered-channel-semaphore pattern (`spec_reverifier.go:109`). Preferred.
- **Size cap** — above ~1 MB fall back to raw-byte hashing (precedented by `maxResponseSizeForBlockExtraction`). Trade-off: large reordered responses would false-disagree again, so weaker on its own; fine as defense-in-depth.

Benchmark was a throwaway `_test.go` in `protocol/relaycore` (decode-into-`interface{}` + `json.Marshal`, `-benchmem`); happy to share.

avitenzer:
The analysis is correct and I verified each claim against the code (the CV gate at relay_processor.go:440, the unbounded max-participants at protocol_message.go:159, and
  the semaphore precedent at spec_reverifier.go:116).
  
  Agreed it's non-blocking. One framing note for the record: the ~9× is the per-response allocation factor (the interface{} tree), and because handleResponse serializes responses through the channel, it's one
  tree at a time per request — providers don't multiply it. Provider count only scales the retained comparison buffers, and our normal setup is 3–5 providers, so the realistic per-request transient on
  today's CV traffic (small identity methods like getGenesisHash/eth_chainId, ~100 B) is sub-KB. The hundreds-of-MB regime requires CV on multi-MB methods (eth_getLogs/debug_trace) and many concurrent CV
  requests — exactly the trigger you gated on.

AnnaR-prog (approved):

nimrod-teich:

```bash
git fetch origin
```

PR #132, opened 2026-06-21
fix(provider-optimizer): use per-endpoint block for sync-score (MAG-1748)

AnnaR-prog:
## Summary
Fixes **MAG-1748**: the provider optimizer's per-provider sync-score is never decremented for a stale provider on methods whose response body carries no block height (`eth_getBalance`, `eth_call`, `eth_estimateGas` — most production traffic).

## Root cause
For no-block-in-body methods, `extractBlockHeightFromJSONResponse` returns 0, so the router falls back to the **global** chain-tracker head (`rpcsmartrouter_server.go` → `chainTracker.GetLatestBlockNum()`) and writes it to `Reply.LatestBlock` for every provider. `OnSessionDone` then fed that global value to the optimizer, so a stale provider's per-provider `SyncBlock` ratcheted to the global head, `calculateSyncLag` returned 0, and its sync selection-score stayed `1.0`. The correct per-endpoint signal (`Endpoint.LatestBlock`, polled by `endpoint_chain_tracker_manager`, already used by the consistency filter and `syncGap`/`qosManager`) was simply not plumbed into the optimizer's sync feed.

## Fix
`protocol/lavasession/consumer_session_manager.go` (`OnSessionDone`): feed the optimizer the per-endpoint `Endpoint.LatestBlock` instead of `latestServicedBlock` (guarded fallback for non-direct sessions).

## Verification — what this rests on (deterministic)
- **Code fact:** pre-fix, for no-block-body methods the optimizer is fed the *global* head for every provider, so a stale provider's sync lag is structurally always 0 — it can **never** be demoted on these methods.
- **Unit tests** (`protocol/provideroptimizer/sync_demotion_repro_test.go`): the optimizer demotes a provider when fed a per-endpoint block that lags the head, and does not when fed the global head. `go build ./...` clean; `lavasession` tests pass.

## Verification — live cluster (fix engages; full demonstration inconclusive)
On the eth-sim cluster the fix produced, for the stale provider, `raw_sync_sec=1200 / composite 0.7153` (vs `0.915` healthy). That value is **mechanically impossible on the pre-fix binary** (which feeds the global head → `raw_sync_sec≈0` → `composite 0.915`), so it confirms the fix *engages* and feeds the per-endpoint stale block. However, the sim harness cannot *reliably* reproduce the stale condition end-to-end — at least four confounds interact: the `eth_blockNumber` **response cache**, the **forward-only** per-endpoint chain tracker (`chain_tracker.go gotNewBlock`), the test's `deep_reset`→healthy→`set_stale` ordering, and chain-tracker poll propagation — so most runs show no demotion (the optimizer ends up seeing every provider at the same block). **Live sim verification is therefore inconclusive; the fix rests on the code analysis + unit tests above.** (An earlier revision of this PR over-claimed a clean live before/after — retracted.)

## Metrics note (investigated)
The `selection_score{score_type="sync"}` gauges are **not mis-computed** — they read the normalized sync field and agree with the optimizer's score-breakdown. Two are emitted: `lava_rpc_endpoint_selection_score` is **reactive** (set only on live selection via `SetProviderSelected`) so it lags when traffic is sparse; `lava_rpc_optimizer_selection_score` is **periodic**. MAG-1748's AC#2 is phrased against the reactive one — it's better asserted against the periodic gauge (or the reactive one should be refreshed periodically).

## Follow-up (separate, deterministic — a code question)
Whether the per-endpoint chain tracker's fetcher polls providers **directly** or **through the `eth_blockNumber` response cache** decides whether the sim (and arguably production, for cache hits) can ever observe per-provider staleness. Worth a look.

## Related
- MAG-1748 (parent), Scenario 3; smart-router-automation PR #116 (test side).

### DIFF
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -2016,8 +2016,20 @@ func (csm *ConsumerSessionManager) OnSessionDone(
 	// calculate QoS - syncGap is the difference between expected and actual block height (0 if not tracked)
 	csm.qosManager.CalculateQoS(csm.atomicReadCurrentEpoch(), consumerSession.SessionId, consumerSession.Parent.PublicLavaAddress, currentLatency, expectedLatency, syncGap, numOfProviders, int64(providersCount))
 	if !isHangingApi {
+		// MAG-1748: latestServicedBlock is the GLOBAL chain-tracker head for methods whose
+		// response body carries no block height (eth_getBalance/eth_call/eth_estimateGas) —
+		// it is identical for every provider, so feeding it here leaves the optimizer's
+		// per-provider sync-score blind to a stale provider. Prefer the per-endpoint
+		// ChainTracker block (endpoint_chain_tracker_manager keeps Endpoint.LatestBlock
+		// current regardless of method), so a lagging provider is actually demoted.
+		syncBlock := uint64(latestServicedBlock)
+		if drsc, ok := consumerSession.Connection.(*DirectRPCSessionConnection); ok && drsc.Endpoint != nil {
+			if endpointBlock := drsc.Endpoint.LatestBlock.Load(); endpointBlock > 0 {
+				syncBlock = uint64(endpointBlock)
+			}
+		}
 		// append relay data only for non hanging apis
-		go csm.providerOptimizer.AppendRelayData(consumerSession.Parent.PublicLavaAddress, currentLatency, specComputeUnits, uint64(latestServicedBlock))
+		go csm.providerOptimizer.AppendRelayData(consumerSession.Parent.PublicLavaAddress, currentLatency, specComputeUnits, syncBlock)
 	}
 
 	csm.updateMetricsManager(consumerSession, currentLatency, !isHangingApi) // apply latency only for non hanging apis
--- a/protocol/provideroptimizer/sync_demotion_repro_test.go
+++ b/protocol/provideroptimizer/sync_demotion_repro_test.go
@@ -1,0 +1,128 @@
+package provideroptimizer
+
+// Optimizer-level characterization for the MAG-1748 sync-demotion bug:
+//
+//	bug(provider-optimizer): per-provider sync-score not decremented for relays whose
+//	response body lacks a block height
+//
+// The optimizer itself is correct: a provider whose reported syncBlock lags the cluster
+// head IS demoted (lower sync-score => lower first-pick share). The production bug was that
+// the feed (consumer_session_manager.OnSessionDone) handed the optimizer the GLOBAL chain
+// head for every provider on methods whose response body carries no block height
+// (eth_getBalance/eth_call/…), so the lag was invisible.
+//
+// These two tests pin both sides of that contract:
+//   - PerEndpointBlock_DemotesLaggingProvider: feed per-endpoint blocks (the fix) => demoted.
+//   - GlobalHeadForAll_LeavesLaggingProviderUndetected: feed the global head to all (the bug)
+//     => not demoted.
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	spectypes "github.com/magma-Devs/smart-router/types/spec"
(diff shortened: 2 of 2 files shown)


avitenzer (approved):

AnnaR-prog:
**Correction (verification).** I earlier cited a live before/after of *35.3% → 28.7%* on the eth-sim cluster — I'm **retracting that**; it was within noise. The simulator's `set_stale` did not produce an optimizer-visible stale block in my runs: `lava_rpc_endpoint_latest_block{simprovider2}` stayed at the head (`2e7`) and `selection_score{score_type="sync"}` stayed `1.0` on **both** the old and fixed binaries, so the first-pick numbers carried no signal. The fix's evidence is the unit tests + code analysis; live sim verification is pending confirmation that `set_stale` actually makes `eth_blockNumber` return a stale head (the ticket's original `latest_block=19,999,900` shows it can). Description updated. The per-endpoint chain tracker is also forward-only, which correctly handles real prod staleness (provider freezes, head advances) but not the sim's artificial backward jump.

AnnaR-prog:
**Resolution to the correction above.** Re-verified using the optimizer's **score-breakdown** (the actual selection computation in the debug log) instead of the laggy gauge / noisy first-pick share. Under `set_stale(simprovider2, 100)` + `eth_getBalance`, the stale provider's composite goes **0.9154 (pre-fix, sync looks perfect: normalized_sync=1) → 0.7153 (fix, normalized_sync=0)** — i.e. the fix demotes it. The fix is confirmed. Two caveats: (1) the `selection_score{sync}` gauge stayed at 1.0 even while the real computation showed sync=0 — it's decoupled/laggy, so AC#2 (phrased against that gauge) likely needs a separate metrics-reporting fix; (2) demotion has a propagation delay (the per-endpoint ChainTracker must poll the stale head first), so first-pick bursts right after set_stale look noisy. Details in the updated description.

AnnaR-prog:
**Gauge investigation (the `selection_score{sync}` question).** Conclusion: **no value bug.** Both gauges read the correct normalized sync field and agree with the optimizer's score-breakdown. The wrinkle is design, not correctness: `lava_rpc_endpoint_selection_score` (the gauge AC#2 cites) is **reactive** — set only on live selection (`SetProviderSelected`, consumer_session_manager.go:1285/1324) — so it lags when traffic is sparse; `lava_rpc_optimizer_selection_score` is **periodic**. Recommendation: assert AC#2 against the periodic gauge (or refresh the reactive one). I've also made the description's live-verification claim conservative — the fix's `raw_sync_sec=1200/composite 0.715` is impossible on the pre-fix binary (so it confirms the fix *engages*), but the sim can't reliably reproduce demotion (eth_blockNumber response cache + forward-only chain tracker + deep_reset ordering + poll propagation), so live sim verification is **inconclusive** and the proof rests on code + unit tests. Open follow-up: does the per-endpoint chain-tracker fetcher poll directly or via the response cache?

AnnaR-prog:
**Follow-up resolved — chain-tracker fetcher: direct, not cache.** Traced it: the per-endpoint tracker polls each provider **directly** — `EndpointChainFetcher.FetchLatestBlockNum` → `sendRawRequest` → `directConnection.SendRequest`/`DoHTTPRequest` (endpoint_chain_fetcher.go:52/75/278/315), with no relay-cache lookup in the path. So the `eth_blockNumber` response cache is client-facing only and does **not** affect the fix's input (`Endpoint.LatestBlock` is the true per-provider block). The sim couldn't demote because the tracker is **forward-only** (`chain_tracker.go gotNewBlock`, accepts only block > current) and `set_stale` is a *backward* jump on a *static* head (all providers pinned at 2e7) — the tracker won't lower below the block it already saw. This does **not** affect production, where real staleness is a provider *freezing* while the head *advances*: the tracker correctly holds the frozen block, the global advances past it, lag opens, and the fix demotes. To reproduce in the sim, advance the head while keeping one provider lagging (a sim/test change, not a code change).

PR #143, opened 2026-06-23
feat: ChainTracker & Probing redesign (MAG-2157): per-endpoint observations → ChainState consensus → proactive prober

avitenzer:
## ChainTracker & Probing redesign (MAG-2157)

Replaces the global single-tip ChainTracker + synthetic probe with a two-plane
architecture: a **data plane** that observes and aggregates per-endpoint block
state, and a **decision plane** (probing + relay) that reads it to drive QoS and
endpoint health — with a strict one-directional flow (upstream → monitor →
ChainState → probe/relay → QoS/Enabled → selection) and **no back-arrows**.

### Topics (all subtasks of MAG-2157, all Done)

- **A — Per-endpoint observation contract** (MAG-2158): side-effect-free
  telemetry per endpoint (latest block, poll-health, last-successful poll),
  distinguishing poll vs relay-harvest sources.
- **B — Polling cadence & quota** (MAG-2159): flattened per-endpoint poll cadence
  (`avgBlockTime/2`), relay-block harvest, and a traffic gate that skips a poll
  when a fresh relay observation already covers the endpoint.
- **C — Per-chain ChainState tip + global-tracker removal** (MAG-2160): in-tree
  `chainstate.ChainState` with a monotonic, outlier-guarded tip and internal
  pull-based strict-majority consensus; the global ChainTracker, the estimator,
  and the dead gRPC service are deleted.
- **E — QoS + endpoint.Enabled contract** (MAG-2162): both relay and probe feed
  all three QoS dimensions (availability/latency/sync) into per-provider EWMAs
  (relay:probe = 4:1); one probe sample per provider per cycle; a single
  `endpoint.Enabled` state machine (relay disables fast, probe re-enables).
- **D — Proactive health prober** (MAG-2161): a per-chain loop on its own
  validated cadence reads stored telemetry + the consensus baseline (zero
  upstream calls), scores every endpoint (regular + backup), feeds one QoS sample
  per provider, and proactively re-enables recovered endpoints in seconds.

### Post-merge review corrections (F1–F7)

- **F1** probe re-enables only from DISTINCT successful polls produced after the
  disable instant (edge-triggered `disabledAt`, stamped at both disable sites).
- **F2** a re-enable also restores the provider to routing
  (`RestoreRecoveredProvider`) — not just `Enabled`.
- **F3** synchronized `endpoint.Enabled` reads; synthetic direct-RPC probe removed
  for static providers.
- **F4** per-interface, per-sample sync reference (no shared optimizer getter).
- **F5** no fresh consensus majority ⇒ sync omitted, never max-across-providers.
- **F6** dedicated validated `--probe-loop-interval`.
- **F7** synthetic direct-RPC liveness feed dropped for static providers.

### Concurrency hardening (`go test -race`)

Made `lavasession`, `rpcsmartrouter`, `provideroptimizer`, `chainlib`,
`chainproxy`, `rpcclient`, `tracing` race-clean. Notable real production bugs
fixed along the way:
- metrics goroutine reading a session after `OnSessionDone` freed it,
- rpcclient `handleResponse` writing `op.err` after signaling the waiter,
- gRPC method names missing the leading `/` (empty responses under newer gRPC),
- gofiber WS use-after-recycle on graceful shutdown.

### Endpoint recovery flow (disable → re-enable)

How an endpoint that hit 50 relay failures and was disabled comes back into
selection — automatically, in seconds, without an epoch wait:

1. **Disable.** The relay path disables the endpoint at
   `ConnectionRefusals >= MaxConsecutiveConnectionAttempts` (50) and stamps
   `disabledAt` (edge-triggered).
2. **It keeps being polled.** The per-endpoint ChainTracker is keyed by URL and is
   independent of `endpoint.Enabled` (the poller "never writes QoS or
   endpoint.Enabled") — trackers are only torn down on epoch/pairing changes, not
   on disable. So a disabled endpoint still polls upstream on its own cadence.
3. **Upstream recovers** → the next poll succeeds → the endpoint's observation
   advances `LastSuccessfulPoll` / `LatestBlock`.
4. **The prober** (`runProbeLoop`, default `--probe-loop-interval` 5s) reads that
   telemetry + the consensus baseline (zero upstream calls) and renders a verdict:
   alive (fresh successful poll) and keeping up (within lag tolerance of consensus).
5. **`Endpoint.RecordProbeVerdict`** re-enables it (`Enabled = true`,
   `ConnectionRefusals = 0`) only after **K=3 DISTINCT successful polls produced
   strictly after `disabledAt`** (F1) — a stale pre-disable observation can never
   re-enable, and a failed poll resets the streak.
6. **`csm.RestoreRecoveredProvider`** restores routing in the same step (F2):
   regular → back into `validAddresses` (out of `currentlyBlockedProviderAddresses`),
   backup → out of `blockedBackupProviders`. Without it the endpoint would be
   `Enabled` yet still unroutable.

**Timeline:** ~K × cadence ≈ **15s** at the default cadence, versus the old
behavior of waiting up to the 15-minute epoch.

Backstops still present: a successful relay re-enables via `Endpoint.ResetHealth`;
a successful reconnect (`retryDisabledEndpoints=true`) re-enables on the connect
path; the epoch transition re-probes previously-blocked providers
(`checkAndUnblockHealthyReBlockedProviders`).

Anti-flap: the fast relay disable (50 consecutive failures) and the cautious probe
re-enable (K=3 distinct healthy polls) use deliberately distinct thresholds, and
the probe never touches an already-enabled endpoint.

### Verification

- `go build ./...`, `go vet ./...`, `gofmt` — clean.
- Full non-race `make test` — **zero failures** (33 packages).
- `go test -race` — green for lavasession, rpcsmartrouter, provideroptimizer,
  chainlib, chainproxy, rpcclient, tracing, chainstate, endpointstate, score, rand.

Closes MAG-2158, MAG-2159, MAG-2160, MAG-2161, MAG-2162 (parent: MAG-2157).


avitenzer:
## Reviewer summary — latest two commits + merge

The core redesign (per-endpoint observations → ChainState consensus → proactive prober) is unchanged. Two commits were added on top, plus a merge of `main`. This note focuses review attention on those.

### `3ca6a23` — five review findings
| # | Area | Fix |
|---|------|-----|
| P1 | `dyncodec` reflection | Callback/parser paths stay slash-less; `/` is added only at the `grpc.ClientConn.Invoke` boundary. |
| P1 | Relay tip eligibility | Harvest a relay's block as the current tip only by **spec function tag** (`GET_BLOCKNUM`, or `GET_BLOCK_BY_NUM` requesting latest). `RequestedBlock==LATEST` and `GetUsedDefaultValue()` (non-deterministic for `eth_blockNumber`) cannot discriminate receipt/by-hash historical responses. |
| P2 | `chainstate` baseline age | Split the overloaded `baselineAt` into TTL-freshness (refreshed every confirming recompute) vs establishment time (preserved while the block is unchanged). Sync-lag now reflects a stuck baseline's true age; a reconfirmed baseline still never expires. |
| P2 | Tracker generation | Register the tracker + allocate its observation generation **synchronously** before the relay dispatch captures it (the poll loop stays async). Previously an early relay captured generation 0 and its harvested tip was silently dropped. |
| P2 | Optimizer `SyncBlock` | Per-provider striped lock + an authoritative monotonic floor read under it. The ristretto cache is async/eventually-consistent and can't enforce "never decreases"; a stale read (incl. the legacy availability-only writer) could regress the block. |

### `e2dc791` — merge of `main` (resolved 3 conflicts)
- **`consumer_types.go`** — kept the branch's probe recovery-flow fields; followed main's repo-wide removal of `Endpoint.Geolocation`.
- **`consumer_session_manager.go`** — kept the race-safe `DebugProbesEnabled()` atomic accessor; **the one resolution worth a close look:** the `OnSessionDone` sync feed now *composes* both fixes — main's MAG-1748 per-endpoint block (so a persistently-lagging provider is demoted) is passed as the `syncBlock` argument to the branch's `AppendRelayDataConsensus` + `resolveSyncReference` (F4/F5: measure lag against the consensus baseline, omit when no fresh majority). Taking either side alone would have silently dropped a shipped fix.
- **`rpcsmartrouter.go`** — kept the branch's removal of the redundant global ChainTracker (per-chain tip now comes from ChainState).
- Reconciled main's MAG-1748 characterization test to the branch's `appendRelayData` signature via `SyncReference{}` (the legacy max-across-providers path it was written against).

### Verification
`go build ./...`, `go vet ./...`, and the full `make test` suite all pass. Targeted `-race` green for `lavasession`, `rpcsmartrouter`, `provideroptimizer`, `chainstate`, `endpointstate`, and `dyncodec`. CI matrix on the PR is fully green.

### Suggested review focus
1. The MAG-1748 + F4/F5 composition in `OnSessionDone` (is per-endpoint-block-vs-consensus-baseline the intended pairing?).
2. The `chainstate` two-timestamp split (does any other consumer read the returned time expecting freshness rather than establishment? — the two in-tree consumers both want establishment).
3. The `SyncBlock` floor's interaction with MAG-1748 demotion (monotonicity predates this PR; Finding 5 only made it race-safe — `TestSyncDemotion_PerEndpointBlock_DemotesLaggingProvider` passes against the merged signature).

protocol/rpcsmartrouter/rpcsmartrouter_server.go
```diff
@@ -1677,6 +1726,325 @@ func (rpcss *RPCSmartRouterServer) filterEndpointsByConsistency(
...
+	endpointAddress string,
+) {
+	if targetEndpoint == nil {
+		return
+	}
+	tip, ok := rpcss.tipBlockFromRelay(chainMessage, reply)
+	if !ok {
+		return
+	}
+	rpcss.recordRelayBlockObservation(targetEndpoint, harvestGen, tip)
```
sotskov-do: **P1 (correctness) — stale relay still mutates tip state after the generation gate rejects it.**

`recordRelayBlockObservation` is generation-gated, but the four writes that follow (`LatestBlock.Store`, `LastBlockUpdate`, `updateLatestBlockHeight`, the latest-block metric, lines 1945-1957) run unconditionally. The doc comment above lists all of these as tip state that only a tip-eligible response "may move." After a same-URL tracker replacement bumps the generation, a stale relay's observation is correctly dropped from the store but still poisons every *other* tip-state field — exactly the poisoning the harvest gate exists to prevent.

Fix: have `recordRelayBlockObservation` return an `accepted bool` and gate all four tip-state writes on it.

protocol/chainstate/chain_state.go
```diff
@@ -0,0 +1,348 @@
...
+	// wider than BucketWidth. The widest such window is the largest set of endpoints that agree
+	// on the tip within tolerance. blocks[j] is the window max (sorted), so it is the candidate
+	// baseline for that window.
+	bestCount, bestMax := 0, int64(0)
+	i := 0
+	for j := 0; j < len(blocks); j++ {
+		for blocks[j]-blocks[i] > cfg.BucketWidth {
+			i++
+		}
+		if count := j - i + 1; count > bestCount {
```
sotskov-do: **P2 (correctness) — `computeMajorityBaseline` keeps only the first max-size window, understating the baseline on ties.**

`count > bestCount` is strict, so a later equal-size window never replaces `bestMax`. With observations `[1000,1002,1004]` and `BucketWidth=2`, both `[1000,1002]` and `[1002,1004]` are strict 2-of-3 majorities, but the loop returns `1002`. This understates the consensus baseline and affects sync scoring / outlier decisions.

Fix: prefer the highest tied majority window (`>=` with a max guard), or treat overlapping tied majorities as ambiguous.

protocol/chainstate/chain_state.go
```diff
@@ -0,0 +1,348 @@
...
+			bestCount = count
+			bestMax = blocks[j]
```
sotskov-do, line 421: **(correctness, PLAUSIBLE) — baseline is the cluster MAX, charging majority endpoints a sync-gap against a block only one node reported.**

With `[1000,1000,1002]` / `BucketWidth=2`, baseline = `bestMax` = `1002`, so the two endpoints at the true-majority `1000` get charged `syncGap = 2` (rpcsmartrouter_server.go:1519). This is the design's own "fast node penalizes the pod" mode, bounded to `BucketWidth`. Low severity and deliberate, but it's the same selection logic as the P2 note above — worth fixing together (e.g. baseline = window min/median rather than max).

protocol/chaintracker/chain_tracker.go
```diff
@@ -472,14 +475,30 @@ func (cs *ChainTracker) start(ctx context.Context, pollingTime time.Duration) er
...
+			cs.timer.Stop()
+		}
 		return err
 	}
+	if cs.flatPollInterval > 0 {
+		// First periodic poll is a full interval after the final init fetch.
+		cs.timer = time.NewTimer(cs.flatPollInterval)
+	}
 	utils.LavaFormatDebug("ChainTracker fetched init data successfully")
 	blockGapTicker := time.NewTicker(pollingTime) // initially every block we check for a polling time
```
sotskov-do: **(efficiency) — flat per-endpoint trackers still run the adaptive block-gap machinery whose result is discarded.**

For every per-endpoint tracker (`flatPollInterval > 0`), this `blockGapTicker` keeps firing `updatePollingTimeBasedOnBlockGap` (a `Percentile`+`Stability` sweep over `blockEventsGap`) and `AddBlockGap` keeps appending per block (line 442), but `computePollInterval` returns at line 558 without ever consulting the result. Pure wasted CPU/allocation multiplied across every direct-RPC endpoint.

Fix: skip starting `blockGapTicker` (and the `AddBlockGap` append) when `flatPollInterval > 0`.

protocol/lavasession/consumer_types.go
```diff
@@ -179,9 +180,24 @@ type Endpoint struct {
...
+	// lastRecoveryPoll is the LastSuccessfulPoll value already counted toward the hysteresis streak,
+	// so a probe cadence faster than the poll cadence cannot count one successful poll twice — only a
+	// strictly newer successful poll advances the streak (F1: "distinct post-disable polls").
+	lastRecoveryPoll time.Time
+	Addons           map[string]struct{}
+	Extensions       map[string]struct{}
+	mu               sync.RWMutex // Protects Connections, ConnectionRefusals, Enabled, consecutiveHealthyProbes, disabledAt, lastRecoveryPoll
 
 	// Per-endpoint sync tracking (for direct RPC QoS)
 	LatestBlock     atomic.Int64 // Latest block seen from this endpoint
```
sotskov-do: **(data race) — `LastBlockUpdate` (next line, 204) is a racy, write-only field.**

Plain `time.Time` (not atomic, not under `mu` — the mutex doc at line 200 lists what it protects and excludes this field), written concurrently by the relay-harvest fan-out goroutines (rpcsmartrouter_server.go:1946) and the tracker `NewLatestCallback` (endpoint_monitor.go:281,288), with **no production reader** anywhere. A `time.Time` is multi-word, so this is a real data race (would trip `-race`, which this PR targets). The field predates the PR but the harvest writer is new.

Fix: drop the field (no readers), or make it atomic / cover it with `mu`.

protocol/probing/verdict.go
```diff
@@ -0,0 +1,92 @@
...
+// is healthy this cycle. It makes NO upstream call and does not touch the data plane — the prober
+// loop (in rpcsmartrouter) reads the telemetry/baseline, calls this, then applies the verdicts via
+// the Topic E contract (AggregateProviderSample / AppendProbeData / Endpoint.RecordProbeVerdict).
+
+const (
+	// defaultStalenessMultiplier × averageBlockTime is the default "alive" horizon: an endpoint is
+	// alive if it produced a fresh observation (poll OR relay) within this window. A multiple of the
+	// block time so fast chains get a tighter horizon. Mirrors the chainstate consensus window
+	// concept but is judged per endpoint.
+	defaultStalenessMultiplier = 10
```
sotskov-do: **(cleanup) — duplicated `10 × averageBlockTime` staleness derivation across packages.**

This `defaultStalenessMultiplier = 10` (floor `minProbeStaleness`) and `chainstate.DefaultStalenessMultiplier = 10` (floor `minStalenessWindow`, chain_state.go:38) independently encode the same "fresh/alive horizon" concept; the comment here even says it mirrors the chainstate consensus window. Tuning one without the other silently diverges the consensus window from the liveness horizon. Give it one named source of truth.

protocol/lavasession/consumer_types.go
```diff
@@ -235,6 +275,74 @@ func (e *Endpoint) MarkUnhealthy() {
...
+	}
+	e.lastRecoveryPoll = recoveryPoll
+	e.consecutiveHealthyProbes++
+	if e.consecutiveHealthyProbes < reEnableAfterK {
+		return false
+	}
+
+	e.Enabled = true
+	e.ConnectionRefusals = 0
+	e.consecutiveHealthyProbes = 0
```
sotskov-do: **(cleanup) — recovery-field reset is copy-pasted across disable/enable sites.**

The `consecutiveHealthyProbes = 0` / `disabledAt = time.Time{}` / `lastRecoveryPoll = time.Time{}` block is duplicated here and at lines 262-263, 310-311, 319, 357-359. The invariant "every `Enabled` transition resets the recovery fields" is enforced only by remembering to copy the block; a new path that misses one silently breaks F1 hysteresis with no compile error.

Fix: extract a private `clearRecoveryTrackingLocked()` so the invariant is a property of the type.

protocol/endpointstate/endpoint_monitor.go
```diff
@@ -477,20 +579,29 @@ func (m *EndpointChainTrackerManager) Stop() {
...
+	utils.LavaFormatInfo("stopped EndpointMonitor",
 		utils.LogAttr("chainID", m.chainID),
 		utils.LogAttr("trackersStopped", trackerCount),
 	)
 }
 
 // ValidateEndpointSync checks if an endpoint is synced within the given threshold.
 // Returns true if the endpoint's latest block is within threshold of the reference block.
-func (m *EndpointChainTrackerManager) ValidateEndpointSync(endpointURL string, referenceBlock int64, threshold int64) bool {
+func (m *EndpointMonitor) ValidateEndpointSync(endpointURL string, referenceBlock int64, threshold int64) bool {
```
sotskov-do: **(cleanup) — `ValidateEndpointSync` / `GetSyncGap` (line 617) are dead after this PR.**

Sync-gap is now computed inline against `chainState.GetConsensusBaseline` (rpcsmartrouter_server.go:~1516); these two methods are callerless in production and survive only in their tests. Per CLAUDE.md's "minimise exported surface area," delete them (and their tests).

protocol/chainstate/chain_state.go
```diff
@@ -0,0 +1,348 @@
...
+			cs.latestBlock = baseline
+			cs.lastObservedAt = now
+		}
+	}
+}
+
+// HasConsensusBaseline reports the last computed baseline and whether a fresh majority existed
+// at the last Recompute, WITHOUT applying the TTL freshness check. Exposed for telemetry/tests;
+// not the sync-scoring path (that is GetConsensusBaseline, which enforces TTL).
+func (cs *ChainState) HasConsensusBaseline() (int64, bool) {
```
sotskov-do: **(cleanup) — `HasConsensusBaseline` is exported but has only test callers.**

The real read paths use `GetConsensusBaseline` / `GetConsensusBaselineWithTime` (TTL-gated). This exported variant bypasses the TTL freshness gate, inviting future callers to read a stale baseline — the exact bug the design removed. Unexport it for telemetry-only use, or drop it.

sotskov-do (commented):
Automated review of the ChainTracker & Probing redesign. The heaviest correctness candidates were refuted on verification (notUpdated nil-deref unreachable; outlier-guard tip-freeze can't occur; consistency stale-read bounded by design; probe re-enable gated on poll-only LastSuccessfulPoll). The findings below survived. P1 is the one I'd block merge on; the two chain_state.go baseline notes are a single fix.

avitenzer:
### Follow-up: per-endpoint tip consolidated into one shared store (`66a4f2b`)

**Why:** the per-endpoint observed tip was written to two independent places — `Endpoint.LatestBlock` (an atomic on the lavasession struct) and the `EndpointObservation` block triple (endpointstate's observation map). Both writes to `Endpoint.LatestBlock` (poll-path `NewLatestCallback` **and** the relay harvest) were **ungated**, so a stale/replaced-tracker observation could move the atomic while being correctly rejected by the gated store — leaving the two copies diverged.

**What changed:**
- New leaf package **`protocol/endpointtip`** — stdlib-only, holds the single source of truth for the tip triple `{Block, ObservedAt, Source}`, keyed `chain|apiInterface|url`. Being a leaf, it's importable by **both** the high layer (endpointstate observers) and the low layer (lavasession QoS reader), dissolving the import cycle that previously forced the value down onto `Endpoint`.
- **`Endpoint.LatestBlock` + `LastBlockUpdate` deleted** along with both ungated writes. The three readers (lavasession syncBlock, rpcsmartrouter syncGap + consistency fallback) now read the store through one nil-safe helper.
- The generation gate and the **time-monotonic** guard are preserved exactly (the guard relocated into `Store.Set`; lock order is `obsMu → store lock`; `onTipObservation` still fires after `obsMu` is released).
- Scope note: the ChainTracker's own internal poll head (`GetLatestBlockNum`) is intentionally kept as the *primary* read in consistency validation — it's the vendored library's read-only poll bookkeeping, not a tip we write, so it can't participate in the divergence this fixes.

**⚠️ Re-review ask:** `protocol/endpointstate/observation.go` was **restructured** — the observation map now stores only poll-health and `GetObservation`/`SnapshotObservations` compose the block triple back from the store (delegation, not a second copy). Its prior review is stale; please re-look at that file specifically.

**Verification:** `go build ./...`, `go vet`, and `-race` all green; full `rpcsmartrouter` + `lavasession` + downstream `probing`/`chainstate` suites pass; zero `.LatestBlock.Store(` remain anywhere.

protocol/rpcsmartrouter/rpcsmartrouter_server.go
avitenzer: Resolved

protocol/chainstate/chain_state.go
avitenzer: Resolved

protocol/chainstate/chain_state.go
avitenzer: Resolved

protocol/chaintracker/chain_tracker.go
avitenzer: Resolved

protocol/lavasession/consumer_types.go
avitenzer: The issue is outdated so can be skipped.

protocol/probing/verdict.go
avitenzer: Resolved

protocol/lavasession/consumer_types.go
avitenzer: resolved

protocol/endpointstate/endpoint_monitor.go
avitenzer: resolved

sotskov-do (approved):

PR #144, opened 2026-06-23
feat(smart-router/debug): recover endpoint health in /debug/reset-all + add /debug/reset-endpoint-health

AnnaR-prog:
## What

Adds `POST /debug/reset-everything` to the smart-router debug HTTP server: one synchronous call that returns the router to a just-restarted state, so stress tests that drive all providers down can run back-to-back without a pod restart.

## The real cause (the ticket was misdiagnosed)

Verified against the code + git history:

- The ticket says `/debug/reset-all` doesn't drain session-manager state (blocked / reported / sticky / retry-ban). **It does** — wired clearing in `consumer_session_manager.go:2316-2395`, landed May 2026.
- The ticket says a per-request `usedProviders` / `ignoredProviders` struct leaks across requests. **No such leak** — it's created per request and always freed on success and failure.

The actual reason the 500 sticks: an endpoint disables at `ConnectionRefusals >= MaxConsecutiveConnectionAttempts` (`consumer_types.go:224`). All-providers-down + 50 relays disables every endpoint → "no pairings available". `Endpoint.ResetHealth()` is the only re-enable, and the only caller is the ~15-minute epoch tick `updateEpoch`. No debug endpoint called it, so the only recovery was a pod restart. `/debug/reset-all` and `/debug/reset-pairing` clear everything *except* this endpoint-health flag — `rebuildPairingFromConfig` only rebuilds providers *absent* from the pairing, so a present-but-disabled provider is left untouched.

## What it does

`POST /debug/reset-everything` composes, cold (no network re-probing, no epoch bump):

1. The full `/debug/reset-all` clearing (factored into a shared `resetAllState` helper).
2. Chain-tracker latest-block flush.
3. Cold pairing rebuild from config.
4. **New:** `ResetEndpointHealth` on every session manager — re-enable disabled endpoints.
5. Prometheus health-gauge mirror so operators see recovery immediately rather than at the next epoch.

`POST` → 200 with a JSON summary (`endpoints_reenabled`, `trackers_reset`, `pairing_restored`, `cache_be_flushed`); any other method → 405. Registered only under `--debug-address`; absent from production builds.

## Testing

- `lavasession.TestEndpointHealthRecovery` — drives a CSM to the stuck state, proves reset-all's session work does **not** recover, proves `ResetEndpointHealth` does. Real behavior, not a round-trip.
- Handler tests — 200 + JSON shape, 405, nil-router safe, chain-tracker reset count.
- Full `lavasession` + `rpcsmartrouter` suites, `go build ./...`, `go vet ./...` (== `make lint`), gofmt — all pass. `/debug/reset-all`'s existing tests pass unchanged after the refactor.

## Notes / scope

- The handler→`ResetEndpointHealth` wiring (`endpoints_reenabled`) is proven at the session-manager layer; the full handler path through a wired router is the behavioral end-to-end (simulator: all-providers-down → `reset-everything` → next relay 200).
- README documents the endpoint (acceptance requires docs).

Part of MAG-1939.

### DIFF
--- a/protocol/rpcsmartrouter/rpcsmartrouter.go
+++ b/protocol/rpcsmartrouter/rpcsmartrouter.go
@@ -526,6 +526,52 @@ func resetAllConsistencies(m *common.SafeSyncMap[string, relaycore.Consistency])
 	})
 }
 
+// resetEndpointHealthAndGauge re-enables every endpoint across all session managers
+// (ConsumerSessionManager.ResetEndpointHealth — clears ConnectionRefusals, sets
+// Enabled=true) and mirrors the reset onto the per-endpoint Prometheus health gauge for
+// every provider (primary + backup), matching what the epoch tick does in updateEpoch.
+// Returns the number of endpoints actually re-enabled.
+//
+// Endpoints disabled after MaxConsecutiveConnectionAttempts consecutive failures (e.g. an
+// all-providers-down stress burst) otherwise stay disabled until the next epoch tick, the
+// only other ResetHealth caller — contaminating later tests. Shared by /debug/reset-all
+// and /debug/reset-endpoint-health.
+func resetEndpointHealthAndGauge(deps debugMuxDeps) int {
+	if deps.router == nil {
+		return 0
+	}
+	deps.router.mu.Lock()
+	defer deps.router.mu.Unlock()
+	total := 0
+	for chainKey, csm := range deps.router.sessionManagers {
+		if csm == nil {
+			continue
+		}
+		total += csm.ResetEndpointHealth()
+
+		// Mirror the struct reset onto the Prometheus health gauge so operators see
+		// providers recover immediately rather than at the next epoch tick — without it
+		// the gauge stays stuck at 0 until a successful relay, one a rarely-used backup
+		// may never receive.
+		server := deps.router.rpcServers[chainKey]
+		if server == nil || server.smartRouterEndpointMetrics == nil || server.listenEndpoint == nil {
+			continue
+		}
+		for _, sessions := range []map[uint64]*lavasession.ConsumerSessionsWithProvider{
+			deps.router.providerSessions[chainKey],
+			deps.router.backupProviderSessions[chainKey],
+		} {
+			for _, cswp := range sessions {
+				if cswp != nil {
+					server.smartRouterEndpointMetrics.SetEndpointOverallHealth(
+						server.listenEndpoint.ChainID, server.listenEndpoint.ApiInterface, cswp.PublicLavaAddress, true)
+				}
+			}
+		}
+	}
+	return total
+}
+
 // buildDebugMux constructs the /debug/time-warp, /debug/time, /debug/reset-scores,
 // and /debug/reset-all HTTP handlers.
 //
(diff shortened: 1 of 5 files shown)


protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -858,6 +848,73 @@ func buildDebugMux(deps debugMuxDeps) *http.ServeMux {
...
+	// (MAG-1939). Runs the full /debug/reset-all clearing, flushes the per-endpoint chain
+	// trackers, cold-rebuilds pairing from config (like /debug/reset-pairing), and — the
+	// piece no other debug endpoint covers — re-enables endpoint health on every provider.
+	//
+	// After an all-providers-down burst every endpoint hits MaxConsecutiveConnectionAttempts
+	// and is disabled; the only paths back are a successful relay (impossible while every
+	// endpoint is disabled) or the ~15-minute epoch tick, so reset-all + reset-pairing leave
+	// the router returning "no pairings available" until a pod restart. ResetEndpointHealth
+	// closes that gap. Stays cold like reset-pairing: no network re-probing, no epoch bump.
+	mux.HandleFunc("/debug/reset-everything", func(w http.ResponseWriter, r *http.Request) {
```
VicSheCodes: Surfacing a test-suite-side concern: 305 references in tests/, src/, scripts/ to /debug/reset-all or the deep_reset_router helper. With this PR as shipped, /debug/reset-all doesn't touch the new Endpoint.Enabled reset disabled endpoints stay disabled, blocking the very contamination MAG-2186 was filed to fix (MAG-2179 nightly contamination). Tests would have to be migrated to
/debug/reset-everything to benefit. If /debug/reset-all is expanded instead, every site inherits the fix free, no migration is required.

Could we deliver:

1. Add a focused /debug/reset-endpoint-health endpoint that calls ResetEndpointHealth() and updates the corresponding Prometheus health gauge. The name mirrors Endpoint.ResetHealth() in the source, making it easy to discover and grep.
2. Keep existing /debug/reset-pairing as-is.
3. Expand /debug/reset-all to additionally call ResetEndpointHealth(), update the Prometheus health gauge, and invoke rebuildPairingFromConfig() on top of its existing drain behavior. The "leave pairing intact" comment in /debug/reset-all doesn't apply here - rebuildPairingFromConfig performs a cold rebuild per the /debug/reset-pairing documentation.

PR keeps the genuinely new code (ResetEndpointHealth() + its unit test). Drops the bundled handler + helper + integration tests + README docs.

AnnaR-prog: Addressed via expand-reset-all + reset-endpoint-health, dropped the new endpoint. `/debug/reset-all` now also re-enables endpoint health (+ Prometheus gauge) and cold-rebuilds pairing, so every existing call site inherits the fix with no migration; added a focused `POST /debug/reset-endpoint-health`; kept `ResetEndpointHealth()` + its unit test. Pushed as b8be80f.

avitenzer:
**`TestEndpointHealthRecovery`: the trailing `csm.ResetBlockedProviders()` is redundant — `ResetEndpointHealth` alone recovers.**

The blocked-provider list self-heals: every `GetSessions` starts with `validatePairingListNotEmpty` → `setValidAddressesToDefaultValue` (`consumer_session_manager.go:816-817`, `750-758`), which refills `validAddresses` and clears the blocked list when `validAddresses` is empty. So once endpoints are re-enabled, the next selection serves regardless of blocked state. `Endpoint.Enabled=false` was the only un-recoverable state, and `ResetEndpointHealth` fixes it.

Verified: dropped the line locally → still `--- PASS: TestEndpointHealthRecovery`.

Suggest removing it (and the now-overstated comment below it) so the test proves the sharper claim — endpoint-health reset is *by itself* the recovery ingredient, matching what `/debug/reset-endpoint-health` actually does. Non-blocking; keeping it for test determinism is also fine if the comment is reworded to say so.

AnnaR-prog:
Good catch — fixed in f8c3fcc.

Removed the trailing `csm.ResetBlockedProviders()` and reworded the comment. You're right on every step: I traced the self-heal path and confirmed it — `GetSessions` → `validatePairingListNotEmpty` → `resetValidAddresses` → `setValidAddressesToDefaultValue`, which clears `currentlyBlockedProviderAddresses` and refills `validAddresses` from the pairing pool. So once `ResetEndpointHealth` re-enables the endpoints, the next `GetSessions` recovers on its own. `Endpoint.Enabled=false` was indeed the only un-recoverable state.

The test now proves the sharper claim, which also matches what `/debug/reset-endpoint-health` actually does (re-enable endpoints + gauge, nothing else). Verified `gofmt`/`go vet` clean and `TestEndpointHealthRecovery` still passes — the run log even shows the self-heal firing: `NO VALID PROVIDERS - TRIGGERING RESET` → `RESET COMPLETED ... newValidAddressesCount=3` → selection succeeds.

avitenzer (approved):

PR #155, opened 2026-06-25
feat(smart-router/debug): add endpoint-state, chain-state, provider-routing read endpoints (MAG-2202)

AnnaR-prog:
## What

Adds the first three of the four read-only debug endpoints requested in [MAG-2202](https://magmadevs.atlassian.net/browse/MAG-2202), so the MAG-2201 automation suite can read ChainTracker/Probing-redesign state from the wire instead of scraping logs or counting relays:

- `GET /debug/endpoint-state` — per-endpoint poll-health + enable/recovery (F1)
- `GET /debug/chain-state` — per-chain consensus/tip state (anti-lie, TTL, realignment)
- `GET /debug/provider-routing` — per-CSM valid / blocked / blocked-backup addresses (F2)

The backing state already exists on this base branch (Topics A/C + the F1 hysteresis fields), so this PR only adds **read accessors + HTTP handlers** — no behavior change to the data plane.

## ⚠️ Stacked on `chaintracker-redesign` (PR #143)

Base branch is **`chaintracker-redesign`, not `main`** — these endpoints read `protocol/chainstate` and `protocol/endpointstate`, which only exist there. This merges **after #143**.

`/debug/probe-loop` (endpoint 4) is intentionally **not** here — it needs counter instrumentation on the existing `runProbeLoop` rather than new read plumbing, so it lands in a separate fast-follow PR (see the corrected ticket comment).

## Response shape — please sanity-check, @victoria

The ticket specifies the per-record fields but not the envelope. I read *"flat top-level keys per endpoint, no nesting"* literally and made each endpoint return a **flat array of self-describing records** (identified by `ChainID` + `ApiInterface`, plus `NetworkAddress` for endpoint-state), deterministically sorted, no object nesting. Concrete examples:

**`/debug/endpoint-state`**
```json
[
  {
    "ChainID": "ETH1", "ApiInterface": "jsonrpc", "NetworkAddress": "http://node-a:8545",
    "Enabled": true, "DisabledAt": "", "ConsecutiveHealthyProbes": 0,
    "ConsecutivePollFailures": 0, "LastSuccessfulPoll": "2026-06-25T09:30:00Z",
    "LatestBlock": 21000000, "ObservedAt": "2026-06-25T09:30:01Z", "Source": "poll"
  }
]
```
(`ConsecutiveHealthyProbes` added beyond the ticket's list — it's the literal F1 quantity; field names are "suggestions" per the ticket.)

**`/debug/chain-state`**
```json
[
  {
    "ChainID": "ETH1", "ApiInterface": "jsonrpc",
    "ObservedTip": 21000000, "LastObservedAt": "2026-06-25T09:30:01Z",
    "ConsensusBaseline": 21000000, "HasBaseline": true,
    "BaselineSince": "2026-06-25T09:29:40Z", "Initialized": true
  }
]
```

**`/debug/provider-routing`**
```json
[
  {
    "ChainID": "ETH1", "ApiInterface": "jsonrpc",
    "ValidAddresses": ["lava@a", "lava@b"],
    "CurrentlyBlockedProviderAddresses": [],
    "BlockedBackupProviders": []
  }
]
```

Conventions honored: Go-identifier keys; timestamps RFC3339 UTC matching `/debug/time` (empty string for a zero time, e.g. `DisabledAt` on an enabled endpoint); millisecond integers N/A here (no durations in these three); address slices non-nil so empties encode as `[]` not `null`; `ChainID`/`ApiInterface` sourced from `listenEndpoint` / `csm.RPCEndpoint()`, never by splitting the chain key.

If the suite would rather have a map keyed by `chainID-apiInterface` (or chain→endpoint), say so and I'll switch — easier to change now than after the tests build to it.

## Acceptance criteria

- [x] Each endpoint returns HTTP 200 with the fields above, reflecting live process state
- [x] Any method other than GET returns HTTP 405
- [x] Registered only when `--debug-address` is set (added inside `buildDebugMux`); production builds unaffected
- [x] nil-router safe (returns `[]`) so a partial test fixture can probe unconditionally

## Test plan

- Accessor unit tests in home packages: `Endpoint.HealthSnapshot` (disable → K=3 hysteresis → re-enable), `ConsumerSessionManager.ProviderRoutingSnapshot` (copies, sorted, non-nil), `ChainState.DebugSnapshot` (raw fields incl. TTL-expired tip still visible).
- Handler tests: 405 (all three), nil-router safety, plus populated-row shape for chain-state and provider-routing and a wiring/skip test for endpoint-state.
- `go build ./...`, `gofmt`, and `go test -race` over `chainstate` / `lavasession` / `rpcsmartrouter` all green.

[MAG-2202]: https://magmadevs.atlassian.net/browse/MAG-2202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/protocol/rpcsmartrouter/rpcsmartrouter.go
+++ b/protocol/rpcsmartrouter/rpcsmartrouter.go
@@ -23,6 +23,7 @@ import (
 	"math"
 	"net/http"
 	"os/signal"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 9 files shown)


protocol/rpcsmartrouter/rpcsmartrouter.go
avitenzer, line 1: protocol/rpcsmartrouter/rpcsmartrouter.go:883 — /debug/endpoint-state can emit duplicate rows for one endpoint identity. (request fix)                                                                     
  GetAllDirectRPCEndpoints() (consumer_session_manager.go:504-505) collects from both csm.pairing and csm.backupProviders with no dedup. The codebase's own comment (consumer_session_manager.go:2274-2279)     
  states a provider address "can legitimately appear in both." When it does, the handler appends two rows sharing the same (ChainID, ApiInterface, NetworkAddress) — the very tuple the PR documents as each    
  row's uniqueness key. Because pairing[P] and backupProviders[P] are distinct *Endpoint structs, their HealthSnapshot() can differ (e.g. one Enabled=true, one false), and debugRowKey collides so sort.Slice  
  leaves their order non-deterministic. The MAG-2201 suite keying results by that tuple would silently overwrite/miscount. Fix: dedup by debugRowKey (or per-NetworkAddress within the chain loop) before       
  writeDebugRows.

avitenzer, line 1: protocol/rpcsmartrouter/rpcsmartrouter.go:979 — new timestamps use wall-clock, not the /debug/time-warp clock. (confirm with author)                                                                       
  DisabledAt, LastSuccessfulPoll, ObservedAt, BaselineSince, LastObservedAt all render real wall-clock instants. The time-warp offset is applied only to the optimizer clock (opt.NowFunc), not to              
  chainstate/endpoint health. If the automation suite manipulates the clock via /debug/time-warp (as chaintracker tests tend to), comparing these fields against /debug/time's effective_time will be off by the
  offset — they must compare against real_time. The format claim ("matches /debug/time convention") is accurate; only the clock source differs. Worth a one-line doc note so the suite author isn't surprised.

avitenzer, line 1: protocol/rpcsmartrouter/rpcsmartrouter.go:866-1007 — the three handlers are structural clones. (optional cleanup)
Identical method!=GET→405 / rows := []map[string]any{} / if deps.router != nil { Lock; iterate; Unlock } / writeDebugRows scaffolding, differing only in the per-row body. Extracting a handleDebugRows(deps, populate) helper would remove the drift risk where a 4th endpoint omits the nil-router guard and reintroduces the panic TestDebugStateEndpoints_NilRouterIsSafe guards against.

avitenzer (approved):

PR #170, opened 2026-06-28
fix(smart-router): fail over from a pinned provider on retry (MAG-2228)

AnnaR-prog:
## Problem

A request pinning a provider with `lava-select-provider` should apply the pin to the **first attempt only** and fail over to a different provider on retry. It didn't — via two distinct mechanisms producing the same "pinned provider doesn't fail over" symptom (MAG-2228).

- **Mechanism A** (no-toggle / already-archive requests): the pin was only stripped as a side effect of the archive add/remove rebuild (`preserveRetrySafeDirectives`). On a retry without a toggle the pin survived → the retry re-read it and dead-ended on `Selected provider already failed in this request` with no failover (HTTP 500 when no node-error result was available to return).
- **Mechanism B** (plain `eth_blockNumber`): the first retry speculatively adds the `archive` extension, which strips the pin (as designed) but changes the `routerKey`. The used/unwanted-provider exclusion is keyed by `routerKey`, so the just-failed provider — recorded under the no-extension key — was not excluded under the archive key and got re-selected ~50% of the time on a QoS-unbiased router.

## Fix

- **Fix 1** (`rpcsmartrouter_server.go`): honor `lava-select-provider` / `lava-stickiness` only on the first attempt (`usedProviders.BatchNumber() == 0`); drop them on retries so selection can fall through to a different provider. Extracted `resolvePinDirectives` for clarity and unit coverage.
- **Fix 2a** (`used_providers.go` + `unified_relay_state_machine.go`): new `UsedProviders.MigrateUnwantedProviders(src, dst)`, called from `stateTransition` when an extension toggle changes the `routerKey`, so a provider already tried under the old key stays excluded under the new one.

## Testing

Unit: `TestResolvePinDirectives`, `TestMigrateUnwantedProviders`; full `lavasession` / `relaycore` / `rpcsmartrouter` suites + `gofmt` + `go vet` all green.

Cluster verification (dev-anna, fix binary on eth-sim-router only, reverted after):

| scenario | before | after |
|---|---|---|
| pinned `eth_blockNumber` + retryable `-32603` (Mechanism B) | ~50% `Lava-Retries: 2`, re-hits the same provider | **10/10** `Lava-Retries: 1`, fails over to a different provider |
| pinned + `lava-extension: archive` (Mechanism A) | deterministic `Lava-Retries: NONE`, node error returned, no failover | **4/4** `Lava-Retries: 1`, success result |

## Known limitation

Fix 2a keys off the state-machine message. `updateProtocolMessageIfNeededWithNewEarliestData` can add `archive` *after* the migrate, but only when `earliestBlockHashRequested != NOT_APPLICABLE` (a no-op for `eth_blockNumber`, so the reported bug is covered). The narrower earliest-block→archive path is not covered; making `GetUnwantedProvidersToSend` union the tried set across all routerKeys ("2b") would close it fully — deferred here to keep the change surgical.

Ticket: MAG-2228

### DIFF
--- a/protocol/lavasession/used_providers.go
+++ b/protocol/lavasession/used_providers.go
@@ -367,3 +367,32 @@ func (up *UsedProviders) GetUnwantedProvidersToSend(routerKey RouterKey) map[str
 	}
 	return unwantedProvidersToSend
 }
+
+// MigrateUnwantedProviders copies every provider already tried under the src
+// routerKey (whether still in-flight in providers, or already responded and moved
+// to unwantedProviders) into the unwanted set of the dst routerKey.
+//
+// The used/unwanted exclusion is partitioned per routerKey, and routerKey is derived
+// from the request's extensions. When a retry toggles an extension (e.g. the
+// speculative archive add/remove) the routerKey changes, so a provider that just
+// failed under the old key would not be excluded under the new one and could be
+// re-selected for the retry (MAG-2228). Calling this on a toggle keeps an
+// already-tried provider excluded across the change.
+func (up *UsedProviders) MigrateUnwantedProviders(src, dst RouterKey) {
+	if up == nil {
+		return
+	}
+	if src.String() == dst.String() {
+		return
+	}
+	up.lock.Lock()
+	defer up.lock.Unlock()
+	srcProviders := up.createOrUseUniqueUsedProvidersForKey(src)
+	dstProviders := up.createOrUseUniqueUsedProvidersForKey(dst)
+	for provider := range srcProviders.providers {
+		dstProviders.unwantedProviders[provider] = struct{}{}
+	}
+	for provider := range srcProviders.unwantedProviders {
+		dstProviders.unwantedProviders[provider] = struct{}{}
+	}
+}
--- a/protocol/lavasession/used_providers_mag2228_test.go
+++ b/protocol/lavasession/used_providers_mag2228_test.go
@@ -1,0 +1,48 @@
+package lavasession
+
+import "testing"
+
+// TestMigrateUnwantedProviders covers MAG-2228 Mechanism B: a provider already tried
+// under one routerKey (e.g. the no-extension lane) must stay excluded after a retry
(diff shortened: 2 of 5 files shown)


avitenzer (approved):
Approving — both fixes are correct and appropriately surgical. Verified locally: `go vet` clean on lavasession/relaycore/rpcsmartrouter, and both new unit tests pass.

**Confirmed correct:**
- **Fix 1** (`resolvePinDirectives`, `BatchNumber()==0`): timing is sound — `AddUsed` is deferred inside `GetSessions` and the pin is resolved before it, so the read reflects only prior batches. The `PairingListEmpty` re-honor case holds (`batchNumber` only increments on a successful dispatch). Pin directives are consumed only via `GetSessions`, so dropping them on retry is complete.
- **Fix 2a** (`MigrateUnwantedProviders`): keying is consistent — the migrate reconstructs keys extensions-only via `NewRouterKeyFromExtensions`, matching the selection-time exclusion lookup at `consumer_session_manager.go:970-971`. Lock usage is safe (use-when-locked helper, no re-entrant deadlock).

**Non-blocking notes:**
1. `MigrateUnwantedProviders` writing into a freshly-created `dst` key mutates the shared `originalUnwantedProviders` (per-key `unwantedProviders` aliases it in `createOrUseUniqueUsedProvidersForKey`). Verified benign: `UsedProviders` is created per-request (`sendRelayWithRetries`, `ProcessRelaySend`), so pollution is scoped to one relay's retries, and it's a pre-existing property of the `AddUnwantedAddresses` path.
2. Coverage is unit + manual-cluster, not integration — no automated test that `stateTransition` actually fires the migrate at the toggle point or that failover happens end-to-end. The cluster table is reasonable evidence; a state-machine-level test would be a nice follow-up.
3. The documented "2b" limitation is acceptable (no-op for the reported `eth_blockNumber` class).
4. Behavioral contract change: `lava-select-provider` is now effectively first-attempt-only — worth a release-note callout.

PR #198, opened 2026-07-06
feat: add Union testnet spec + gRPC example config

nimrod-teich:
## Summary

- Add a **UNION mainnet** spec (chain-id `union-1`) as the base, carrying the tokenfactory api catalog.
- Add a **UNIONT testnet** spec that inherits from it (`imports: ["UNION"]`) and overrides only the chain-id verification (`union-testnet-8`) — same mainnet/testnet pattern as `COSMOSHUB`/`COSMOSHUBT`.
- Add a **gRPC example config** (`smartrouter_union_grpc.yml`) that serves Union over gRPC via the Lava public gateway `union.grpc.lava.build:443`.

## Why the split

The only live Lava gateway endpoint, `union.grpc.lava.build:443`, is **network `union-1` (Union mainnet, `uniond v1.2.3`)** — not `union-testnet-8` as the original single-spec version declared. Making `UNION` the base and having `UNIONT` inherit keeps one api catalog while letting each network verify its own chain-id.

## Verified end-to-end (locally)

```
make build
./build/smartrouter config/smartrouter_examples/smartrouter_union_grpc.yml --use-static-spec specs/
grpcurl -plaintext -d '{}' localhost:3361 cosmos.staking.v1beta1.Query/Params
```

- ✅ Router builds, boots, and listens on `:3361` ("ready for requests")
- ✅ **Serves live gRPC** — `cosmos.staking.v1beta1.Query/Params` returns real `union-1` data (`bondDenom: "au"`) through the router
- ✅ Reflection lists the full Cosmos SDK / IBC / CosmWasm / feemarket service set

## Known spec gaps (documented, not blocking gRPC serving)

These are spec-completeness issues against this particular public gateway, not gRPC-path defects:

1. **No `GET_BLOCKNUM` parse-directive** in the grpc collection → chain-tracker logs `GET_BLOCKNUM missing function template` on startup. Serving still works; block-height tracking doesn't.
2. **Gateway doesn't serve tokenfactory** — it serves the standard Cosmos SDK set. The spec's `tokenfactory.v1beta1.Query/*` methods aren't resolvable via reflection on this upstream, so the reflection-based startup verify is best-effort (`tx-indexing` skip-verified in the example). A Union node that serves tokenfactory would verify those cleanly.

## Test plan

```
make build
smartrouter health config/smartrouter_examples/smartrouter_union_grpc.yml --use-static-spec specs/ 2>/dev/null | jq
# serve + live query
smartrouter config/smartrouter_examples/smartrouter_union_grpc.yml --use-static-spec specs/
grpcurl -plaintext -d '{}' localhost:3361 cosmos.staking.v1beta1.Query/Params
```

### DIFF
--- a/config/smartrouter_examples/smartrouter_union_grpc.yml
+++ b/config/smartrouter_examples/smartrouter_union_grpc.yml
@@ -1,59 +1,41 @@
-# Smart Router gRPC Configuration — Union Testnet (UNIONT)
-# Mode: Direct connections to Union testnet endpoints on the Lava public gateway
+# Smart Router gRPC Configuration — Union Mainnet (UNION)
+# Mode: Direct gRPC connection to Union mainnet on the Lava public gateway
 #
-# Union is a Cosmos SDK chain (imports COSMOSSDK) and exposes three interfaces:
-#   - REST API          (/tokenfactory endpoints)
-#   - gRPC API          (tokenfactory.v1beta1.Query methods)
-#   - Tendermint RPC    (consensus / block queries, event subscriptions)
+# Union is a Cosmos SDK chain (UNION imports COSMOSSDK). This example serves the
+# gRPC interface, pointed at the real Lava-hosted endpoint:
+#   union.grpc.lava.build:443  (network: union-1, uniond v1.2.3)
 #
-# This example focuses on the gRPC leg, pointed at the real Lava-hosted
-# endpoint (union.grpc.lava.build:443). The REST and Tendermint legs follow
-# the same lava.build gateway naming convention.
-#
-# NOTE: Union is a testnet. A single upstream is enough to demonstrate serving
-# gRPC; add a second, DISTINCT source per interface (different group-label) if
-# you want cross-validation / group-diversity policies to have a quorum to span.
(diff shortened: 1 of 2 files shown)


nimrod-teich:
Closing after parity-checking against the existing `smartrouter_cosmos.yml` gRPC example.

Ran identical checks on both configs:

| Check | smartrouter_cosmos.yml | smartrouter_union_grpc.yml |
|-------|------------------------|----------------------------|
| Endpoint reflection `list` | ✅ | ✅ |
| `health` probe | ❌ times out 30s | ❌ times out 30s |
| Router boots + listens | ✅ | ✅ |
| Serves live gRPC query | ✅ (`uatom`) | ✅ (`au`) |

The `health` probe timing out against these public gateways is **pre-existing behavior** — the already-merged cosmos example fails its health probe the same way (its own header comments note the gRPC startup verify is best-effort against rate-limited public endpoints). Both configs serve live gRPC queries through the router fine.

Since Union's gRPC behaves identically to the existing cosmos example and adds no new capability, closing rather than merging. Verified takeaway stands: **smart-router serves gRPC upstreams like `union.grpc.lava.build` / `lava.grpc.lava.build`** end-to-end.

PR #209, opened 2026-07-09
fix(chaintracker-redesign): address code-review findings (9 correctness fixes)

avitenzer:
## Summary

Follow-up to #143. A multi-agent code review of the chaintracker/probing redesign surfaced 9 correctness/robustness issues that were **not** included in the squash-merged PR #143 — verified by inspecting the merged code on `main` (all 9 locations carry the original pre-fix behavior). This PR applies all 9, each with a focused regression test.

### 🔴 Critical — tip poisoning (shared root cause: the anti-lie guard only existed when a consensus baseline did, which no 1–2 endpoint pod ever forms)
- **#1 `chainstate`** — `SetLatestBlock` now anchors the anti-lie guard on the fresh tip when there is no baseline, and re-adopts a fresh lower block once a poisoned tip goes TTL-stale (self-heal). Bounds a poisoning of the tip to ~one TTL instead of the process lifetime.
- **#2 `rpcsmartrouter`** — cache finalization reads the strict consensus baseline (`getLatestBlockForCacheFinalization`), never the optimistic tip or bootstrap atomic — a too-high value there falsely finalizes still-mutable blocks into the long-TTL cache.
- **#3 `rpcsmartrouter`** — the router-wide bootstrap atomic ratchet in `harvestAndUpdateTipFromRelay` is gated on the ChainState verdict, not the per-endpoint store's time-monotonic guard, so a lying-high harvest can't poison archive routing.

### 🟠 High
- **#4** — feed `effectiveBlockTime` (not the raw spec value, which may be 0) to the probe `VerdictConfig` and consensus loop, so healthy endpoints aren't scored unavailable on chains without `average_block_time`.
- **#5** — re-register `--chain-tracker-polling-multiplier` as a deprecated no-op so existing deployments don't fail process start on upgrade.

### 🟡 Medium
- **#6** — sync-gap scoring falls back to the guarded observed tip when no majority exists (`endpointSyncGap` / `syncGapAgainstReference`), restoring lag demotion on 2-endpoint pods. (Reporting/reputation QoS only — the optimizer's selection path keeps the stricter F5 omit-on-no-majority contract, unchanged.)
- **#7** — consistency pre-validation reads the freshest of the poll atomic and the endpointtip store (`freshestEndpointTip`), so a traffic-gate-frozen atomic no longer false-rejects a live relay-fed endpoint.
- **#8** — a traffic-gate skip is reported (`skipped bool`) and PRESERVES the poll-failure backoff (`nextFetchFails`) instead of resetting it, so a broken poll path backs off and its recovery evidence accrues.
- **#9** — the probe's lag tolerance is sourced from the per-chain `EndpointLagThreshold` (`probeVerdictConfigFor`), matching consistency pre-validation instead of a hardcoded 10.

## Testing

Each fix extracts its decision logic into a small pure helper unit-tested against the specific failure scenario from the review. Touched packages green on top of `main`:

- `protocol/chainstate` ✅
- `protocol/chaintracker` ✅
- `protocol/probing` ✅
- `protocol/rpcsmartrouter` ✅

`go build ./...` clean.

---

### 🟡 Additional (10th finding, from follow-up review) — reset must clear the endpointtip store
- **`endpointstate`** — `ResetAllLatestBlocks` zeroed only each tracker's poll atomic, leaving the shared endpointtip store populated with the pre-reset block. Consistency pre-validation reads the freshest of the two sources (`freshestEndpointTip`, #7), so after a reset it resurfaced the stale value via `max()` and kept gating against the block the reset asked to discard — violating `ResetLatestBlock`'s documented "consistency sees <= 0 and skips" contract. Now removes the store entry in lockstep with the atomic reset (`endpointtip.Remove`), so both sources read 0 together. Regression test asserts the store reads 0 after a reset (fails without the fix); package is race-clean.

_Note: this gap is orthogonal to #7 and pre-existed on `main` — #7 only changed the both-positive (traffic-gate-frozen) case; the reset case returns the same stale value main's original fallback did. Fixing it at the reset (not the read) is correct because a timestamp-aware read would still lose to the store's real timestamp (reset zeroes the atomic's changeTime)._

### DIFF
--- a/protocol/chainstate/chain_state.go
+++ b/protocol/chainstate/chain_state.go
@@ -140,7 +140,7 @@ type ChainState struct {
 	now func() time.Time
 
 	mu          sync.RWMutex
-	latestBlock int64 // observed tip; only SetLatestBlock raises it, only Recompute lowers it
+	latestBlock int64 // observed tip; raised by SetLatestBlock, lowered only by Recompute's realign or SetLatestBlock's stale-tip re-adoption
 	// lastObservedAt is the wall-clock of the last ACCEPTED observation of the current tip,
 	// refreshed even when the block is unchanged (an equal-block confirmation re-proves the tip
 	// is live). TTL freshness is measured from this, so a stable-but-confirmed tip does not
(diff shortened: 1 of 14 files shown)


nimrod-teich:
@avitenzer Why all of these merge commits? 
giphy

nimrod-teich:
Please rebase and get rid of them.

nimrod-teich (approved):

PR #213, opened 2026-07-15
fix(relaycore): single provider address on all-transport-errors failure result (MAG-2351)

avitenzer:
## Summary

[MAG-2351](https://magmadevs.atlassian.net/browse/MAG-2351): `Lava-Provider-Address` listed roughly **2× the provider names** (with repeats) whenever every attempt failed at the transport level, disagreeing with `Lava-Retries`. Two linked causes, both in `relaycore/relay_processor.go`, fixed here:

1. **Root cause**: `buildFailureResult` packed the comma-joined unwanted-provider list into `ProviderInfo.ProviderAddress`, a field that holds a single provider everywhere else. `appendHeadersToRelayResult` then appended that whole blob as its resolver entry after the per-attempt names. The result now keeps the **single** provider whose error body is returned — the best node/protocol error's own `RelayResult` — restoring the header contract on this path: entry count == `Lava-Retries` + 1, last entry == response source. The attempted list is no longer written here; it stays recoverable from the `relayProcessor` dump logged on the `SendParsedRelay` error path. The header path itself needed **no changes**.
2. **Exposed by the new test — same file**: `processStatefulResult` called `buildFailureResult` with a hardcoded `protocolErrorCount` of `0`, so a Stateful relay in which every attempt transport-failed returned a bare 500 with no error message and no provider attribution. Stateful and Stateless result selection were otherwise byte-identical, so they collapse into `processNonCrossValidationResult`, which passes the real count for both. Stateful all-transport-failure responses now carry the responsible provider's error text, as Stateless already did (maskable via `ReturnMaskedErrors`).

A third fix originally in this PR — the `lavasession/used_providers.go` blocklist map-aliasing / dedup change — was split out in `0d8fedb8` and is tracked separately as [MAG-2442](https://magmadevs.atlassian.net/browse/MAG-2442). It's a real but independent latent bug (each new router key aliases the user's original blocklist map, mutating it across keys); once the joined list is gone from the header it is no longer load-bearing for MAG-2351, so it is not touched here.

## Tests

- `TestRelayProcessorExhaustedTransportErrors` (relaycore) — drives the real `ProcessingResult()` with all-protocol-errors under **both** Stateless and Stateful; asserts a single real provider address on the failure result. Fails before this fix (and caught cause 2).
- `TestAppendHeadersToRelayResult_ExhaustedTransportErrors` (rpcsmartrouter) — header contract on the exhausted path: `Lava-Retries: 2` with exactly `provider1,provider2,provider3`, resolver last.
- Full suites for `relaycore` and `rpcsmartrouter` pass; `go vet` + `gofmt` clean.

[MAG-2351]: https://magmadevs.atlassian.net/browse/MAG-2351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAG-2442]: https://magmadevs.atlassian.net/browse/MAG-2442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


avitenzer:
## Max-effort review response (commit `0d8fedb`)

A deep multi-agent review surfaced correctness issues my original commit introduced. The follow-up commit trims the PR to the minimal correct header fix.

**Kept (the actual fix):** `buildFailureResult` no longer packs the comma-joined attempted list into the single-provider `ProviderInfo.ProviderAddress` — that's what fixes the ~2× `Lava-Provider-Address` on the exhausted path.

**Removed / corrected:**
- **`attemptedProviders` log attribute** — `LavaFormatError` serializes attributes into the *returned* error string, which reaches client bodies when `ReturnMaskedErrors` is off; and it used the wrong set (`AllUnwantedAddresses` includes never-contacted blocklist providers). Removed rather than "corrected" since no accurate attempted set exists at that point.
- **Two `ProviderInfo` fallback stanzas** — provable no-ops (the best node/protocol error's `Response.RelayResult` already carries its single provider; verified — the exhausted-path tests pass without them), and they introduced an unlocked write into shared results-manager state.
- **Eager `AllUnwantedAddresses()`** in `ProcessingResult` — no consumer remained.
- `processStatefulResult` / `processStatelessResult` were byte-identical after the fix → collapsed into `processNonCrossValidationResult`.

**One deliberate behavior change — flagging for sign-off:** the merged function passes the real `protocolErrorCount` for **Stateful** too, so an all-transport-errors Stateful failure now names the single responsible provider. On `main` that path returned an empty provider → `Lava-Provider-Address: Cached` (misleading — the same dishonest-header class this ticket fixes). Consequence: Stateful failure bodies now carry the provider's error text as Stateless already does (maskable via `ReturnMaskedErrors`). If the team wants Stateful errors anonymized, that's a separate policy — call it here.

**Split out:** the lavasession blocklist map-aliasing / dedup change moved to **MAG-2442** — a real but separate latent bug, no longer load-bearing here once the joined list is gone from the header, and it interacts with in-flight MAG-2392 work.

Tests: `relaycore` / `lavasession` / `rpcsmartrouter` suites pass; `go vet` + `gofmt` clean.

AnnaR-prog (commented):
Reviewed the diff and verified the key claims against the sources on this branch and `main`. **The code is correct and I'd approve it as-is — but the PR description is stale in a way that matters**, so flagging that before merge.

## What I verified

- The `Lava-Provider-Address` chain builder in `rpcsmartrouter_server.go` (unchanged here) walks protocol errors → node errors → successes, skips the resolver, and appends it last. With `buildFailureResult` now returning the best error's own `RelayResult` (single address), the contract — one entry per attempt, resolver last, count == `Lava-Retries` + 1 — holds structurally. "The header path itself needed no changes" is accurate.
- The final metadata append is guarded (`relayResult.Reply` is synthesized when nil), so returning a protocol-error `RelayResult` with a nil `Reply` is safe.
- Bonus fix worth noting: `returnedResult` aliases the results manager's *stored* error entry, so the old `strings.Join` write was also mutating that stored entry — polluting the very slice the header builder walks (this is what produced the intra-blob repeats). Removing the write fixes both ends.
- The `processStatefulResult`/`processStatelessResult` merge is a genuine improvement — the two were identical except for the hardcoded `protocolErrorCount: 0`, which is exactly the kind of divergence the copy-paste already caused once. Good catch by the new test that drives both selections through the real `ProcessingResult()`.

## Findings

**1. The PR body describes a third fix that is no longer in the diff — and the underlying bug is real and still live on `main`.**
The body claims a `lavasession/used_providers.go` fix (per-key copy of the original blocklist + `AllUnwantedAddresses` dedup) and a `TestOriginalBlocklistIsCopiedPerRouterKey` test. Commit `0d8fedb8` dropped these to keep the PR scoped, but the body wasn't updated. Meanwhile on `main`, `createOrUseUniqueUsedProvidersForKey` (`protocol/lavasession/used_providers.go:176`) still assigns `unwantedProviders: up.originalUnwantedProviders` — the same map reference — and `AddUnwantedAddresses` writes into it, so a transport failure under one extension router key mutates the request's original user blocklist and every later-created router key shares the polluted map (request-scoped over-blocking). Ironically `NewUsedProviders` ten lines up already does the copy, with a comment explaining why. Please update the body to the reduced scope and link the follow-up ticket so this doesn't get closed as fixed under MAG-2351.

**2. The body's "attempted list moved into the failure log (`attemptedProviders`)" claim is also stale** — there is no such log attribute in the final code; the failure log carries only the GUID. In practice the info is still recoverable (the `SendParsedRelay` error path logs the whole `relayProcessor`, whose dump includes the unwanted/used lists), so nothing is lost, but the description shouldn't claim a field that doesn't exist.

**3. Coordination with #212:** both PRs insert a new test at the same anchor in `rpcsmartrouter_server_test.go` (right after `TestAppendHeadersToRelayResult_DisagreeingOnQuorumFailure`), so whichever lands second gets a trivial rebase conflict there. `relay_processor.go` hunks don't overlap and will auto-merge.

## Client-visible behavior (all strictly toward correctness)

- Exhausted-transport-errors responses now honor the documented header contract; anyone parsing the old doubled header was already getting data that disagreed with `Lava-Retries`, so breakage risk is negligible.
- Stateful all-transport-failure responses now carry a real error body and provider attribution instead of a bare 500.
- Small information-hygiene win: the header no longer leaks user-blocklisted providers that were never contacted.

**Verdict: code LGTM. Requesting a PR-body update + follow-up ticket for the `used_providers.go` aliasing bug before merge.**

AnnaR-prog:
Follow-up on finding 1 in my review: the dropped `lavasession` blocklist-aliasing fix is already tracked as **[MAG-2442](https://magmadevs.atlassian.net/browse/MAG-2442)** — "per-router-key unwanted set aliases the user blocklist map, mutating it across keys" (Bug, split out of this PR during review). So the "filed separately" note in commit `0d8fedb8` checks out; nothing to file. The only remaining nit is that this PR's description still lists that fix (and `TestOriginalBlocklistIsCopiedPerRouterKey`) as present — worth trimming the body to the two fixes actually in the diff and linking MAG-2442.

[MAG-2442]: https://magmadevs.atlassian.net/browse/MAG-2442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

avitenzer:
Thanks Anna — trimmed the body to the two `relay_processor.go` fixes and linked MAG-2442 for the split-out `used_providers.go` aliasing bug (agreed, nothing to file — it's already the tracker). Also fixed the stale `attemptedProviders` line. On #212: yep, both anchor a new test after `TestAppendHeadersToRelayResult_DisagreeingOnQuorumFailure` — whichever lands second takes the trivial rebase; the `relay_processor.go` hunks don't overlap.

AnnaR-prog (approved):

PR #214, opened 2026-07-15
fix: serve stale data when all providers fail consistency validation (MAG-2392)

avitenzer:
## What changed

- keep consistency-only pre-dispatch retries inside a single relay task so the router can inspect the full candidate set
- when normal selection is exhausted, reopen only providers rejected by consistency validation and allow one stale fallback pass
- preserve fresh-provider preference and keep providers excluded for unrelated failures out of the fallback
- release pre-dispatch sessions and CU reservations without recording false QoS/provider failures
- keep cross-validation strict; it does not use the stale fallback
- add regression coverage for mixed fleets, all-stale fleets, one-shot activation, provider exclusion, and session accounting

## Why

MAG-2392 occurs when the consumer's seen block is ahead of every available endpoint. Pre-request consistency validation rejects each selected endpoint, marks it unwanted, and eventually returns `No pairings available`, causing an HTTP 500 even though the endpoints can still serve data.

The new request-local exhaustion fallback continues preferring a fresh endpoint, but serves from a previously consistency-rejected endpoint once the router has established that normal selection has no candidates left.

## Impact

Requests no longer fail solely because every available provider is behind the consumer's seen block. The fallback is one-shot and limited to the exact consistency-rejected provider set, preventing retry loops or accidental reopening of transport-failed providers.

## Validation

- `go test ./protocol/rpcsmartrouter -count=1`
- `go test ./protocol/lavasession -count=1`
- targeted race-detector tests for both changed packages
- `go vet ./protocol/rpcsmartrouter ./protocol/lavasession`
- `git diff --check`

## Issue

[MAG-2392](https://magmadevs.atlassian.net/browse/MAG-2392)

## Deployment verification

Re-run the production simulator's all-stale Solana consistency scenario after deploying an image containing this commit.

[MAG-2392]: https://magmadevs.atlassian.net/browse/MAG-2392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -1928,6 +1928,24 @@ func (csm *ConsumerSessionManager) blockProvider(ctx context.Context, address st
 	return nil
 }
 
+// OnSessionDiscarded releases a session that was selected but intentionally
+// dropped before any relay was dispatched. It returns the reserved compute
+// units and unlocks the session without recording a QoS failure or adding a
+// consecutive provider error: no upstream request was made, so availability
+// was not actually tested.
+func (csm *ConsumerSessionManager) OnSessionDiscarded(consumerSession *SingleConsumerSession, reason error) error {
+	if err := consumerSession.VerifyLock(); err != nil {
+		return fmt.Errorf("OnSessionDiscarded, consumerSession.lock must be locked before accessing this method: %w", err)
+	}
+
+	cuToDecrease := consumerSession.LatestRelayCu
+	parentConsumerSessionsWithProvider := consumerSession.Parent
+	consumerSession.LatestRelayCu = 0
+	consumerSession.Free(reason)
+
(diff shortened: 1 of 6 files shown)


avitenzer:
Heads-up for review — interaction with a newly-filed bug: this PR's `RemoveUnwantedAddresses` deletes from every router key's `unwantedProviders` map, which today all alias one shared `originalUnwantedProviders` reference. That aliasing is a separate latent bug, **MAG-2442** (split out of #213). Its copy-fix will make each key hold its own map, changing `RemoveUnwantedAddresses` from mutating one shared map to per-key deletes. Functionally fine for the one-shot fallback here, but the two should be reconciled when MAG-2442 lands so the fallback keeps reopening exactly the intended set.

AnnaR-prog (commented):
—

AnnaR-prog (approved):

PR #218, opened 2026-07-16
refactor(spec): remove 15 unused spec fields from the Go model (PR 1/3)

avitenzer:
## What

Removes 15 spec fields that have no runtime consumer in smart-router, plus the dead code that existed only to support them. **No `specs/*.json` are modified** — the bundled mirror is synchronised in a later PR.

### Removed from `types/spec`
- **Governance leftovers**: `min_stake_provider`, `providers_types`, `contributor`, `contributor_percentage`, `shares`, `identity`, `block_last_updated`, `reliability_threshold`, `data_reliability_enabled` — plus the `Coin` / `FlexFloat64` custom unmarshalers and the `Spec_ProvidersTypes` enum that existed only for them.
- **Proposal envelope metadata**: `proposal.title`, `proposal.description`, `deposit`. The `{proposal:{specs}}` envelope is retained (both loaders read `proposal.specs`).
- **Api-level**: `extra_compute_units` (de-wired from JSON-RPC/Tendermint batch CU summation, `Api.Equal`, and the default-API initializer), `category.local`, `category.subscription`. Subscription behaviour stays driven by `FUNCTION_TAG_SUBSCRIBE` parse directives.

## Why it's safe

Go's `encoding/json` silently ignores unknown keys, so legacy spec JSON — and the current `specs/*.json`, which still carry these keys — keeps loading unchanged. No production code reads the removed fields, and nothing serialises specs, so dropping non-`omitempty` `min_stake_provider` changes no output. `extra_compute_units` is `0` in every spec, so removing it from batch CU summation and `Api.Equal` is inert.

## Commits

1. `test(spec)` — backward-compat coverage added **before** any removal: a retained legacy fixture carrying every removed key in its rich shape, decoded through both the `keeper` and `specfetcher` loader paths; a bundled-spec load+expand guard; batch-CU-sum assertions. Written green against the pre-removal code, so they lock in behaviour rather than describe it.
2. `refactor(spec)` — the removal + coupled test updates (tests now assert active fields; legacy JSON bodies kept as fixtures).

## Verification

- `go build ./...` clean; **full suite: 34 packages OK, 0 FAIL**.
- **Live boot**: the built binary boots against real ETH1 upstreams, loads/expands the bundled specs, `/lava/health` → 200, `eth_blockNumber` returns a live block, `eth_chainId` → `0x1`.

## Sequence — PR 1 of 3

1. **(this)** smart-router Go cleanup + compat tests — bundled specs untouched.
2. lava-specs: remove the keys from the catalog + update the spec skills.
3. smart-router: sync the bundled `specs/` mirror from the cleaned lava-specs revision.

PR 1 lands first so a deployed router carries the tested compatibility guarantee before the data is mutated.


utils/keeper/spec_aggregation_test.go
```diff
@@ -471,13 +470,10 @@ func TestSpecAggregation_FullReplacement(t *testing.T) {
 	require.False(t, finalSpec.Enabled)
 
 	// min_stake_provider should be from source2 (lower amount)
```
AnnaR-prog: These three comments (472, 474, 476) now describe assertions that were deleted in this PR — `min_stake_provider`, `reliability_threshold`, and `shares` are no longer checked here, so "COMPLETELY from source2" is actually verified only for `Name`, `Enabled`, `average_block_time`, and the `api_collections` below. Suggest deleting the three orphaned comments so the test doesn't imply coverage it no longer has.

protocol/chainlib/jsonRPC_test.go
```diff
@@ -349,6 +348,21 @@ func TestJsonRpcBatchCall(t *testing.T) {
...
+	for _, single := range []string{
+		`{"jsonrpc":"2.0","id":1,"method":"eth_chainId","params":[]}`,
+		`{"jsonrpc":"2.0","id":2,"method":"eth_accounts","params":[]}`,
+		`{"jsonrpc":"2.0","id":3,"method":"eth_blockNumber","params":[]}`,
+	} {
+		singleMsg, errSingle := chainParser.ParseMsg("", []byte(single), http.MethodPost, nil, extensionslib.ExtensionInfo{LatestBlock: 0})
+		require.NoError(t, errSingle)
+		sumCU += singleMsg.GetApi().ComputeUnits
+	}
+	require.Equal(t, sumCU, chainMessage.GetApi().ComputeUnits, "batch CU must equal the sum of member compute_units")
```
AnnaR-prog, line 365: Good regression guard. One robustness nit: `sumCU` is re-derived from the same parser and `compute_units` is `omitempty`, so if a future spec omits it for these methods `sumCU` becomes 0 and this assertion degrades to `0 == 0` — passing while proving nothing. A one-line `require.Greater(t, sumCU, uint64(0))` (or asserting the concrete expected total) keeps it honest. Same pattern in `tendermintRPC_test.go:250` and `:334`. Non-blocking.

protocol/chainlib/tendermintRPC_test.go
```diff
@@ -235,6 +234,21 @@ func TestTendermintRpcBatchCall(t *testing.T) {
...
+	sumCU := uint64(0)
+	for _, single := range []string{
+		`{"jsonrpc":"2.0","id":1,"method":"block","params":{"height":"99"}}`,
+		`{"jsonrpc":"2.0","id":2,"method":"block","params":{"height":"100"}}`,
+	} {
+		singleMsg, errSingle := chainParser.ParseMsg("", []byte(single), "", nil, extensionslib.ExtensionInfo{LatestBlock: 0})
+		require.NoError(t, errSingle)
+		sumCU += singleMsg.GetApi().ComputeUnits
+	}
+	require.Equal(t, sumCU, chainMessage.GetApi().ComputeUnits, "batch CU must equal the sum of member compute_units")
```
AnnaR-prog, line 251: Same vacuous-assertion nit as `jsonRPC_test.go:364`: if `block`'s `compute_units` were ever omitted, `sumCU` is 0 and this becomes `0 == 0`. Add `require.Greater(t, sumCU, uint64(0))` to guard it. Non-blocking.

protocol/chainlib/tendermintRPC_test.go
```diff
@@ -304,6 +318,21 @@ func TestTendermintRpcBatchCallWithSameID(t *testing.T) {
...
```
AnnaR-prog, line 336: Same as `:250` and `jsonRPC_test.go:364` — add `require.Greater(t, sumCU, uint64(0))` so the batch-CU check can't silently go vacuous if `compute_units` is ever 0 or omitted. Non-blocking.

AnnaR-prog (commented):
Reviewed by building the branch in a clean worktree: `go build ./...` and a full module-wide test-compile (`go test ./... -run '^$'`) are both clean, and the types/spec, utils/keeper, utils/specfetcher, and chainlib tests pass.

Verified the safety claim independently rather than from the description: no production reader of any removed field/getter, nothing marshals `Spec` (so dropping non-`omitempty` `min_stake_provider` changes no output), `extra_compute_units` is `0` across all bundled specs **and** was never consumed (the batch path summed it into a separate field that had no reader; the `ComputeUnits` sum line is unchanged), and subscription routing keys off the `FUNCTION_TAG_SUBSCRIBE` parse directive, not the removed `category.subscription`. Cross-validation is untouched — `data_reliability_enabled`/`reliability_threshold` have no readers and CV is header-driven.

No correctness issues found. Two minor test-cleanliness notes inline. Nice test-first sequencing.

AnnaR-prog (approved):

PR #222, opened 2026-07-20
fix(rpcsmartrouter): wire /debug/time-warp into ChainState TTL/staleness (MAG-2307)

AnnaR-prog:
## What & why

`POST /debug/time-warp` is meant to age out stale router state without waiting real time. It moved the provider-optimizer clock but **not** the per-chain `ChainState` clock, so ChainState's TTL / staleness / consensus windows could only be exercised by waiting *real* time — 130 s on ETH (10× average block time), ~4 s on Solana. The warp never reached ChainState because it was built with a hard-wired `time.Now` (`chainstate.New` → `NewWithClock(..., time.Now)`).

Fixes **[MAG-2307](https://magmadevs.atlassian.net/browse/MAG-2307)**. Per the reporter's own investigation this is a new debug-surface gap, not a regression.

## Approach

- **ChainState gets a debug-only warp offset** (`warpOffset atomic.Int64`) added on top of the base clock in a new `effectiveNow()`. The four clock reads (`SetLatestBlock`, `GetLatestBlock`, `GetConsensusBaseline` TTL, `Recompute` staleness) now go through `effectiveNow()`.
- **The `/debug/time-warp` and `/debug/reset-all` handlers set the offset per chain** (`setAllChainStateDebugOffset`), mirroring how they already set/clear each optimizer's `NowFunc`. `/debug/reset-scores` deliberately does **not** touch it — it documents itself as leaving the clock alone.
- **Read-clock-only, by design.** Only ChainState's own clock is warped; the external endpoint-observation write stamps (relay/poll) stay on real time. That's exactly what ages observations out under a forward warp when endpoints are quiesced (the age-out use case), without perturbing the live relay-gate freshness path.

## Production is inert

`warpOffset` defaults to 0, so `effectiveNow() == now()` on every relay path — byte-for-byte unchanged behavior. Only the debug handler ever stores a non-zero offset. Concurrency is safe by construction (atomic offset; the base `now` func is still never reassigned).

## ⚠️ Note on `observed_tip` (avoids a ticket bounce)

The ticket flagged both `has_baseline` staying `true` **and** `observed_tip` unchanged. This fix clears the former, but **raw `observed_tip` / `has_baseline` in `/debug/chain-state` are intentionally the non-TTL-gated view** (so black-box tests can still assert raw transitions). Warp-driven expiry is exposed through two **new computed fields, `TipFresh` / `BaselineFresh`**, which flip immediately on a warp (the raw fields only change on the next `Recompute`). **Assert `TipFresh` / `BaselineFresh`** (or poll `has_baseline` against the live router) to observe the warp taking effect.

## Testing

- `chainstate` unit test: a warp offset ages TTL exactly like real time; pins the raw-vs-gated snapshot split; resetting the offset restores freshness.
- rpcsmartrouter HTTP end-to-end test: `POST /debug/time-warp {offset_seconds:200}` → `/debug/chain-state` `TipFresh`/`BaselineFresh` flip to `false` → `offset_seconds:0` restores them.
- `go build ./...`, full `chainstate` + `rpcsmartrouter` suites, and `-race` on `chainstate` all pass.

[MAG-2307]: https://magmadevs.atlassian.net/browse/MAG-2307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/protocol/chainstate/chain_state.go
+++ b/protocol/chainstate/chain_state.go
@@ -20,6 +20,7 @@ package chainstate
 import (
 	"sort"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
@@ -135,9 +136,16 @@ func DefaultConfig(averageBlockTime time.Duration) Config {
 type ChainState struct {
 	chainID string
 	cfg     Config
-	// now is the clock; time.Now in production, overridable in tests for deterministic TTL /
+	// now is the BASE clock; time.Now in production, overridable in tests for deterministic TTL /
 	// staleness. Set once at construction and never reassigned, so it is read without the lock.
 	now func() time.Time
+	// warpOffset is a debug-only forward shift added on top of `now` (see effectiveNow). It exists
+	// so the /debug/time-warp HTTP hook can age this chain's TTL/staleness/consensus windows the
+	// same way it ages the provider-optimizer clock (MAG-2307) — without it the warp never reached
+	// ChainState and per-chain expiry could only be exercised by waiting real time. Atomic, so it is
+	// read lock-free in effectiveNow and written lock-free by SetDebugClockOffset. Zero in production
+	// (only the debug handler ever stores a non-zero value), so it is a no-op on the relay hot path.
+	warpOffset atomic.Int64 // nanoseconds
 
 	mu          sync.RWMutex
 	latestBlock int64 // observed tip; raised by SetLatestBlock, lowered only by Recompute's realign or SetLatestBlock's stale-tip re-adoption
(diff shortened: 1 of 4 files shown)


avitenzer (approved):

VicSheCodes:
Thanks for adding this, @AnnaR-prog Anna. We do want the capability: aging ChainState TTL over HTTP so black-box tests don't need to wait ~130 seconds. Two things we need before this can merge, though, not suggestions.

1. Give ChainState its own dedicated endpoint, separate from /debug/time-warp, that we can drive from our tests.
/debug/time-warp is the optimizer/QoS clock (the flag help says "shift QoS clock", and offset=0 calls opt.ResetState()). We run warp-then-reset-to-0 routinely to reset QoS state, so if the same knob also moves ChainState it's easy to corrupt ChainState by accident.
What we need from the endpoint: from a black-box test, set a forward offset on ChainState's staleness/TTL clock (offset_seconds relative to real time, 0 = back to real time), so a stale entry crosses its TTL in seconds instead of the real ~130s. It must be fully independent of /debug/time-warp, so a QoS reset never touches ChainState. The exact endpoint shape and name are yours to pick.

2. Fix the timestamp writes (applies to whichever clock ChainState uses).
SetLatestBlock and Recompute currently save timestamps with the warped clock, not real time.

```
warp forward:            fake time = 12:00
provider sends a block:  stored received_at = 12:00
reset the warp:          real time = 10:05
freshness check:         10:05 - 12:00 = negative age  (treated as fresh)
```

The router treats the data as received in the future and marks it fresh; stale data can stay fresh until real time catches up. Fix: store timestamps on the real clock; use the injected clock only when calculating whether data is stale. Stored timestamp uses real time; freshness check uses the injected time.

The relevant code is line 257 (`now := cs.effectiveNow()`), where the shared clock currently reaches the write path. Happy to talk through the endpoint design.

AnnaR-prog:
Follow-up in **#224** addresses this review — since #222 was already merged, the fixes went into a new PR against `main`.

@VicSheCodes — both required changes are in:

1. **Dedicated ChainState endpoint, decoupled from `/debug/time-warp`.** New `POST /debug/chain-state-time-warp` drives *only* ChainState's TTL/staleness clock (`offset_seconds` relative to real time, `0` = real; same finite / `>= 0` / `<= 24h` validation). `/debug/time-warp` and `/debug/reset-all` no longer touch ChainState, so a routine QoS warp-then-reset can't disturb it. ⚠️ **Breaking for any black-box test that used `/debug/time-warp` to age ChainState** — those need to switch to the new endpoint.
2. **Timestamp writes on the real clock.** `SetLatestBlock` / `Recompute` now *evaluate* freshness against the warped clock but *store* `lastObservedAt` / `baselineAt` / `baselineSince` on the real base clock, so a write under a warp no longer stamps a future time (no more `real - future = negative age` "fresh forever").

Copilot's three comments are also handled in #224: `chain_state.go:258` and `:377` are the same real-clock-write fix as (2); `:408` corrects the `ChainStateSnapshot` doc (`GetConsensusBaselineWithTime` returns `baselineSince`, not `baselineAt`).

PR #223, opened 2026-07-21
feat(smart-router/debug): add /debug/reset-probe-backoff + /debug/reset-chaintracker-rows (MAG-2395)

AnnaR-prog:
## What & why

Two pieces of per-endpoint ChainTracker state cannot be cleared by any debug endpoint today, so tests that reset the router and then measure poll timing or provider health start from a dirty state and can silently give a wrong result. This adds two focused reset endpoints. **reset-all and reset-scores are unchanged** (the ticket's explicit requirement).

Implements **[MAG-2395](https://magmadevs.atlassian.net/browse/MAG-2395)**.

### Architecture note
The ticket's pointers reference the pre-redesign `protocol/chaintracker/chain_tracker.go`; on current `main` the old ChainTracker is still live but owned **per-endpoint** by `endpointstate.EndpointMonitor` (one `ChainTracker` per URL). The two capabilities are mapped onto that current structure.

## `POST /debug/reset-probe-backoff`
Clears the per-endpoint failure backoff so each provider returns to base poll cadence. Nothing else.
- `fetchFails` is goroutine-local and the poll timer can be stretched up to `BACKOFF_MAX_TIME` (10m); neither is safely resettable from outside the poll goroutine. So `ChainTracker` gets a buffered `resetBackoffCh`; a new `select` case zeroes `fetchFails` and reschedules the next poll at base cadence. `ResetBackoff()` is a non-blocking send, so the handler never waits on the poll goroutine.
- **Observability:** `/debug/endpoint-state` had no interval/backoff field, so the ticket's acceptance ("poll interval visibly stretches … next poll on the BASE interval") wasn't observable. `updateTimer` now publishes the live interval to an atomic, exposed as **`PollIntervalMs`** — it goes base → stretched under failures and back to base after the reset.

## `POST /debug/reset-chaintracker-rows`
Re-registers every configured endpoint that lost its ChainTracker row. Nothing else.
- The epoch cleanup (`cleanupStaleTrackers`) can delete a briefly-unhealthy provider's row (MAG-2445); no reset recreates it, so it returns only on a ~15-minute rebuild or a pod restart.
- Re-registration reuses `GetOrCreateTracker` over `GetAllDirectRPCEndpoints()` — the **config/pairing** set startup registers from, **not** the health-filtered live set the cleanup deletes from — so a temporarily-disabled provider is restored while a genuinely config-removed one is correctly left out. `GetOrCreateTracker` is idempotent and non-blocking, so a still-down provider's row reappears immediately and the handler never blocks on init fetch.

## Relationship to MAG-2445
This is the **recovery tool**; MAG-2445 is the complementary root fix (stop deleting the rows). They ship independently.

## Testing
- **chaintracker:** interval-publish, non-blocking-coalesce, and an **end-to-end reset-returns-to-base-cadence** test (run 3× under `-race`, no flakiness).
- **endpointstate:** `ResetAllBackoff` dispatch + `BackoffSnapshot`.
- **rpcsmartrouter:** handler contracts (POST-only → 405, JSON shape).
- `go build ./...`, `rpcsmartrouter` full suite, `endpointstate` `-race`, and `chaintracker` all pass.
- Pre-existing/unrelated: `TestChainTrackerCallbacks/one_long_test` trips `-race` on clean `main` too (reproduced with these changes stashed) — not introduced here.

[MAG-2395]: https://magmadevs.atlassian.net/browse/MAG-2395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/protocol/chaintracker/chain_tracker.go
+++ b/protocol/chaintracker/chain_tracker.go
@@ -38,6 +38,15 @@ type IChainTracker interface {
 	// repopulates the value on its next successful fetch.
 	ResetLatestBlock()
 
+	// ResetBackoff clears the failure backoff so the next dedicated poll fires at base cadence
+	// (debug recovery for /debug/reset-probe-backoff, MAG-2395).
+	ResetBackoff()
+
+	// CurrentPollInterval returns the poll interval most recently scheduled — base cadence when
+	// healthy, exponentialBackoff-stretched when the endpoint has been failing. Debug telemetry
+	// only (MAG-2395).
+	CurrentPollInterval() time.Duration
+
 	// StartAndServe starts the chain tracker and serves gRPC if configured
 	StartAndServe(ctx context.Context) error
 
@@ -137,6 +146,17 @@ type ChainTracker struct {
 
 	// allows us to mock the chain fetcher for different use cases for example: Solana needs slot to block number
 	iChainFetcherWrapper IChainFetcherWrapper
+
+	// currentPollIntervalNanos publishes the interval updateTimer most recently scheduled: base
+	// cadence when healthy, stretched by exponentialBackoff when fetchFails > 0. fetchFails is
+	// goroutine-local, so this atomic is the only lock-free window into the live backoff state,
+	// surfaced by /debug/endpoint-state as PollIntervalMs (MAG-2395). Written only by the poll
+	// goroutine (updateTimer), read by any goroutine.
+	currentPollIntervalNanos atomic.Int64
+	// resetBackoffCh signals the poll goroutine to clear the failure backoff — zero fetchFails and
+	// reschedule the next poll at base cadence. Buffered(1) with a non-blocking send in
+	// ResetBackoff, so /debug/reset-probe-backoff never blocks on the poll goroutine (MAG-2395).
+	resetBackoffCh chan struct{}
 }
 
 func (cs *ChainTracker) IsDummy() bool {
@@ -583,6 +603,16 @@ func (cs *ChainTracker) start(ctx context.Context, pollingTime time.Duration) er
 				if enoughSamples {
 					blockGapTicker.Reset(pollingTime * 10)
 				}
+			case <-cs.resetBackoffCh:
+				// /debug/reset-probe-backoff (MAG-2395): clear the failure backoff and reschedule the
+				// next poll at base cadence. Cadence only — block data, observations, and the
+				// traffic-gate skip counter (relaySkipsSinceRealPoll) are untouched. Stop the old
+				// (possibly minutes-out) timer before updateTimer reassigns cs.timer so it does not leak.
+				fetchFails = 0
+				if cs.timer != nil {
+					cs.timer.Stop()
+				}
+				cs.updateTimer(pollingTime, fetchFails)
 			case <-ctx.Done():
 				cs.timer.Stop()
 				return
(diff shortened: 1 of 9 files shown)


avitenzer:
## Code Review

### Overview

Adds two focused debug reset endpoints: one clears per-endpoint ChainTracker failure backoff (via a channel signal into the poll goroutine, plus a new observable `PollIntervalMs` in `/debug/endpoint-state`), the other re-registers ChainTracker rows deleted by the epoch cleanup (the MAG-2445 recovery tool). Roughly half the diff is tests. `reset-all`/`reset-scores` untouched, as the ticket requires.

### Correctness — verified ✅

Each concurrency claim in the PR description was traced through the code:

- **The channel-based reset is sound.** `fetchFails` is goroutine-local to the poll loop (`chain_tracker.go:577`), so the `resetBackoffCh` select case is the only safe way to zero it. The case correctly stops the possibly-minutes-out old timer before `updateTimer` reassigns `cs.timer`, and `cs.timer` is only ever written by the constructor path (happens-before the goroutine spawn) and the poll goroutine itself — no data race. Passing the loop-local `pollingTime` also does the right thing for both the flat and legacy adaptive trackers.
- **`ResetBackoff` is genuinely non-blocking**, and the select-with-default even degrades safely to a no-op on a nil channel (struct-literal trackers in tests).
- **`currentPollIntervalNanos` is consistent.** One subtle path is covered correctly: after init, flat trackers create the first timer directly (`chain_tracker.go:558`) without going through `updateTimer` — but the constructor pre-stores `FlatPollInterval`, so the published value matches. The external test's immediate `require.Equal(base, ...)` after `StartAndServe` relies on exactly this, and it holds.
- **The handler really never blocks on a down endpoint.** `NewChainTracker` does no I/O, and `GetOrCreateTracker` registers synchronously and defers the init fetch to the `startTrackerWithRetry` goroutine (`endpoint_monitor.go:339`).
- **Lock ordering is safe.** The handlers take `router.mu` → `EndpointMonitor.mu`, the same order the existing `/debug/endpoint-state` handler and `resetAllChainTrackers` use, and nothing in `endpointstate` takes `router.mu`, so no inversion. Tracker callbacks (`onNewBlock` etc.) fire without `m.mu` held, so re-entrancy is fine too.

### Findings (all minor — debug-plane only)

1. **TOCTOU with the epoch cleanup** — `cleanupStaleTrackers` deliberately runs *outside* `router.mu`, so holding `router.mu` in `reregisterChainTrackerRows` does not exclude a concurrent cleanup pass; an in-flight cleanup that already built its keep-set can delete a row the handler just recreated. Acceptable for a recovery tool (the operator re-POSTs, and MAG-2445 is the root fix), but worth documenting so a future reader doesn't assume the lock provides that exclusion. → **Addressed in 2eee050** (comment added).
2. **`created` could be skewed or negative** — it was computed as a `GetEndpointCount()` before/after delta, which a concurrent `RemoveTracker` between the two reads could corrupt. → **Addressed in 2eee050**: creations are now counted against a snapshot of the pre-existing row URLs, which also prevents double-counting on a duplicate config URL.
3. **`PollIntervalMs: 0` is doing double duty** — for a per-endpoint row it can only mean "no tracker row exists" (the constructor pre-stores base cadence otherwise), which is actually a useful signal for diagnosing MAG-2445, but it's undocumented. One line in the field comment ("0 ⇒ no live tracker row") would make the two endpoints compose nicely for operators. Non-blocking.
4. **Naming nit:** "probe-backoff" resets the ChainTracker *poll* backoff, while "probe" elsewhere in this codebase means the health-probe machinery (`probeProviders`, `ConsecutiveHealthyProbes`). If the ticket fixed this name, keep it — otherwise `reset-poll-backoff` would be less confusable. Non-blocking.

### Test coverage — good

Layered sensibly: an internal unit test pins the atomic-publish math (`flat*8` matches `exponentialBackoff(flat, 3)` — verified), an external end-to-end test drives real backoff and recovery through the public API, monitor-level tests cover dispatch/snapshot with fakes, and handler tests pin the HTTP contract (405s, JSON shape). The only latent flake vector is the 40ms base cadence under a loaded CI runner, but the 3s `Eventually` windows give ~75× headroom.

### Conventions, performance, security — fine

- Handler style (manual `fmt.Fprintf` JSON, POST-only guards, nil-safe helpers mirroring `resetAllChainTrackers`), comment density, and MAG-ticket annotations all match the file's existing idiom.
- Steady-state cost is one atomic store per poll and one map build per `/debug/endpoint-state` call — negligible.
- Both endpoints are on the debug mux (registered only in debug mode, consistent with the existing reset endpoints), state-clearing only, no input parsing. No new security surface.

### Verdict

**Approve.** The hard part — mutating goroutine-local backoff state without locks or races — is done correctly and well-tested. The two blocking-ish items (1–2) are already fixed in `2eee050`; items 3–4 are optional polish.

avitenzer (approved):
Approving per the review above — concurrency design verified, the two substantive findings were fixed in 2eee050, and the remaining notes are non-blocking polish.

PR #224, opened 2026-07-21
fix(rpcsmartrouter): dedicated ChainState time-warp endpoint + real-clock timestamps (MAG-2307 review)

AnnaR-prog:
## What & why

Review follow-up to **PR #222** (MAG-2307, merged as `c011a80`) — addresses @VicSheCodes's two required changes and all three Copilot comments. Since #222 is already merged, this is a new PR against `main`.

### 1. Dedicated ChainState endpoint, decoupled from `/debug/time-warp` (VicSheCodes #1)
`/debug/time-warp` is the optimizer/QoS clock, and the black-box suite runs `warp → reset-to-0` routinely to reset QoS state. #222 coupled ChainState's warp to it, so a routine QoS reset would disturb ChainState.
- **New `POST /debug/chain-state-time-warp`** drives *only* ChainState's TTL/staleness clock (`offset_seconds` relative to real time, `0` = real; same finite / `>= 0` / `<= 24h` validation). Effect is observable via `TipFresh` / `BaselineFresh` in `/debug/chain-state`.
- `/debug/time-warp` and `/debug/reset-all` **no longer touch ChainState** — fully independent, so a QoS reset can never disturb it.

### 2. Store timestamps on the real clock (VicSheCodes #2 = Copilot `chain_state.go:258`, `:377`)
`SetLatestBlock` / `Recompute` stored `lastObservedAt` / `baselineAt` / `baselineSince` using the **warped** clock. A write while a warp is active stamped a *future* time → clearing the warp gave `real − future = negative age` → the entry read as fresh forever until real time caught up. Fix: **freshness is evaluated against the warped clock, but stored timestamps always use the real base clock** (`cs.now()`). (I'd flagged this skew as a "limitation" in #222 — the reviewers are right that it should be fixed.)

### 3. Doc fix (Copilot `chain_state.go:408`)
Corrected the `ChainStateSnapshot` comment: `GetConsensusBaselineWithTime` returns `baselineSince` (establishment time), not `baselineAt` (the internal TTL-freshness stamp, never exposed).

## Copilot assessment
All three comments are valid — `:258` / `:377` are the same real issue as VicSheCodes #2 (fixed), and `:408` is a genuine (pre-existing) doc inaccuracy (fixed).

## Testing
- `TestDebugChainStateTimeWarp_AgesChainState` — the new endpoint ages ChainState (`TipFresh`/`BaselineFresh` → false), reset restores them.
- `TestDebugTimeWarp_DoesNotTouchChainState` — `/debug/time-warp` (forward warp **and** reset-to-0) leaves ChainState untouched.
- `TestSetLatestBlock_TimestampStaysOnRealClockUnderWarp` / `TestRecompute_TimestampsStayOnRealClockUnderWarp` — a write under a warp stamps the real clock, and the entry still expires on real time once the warp is cleared (no negative-age freshness).
- Full `chainstate` (`-race`) and `rpcsmartrouter` suites pass; `go build ./...` and `gofmt` clean.

### DIFF
--- a/protocol/chainstate/chain_state.go
+++ b/protocol/chainstate/chain_state.go
@@ -198,20 +198,30 @@ func NewWithClock(chainID string, cfg Config, clock func() time.Time) *ChainStat
 	return &ChainState{chainID: chainID, cfg: cfg, now: clock}
 }
 
-// effectiveNow is the clock every TTL/staleness/consensus read uses: the base clock plus the
+// warpOffsetDuration is the current debug warp offset as a Duration (0 in production). Split out so
+// effectiveNow and the "store real / compare warped" split in SetLatestBlock and Recompute read the
+// same lock-free value.
+func (cs *ChainState) warpOffsetDuration() time.Duration {
+	return time.Duration(cs.warpOffset.Load())
+}
+
+// effectiveNow is the clock every TTL/staleness/consensus COMPARISON uses: the base clock plus the
 // debug warp offset. In production the offset is always zero, so this is exactly the base clock;
-// under /debug/time-warp it shifts forward so a warp ages this chain's windows just like it ages
-// the provider-optimizer clock (MAG-2307). The atomic load is lock-free, matching the base clock's
-// no-lock read contract.
+// under the debug ChainState time-warp it shifts forward so a warp ages this chain's windows without
+// waiting real time (MAG-2307). It is used ONLY to EVALUATE freshness — stored timestamps
+// (lastObservedAt / baselineAt / baselineSince) always use the real base clock cs.now(), so clearing
+// the warp can never leave a future-dated timestamp that reads as artificially fresh (MAG-2307
+// review). The atomic load is lock-free, matching the base clock's no-lock read contract.
 func (cs *ChainState) effectiveNow() time.Time {
-	return cs.now().Add(time.Duration(cs.warpOffset.Load()))
+	return cs.now().Add(cs.warpOffsetDuration())
 }
 
 // SetDebugClockOffset shifts this chain's effective clock forward by d, aging its TTL/staleness/
-// consensus windows without waiting real time. Debug-only: the /debug/time-warp and /debug/reset-all
-// handlers call it (with d and 0 respectively), mirroring how they set/clear the optimizer's NowFunc.
-// It is NOT reachable from any relay path. The write is atomic and takes no lock, so it is safe to
-// call concurrently with live reads.
+// consensus windows without waiting real time. Debug-only: driven by the dedicated
+// /debug/chain-state-time-warp endpoint, which is INDEPENDENT of /debug/time-warp (the optimizer/QoS
+// clock) so a routine QoS warp-then-reset never disturbs ChainState. It changes only freshness
+// EVALUATION (effectiveNow), never a stored timestamp. Not reachable from any relay path. The write
+// is atomic and takes no lock, so it is safe to call concurrently with live reads.
 func (cs *ChainState) SetDebugClockOffset(d time.Duration) {
 	cs.warpOffset.Store(int64(d))
 }
@@ -254,8 +264,11 @@ func (cs *ChainState) SetLatestBlock(block int64) (latest int64, at time.Time, a
 	if block <= 0 {
 		return cs.latestBlock, cs.lastObservedAt, false
 	}
-	now := cs.effectiveNow()
-	_, tipFresh := cs.freshLatestLocked(now)
+	// Freshness is EVALUATED against the warped clock, but the stored lastObservedAt uses the real
+	// base clock — a debug warp must never write a future-dated timestamp that reads as artificially
+	// fresh once the warp is cleared (real - future = negative age; MAG-2307 review).
+	nowReal := cs.now()
+	_, tipFresh := cs.freshLatestLocked(nowReal.Add(cs.warpOffsetDuration()))
 	if cs.initialized {
 		if block < cs.latestBlock {
 			if tipFresh {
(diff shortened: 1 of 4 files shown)


avitenzer:
## Review

**TLDR: The core changes are correct and well-tested, and the doc fix checks out against the code — but there's one real leftover: `/debug/time-warp` still responds with `"applied_to_chains":true` even though it no longer touches any ChainState. That response field is now false advertising and should be removed or changed before merge. Beyond that, one missing test and one minor duplication nit.**

### Correctness — verified ✅

- **The real-clock/warped-clock split is right.** In the full `SetLatestBlock` on this branch: `cs.mu` is held before `nowReal := cs.now()` and the atomic warp-offset read, every store site uses `nowReal`, and every freshness evaluation uses `nowReal.Add(warpOffset)`. `Recompute` follows the same pattern. No lock-ordering or torn-read issues.
- **The doc fix is accurate.** `freshBaselineLocked` does return `cs.baselineSince`, so the corrected `ChainStateSnapshot` comment matches the code.
- **The new handler's validation** (finite / `>= 0` / `<= 24h`, 1 KiB body cap) exactly mirrors the existing endpoint, and returning a count in `applied_to_chains` is a nice observability upgrade over the old hardcoded `true`.
- **No downstream breakage found**: grepped the internal black-box repo for `time-warp` / `applied_to_chains` references — none, so decoupling the existing endpoint's behavior shouldn't break an existing consumer.

### Issues

**1. Stale `applied_to_chains:true` in the `/debug/time-warp` response (should fix)** — `protocol/rpcsmartrouter/rpcsmartrouter.go:841`. The handler still ends with:

```go
fmt.Fprintf(w, `{"offset_seconds":%v,"applied_to_chains":true}`, body.OffsetSeconds)
```

That field was added in #222 precisely to signal "the warp reached the ChainStates" — which this PR deliberately made untrue. A test author reading the response would reasonably conclude the coupling still exists. Drop the field entirely (cleanest, since the new endpoint owns that signal now) or emit `"applied_to_chains":0` for symmetry with the new endpoint's count.

**2. `/debug/reset-all` decoupling is claimed but not tested (should add)** — The PR body says both `/debug/time-warp` *and* `/debug/reset-all` no longer touch ChainState, but only the former is pinned (`TestDebugTimeWarp_DoesNotTouchChainState`). The reset-all change (removing `setAllChainStateDebugOffset(deps, 0)` at `rpcsmartrouter.go:945`) has no test. A regression re-adding that line would pass the whole suite. Suggested shape: apply a chain-state warp, POST `/debug/reset-all`, assert `TipFresh` is still `false` (the warp survived the reset).

**3. Semantic consequence worth a doc sentence (optional)** — Storing real timestamps while evaluating on the warped clock means a write *during* an active warp is immediately stale-by-offset: live traffic cannot refresh `TipFresh` while a >TTL warp is active, and equal-block confirmations won't restore freshness either. This is the intended "age out state while endpoints are quiesced" semantic and it's the right trade-off (the negative-age bug was strictly worse), but it inverts #222's behavior where writes under a warp read fresh. Since black-box tests may warp while relays are still flowing, one sentence on the new endpoint's doc comment ("writes during an active warp read as already aged") would prevent confusion.

**4. Handler duplication (nit)** — The new endpoint duplicates ~25 lines of body-cap/decode/NaN/range validation from `/debug/time-warp`. A shared `parseDebugOffsetSeconds(w, r) (time.Duration, bool)` helper would keep the two endpoints' validation from drifting. Fine to defer given this codebase leans toward explicit, heavily-commented handlers.

### Test coverage

Good overall: the four new tests pin exactly the two behaviors that changed (decoupling in each direction, real-clock stamping in both write paths), and the `TestRecompute_TimestampsStayOnRealClockUnderWarp` choice of a *sub-TTL* warp so the baseline can still form under the warped evaluation is careful work. The one gap is item 2 above. The `newAgedChainStateRouter` fixture extraction is a clean refactor of the previous inline test.

**Verdict:** Approve with one required change (the stale `applied_to_chains:true` response field) and one requested test (`/debug/reset-all` leaves the ChainState warp intact). Everything else is solid.

AnnaR-prog:
Thanks for the thorough review @avitenzer — all addressed in `edcd97e`.

**1. `applied_to_chains` response field.** One correction on the premise: this field **predates #222** — it's in the pre-#222 handler (`c011a80^`, line 812) and #222's diff never touched it. It reports the *optimizer*-chain warp (the handler ranges `optimizers`), which is still accurate after the decoupling, so I left `/debug/time-warp` unchanged.

That said, the real issue is the one Copilot flagged inline: my new endpoint reused the key `applied_to_chains` as an **int** while time-warp returns a **bool** — a same-key / different-type collision. Fixed by having `/debug/chain-state-time-warp` return a distinct **`chain_states_warped`** count instead. So `applied_to_chains` is now bool-only (time-warp), and the new endpoint's count lives under its own key — no cross-endpoint type collision.

**2. `/debug/reset-all` test — added.** `TestDebugResetAll_DoesNotTouchChainStateWarp` applies a ChainState warp, POSTs `/debug/reset-all`, asserts `TipFresh` is still `false` (the warp survived), then clears via the dedicated endpoint. Good catch — a real regression guard for the removed `setAllChainStateDebugOffset(deps, 0)` call.

**3. Semantic doc note — added.** The new endpoint's comment now documents that stored timestamps use the real clock, so a write arriving during an active >TTL warp reads as already aged (the intended age-out-while-quiesced behavior; warp with endpoints quiesced for a clean result).

**4. Validation duplication — deferred**, per your suggestion and the codebase's explicit-handler leaning.

Full `rpcsmartrouter` suite + `go build ./...` / `gofmt` clean.

VicSheCodes (approved):

avitenzer (approved):

PR #225, opened 2026-07-21
fix(consistency): measure endpoints against the guarded chain tip; retire per-user seenBlock (Topic C T1+T2)

Tomelia1999:
## What

Topic C **T1 + T2 (C-G)**. Fixes the confirmed production bug where one provider reporting a fake-high block captured all traffic on a multi-provider pod and drove a single-provider pod to **No pairings** until a manual reset.

**Root cause:** consistency validation compared each endpoint against a **per-user `seenBlock`** fed straight from the served provider's `Reply.LatestBlock`, guarded only by monotonic-increase — no anti-lie cross-check. A lying-high block poisoned it, honest providers then measured as "behind" and were filtered out, and the value could never descend (monotonic-max, TTL refreshed by traffic).

## The change

**T1 — one tip:**
- `Recompute` applies the D3 edge correction: snap **down** when the tip leads consensus by more than `OutlierThreshold` (kept), and pull **up** when it trails (new). Within-band optimistic leads are preserved (read-your-writes).
- Added `latestSince` + `GetLatestBlockWithTime` (the tip's equivalent of `baselineSince`) so time-based sync scoring measures tip *establishment*, not last confirmation.
- Unexported the consensus baseline; every production reader (optimizer sync, cache finalization, `endpointSyncGap`, probe cycle) now reads the one tip.

**T2 — consistency reads the tip (C-G):**
- `ValidateEndpointCapability` takes `chainTip` instead of `seenBlock`; the per-endpoint operand is unchanged.
- Removed the `seenBlock` write path entirely, and **deleted the per-user store** (the `Consistency` interface, ristretto cache, TTL, and the gen/tombstone reset machinery). Only `UserDataKey` survives, for the shared-state cache id.

## Accepted residuals (please review)

- **Cache finalization** now measures against the tip. This *restores* the behavior that predates a recent baseline-only change, and is safer than it was then (tip-or-0, no bootstrap-atomic fallback, plus the outlier guard).
- **Sync scoring** can charge endpoints at the majority for a within-threshold optimistic lead, bounded by `OutlierThreshold`.
- The cross-instance shared-state seen-block adopt path is removed — it was already dead (write/read used mismatched `SharedStateId`s, so the cache key never matched). Rebuilding it on the tip is tracked as a follow-up.

## Reconciliation with MAG-2307 (heads-up to @annarayer)

This rebased cleanly onto `main` including your **MAG-2307** ChainState warp-clock. They're additive: the baseline getter is now **both** unexported (this PR's single-reader rule) **and** warp-aware (your `effectiveNow`). Two auto-merge gaps were fixed by hand — `GetLatestBlockWithTime` now routes through `effectiveNow`, and your warp test's baseline call was renamed to the unexported form.

## Testing

- `go build ./...` + full `go test ./...` green (34 packages).
- Adds `consistency_liar_repro_test.go` — a Go-level stand-in for the two Python simulator tests. 4 of its 5 cases were verified to **fail against the old `seenBlock` reference and pass against the tip**.

> ⚠️ **Acceptance gate not yet run.** The stated definition-of-done is the two Python simulator tests (`test_check_timestamp_and_liar_score_isolation.py`, `test_lying_provider_tip_selfheal.py`), which live outside this repo. Please run them before merging to main — until then the accurate claim is "mechanism fixed and unit-gated," not "production bug confirmed fixed end-to-end."


avitenzer:
# Review

## Overview

This PR fixes the confirmed F14 production bug: consistency validation compared endpoints against a per-user `seenBlock` fed straight from the served provider's `Reply.LatestBlock` with only a monotonic-increase guard — one lying-high provider poisoned the reference, filtered out every honest endpoint, and the value could never descend. The fix unifies all "what block is the chain at" readers onto the anti-lie-guarded `ChainState` tip (D3 edge correction: snap down beyond `OutlierThreshold`, pull up below baseline), unexports the consensus baseline, and deletes the entire per-user store (~920 lines removed).

Verified locally against the PR head (`db2dca1`): `go build ./...` passes and the four affected packages' tests (`chainstate`, `relaycore`, `chaintracker`, `rpcsmartrouter`) are all green. The dead-code claim about the cross-instance shared-state adopt path **checks out**: the write side stamps `SharedStateId = listenEndpoint.Key()` (`rpcsmartrouter_server.go:2931`) while the read side used `dappId__consumerIp`, so the cache-server key (`chainID + "_" + sharedStateId`, `ecosystem/cache/handlers.go:556`) could never match — the removed adopt branch was genuinely unreachable.

The core design is sound and the direction is right. Findings below, ordered by severity.

## Findings

### 1. A within-band liar can still capture a multi-provider pod — the sharpest residual, and it isn't named in the PR

The PR names within-band residuals for sync scoring and cache finalization, but not for the consistency filter itself, where the original bug bit. The thresholds are asymmetric:

- The tip may lead consensus by up to `OutlierThreshold` — floor **32 blocks**, ≈**100 blocks** for a 12s chain like ETH1 (`chain_state.go`).
- `filterEndpointsByConsistency` drops endpoints more than `EndpointLagThreshold` behind the tip — commonly the floor of **10 blocks** (`consistency_config.go`).

So on a 3-provider ETH1 pod with honest majority at 1000: a liar polling `1099` is accepted by `SetLatestBlock` (within `baseline+100`), the tip settles at 1099, Recompute deliberately preserves the in-band lead, and the liar's equal-block confirmations keep the tip TTL-fresh **indefinitely**. Honest endpoints read lag 99 > 10 and are filtered — the takeover recurs, bounded in magnitude (~100 blocks) but *unbounded in duration*. `TestLiar_MultiProviderHonestEndpointsKeepServing` only exercises the out-of-band lie (~20M blocks); change `liarClaim` to `honestBlock + 99` and the honest endpoints get filtered again. Compounding it, sync scoring now also charges honest endpoints that same in-band gap, degrading their QoS ranking in the liar's favor.

**Suggestions:** at minimum, name this residual in the PR body/plan and add a within-band-liar case to the Python acceptance suite. A code-level mitigation that preserves the single-reader rule everywhere else: when a baseline exists, clamp the *consistency reference only* to `min(chainTip, baseline + EndpointLagThreshold)` — honest endpoints at the majority then always pass, while a genuinely-behind endpoint is still caught.

### 2. `/debug/reset-all` can no longer flush a poisoned tip, but claims it can

Two related inaccuracies in `rpcsmartrouter.go`:

- The comment at the removed flush site says "The chain tip itself is reset via the tracker reset below" — it isn't. `resetAllChainTrackers` calls `endpointChainTrackerManager.ResetAllLatestBlocks()`, which zeroes the *per-endpoint monitor* values; `ChainState.latestBlock` is untouched, and nothing else in reset-all touches it. A poisoned in-band tip survives reset-all until TTL (~2 minutes on ETH1) — or indefinitely under finding 1, since the liar keeps refreshing it.
- The `"seen-block"` capability is still advertised with nothing behind it — documented as intentionally stale (T11), but the external test framework probes exactly this key to decide it doesn't need a process restart.

**Suggestion:** add a `ChainState` reset method and call it from reset-all. That makes the advertised capability honest again *and* restores the operator big-hammer for finding 1 — currently there is no manual recovery from a within-band capture at all, which is a step backwards from the bug's own workaround.

### 3. Committed code references a git-ignored personal folder

Three new comments point into `agent_docs/bug-reports/chainTracker-architecture/topic-C-action-plan.md`: `relaycore/consistency.go` (~line 28), `rpcsmartrouter.go` (~line 957), and `rpcsmartrouter_server.go` (~line 3193, plus the T10 mention nearby). `agent_docs/` is git-ignored and personal — these pointers dangle for every other developer. Replace with Jira ticket IDs (and file T10/T11 as real tickets if they aren't).

### 4. Stale documentation contradicting the change

- `rpcsmartrouter_server.go:842–847` (`getLatestBlock`'s Site A doc): still says "Cache FINALIZATION does not read this: it needs the strict-majority consensus baseline… the optimistic tip … is not good enough." That's now exactly what finalization reads — this comment argues against the PR's own accepted residual and will mislead the next reader.
- `chain_state.go` comments (lines ~176, 357, 467, 470, 482) still name the exported `GetConsensusBaseline`/`GetConsensusBaselineWithTime`, which no longer exist.
- `rpcsmartrouter.go:1244`: "`BACKOFF_MAX_TIME = 10m`" — stale after the tuning commit changed it to 1m.
- `SetConsensusBaselineGetter` now installs a tip getter; the rename is acknowledged as deferred, but worth a follow-up ticket so it doesn't fossilize.

### 5. Test coverage gaps

- **`latestSince` / `GetLatestBlockWithTime` have no direct tests.** The baseline had `TestRecompute_BaselineSincePreservedWhileBlockUnchanged` pinning exactly these establishment-time semantics; its tip mirror is missing. Sync scoring and the probe cycle now depend on `latestSince` — a regression (e.g. resetting it on an equal-block confirmation, which would make every endpoint's sync lag read ~0 forever) would be invisible to the suite.
- `TestGetLatestBlockForCacheFinalization_TipOrZero`'s third phase asserts the right value for the wrong reason: `cs.SetLatestBlock(1200)` is *rejected by the write guard* (baseline 1040 + threshold 100 = 1140 < 1200), so the tip stays 1050 and the subsequent Recompute at 1060 exercises **pull-up**, not the snap-down the comment claims. Poison the tip in-band first (or cold-start) if snap-down is what this phase should lock.

### 6. Undocumented tuning commit bundled in

The PR includes `db2dca1` (`DefaultBucketWidth` 2→5, `BACKOFF_MAX_TIME` 10min→1min) — neither appears in the PR body. Both are behaviorally meaningful: BucketWidth 5 widens the "agreeing" cluster, and since the baseline is the cluster's *most-advanced* block, a fast-or-lying reporter within 5 blocks of the mode now raises the baseline itself; BACKOFF_MAX_TIME 1min means permanently-dead endpoints are re-polled every minute forever instead of every 10. These deserve their own PR, or at least their own section in this one's description.

### 7. Acceptance gate

The PR is candid that the two Python simulator tests haven't run — seconding that they must run before merge (squash-merge, per repo convention), ideally extended with the within-band case from finding 1.

## Verdict

Solid fix with excellent regression tests for the confirmed bug shape; approve after addressing findings 2 and 3, with finding 1 either mitigated or explicitly ticketed and simulator-covered — it is a bounded re-opening of the exact bug this PR closes.

avitenzer (approved):

PR #226, opened 2026-07-21
refactor(endpointtip): block-monotonic per-endpoint tip with a staleness backstop (Topic C T4)

Tomelia1999:
## What

Topic C **T4 / C-D**. Makes the per-endpoint tip store order by **block**, not by arrival time.

> **Stacked on #225 (T1+T2).** Targets `unify-seenBlock-chainTip`; will auto-retarget to `main` when #225 merges. Review #225 first.

## Root cause (F2)

The per-endpoint store dropped any write whose `ObservedAt` predated the stored one. So a newer-but-**lower** straggler (a relay observed earlier but delivered late) overwrote a higher poll block, **regressing an honest endpoint's tip** for no real reason. That per-endpoint value is the operand consistency measures, an input to sync scoring, and a vote in consensus — so a wrong value leaks widely.

## The change

`endpointtip.Store.Set` is now block-monotonic with a staleness backstop (mirrors `ChainState.SetLatestBlock`):
- a **higher** block always wins;
- an **equal** block advances the freshness stamp;
- a **lower** block is rejected while the stored tip is fresh (kills F2) and accepted once stale (a real reorg down).

Staleness is measured between the stored and incoming **observation stamps** (`> staleAfter`), not wall-clock — deterministic, no clock injection. Resolves **D-FORK** as *staleness-only* (no fork signal), dropping the cross-package `chaintracker→store` wiring the original plan needed.

The horizon reuses the shared **`chainstate.StalenessWindow`** — the same derivation the consensus window / tip TTL uses. `endpointtip` defines no staleness constants of its own; the monitor derives and passes it in. (Note: the probe's liveness horizon shares the same **multiplier** — `DefaultStalenessMultiplier` — but re-derives with its own floor rather than calling `StalenessWindow`, so the shared knob across the codebase is the multiplier, not the function.)

## Accepted trade-off

With no anti-lie guard at the per-endpoint layer (that lives in `ChainState`), a transient lying-**HIGH** observation is now retained until it goes stale, where arrival-time overwrote it instantly. This is **asymmetric in our favour**: F2 harmed *honest* endpoints (wrongly filtered → availability loss); the new cost only inflates a *liar's* own tip (F19, already accepted), and the global tip + consensus stay protected by ChainState's guard and clustering. `TestHarvest_BootstrapAtomicGatedOnChainStateVerdict` documents this.

## Testing

- `go build ./...` + full `go test ./...` green (34 packages).
- New `endpointtip` unit tests cover the straggler rejection, equal-block refresh, and the stale-backstop reorg acceptance.

### DIFF
--- a/protocol/chainstate/chain_state.go
+++ b/protocol/chainstate/chain_state.go
@@ -122,12 +122,22 @@ func outlierThresholdForBlockTime(averageBlockTime time.Duration) int64 {
 	return blocks
 }
 
+// StalenessWindow is the ONE derivation of the "fresh/alive horizon" from a chain's average
+// block time: max(DefaultStalenessMultiplier × avgBlockTime, minStalenessWindow). It is the
+// single source of truth for the concept — the consensus window / tip TTL (DefaultConfig), the
+// probe's liveness horizon, and the per-endpoint tip's staleness backstop (endpointtip, fed via
+// the monitor) all read it, so tuning DefaultStalenessMultiplier moves every one in lockstep. A
+// non-positive block time falls back to the floor.
+func StalenessWindow(averageBlockTime time.Duration) time.Duration {
+	return max(time.Duration(DefaultStalenessMultiplier)*averageBlockTime, minStalenessWindow)
+}
+
 // DefaultConfig derives the freshness/consensus window AND the outlier threshold from a chain's
-// average block time (D6): StalenessWindow = TTL = max(DefaultStalenessMultiplier × avgBlockTime,
-// floor); OutlierThreshold = clamp(OutlierTimeBudget / avgBlockTime, floor, ceiling). BucketWidth
-// stays a block-count constant (clustering tolerance is intrinsically in blocks, not time).
+// average block time (D6): StalenessWindow = TTL = StalenessWindow(avgBlockTime); OutlierThreshold
+// = clamp(OutlierTimeBudget / avgBlockTime, floor, ceiling). BucketWidth stays a block-count
+// constant (clustering tolerance is intrinsically in blocks, not time).
 func DefaultConfig(averageBlockTime time.Duration) Config {
-	window := max(time.Duration(DefaultStalenessMultiplier)*averageBlockTime, minStalenessWindow)
+	window := StalenessWindow(averageBlockTime)
 	return Config{
 		BucketWidth:      DefaultBucketWidth,
 		OutlierThreshold: outlierThresholdForBlockTime(averageBlockTime),
--- a/protocol/chainstate/chain_state.go
+++ b/protocol/chainstate/chain_state.go
@@ -122,12 +122,14 @@ func outlierThresholdForBlockTime(averageBlockTime time.Duration) int64 {
 	return blocks
 }
 
-// StalenessWindow is the ONE derivation of the "fresh/alive horizon" from a chain's average
-// block time: max(DefaultStalenessMultiplier × avgBlockTime, minStalenessWindow). It is the
-// single source of truth for the concept — the consensus window / tip TTL (DefaultConfig), the
-// probe's liveness horizon, and the per-endpoint tip's staleness backstop (endpointtip, fed via
-// the monitor) all read it, so tuning DefaultStalenessMultiplier moves every one in lockstep. A
-// non-positive block time falls back to the floor.
+// StalenessWindow derives the "fresh/alive horizon" from a chain's average block time:
+// max(DefaultStalenessMultiplier × avgBlockTime, minStalenessWindow). The consensus window / tip
+// TTL (DefaultConfig) and the per-endpoint tip's staleness backstop (endpointtip, fed via the
+// monitor) both call it, so they move in lockstep. The shared knob across the codebase is the
+// MULTIPLIER, not this function: the probe's liveness horizon (probing.DefaultVerdictConfig)
+// re-derives with DefaultStalenessMultiplier but a deliberately different floor (minProbeStaleness),
+// so it tracks the multiplier without reading this function. A non-positive block time falls back
+// to the floor.
 func StalenessWindow(averageBlockTime time.Duration) time.Duration {
 	return max(time.Duration(DefaultStalenessMultiplier)*averageBlockTime, minStalenessWindow)
 }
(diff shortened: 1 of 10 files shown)


avitenzer:
## Review

**Verdict: the core logic is correct and well-tested — approve after the doc fixes below.** Ran `endpointtip`, `chainstate`, `endpointstate`, and the touched `rpcsmartrouter` tests on this branch: all green.

### What's done well

- **The rejected-write-must-not-refresh-the-stamp rule is the load-bearing detail**, and it's both implemented and pinned by a dedicated assertion (`store_test.go`: "rejected write must not refresh the stamp"). Without it, a repeating straggler would keep the stale tip fresh forever and the backstop would never fire — the design would livelock. Good catch designing this in.
- **Equal-block refresh advances the stamp but never regresses it**, so a healthy endpoint at a stable head stays "fresh" indefinitely — and the test verifies the refresh actually *extends* the rejection window (the base+17s case). That proves the mechanism, not just the state.
- Measuring staleness as the gap between two observation stamps (instead of against `time.Now()`) keeps the store clock-free and the tests deterministic. The trade — the backstop only matures when a new observation arrives — is fine, since a dead endpoint's tip is irrelevant anyway.
- The `staleAfter <= 0` up-only escape hatch is minimal and its only consumers (test seeds via `Remove`-then-`Set`) are correct.
- Consistent flooring: the monitor floors `avgBlockTime` to `DefaultAverageBlockTime` exactly as `rpcsmartrouter_server.go` does before `DefaultConfig`, so `tipStaleAfter` and ChainState's TTL genuinely agree (`rpcsmartrouter_server.go:156-190`, `endpoint_monitor.go:165-168`).
- The reworked `TestHarvest_BootstrapAtomicGatedOnChainStateVerdict` honestly documents the accepted regression (lie retained until stale) instead of quietly deleting the old recovery assertion.

### Issues to fix (all documentation — but one is an invariant other code may rely on)

The semantics flip left several comments asserting the **old** contract:

1. **`protocol/endpointstate/observation.go:35-37`** — the `EndpointObservation` doc still says *"ObservedAt is monotonic — a later write never moves it backward, and a stale observation (older than the current ObservedAt) is ignored — enforced in endpointtip.Store.Set."* Both halves are now wrong: a higher block with an older stamp **does** move `ObservedAt` backward (the new store test pins this), and older-stamped observations are accepted when the block is higher. Anyone building on this stated invariant will be misled.
2. **`protocol/endpointstate/observation.go:145`** — `RecordRelayObservation` doc: *"honors the monotonic ObservedAt guard"* — stale.
3. **`protocol/rpcsmartrouter/rpcsmartrouter_server_test.go:2670-2673`** — `TestEndpointTipPreferStore`'s rationale says the store *"reflects the newest observation — including a newer LOWER block after a reorg."* Under T4 that reflection is **delayed by up to the staleness window**, not immediate. The prefer-store conclusion still holds (the store eventually self-heals; the atomic never does), but the narrative needs updating.
4. **`protocol/rpcsmartrouter/rpcsmartrouter_server_test.go:2643-2644`** — `seedEndpointTip` doc still says *"ObservedAt = time.Now() keeps the time-monotonic guard happy"* — obsolete; successive seeds now work because of the `Remove`, which the new inline comment already explains. Delete the old sentence.
5. **`protocol/rpcsmartrouter/rpcsmartrouter_server.go:2543`** and **`protocol/endpointstate/endpoint_monitor.go:504`** — both say "time-monotonic" where the guard is now block-monotonic. The surrounding reasoning survives in both places (arguably *more* strongly at 2543, since a lying-high now passes the per-endpoint guard trivially); only the terminology is stale.
6. **`protocol/chainstate/chain_state.go:126-129`** — the new `StalenessWindow` doc overstates its reach: it claims *"the probe's liveness horizon … all read it"*, but `probing/verdict.go:49` does **not** call it — it re-derives from `DefaultStalenessMultiplier` with its own floor (`minProbeStaleness` = 5s vs `minStalenessWindow` = 2s). Either soften the claim to "shares the multiplier", or refactor the probe to `max(chainstate.StalenessWindow(abt), minProbeStaleness)` — since 5s > 2s that expression is algebraically identical to today's, and it would make the "single source of truth" claim actually true.

### Observations (no change requested)

- **Higher-block writes store the incoming stamp wholesale, so `ObservedAt` can regress** (a delayed high observation lands with its older stamp — the first store test pins this at base−1s). Every consequence points the conservative direction: the relay poll-suppression gate expires sooner, the backstop matures sooner, and the endpoint's consensus-vote freshness reads slightly older. Stamps are router-assigned (`time.Now()` at observation), so an endpoint can't manufacture the regression. Fine as is; a clamp (`max(t.ObservedAt, cur.ObservedAt)`) would trade this for a violated "triple moves atomically" doc, which isn't worth it.
- **`/debug/time-warp` will not age this backstop.** ChainState's windows age via its warped `effectiveNow`, but the store compares two real observation stamps. Someone simulating a reorg via warp will see ChainState go stale while the per-endpoint tip still rejects downward moves until real observation time elapses. Deliberate per the design ("no clock injection"), but worth a line in the probing-automation docs so a tester doesn't file it as a bug.
- **The reorg half of the trade-off also touches honest endpoints**, not just liars: after a genuine reorg down, an honest endpoint's stored tip stays stale-high for up to ~10 block times, during which consistency measures it as more caught-up than it is. That fails *open* (no wrongful filtering — the exact inverse of F2's harm) and is bounded, so it's acceptable; noting it because the PR body frames retention purely as a liar-side cost.

### Test coverage

Good: straggler rejection, equal-block refresh (including the window-extension proof and stamp-no-regress), stale-backstop reorg acceptance, `staleAfter<=0`, and the integration-level lie-retention consequence are all pinned. The only untested seam is the monitor actually passing a non-zero `tipStaleAfter` (a unit test asserting `NewEndpointMonitor(...).tipStaleAfter == chainstate.StalenessWindow(...)` would pin the wiring), but the horizon-agreement property is low-risk given both sides floor identically.

### Security

No concerns. The change strictly tightens what a late/reordered network response can do to state; the lying-high retention it introduces is confined to the liar's own per-endpoint value, with the global tip still behind ChainState's anti-lie guard and the bootstrap atomic gated on ChainState's verdict (verified unchanged at `rpcsmartrouter_server.go:2553-2560`).

avitenzer (approved):

avitenzer (approved):

PR #228, opened 2026-07-22
fix(shared-state): rebuild cross-pod consistency on the guarded chain tip (T10)

Tomelia1999:
## What

Rebuilds the cross-pod shared-state consistency mechanism on the **guarded chain tip** instead of the retired per-user seenBlock. Stacked on #226 (T4).

### The bug it fixes
Shared-state mode wrote the seen block under a **chain-level** key (`listenEndpoint.Key()` = `ChainID+ApiInterface`) but **read** it under a **per-user** key (`dappId__consumerIp`). The two never collided, so `getSeenBlockForSharedStateMode` missed every time and returned 0 — the feature has been dead since the initial commit (`22cbf32`). T2 (#225) removed the dead read half; this rebuilds it correctly.

### Changes
- **Read id → `listenEndpoint.Key()`** (matches the write; chain-level) so the round-trip closes.
- **`adoptSharedStateTip()`** feeds the fleet-max peer tip through `ChainState.SetLatestBlock` — the same outlier + Recompute anti-lie guards as a local observation — instead of raw into request resolution. A poisoned peer value is snapped down, not trusted. No-op unless shared state is on, ChainState is wired, and the peer is genuinely ahead.
- **Cache TTL**: the shared tip now expires after `SharedStateTipBlockMultiplier(10) × averageBlockTime` (`SharedStateTipExpiration`, floored at `ExpirationNonFinalized`), replacing the flat ~1h finalized TTL. Mirrors the router's own `StalenessWindow`, so a dead pod's tip evaporates on the same horizon the router considers a tip stale.
- The write side needed **zero changes** — it already publishes the guarded tip under the chain-level key.

## Verification

**Live**, on a 2-replica `base-router` + shared `router-cache` (dev cluster, `--shared-state` on):
- Reads hit the aligned key `BASE_BASEjsonrpc` at **~99.4%** (328 hits / 2 cold-start misses in 90s).
- Writes land under that same key (`Got Cache Set … latestKnownBlock` == the read value).
- Peer-ahead tips adopt through the guard on **both** replicas (`peer=local+1`), 0 crashes, 0 errors.

**Unit tests**:
- `ecosystem/cache/shared_state_tip_test.go` — TTL math (incl. zero-block-time floor); key round-trip + per-user-id isolation (encodes the bug); max-merge.
- `protocol/rpcsmartrouter/shared_state_tip_adopt_test.go` — adopt glue: no-op when off / not ahead / nil ChainState; adopts a plausible peer tip; **snaps down an over-threshold (lying-high) peer tip**.

Full `./protocol/rpcsmartrouter/...` and `./ecosystem/cache/...` suites pass; `go vet` + full build clean.

### DIFF
--- a/ecosystem/cache/handlers.go
+++ b/ecosystem/cache/handlers.go
@@ -320,13 +320,13 @@ func (s *RelayerCacheServer) performInt64WriteWithValidationAndRetry(
 	}
 }
 
-func (s *RelayerCacheServer) setSeenBlockOnSharedStateMode(chainId, sharedStateId string, seenBlock int64) {
+func (s *RelayerCacheServer) setSeenBlockOnSharedStateMode(chainId, sharedStateId string, seenBlock int64, ttl time.Duration) {
 	if sharedStateId == "" {
 		return
 	}
 	key := latestBlockKey(chainId, sharedStateId)
 	set := func() {
-		s.CacheServer.finalizedCache.SetWithTTL(key, seenBlock, 0, s.CacheServer.ExpirationFinalized)
+		s.CacheServer.finalizedCache.SetWithTTL(key, seenBlock, 0, ttl)
 	}
 	get := func() int64 {
 		return s.getSeenBlockForSharedStateMode(chainId, sharedStateId)
@@ -375,7 +375,7 @@ func (s *RelayerCacheServer) SetRelay(ctx context.Context, relayCacheSet *relayt
 		cache.SetWithTTL(string(cacheKey), cacheValue, cacheValue.Cost(), s.getExpirationForChain(time.Duration(relayCacheSet.AverageBlockTime), relayCacheSet.BlockHash))
 	}
 
-	s.setSeenBlockOnSharedStateMode(relayCacheSet.ChainId, relayCacheSet.SharedStateId, latestKnownBlock)
+	s.setSeenBlockOnSharedStateMode(relayCacheSet.ChainId, relayCacheSet.SharedStateId, latestKnownBlock, s.CacheServer.SharedStateTipExpiration(time.Duration(relayCacheSet.AverageBlockTime)))
 	s.setLatestBlock(latestBlockKey(relayCacheSet.ChainId, ""), latestKnownBlock)
 	s.setBlocksHashesToHeights(relayCacheSet.ChainId, relayCacheSet.BlocksHashesToHeights)
 	return &emptypb.Empty{}, nil
--- a/ecosystem/cache/server.go
+++ b/ecosystem/cache/server.go
@@ -32,6 +32,12 @@ const (
 	ExpirationNodeErrorsOnFinalizedFlagName      = "expiration-finalized-node-errors"
 	FlagCacheSizeName                            = "max-items"
 	DefaultExpirationForNonFinalized             = 500 * time.Millisecond
+	// SharedStateTipBlockMultiplier bounds how long a pod's published chain tip lives in the
+	// shared cache: SharedStateTipBlockMultiplier * averageBlockTime. It mirrors the smart
+	// router's own tip-staleness horizon (chainstate.StalenessWindow) so a dead pod's tip
+	// evaporates on roughly the same timescale the router considers a tip stale, rather than
+	// lingering for the full finalized (~1h) TTL.
+	SharedStateTipBlockMultiplier                = 10
 	DefaultExpirationTimeFinalizedMultiplier     = 1.0
 	DefaultExpirationTimeNonFinalizedMultiplier  = 1.0
 	DefaultExpirationBlocksHashesToHeights       = 48 * time.Hour
@@ -201,6 +207,15 @@ func (cs *CacheServer) Serve(ctx context.Context, listenAddr string) {
 func (cs *CacheServer) ExpirationForChain(averageBlockTimeForChain time.Duration) time.Duration {
 	eighthBlock := averageBlockTimeForChain / 8
 	return lavaslices.Max([]time.Duration{eighthBlock, cs.ExpirationNonFinalized})
+}
+
+// SharedStateTipExpiration returns the TTL for a pod's published chain tip in shared-state
+// mode: SharedStateTipBlockMultiplier * averageBlockTime, floored at ExpirationNonFinalized so
+// a chain whose spec omits the average block time still gets a sane, non-zero TTL (a zero TTL
+// would never expire).
+func (cs *CacheServer) SharedStateTipExpiration(averageBlockTimeForChain time.Duration) time.Duration {
+	ttl := averageBlockTimeForChain * SharedStateTipBlockMultiplier
+	return lavaslices.Max([]time.Duration{ttl, cs.ExpirationNonFinalized})
 }
 
 func (cs *CacheServer) GetTotalCacheSize() int64 {
--- a/ecosystem/cache/shared_state_tip_test.go
+++ b/ecosystem/cache/shared_state_tip_test.go
@@ -1,0 +1,58 @@
+package cache
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestSharedStateTipExpiration locks the shared-state tip TTL contract (T10): a pod's published
+// tip lives for SharedStateTipBlockMultiplier * averageBlockTime, floored at ExpirationNonFinalized
+// so a chain with an unknown block time still gets a sane, non-zero TTL instead of never expiring.
+func TestSharedStateTipExpiration(t *testing.T) {
+	cs := &CacheServer{ExpirationNonFinalized: 500 * time.Millisecond}
+
+	// Normal chain: 12s block → 120s TTL (well above the floor, far below the ~1h finalized TTL).
+	require.Equal(t, 120*time.Second, cs.SharedStateTipExpiration(12*time.Second))
+
+	// Fast chain: 400ms block → 4s (still above the floor).
+	require.Equal(t, 4*time.Second, cs.SharedStateTipExpiration(400*time.Millisecond))
+
+	// Unknown block time (spec omits average_block_time): 0*10 = 0 → floored so it is not eternal.
+	require.Equal(t, cs.ExpirationNonFinalized, cs.SharedStateTipExpiration(0))
+}
+
+// TestSharedStateTip_RoundTripAndKeyIsolation is the T10 fix: a tip written under one shared-state
+// id is readable only under the SAME id. The historical bug was that the write used a chain-level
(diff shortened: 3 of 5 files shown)


avitenzer:
## Review

**Verdict: the core fix is correct and well-tested — I verified the key alignment, guard behavior, and both new test files pass on the PR branch. But one claim in the PR body and code comments is inaccurate in a way that matters: the write side does *not* publish a guarded tip, and combined with a confirmed cold-start guard bypass on the adopt side, a single lying endpoint anywhere in the fleet can poison a restarting pod's chain tip for a bounded window. Worth addressing (or explicitly accepting) before merge.**

### What I verified (all pass)

- `RPCEndpoint.Key()` is `ChainID + ApiInterface` — pod-independent, so cross-pod sharing actually works; write side (`protocol/rpcsmartrouter/rpcsmartrouter_server.go:2948`) and read side (`:3050`) now derive the same key.
- On a cache *miss*, `GetRelay` still returns the shared seen block in `cacheReply.SeenBlock`, so adoption works at miss rate too; on a gRPC `cacheError`, `cacheReply` is nil and the nil-safe proto getter returns 0 → clean no-op.
- Equal-value re-publishes refresh the TTL (`existingInfo <= newInfo` in `performInt64WriteWithValidationAndRetry`), so the leading pod keeps the entry alive and a dead pod's tip expires — the design works as described.
- Cache finalization is *not* exposed to a poisoned tip — `isFinalizedForCacheWrite` uses the consensus baseline (`getLatestBlockForCacheFinalization`), not this tip. Good containment.
- Both new test files pass (`go test`, run on the PR head). The round-trip test genuinely encodes the historical bug.

### Finding 1 (medium): the published value is not guarded, and cold-start adoption bypasses the guard

The PR body says "the write side needed zero changes — it already publishes the guarded tip," and the `adoptSharedStateTip` comment says the shared value is "the fleet-max of every pod's guarded tip." Neither is accurate. `SetRelay` publishes `max(Response.LatestBlock, SeenBlock)` (`ecosystem/cache/handlers.go:350`), and `Response.LatestBlock` is the endpoint-reported value from the reply (or the per-endpoint observation store, `rpcsmartrouter_server.go:1891`) — it passes the per-endpoint time-monotonic gate at best, never the chain-level anti-lie guard. The harvest path explicitly treats this scope distinction as the reason the bootstrap atomic needs its own gate (`rpcsmartrouter_server.go:2542`: "a lying-high endpoint passes its own per-endpoint guard").

On its own that's survivable, because receiving pods snap the value down — *except on cold start*. Confirmed empirically with a probe test on this branch: an uninitialized `ChainState` adopts a peer tip of 1,000,000,000 raw (the guard has no first-observation reference — `chain_state.go` documents this caveat itself), and an honest local observation of 23M afterwards **cannot displace it while the tip is fresh** (`SetLatestBlock` returns `advanced=false`, tip stays at 1e9). Since `getLatestBlock()` treats ChainState as the authority (`rpcsmartrouter_server.go:848`), a poisoned restarting pod mislabels archive routing and finality and resolves LATEST cache keys at the lying block until the shared entry (10 block times, max-merge makes it sticky — lower honest publishes can't overwrite it) plus the ChainState staleness window expire.

The chain: lying endpoint on any pod → unguarded publish, sticky via max-merge → any pod that restarts in that window adopts it raw. Two hardening options, either one breaks the chain:

- **At the source (preferred):** gate the publish the way the bootstrap atomic is gated — publish `min(latestKnownBlock, guardedTip)` or skip the shared-state publish when `tip > chainState accepted`. This makes the "fleet-max of guarded tips" comment true and protects every consumer.
- **At adoption:** skip `adoptSharedStateTip` when `chainState.GetLatestBlock()` reports not-ok (uninitialized or stale). Costs the fast warm-up benefit, which is exactly when shared state is most useful — hence preferring the write-side fix.

If the bounded exposure is consciously accepted instead, the comments and PR body should at least stop claiming the published value is guarded, and a cold-start test should encode the actual behavior.

### Minor points

- `adoptSharedStateTip` gates on `peerTip > localRelayData.SeenBlock` (the request-stamped value) rather than the current local tip, so the "shared-state tip is newer" debug log fires even when the peer tip is older than the local ChainState tip. Correctness holds because `SetLatestBlock` is authoritative, but the log will overstate adoption frequency.
- For chains whose spec omits `average_block_time`, the TTL floors at `ExpirationNonFinalized` (default 500ms) — shared state becomes effectively inert for those chains, versus ~1h before. The comment covers the zero case; just confirming this behavior change is intended.
- `getSeenBlockForSharedStateMode` logs "Failed fetching state from cache for this user id" at Info level — the id is now chain-level, so the message is stale, and post-fix this line fires on every cold-start read. Trivial, pre-existing.

### Test coverage

Good: TTL math including the zero floor, key round-trip encoding the historical bug, per-id isolation, max-merge, and all five adopt-glue branches including nil-ChainState panic safety and the snap-down. The one missing case is the cold-start adopt (uninitialized ChainState) — which is precisely where the guard doesn't hold, and the suite's "snapped down by the guard" test only covers a *seeded* ChainState, which may be why the gap went unnoticed.

**Recommendation:** approve after either hardening the write-side publish (preferred — it also makes the comments true) or explicitly downgrading the "guarded tip" claims and adding a cold-start adopt test that encodes the accepted exposure.

Tomelia1999:
Its an hedge case that should be covered, lets not get into it now, lets add it as a subtask to the "Caching" taks @avitenzer 

> ## Review
> **Verdict: the core fix is correct and well-tested — I verified the key alignment, guard behavior, and both new test files pass on the PR branch. But one claim in the PR body and code comments is inaccurate in a way that matters: the write side does _not_ publish a guarded tip, and combined with a confirmed cold-start guard bypass on the adopt side, a single lying endpoint anywhere in the fleet can poison a restarting pod's chain tip for a bounded window. Worth addressing (or explicitly accepting) before merge.**
> 
> ### What I verified (all pass)
> * `RPCEndpoint.Key()` is `ChainID + ApiInterface` — pod-independent, so cross-pod sharing actually works; write side (`protocol/rpcsmartrouter/rpcsmartrouter_server.go:2948`) and read side (`:3050`) now derive the same key.
> * On a cache _miss_, `GetRelay` still returns the shared seen block in `cacheReply.SeenBlock`, so adoption works at miss rate too; on a gRPC `cacheError`, `cacheReply` is nil and the nil-safe proto getter returns 0 → clean no-op.
> * Equal-value re-publishes refresh the TTL (`existingInfo <= newInfo` in `performInt64WriteWithValidationAndRetry`), so the leading pod keeps the entry alive and a dead pod's tip expires — the design works as described.
> * Cache finalization is _not_ exposed to a poisoned tip — `isFinalizedForCacheWrite` uses the consensus baseline (`getLatestBlockForCacheFinalization`), not this tip. Good containment.
> * Both new test files pass (`go test`, run on the PR head). The round-trip test genuinely encodes the historical bug.
> 
> ### Finding 1 (medium): the published value is not guarded, and cold-start adoption bypasses the guard
> The PR body says "the write side needed zero changes — it already publishes the guarded tip," and the `adoptSharedStateTip` comment says the shared value is "the fleet-max of every pod's guarded tip." Neither is accurate. `SetRelay` publishes `max(Response.LatestBlock, SeenBlock)` (`ecosystem/cache/handlers.go:350`), and `Response.LatestBlock` is the endpoint-reported value from the reply (or the per-endpoint observation store, `rpcsmartrouter_server.go:1891`) — it passes the per-endpoint time-monotonic gate at best, never the chain-level anti-lie guard. The harvest path explicitly treats this scope distinction as the reason the bootstrap atomic needs its own gate (`rpcsmartrouter_server.go:2542`: "a lying-high endpoint passes its own per-endpoint guard").
> 
> On its own that's survivable, because receiving pods snap the value down — _except on cold start_. Confirmed empirically with a probe test on this branch: an uninitialized `ChainState` adopts a peer tip of 1,000,000,000 raw (the guard has no first-observation reference — `chain_state.go` documents this caveat itself), and an honest local observation of 23M afterwards **cannot displace it while the tip is fresh** (`SetLatestBlock` returns `advanced=false`, tip stays at 1e9). Since `getLatestBlock()` treats ChainState as the authority (`rpcsmartrouter_server.go:848`), a poisoned restarting pod mislabels archive routing and finality and resolves LATEST cache keys at the lying block until the shared entry (10 block times, max-merge makes it sticky — lower honest publishes can't overwrite it) plus the ChainState staleness window expire.
> 
> The chain: lying endpoint on any pod → unguarded publish, sticky via max-merge → any pod that restarts in that window adopts it raw. Two hardening options, either one breaks the chain:
> 
> * **At the source (preferred):** gate the publish the way the bootstrap atomic is gated — publish `min(latestKnownBlock, guardedTip)` or skip the shared-state publish when `tip > chainState accepted`. This makes the "fleet-max of guarded tips" comment true and protects every consumer.
> * **At adoption:** skip `adoptSharedStateTip` when `chainState.GetLatestBlock()` reports not-ok (uninitialized or stale). Costs the fast warm-up benefit, which is exactly when shared state is most useful — hence preferring the write-side fix.
> 
> If the bounded exposure is consciously accepted instead, the comments and PR body should at least stop claiming the published value is guarded, and a cold-start test should encode the actual behavior.
> 
> ### Minor points
> * `adoptSharedStateTip` gates on `peerTip > localRelayData.SeenBlock` (the request-stamped value) rather than the current local tip, so the "shared-state tip is newer" debug log fires even when the peer tip is older than the local ChainState tip. Correctness holds because `SetLatestBlock` is authoritative, but the log will overstate adoption frequency.
> * For chains whose spec omits `average_block_time`, the TTL floors at `ExpirationNonFinalized` (default 500ms) — shared state becomes effectively inert for those chains, versus ~1h before. The comment covers the zero case; just confirming this behavior change is intended.
> * `getSeenBlockForSharedStateMode` logs "Failed fetching state from cache for this user id" at Info level — the id is now chain-level, so the message is stale, and post-fix this line fires on every cold-start read. Trivial, pre-existing.
> 
> ### Test coverage
> Good: TTL math including the zero floor, key round-trip encoding the historical bug, per-id isolation, max-merge, and all five adopt-glue branches including nil-ChainState panic safety and the snap-down. The one missing case is the cold-start adopt (uninitialized ChainState) — which is precisely where the guard doesn't hold, and the suite's "snapped down by the guard" test only covers a _seeded_ ChainState, which may be why the gap went unnoticed.
> 
> **Recommendation:** approve after either hardening the write-side publish (preferred — it also makes the comments true) or explicitly downgrading the "guarded tip" claims and adding a cold-start adopt test that encodes the accepted exposure.
> 
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

avitenzer (approved):

avitenzer (approved):

PR #233, opened 2026-07-23
refactor(chainstate): delete the bootstrap atomic; readers take the self-healing tip (T3)

Tomelia1999:
## What

Retires the `latestBlockHeight` **bootstrap atomic** — the one tip source that never healed — and points readers at the self-healing `ChainState` tip. Also folds in two cleanups the deletion enables. Stacked on #228.

### The problem
The atomic was monotonic-max with **no downward correction**, so a cold-start or stale lie could stick for the life of the process (F1, F11). Everything else in Topic C got anti-lie guards that heal within seconds; this was the last never-healing store.

### Changes
- **Delete** the atomic, `updateLatestBlockHeight`, and the chain-accepted harvest gate. The router-wide latest-block metric is now sourced from the guarded tip.
- **Add `ChainState.GetLatestBlockAllowStale()`** — the raw, TTL-ignored tip for lenient readers. **Archive routing** (`getLatestBlockAllowStale`) reads it instead of the atomic. The tip self-heals both ways (grows under the outlier guard; re-adopts a lower block once stale), so it can't stay stuck on a lie.
- **Cold-start** is covered by the tip: it's seeded by the first tip-eligible observation (`OnTipObservation → SetLatestBlock`), which ran *before* the atomic ever did — so the atomic covered no window the tip doesn't.
- **Consolidate `getLatestBlockForCacheFinalization` into `getLatestBlock`**: once the atomic was gone the two were byte-identical (both "gated: fresh tip or 0"). Finalization calls `getLatestBlock` directly, with the "must be the gated read" safety note at the call site.
- **Rename** `getLatestBlockBestEffort → getLatestBlockAllowStale` (and the primitive to match) so the name says what you get: the tip, stale allowed.

### Safety
- **Cache finalization is behavior-unchanged** — it keeps the **gated** read (fresh tip or 0). An allow-stale value there could falsely finalize a mutable block into the long-TTL store (durable, global). This is the deliberate carve-out.
- **Archive routing** accepts the residual: a prolonged all-endpoints-silent outage can freeze the tip slightly low → at worst a retryable under-archive, self-healing the instant any endpoint reports. Non-corrupting. (`UpgradeToArchiveIfNeeded` adds the archive extension on retry #1, so an under-archived request that node-errors on a pruning node escalates — the cost is a wasted round-trip, not a wrong answer.)

### ⚠️ Metric semantics change — `smartrouter_latest_block`

Deleting the atomic changes this gauge in ways worth flagging to whoever owns the dashboards. **Nothing here is a bug; it is the self-heal becoming visible.**

- **It is no longer monotonic.** It now follows the guarded tip, which can go **down** — a stale tip re-adopting a lower block, or `Recompute` snapping an over-band tip to the baseline. The old CAS ratchet made it monotonic-max for the process lifetime, so it could only ever climb. **Any panel or alert built on `rate()` / `increase()` over this gauge, or on "latest block never goes backwards", changes behavior.** A decrease is now expected and means the anti-lie guard corrected a bad tip.
- **Its source is the guarded tip, not the raw harvested block.** A lying-high endpoint that ChainState rejects no longer reaches the metric. This is exactly what the deleted chain-accepted gate protected; it now falls out of reading `GetLatestBlock()` directly, and is pinned by a test (below).
- **It is written on every accepted tip-eligible harvest**, not only when the value advances. This sits on the same per-relay path as the existing per-endpoint `rpc_endpoint_latest_block` write, so it is ~2× on an already-per-relay cost rather than a new class of cost.
- **A TTL-stale tip leaves it untouched** — the gauge holds its last reported value rather than dropping to 0, so "no fresh tip" reads as a flat line, not a cliff.

### Getter set is now minimal
`ChainState`: `GetLatestBlock` (fresh), `GetLatestBlockWithTime` (fresh + age), `GetLatestBlockAllowStale` (raw). Router: `getLatestBlock` (gated), `getLatestBlockAllowStale` (stale-tolerant). The finalization duplicate is gone.

## Tests
- Rewrote the atomic tests to assert the **tip / archive-routing reader**: a lying-high harvest stays guarded by ChainState, and the archive tip **heals DOWN** where the atomic stayed stuck.
- `TestHarvest_ArchiveRoutingTipGuardedByChainState` also pins the **metric** contract: after T3 the gated `GetLatestBlock()` read in `harvestAndUpdateTipFromRelay` has exactly one consumer (`smartrouter_latest_block`), and nothing asserted it — the deleted `latestBlockHeight.Load()` used to cover that incidentally. It now asserts the gauge follows honest advances, holds the guarded `1050` when ChainState rejects a lying-high `1_000_000`, and is untouched when the harvest bails at the store gate. Verified to fail (actual `1e+06`) when the source is reverted to the raw `tip`.
- Two server tests now wire a **seeded chainState** instead of the atomic.
- New `TestGetLatestBlockAllowStale` (chainstate) + `TestGetLatestBlockAllowStale_ServesStaleAndHeals` (router).

`go build ./...` + `go vet` clean; full `chainstate`, `endpointstate` and `rpcsmartrouter` suites pass (`chainstate` also under `-race`).

### DIFF
--- a/protocol/chainstate/chain_state.go
+++ b/protocol/chainstate/chain_state.go
@@ -315,12 +315,12 @@ func (cs *ChainState) Reset() {
 // turned a cold-start lie into a process-lifetime one, is retired (T3); every reader goes through
 // this tip. Do not reintroduce a ratcheting mirror of it without its own downward path.
 //
-// Returns the resulting tip, the time it was last confirmed, and whether this call ADVANCED it
-// (equal-block confirmations and downward re-adoptions return advanced=false). advanced is the
-// accept/reject signal for a caller that needs to know whether its observation actually moved the
-// tip: the OnTipObservation hook discards it, adoptSharedStateTip LOGS it (as `adopted`) so a
-// rejected peer tip is visible rather than silent, and the package tests assert guard behaviour
-// through it.
+// Returns the resulting tip, the time it was last confirmed, and whether this call moved the tip
+// UP. advanced is not an accept/reject signal: equal-block confirmations, downward re-adoptions,
+// and rejected observations all return false. The OnTipObservation hook discards it;
+// adoptSharedStateTip logs it as `tip_advanced` and separately compares the resulting tip with the
+// positive peer input to determine whether that value was taken; package tests assert guard
+// behaviour through it.
 func (cs *ChainState) SetLatestBlock(block int64) (latest int64, at time.Time, advanced bool) {
 	cs.mu.Lock()
 	defer cs.mu.Unlock()
--- a/protocol/chainstate/chain_state.go
+++ b/protocol/chainstate/chain_state.go
@@ -181,8 +181,11 @@ type ChainState struct {
 	// would report ~0 lag forever, since every confirming observation refreshes it.
 	latestSince time.Time
 	// initialized is a STICKY flag: true once any positive observation has ever been accepted.
-	// It distinguishes a genuine cold start (no observation yet → bootstrap fallback is allowed)
-	// from a tip that has merely gone stale by TTL (Finding 1). It never resets to false.
+	// SetLatestBlock reads it to know whether a REFERENCE exists to guard the next observation
+	// against — the very first observation has none, which is why no anti-lie check can fire on it.
+	// It also separates a genuine cold start from a tip that has merely gone stale by TTL (Finding
+	// 1): both report (0, false) to readers, but only the latter has a value worth healing. Ordinary
+	// staleness never reverts it; only Reset() clears it back to false.
 	initialized bool
 	baseline    int64 // last computed strict-majority consensus baseline (valid iff hasBaseline)
 	hasBaseline bool  // whether a fresh majority existed at the last Recompute
@@ -264,6 +267,11 @@ func (cs *ChainState) SetDebugClockOffset(d time.Duration) {
 // any peer forms a baseline. Reset is the only way to drop such a value immediately rather than
 // waiting for the liar to stop and a TTL to expire. Debug-only: reached from /debug/reset-all and
 // /debug/reset-scores, never from a relay path.
+//
+// Know one consequence before pulling it: clearing the tip also blanks GetLatestBlockAllowStale, so
+// archive routing loses its last-known head and force-archives EVERY concrete-block request (the
+// conservative answer to an unknown head, archive_parser_rule.go) until the next observation lands —
+// roughly half a block time on a polling pod. Correct, but briefly expensive on the archive pool.
 func (cs *ChainState) Reset() {
 	cs.mu.Lock()
 	defer cs.mu.Unlock()
@@ -300,14 +308,17 @@ func (cs *ChainState) Reset() {
 //     poisoning of THIS tip to ~one TTL instead of the process lifetime.
 //
 // Caveat: the guard cannot fire on the VERY FIRST observation (no reference exists yet), so a
-// cold-start lie is accepted here. This tip self-heals after one TTL, but callers that ratchet a
-// separate monotonic-max store from accepted observations (the rpcsmartrouter bootstrap atomic)
-// do NOT self-heal — a cold-start lie persists there for the process lifetime. That store must add
-// its own bound if it needs one; this tip's self-heal does not cover it.
+// cold-start lie is accepted here. It is bounded, not permanent: the stale-tip re-adoption above
+// drops it within one TTL, and on a pod that forms a baseline Recompute snaps it back sooner. That
+// self-heal is now the WHOLE bound, and it suffices — no consumer keeps a separate monotonic-max
+// copy of this value any more. The rpcsmartrouter bootstrap atomic, which was the one such copy and
+// turned a cold-start lie into a process-lifetime one, is retired (T3); every reader goes through
+// this tip. Do not reintroduce a ratcheting mirror of it without its own downward path.
 //
 // Returns the resulting tip, the time it was last confirmed, and whether this call ADVANCED it
-// (equal-block confirmations and downward re-adoptions return advanced=false — consumers that
-// ratchet a monotonic maximum, e.g. the bootstrap atomic, must only follow advances).
+// (equal-block confirmations and downward re-adoptions return advanced=false). advanced is the
+// accept/reject signal for a caller that needs to know whether its observation actually moved the
+// tip; the production hooks discard it, and the package tests assert guard behaviour through it.
 func (cs *ChainState) SetLatestBlock(block int64) (latest int64, at time.Time, advanced bool) {
 	cs.mu.Lock()
 	defer cs.mu.Unlock()
@@ -440,9 +451,14 @@ func (cs *ChainState) ConsensusBucketWidth() int64 {
 }
 
 // Initialized reports whether ChainState has ever accepted a positive observation. It is sticky
-// — once true it never reverts — so callers can distinguish a genuine cold start (allow a
-// bootstrap fallback) from a tip that has merely gone stale by TTL (do NOT revive a frozen
-// value, Finding 1).
+// within a lifetime — ordinary staleness never reverts it, only Reset() clears it — so a caller can
+// distinguish a genuine cold start (no value has ever existed) from a tip that has merely gone stale
+// by TTL (a real value exists but must not be served frozen, Finding 1).
+//
+// NOTE: no production reader consults this any more. It existed so getLatestBlock could allow a
+// bootstrap fallback on cold start while refusing to revive a frozen value once stale; that fallback
+// (the bootstrap atomic) is retired (T3) and every reader now takes the tip or 0. Retained for tests
+// and diagnostics — delete it if no future consumer appears.
 func (cs *ChainState) Initialized() bool {
 	cs.mu.RLock()
 	defer cs.mu.RUnlock()
--- a/protocol/chainstate/chain_state.go
+++ b/protocol/chainstate/chain_state.go
@@ -318,7 +318,9 @@ func (cs *ChainState) Reset() {
 // Returns the resulting tip, the time it was last confirmed, and whether this call ADVANCED it
 // (equal-block confirmations and downward re-adoptions return advanced=false). advanced is the
 // accept/reject signal for a caller that needs to know whether its observation actually moved the
-// tip; the production hooks discard it, and the package tests assert guard behaviour through it.
+// tip: the OnTipObservation hook discards it, adoptSharedStateTip LOGS it (as `adopted`) so a
+// rejected peer tip is visible rather than silent, and the package tests assert guard behaviour
+// through it.
 func (cs *ChainState) SetLatestBlock(block int64) (latest int64, at time.Time, advanced bool) {
 	cs.mu.Lock()
 	defer cs.mu.Unlock()
--- a/protocol/chainstate/chain_state.go
+++ b/protocol/chainstate/chain_state.go
@@ -9,9 +9,13 @@
 // touches QoS or endpoint.Enabled, and nothing external writes its consensus — there is no
 // SetMajorityBaseline API (T2, inverted: ChainState owns consensus; the probe is read-only).
 //
-// It exposes ONE read API for "what block is the chain at" (Topic C governing decision, D1/D4):
-//   - GetLatestBlock / GetLatestBlockWithTime — the tip. Every consumer reads this: archive
-//     routing, cache finalization, sync scoring, and consistency.
+// It exposes the read API for "what block is the chain at" (Topic C governing decision, D1/D4):
+//   - GetLatestBlock / GetLatestBlockWithTime — the tip when TTL-fresh, else not-found (never a
+//     stale value). Cache finalization, sync scoring, consistency, and the request seenBlock read this.
+//   - GetLatestBlockAllowStale — the same tip but TTL-ignored (0 only if never observed), for archive
+//     routing, which prefers a slightly-stale head to "unknown". Both self-heal via SetLatestBlock /
+//     Recompute (grow under the outlier guard, re-adopt a lower block once stale), so neither can stay
+//     stuck on a lie the way the retired bootstrap atomic could.
 //
 // Consensus is NOT a readable value — it is background machinery that corrects the tip at the
 // band edges only (see Recompute): snap DOWN when the tip leads consensus by more than
@@ -357,6 +361,18 @@ func (cs *ChainState) GetLatestBlock() (int64, bool) {
 	cs.mu.RLock()
 	defer cs.mu.RUnlock()
 	return cs.freshLatestLocked(cs.effectiveNow())
+}
+
+// GetLatestBlockAllowStale returns the current tip value IGNORING TTL freshness — the last accepted,
+// self-healing observation (0 if none yet). It is for lenient readers (archive routing) that
+// prefer a slightly-stale head over an "unknown" that would force every request to the archive
+// pool. The tip still self-heals: SetLatestBlock grows it under the outlier guard and re-adopts a
+// lower block once stale, so this can lag but cannot stay stuck on a lie (unlike the retired
+// bootstrap atomic). Freshness-sensitive readers (cache finalization) use GetLatestBlock instead.
+func (cs *ChainState) GetLatestBlockAllowStale() int64 {
+	cs.mu.RLock()
+	defer cs.mu.RUnlock()
+	return cs.latestBlock
 }
 
 // GetLatestBlockWithTime is GetLatestBlock plus the wall-clock at which the current tip BLOCK was
--- a/protocol/chainstate/chain_state_test.go
+++ b/protocol/chainstate/chain_state_test.go
@@ -28,6 +28,28 @@ func newTestState(t *testing.T) (*ChainState, *fakeClock) {
 	return cs, clk
 }
 
+// TestGetLatestBlockAllowStale covers the raw (TTL-ignoring) getter T3's archive routing relies on: it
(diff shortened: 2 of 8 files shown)


avitenzer (commented):
## Review

**Framing.** The architectural move is worth naming: replacing a **monotonic-max ratchet** with a **self-healing authoritative value**. A monotonic-max store is attractive because it never regresses — but "never regresses" is exactly the bug when the input can lie: one bad high observation is absorbed permanently, with no code path that can ever lower it. `ChainState`'s tip instead accepts a downward correction from two independent sources (`Recompute`'s band snap, and TTL-expiry re-adoption), which bounds any poisoning to ~one TTL rather than the process lifetime. The second pattern — one stored value with two read contracts (`GetLatestBlock` = fresh-or-not-found, `GetLatestBlockAllowStale` = raw) — is better than two stored values, because the divergence lives in the reader's policy rather than in a second copy that can drift.

### Verification run

| Check | Result |
|---|---|
| `go build ./...`, `go vet` on both packages | clean |
| `go test ./protocol/chainstate/...` | pass (0.9s) |
| `go test ./protocol/rpcsmartrouter/...` (full suite, `-count=1`) | pass (83s) |
| Dangling refs to `latestBlockHeight` / `getLatestBlockBestEffort` / `getLatestBlockForCacheFinalization` | none, incl. sibling `lava` and enterprise-internal checkouts |

Two load-bearing claims from the description confirmed independently rather than taken on trust:

- **"The atomic covered no window the tip doesn't."** Verified structurally: the atomic's only writer was `harvestAndUpdateTipFromRelay`, and its gate was `chainAccepted = ok && tip <= accepted`. With `chainState` non-nil (unconditional at `rpcsmartrouter_server.go:189`), `ok == false` on an uninitialized ChainState means the atomic could never hold a non-zero value while the tip was still cold. The cold-start fallback was genuinely unreachable-as-distinct.
- **"Allow-stale can't reach finalization."** Traced the consumer: `ExtensionInfo.LatestBlock` is read only by `ExtensionParser.ExtensionParsing` → `ArchiveParserRule.isPassingRule` (`extension_parser.go:60,70`). No path into the cache key or `isFinalizedForCacheWrite`. The carve-out holds mechanically, not just by convention.

### What's good

- The getter set really is minimal now, and the naming is honest — `getLatestBlockAllowStale` states its contract where `getLatestBlockBestEffort` invited a caller to assume "best" meant "safest".
- Collapsing the finalization duplicate with the safety note moved to the **call site** (`rpcsmartrouter_server.go:2868`) is the right placement: the invariant is "this call must use the gated read", which is a property of the call, not of the function.
- `GetLatestBlockAllowStale`'s `raw > 0` check at `rpcsmartrouter_server.go:874` is not redundant defensive noise — it prevents an `int64 → uint64` wrap producing a colossal "latest", which would silently disable archive routing. Worth keeping.
- The test rewrites are genuine behavioral rewrites, not mechanical renames: `TestGetLatestBlockAllowStale_ServesStaleAndHeals` asserts the heal-*down* property, which is the whole thesis of the change and was untestable before.

---

## 1. Stale documentation — primary finding (5 sites)

This PR is ~200 deleted lines and mostly comment rewriting; the encoded reasoning *is* the deliverable, so a dangling reference costs more here than in a typical refactor. Ranked:

**a. `protocol/chainstate/chain_state.go:292-296` — worst.** This doesn't merely name a dead thing, it *instructs* a future maintainer:

> `// ...callers that ratchet a separate monotonic-max store from accepted observations (the rpcsmartrouter bootstrap atomic) do NOT self-heal — a cold-start lie persists there for the process lifetime. That store must add its own bound if it needs one; this tip's self-heal does not cover it.`

A reader six months out will go hunting for a store that no longer exists, and may conclude a required bound is missing. Should become: the cold-start lie is bounded to ~one TTL, and there are no longer any ratcheting consumers.

**b. `chain_state.go:298-300`** — the *only* stated justification for `SetLatestBlock`'s `advanced` return:

> `// ...consumers that ratchet a monotonic maximum, e.g. the bootstrap atomic, must only follow advances`

That consumer is deleted by this PR. Both remaining non-test callers (`rpcsmartrouter_server.go:222`, `:2767`) discard all three returns, so `advanced` now has test-only consumers. **Don't delete the return** — the tests legitimately use it to assert accept/reject, which is a real signal. Just re-justify it as an accept/reject signal for callers and tests.

**c. `protocol/rpcsmartrouter/chainstate_glue_test.go:307` and `:313`** — self-inflicted here, and the cheapest fix. The doc block still leads with the **old function name** (`TestCacheServedReply_DoesNotPoisonBootstrapAtomic`) while the function below is renamed to `TestCacheServedReply_DoesNotMoveTip`; line 313 still describes "the bootstrap atomic that `getLatestBlock` falls back to on cold start" — the fallback this PR removes.

**d. `protocol/endpointstate/observation.go:164`** — outside the diff, so this is the one that rots silently:

> `// The caller uses this to gate the remaining ungated tip-state writes (router bootstrap atomic, per-endpoint metric)`

Only the per-endpoint metric remains.

**e. `protocol/rpcsmartrouter/rpcsmartrouter_server.go:3172`** — the cache-hit branch still says a cached `reply.LatestBlock` "must NOT feed the bootstrap atomic (it would freeze a stale value during cold start)".

Minor, pre-existing: `crossValidationFinalityLabel` (`:576`) still advertises "chain-tracker/estimator-aware latest block, not just the cached counter" — both the estimator and the counter are gone. Free to fix while you're in here.

## 2. Two assertions downgraded to prose (test coverage)

In `relay_observation_harvest_test.go`, both removed assertions were replaced by comments rather than tip-equivalent checks:

- **`:461-463`** is the one that matters. The invariant *"the generation gate blocks the downstream tip-state write"* now has no assertion at all — only a comment claiming it. That's exactly the class of property that silently stops holding. It is restorable: wire `chainState` plus `OnTipObservation: func(b int64) { cs.SetLatestBlock(b) }` and assert `cs.GetLatestBlockAllowStale() == 0`. The pattern already exists a couple of files over at `chainstate_glue_test.go:357-372` — this test currently builds its monitor without the hook, which is why the check wasn't a straight substitution.
- `TestHarvestAndUpdateTipFromRelay_HistoricalDoesNotPoisonTip` lost its positive assertion ("a tip-eligible response moves the router-wide value"). Lower priority — `TestHarvest_ArchiveRoutingTipGuardedByChainState` covers it directly.

## 3. `smartrouter_latest_block` changed monotonicity — worth a line in the description

`SetRouterLatestBlock` moved from a monotonic CAS-guarded update to a plain gauge `Set` on every accepted harvest (`rpcsmartrouter_server.go:2487-2497`), so the series can now **decrease**. Intended, and arguably the point — but it's an operator-visible change the description doesn't mention.

Blast radius, already checked: the metric is `smartrouter_latest_block`, declared a `GaugeVec` (`metrics/smartrouter_metrics_manager.go:300`). The only in-repo references are `docs/METRICS.md:105` (documents it as a Gauge — still accurate) and a passing mention in `docs/CHAINTRACKER-PROBING-AUTOMATION.md`. **Nothing in-repo applies `rate()`/`increase()`/`deriv()` to it**, and there is no `dashboards/`/`monitoring/` directory. So this is informational: one line in the PR body, plus a glance at any external Grafana panel that graphs it.

**Follow-up, explicitly not a change request for this PR:** the metric is written only from the relay-harvest path, so a low-traffic pod's gauge goes stale even while the poller advances the tip. Now that the tip is the single source, driving the metric from the `OnTipObservation` hook (`rpcsmartrouter_server.go:222`) would cover the poll path *and* remove the slightly awkward re-read-under-a-second-lock immediately after `SetLatestBlock` already ran inside `recordRelayBlockObservation`. The poll gap is pre-existing, and this is a deletion PR — expanding it into metrics plumbing invites scope creep.

## 4. `ChainState.Reset()` now blanks archive routing too — doc note

`Reset()` (reachable from `/debug/reset-all`, `rpcsmartrouter.go:617`) zeroes `latestBlock`, so `getLatestBlockAllowStale()` returns 0 and `ArchiveParserRule` force-archives **every** concrete-block request until the next observation (`archive_parser_rule.go:33`). Previously the atomic survived a Reset and kept archive routing working. The window is short (~half a block time until the next poll) and force-archive-when-unknown is the conservative answer, so the behavior looks correct — it's just undocumented. Worth a sentence in `Reset`'s doc block (`chain_state.go:250-268`), which currently discusses only the tip, not this now-coupled routing consequence.

## 5. PR mechanics

- **Base is `t10-shared-chaintip`, not `main`.** This repo disallows merge commits and squash-merges, so once T10 squashes into `main` this branch's base commit ceases to exist and the PR needs retargeting — after which the diff can look surprising until it's rebased onto the squashed base. Sequence: merge #228 → merge T10 → retarget this PR → re-verify the diff before approving.
- `21d0112 "unit test"` doesn't follow the repo's conventional-commit style. Mostly harmless under squash-merge since the PR title becomes the commit subject, but it's noticeable in the branch history.

---

## Verdict

**Looks good after the doc cleanups.** The deletion is correct and well-reasoned, the gated/lenient split is contained where the description says it is, the cold-start-window argument holds structurally, and the full suite passes. Nothing here is a functional blocker.

The one item worth insisting on is `chain_state.go:292-296`, because it actively misdirects. `chainstate_glue_test.go:307` is a two-word fix, and `endpointstate/observation.go:164` is the one that will rot unnoticed if it isn't caught in this PR.

avitenzer:
**Heads up: I pushed `2624c60` to this branch** — the doc fixes from my review above, so you don't work through items that have already landed. Comment-only, no behaviour change.

### Addressed by `2624c60`

- **Finding 1 (stale docs)** — all five listed sites, plus **two more the sweep turned up that the review didn't list**: `chain_state.go`'s `initialized` field doc and `Initialized()` itself. Both described the retired cold-start-vs-stale fallback decision. I'd grepped for `bootstrap atomic`; these two say "bootstrap **fallback**", which is why the review missed them.
  - Worth knowing while you're in there: **`Initialized()` now has zero production callers.** Its only consumer was `getLatestBlock`'s cold-start branch, which this PR deletes. The underlying `initialized` *field* is still load-bearing (`SetLatestBlock` reads it to know whether a guard reference exists), so the field stays — but the exported accessor is test-only now. I left a note saying as much rather than deleting the method; your call.
  - `advanced` is **kept**, not deleted — the package tests legitimately assert guard behaviour through it. Only its justification changed.
- **Finding 4** — `Reset()`'s doc block now records that clearing the tip also blanks `GetLatestBlockAllowStale`, so archive routing force-archives every concrete-block request until the next observation (~half a block time).
- **The minor `crossValidationFinalityLabel` note** — dropped the retired chain-tracker/estimator/"cached counter" description.

### Two things in that commit to look at deliberately

1. I added *"delete it if no future consumer appears"* to `Initialized()`'s doc — a suggestion embedded as a comment on a method this PR doesn't otherwise touch. Remove it if that reads as overreach.
2. Both `initialized` docs claimed the flag "never resets to false", but `Reset()` clears it and `chain_state_test.go:654` asserts exactly that. Corrected. **This is a pre-existing inaccuracy unrelated to T3** — please don't read it as fallout from the atomic deletion.

### Still open (not in `2624c60`)

- **Finding 2** — `relay_observation_harvest_test.go:461`. The generation-gate invariant is still asserted only in prose. That needs a real assertion (wire `chainState` + `OnTipObservation`, assert `GetLatestBlockAllowStale() == 0`), i.e. a code change rather than a doc edit, so I left it to you. Happy to push it if you'd rather.
- **Finding 3** — the one-line note in the PR description about `smartrouter_latest_block` becoming non-monotonic. Yours to edit.
- **Finding 5** — base is still `t10-shared-chaintip`; the merge-T10-then-retarget sequence still applies.

Verification on `2624c60`: `gofmt` clean on touched files, `go build ./...` + `go vet` clean, and `chainstate`, `endpointstate` and the full `rpcsmartrouter` suite (84s) all pass.

Tomelia1999:
## Post-review changes — `e428a01` + `3adace8`

Thanks for the thorough pass. Everything from Finding 1 and Finding 4 verified as landed, Finding 3 is now in the description, and I pushed two commits. Two of your items I did **not** implement as specified — evidence below, happy to be overruled.

### First, a heads-up on the SHAs

`2624c60` and `21d0112` no longer exist on this branch. It was rebased on 2026-07-27T10:07, so they're unreachable from `origin/t3-delete-bootstrap-atomic` (they survive locally as dangling objects, which is why they still resolve for you). The content landed as:

| Your SHA | On branch | Patch |
|---|---|---|
| `2624c60` (docs) | `cb2b640` | identical — only index hashes and `@@` offsets differ, same diffstat |
| `21d0112` ("unit test") | `c573992` | byte-identical |

Flagging so nobody goes hunting for them or force-pushes over the rebase.

### `advanced` is not test-only — and the doc commit repeated the error (fixed in `3adace8`)

> Both remaining non-test callers (`rpcsmartrouter_server.go:222`, `:2767`) discard all three returns, so `advanced` now has test-only consumers.

`:222` discards. `:2766` does not:

```go
tip, _, advanced := rpcss.chainState.SetLatestBlock(peerTip)
utils.LavaFormatDebug("shared-state tip fed to chain state", ..., utils.LogAttr("adopted", advanced), ...)
```

That's `adoptSharedStateTip`, added by `4efdb67` — a sibling commit on this branch, one before the T3 commit. So `advanced` has a live production consumer.

The replacement text in `cb2b640` inherited it: `chain_state.go:321` read *"the production hooks discard it."* Since the whole point of that commit was retiring inaccurate comments, I've re-justified it to name all three consumers (hook discards, `adoptSharedStateTip` logs, tests assert).

### Finding 2 — I did not add the tip assertion. It's the wrong assertion at this seam.

Your instinct that prose-replacing-an-assertion deserves scrutiny was right, and it led somewhere real — but one line further down. Three findings, each mutation-tested rather than reasoned:

**1. The tip invariant is already asserted, one layer down, better.** `endpointstate/chainstate_feed_test.go`:

```go
// A stale-generation relay is rejected by the gate → no tip feed.
m.RecordRelayObservation(url, gen+1, 999, now.Add(3*time.Millisecond))
require.Equal(t, []int64{100, 200}, sink.snapshot(), "a rejected (stale-gen) relay feeds no tip")
```

Disabling the generation gate fails that test with exactly that message. It records hook invocations via a sink, so it asserts *the hook did not fire* rather than *a downstream value is still 0* — and it's non-vacuous by construction, proving the hook live via two prior accepted observations before the negative case.

**2. The proposed probe is vacuous.** I implemented it exactly as specified — wired `chainState` + `OnTipObservation`, asserted `GetLatestBlockAllowStale() == 0`. Then I removed the `OnTipObservation` wiring entirely, so nothing could feed ChainState under any circumstances. **It still passed.** You deprioritised the positive assertion as "lower priority", but that's precisely the control that makes the negative one sound — they aren't separable.

**3. It adds no detection power here.** With the gate disabled, the existing store assertion fires first and `require` aborts before the new one runs. They're also logically equivalent: the tip is fed iff `endpointtip.Set()` returns true, which is exactly what `GetObservation().LatestBlock` reflects. tip-moved ⟺ store-accepted.

**The comment was also inverted, not just stale.** `OnTipObservation → SetLatestBlock` fires from a `defer` registered *inside* `RecordRelayObservation` — it does not run after the early return in `harvestAndUpdateTipFromRelay`. The tip doesn't move because the gate left `tipBlock` at 0, not because the harvest "bailed before" it. **The early return guards the metrics, not the tip.** Implementing the fix would have encoded that inverted story into an assertion.

### What I pinned instead

The per-endpoint metric write (`SetEndpointLatestBlock`, raw `tip`) sits after the early return and is the only thing it actually protects. I hoisted that write above the gate — the mutation that lets a stale-generation relay publish `rpc_endpoint_latest_block` for an endpoint that never reported the block — and the **entire** `rpcsmartrouter` + `endpointstate` + `chainstate` suite passed. Genuinely undetected.

`3adace8` pins it, with a positive control after the live-generation harvest so the rejection can't pass vacuously (verified: unwiring the metrics manager fails the control). Corrected comment now points at `chainstate_feed_test.go` as the owner of the tip invariant.

### `e428a01` — the router-wide metric

Same class of gap, found before your review landed: after T3 the gated `GetLatestBlock()` read in `harvestAndUpdateTipFromRelay` has exactly one consumer — `smartrouter_latest_block` — and nothing asserted it. The rewritten test checks `getLatestBlockAllowStale()`, a different getter, so a regression reporting the raw `tip` would publish a lying-high block router-wide unnoticed. Now asserted; verified to fail with `actual: 1e+06` when the source is reverted to raw `tip`.

### Two smaller notes

- **Merge policy:** actual settings are `merge: false`, `squash: true`, `rebase: true`. Merge commits are disallowed, but squash *is* allowed — "disallows merge commits and squash-merges, so once T10 squashes into main" is self-contradictory. Your sequencing advice stands regardless, since squash is the likely path.
- **`raw > 0`:** it can't prevent an `int64 → uint64` wrap, because `latestBlock` is never negative — `SetLatestBlock` rejects `block <= 0`, `Reset` writes 0, `Recompute` writes `baseline`. All five write sites are non-negative. Worth keeping as defence, just not on that rationale.

### Agreed and adopted

- Your follow-up that the metric is written only from the relay-harvest path is correct — a low-traffic pod's gauge goes stale while the poller advances the tip. Agreed it's out of scope for a deletion PR; worth its own ticket.
- The `Initialized()` note reads fine to me, not overreach. "Test-only" is accurate — `chain_state_test.go` uses it at 217/221/224/231/698/705, plus the debug-snapshot field.
- Finding 5 sequencing (merge #228 → merge T10 → retarget → re-verify diff) still applies; nothing done here changes that.

**Verification on `3adace8`:** gofmt clean, `go build ./...` + `go vet` clean, and `chainstate` (also under `-race`), `endpointstate` and the full `rpcsmartrouter` suite all pass.

avitenzer (approved):

PR #234, opened 2026-07-27
fix(stack): land T4 (#226) and T10 (#228) on main after out-of-order stack merges

Tomelia1999:
## What

Lands **#226 (T4)** and **#228 (T10)** on `main`. Both are already approved and marked Merged, but neither reached `main` — this PR carries their combined work the last step.

No new code. Every commit here has already been reviewed and merged in #226 or #228.

## Why this is needed

The stack was merged **bottom-up without retargeting**, so each merge landed in a branch that had already been copied upward:

| Time | PR | Merged into | Reached main? |
|---|---|---|---|
| 11:14:49 | #225 | `main` | ✅ |
| 11:15:28 | #226 | `unify-seenBlock-chainTip` | ❌ — `unify` had gone to main 39s earlier |
| 11:15:56 | #228 | `t4-block-monotonic-endpoint-tip` | ❌ — `t4` had gone to `unify` 28s earlier |

Each PR still pointed at its original stacked parent rather than `main`, so #226 and #228 merged correctly *into the stack* and never propagated. Ancestry confirms it — no branch is an ancestor of any other:

```
main   in-main      NOT-in-unify  NOT-in-t4
unify  NOT-in-main  in-unify      NOT-in-t4
t4     NOT-in-main  NOT-in-unify  in-t4
```

`t4-block-monotonic-endpoint-tip` already holds **both** T4 and T10 (because #228 merged into it), so a single PR from it closes the gap. #226 and #228 cannot be reused — GitHub will not re-target or re-merge a merged PR.

## Contents — 6 commits

**T4** (from #226)
- `bf904d5` refactor(endpointtip): block-monotonic per-endpoint tip with a staleness backstop
- `5869971` refactor(endpointtip): anchor freshness stamp forward; fix staleness-window docs
- `45ac2ff` docs(endpointtip): correct stale time-monotonic comments to block-monotonic (T4)
- `1ab55ee` refactor(probing): derive liveness horizon via chainstate.StalenessWindow

**T10** (from #228)
- `31bd069` fix(shared-state): rebuild cross-pod consistency on the guarded chain tip (T10)
- `936b04f` refactor(shared-state): log adopt outcome so a rejected peer tip is visible

`14 files changed, 432 insertions(+), 132 deletions(-)`

## ⚠️ The commit list shows 9, not 6 — this is expected

[...]

PR #235, opened 2026-07-27
fix(stack): land T4 (#226) and T10 (#228) on main after out-of-order stack merges

Tomelia1999:
## What

Lands **#226 (T4)** and **#228 (T10)** on `main`, **plus 3 commits addressing this PR's own review**. Both stacked PRs are already approved and marked Merged, but neither reached `main` — this PR carries their combined work the last step.

> **Updated:** this PR originally carried no new code. It now also contains 3 review-fix commits (`ee754cf`, `708b810`, `be4b2fc`) — see [Review fixes](#review-fixes) below. Commits `ac2165c`…`3e626e7` are still exactly as reviewed in #226/#228.

## Why this is needed

The stack was merged **bottom-up without retargeting**, so each merge landed in a branch that had already been copied upward:

| Time | PR | Merged into | Reached main? |
|---|---|---|---|
| 11:14:49 | #225 | `main` | ✅ |
| 11:15:28 | #226 | `unify-seenBlock-chainTip` | ❌ — `unify` had gone to main 39s earlier |
| 11:15:56 | #228 | `t4-block-monotonic-endpoint-tip` | ❌ — `t4` had gone to `unify` 28s earlier |

Each PR still pointed at its original stacked parent rather than `main`, so #226 and #228 merged correctly *into the stack* and never propagated. Ancestry confirms it — no branch is an ancestor of any other:

```
main   in-main      NOT-in-unify  NOT-in-t4
unify  NOT-in-main  in-unify      NOT-in-t4
t4     NOT-in-main  NOT-in-unify  in-t4
```

`t4-block-monotonic-endpoint-tip` already holds **both** T4 and T10 (because #228 merged into it), so a single PR from it closes the gap. #226 and #228 cannot be reused — GitHub will not re-target or re-merge a merged PR.

## Contents — 9 commits

**T4** (from #226, unchanged)
- `ac2165c` refactor(endpointtip): block-monotonic per-endpoint tip with a staleness backstop
- `42be4f8` refactor(endpointtip): anchor freshness stamp forward; fix staleness-window docs
- `8fe9926` docs(endpointtip): correct stale time-monotonic comments to block-monotonic (T4)
- `739a755` refactor(probing): derive liveness horizon via chainstate.StalenessWindow

**T10** (from #228, unchanged)
- `d374308` fix(shared-state): rebuild cross-pod consistency on the guarded chain tip (T10)
- `3e626e7` refactor(shared-state): log adopt outcome so a rejected peer tip is visible

**Review fixes** (new — see below)
- `ee754cf` fix(cache): floor the shared-state tip TTL at a hard minimum
- `708b810` docs(rpcsmartrouter): fix the T4 doc the sweep missed; separate adopt outcomes
- `be4b2fc` docs(endpointtip): document the one write that stores a hybrid tip triple

`14 files changed, 546 insertions(+), 146 deletions(-)`

<a name="review-fixes"></a>
## Review fixes

Three commits addressing findings from the review on this PR. All three concern code inherited from #226/#228 — none was introduced by the re-land. **Only the first changes behavior.**

### `ee754cf` — shared-state tip TTL floor (behavior change)

`SharedStateTipExpiration` floored the tip TTL at `ExpirationNonFinalized`, justified in its docstring as guaranteeing "a sane, non-zero TTL (a zero TTL would never expire)". `ExpirationNonFinalized` cannot underwrite either half:

- **Not guaranteed non-zero.** It is `flag * multiplier` (`server.go:75`) and *both* are operator-settable, so `--expiration-non-finalized=0` makes it `0`. Ristretto reads a zero TTL as *never expires* — a dead pod's tip would be stranded in the shared cache permanently, the exact failure the floor exists to prevent.
- **Not guaranteed useful.** At its 500ms default it is ~3 orders of magnitude short for a chain tip. A spec omitting `average_block_time` yields `0 * 10 = 0`, so the published tip evaporates between relays and shared state silently degrades back to off — the condition T10 exists to fix.

The root cause is borrowing a floor from an unrelated concept: `ExpirationNonFinalized` answers "how long may a cached RPC *response* be reused?" (short on purpose), while this answers "how long is a peer pod's *chain tip* worth knowing?" (should track the staleness horizon).

Adds `MinSharedStateTipExpiration = 2s` as a third, independent `max()` term, mirroring the router-side floor (`chainstate.minStalenessWindow`) without `ecosystem/cache` depending on `protocol/`. Deliberately a constant rather than a flag — a tunable minimum reintroduces the same hole.

**Blast radius is one case: a spec with no `average_block_time`.** Since `abt * 10 >= 2s` for any `abt >= 200ms`, every real chain (ETH 12s, Solana 400ms) is unchanged.

### `708b810` — T4 doc the sweep missed + adopt-outcome logging (no behavior change)

- `endpointTipPreferStore`'s doc still asserted the retired contract: *"fed … under a TIME-monotonic guard, so it always reflects the newest observation — including a newer LOWER block after a reorg/rollback."* Under T4 that second clause is false — a lower block is held off while the stored higher tip is fresh, so the downward correction lags by up to the staleness window. This is the same correction `8fe9926` applied to `TestEndpointTipPreferStore`'s rationale; the test-side twin was fixed while the production doc was not, leaving the only surviving `time-monotonic` reference in the tree.
- `adoptSharedStateTip` logged `adopted=advanced`, but `SetLatestBlock` returns `advanced=false` in **two opposite** situations: the guard rejected the peer value, *and* a stale local tip was re-adopted **downward** to it (`chain_state.go:324-328` sets the tip, then returns false). So the field `3e626e7` added specifically to make a rejected peer tip visible could not distinguish rejection from acceptance. Now logs `peer_tip_taken` (`tip == peerTip`), which maps correctly across all five `SetLatestBlock` outcomes, alongside `tip_advanced` under a name that says what it means. `TestAdoptSharedStateTip_StaleTipTakesLowerPeerValue` pins the distinction against the real `ChainState`.
- The same function's doc credited `SetLatestBlock` with "outlier/**Recompute** guards" — it never calls `Recompute`.

### `be4b2fc` — hybrid tip triple (docs only)

`Tip` is documented as atomic — "a write either advances all three or none" — but `Set`'s higher-block branch keeps the incoming `Block`/`Source` while retaining the **stored** `ObservedAt` (the anchoring from `42be4f8`). The anchoring is correct and unchanged; what was undocumented is the effect on `freshRelayTip`, the one reader keying off `Source` *and* `ObservedAt` together — it can see `Source=Relay` stamped with a poll's timestamp. The window is bounded by lock acquisition (both paths stamp `time.Now()` immediately before the store write), which is why this is a comment rather than a second stamp field.

## Branch note

This PR is opened from `stack-repair-t4-t10`, a clean branch created by rebasing `t4-block-monotonic-endpoint-tip` onto `main`.

A PR directly from `t4-block-monotonic-endpoint-tip` does **not** work (attempted in #234, now closed). Because this repo rebase-merges, `main` holds #225's content under *rewritten* SHAs while `t4` still carries the originals. Their merge-base is therefore `fd7b054` — before #225 landed — so GitHub's three-dot diff pulled in #225's changes too (30 files, +1413/-1067) and the merge reported `CONFLICTING`, with 4 both-changed hunks where #225's original commits collide with the already-applied copies on `main`.

Rebasing drops those three by patch-id, which a merge cannot do:

```
warning: skipped previously applied commit 0978c48
warning: skipped previously applied commit 5f72efd
warning: skipped previously applied commit 9b900ab
Successfully rebased and updated detached HEAD.
```

The result is this branch: merge-base is the current `main` tip, so the three-dot diff equals the true content difference, and it merges cleanly. No shared branch was rewritten; `t4-block-monotonic-endpoint-tip` is untouched.

## Verification

The re-land was audited by tree identity: before the review-fix commits, this branch's tree was **byte-identical** (`13a6d924…`) to `t4-block-monotonic-endpoint-tip`, the branch that held both #226 and #228 — proving nothing was added or dropped by the rebase. The merge-base with `main` is `main`'s own tip, so the three-dot diff equals the true content difference.

With the review fixes applied: `go build ./...` clean, `go vet ./...` clean (module-wide — this matters, since `Store.Set` gained a required parameter and `go build` skips `_test.go`), `gofmt` clean, and `chainstate`, `endpointstate`, `endpointtip`, `probing`, `ecosystem/cache` and the full `rpcsmartrouter` suite all pass.

## After this merges

#233 (T3) is the last of the stack and currently based on `t10-shared-chaintip`. Once this lands it will be rebased onto `main` and retargeted. Dry-run already done: 9 already-applied commits auto-drop (including the two T10 commits #233 carries), 6 T3 commits survive, zero conflicts. Note that dry-run predates the three review-fix commits here, so it is worth re-running against the merged `main`.

### DIFF
--- a/ecosystem/cache/server.go
+++ b/ecosystem/cache/server.go
@@ -32,6 +32,12 @@ const (
 	ExpirationNodeErrorsOnFinalizedFlagName      = "expiration-finalized-node-errors"
 	FlagCacheSizeName                            = "max-items"
 	DefaultExpirationForNonFinalized             = 500 * time.Millisecond
+	// SharedStateTipBlockMultiplier bounds how long a pod's published chain tip lives in the
+	// shared cache: SharedStateTipBlockMultiplier * averageBlockTime. It mirrors the smart
+	// router's own tip-staleness horizon (chainstate.StalenessWindow) so a dead pod's tip
+	// evaporates on roughly the same timescale the router considers a tip stale, rather than
+	// lingering for the full finalized (~1h) TTL.
+	SharedStateTipBlockMultiplier                = 10
 	DefaultExpirationTimeFinalizedMultiplier     = 1.0
 	DefaultExpirationTimeNonFinalizedMultiplier  = 1.0
 	DefaultExpirationBlocksHashesToHeights       = 48 * time.Hour
@@ -201,6 +207,15 @@ func (cs *CacheServer) Serve(ctx context.Context, listenAddr string) {
 func (cs *CacheServer) ExpirationForChain(averageBlockTimeForChain time.Duration) time.Duration {
 	eighthBlock := averageBlockTimeForChain / 8
 	return lavaslices.Max([]time.Duration{eighthBlock, cs.ExpirationNonFinalized})
+}
+
+// SharedStateTipExpiration returns the TTL for a pod's published chain tip in shared-state
+// mode: SharedStateTipBlockMultiplier * averageBlockTime, floored at ExpirationNonFinalized so
+// a chain whose spec omits the average block time still gets a sane, non-zero TTL (a zero TTL
+// would never expire).
+func (cs *CacheServer) SharedStateTipExpiration(averageBlockTimeForChain time.Duration) time.Duration {
+	ttl := averageBlockTimeForChain * SharedStateTipBlockMultiplier
+	return lavaslices.Max([]time.Duration{ttl, cs.ExpirationNonFinalized})
 }
 
 func (cs *CacheServer) GetTotalCacheSize() int64 {
--- a/ecosystem/cache/server.go
+++ b/ecosystem/cache/server.go
@@ -37,14 +37,25 @@ const (
 	// router's own tip-staleness horizon (chainstate.StalenessWindow) so a dead pod's tip
 	// evaporates on roughly the same timescale the router considers a tip stale, rather than
 	// lingering for the full finalized (~1h) TTL.
-	SharedStateTipBlockMultiplier                = 10
-	DefaultExpirationTimeFinalizedMultiplier     = 1.0
-	DefaultExpirationTimeNonFinalizedMultiplier  = 1.0
-	DefaultExpirationBlocksHashesToHeights       = 48 * time.Hour
-	DefaultExpirationTimeFinalized               = time.Hour
-	DefaultExpirationNodeErrors                  = 250 * time.Millisecond
-	CacheNumCounters                             = 100000000 // expect 10M items
-	unixPrefix                                   = "unix:"
+	SharedStateTipBlockMultiplier = 10
+	// MinSharedStateTipExpiration is the absolute floor for that TTL, mirroring the router-side
+	// floor (chainstate.minStalenessWindow) rather than importing it — ecosystem/cache is a
+	// standalone binary and must not depend on protocol/. It is a hard-coded constant, NOT an
+	// operator-tunable duration, because it underwrites two guarantees the tunables cannot:
+	//   - non-zero: ristretto treats a zero TTL as "never expires", so a dead pod's tip would
+	//     linger forever. ExpirationNonFinalized alone cannot prevent this — it is
+	//     flag * multiplier, both operator-settable, and either being 0 makes it 0.
+	//   - long enough to be useful: a chain tip must outlive the gap between requests. At the
+	//     500ms ExpirationNonFinalized default a tip published for a spec with no
+	//     average_block_time would evaporate between relays, silently disabling shared state.
+	MinSharedStateTipExpiration                 = 2 * time.Second
+	DefaultExpirationTimeFinalizedMultiplier    = 1.0
+	DefaultExpirationTimeNonFinalizedMultiplier = 1.0
+	DefaultExpirationBlocksHashesToHeights      = 48 * time.Hour
+	DefaultExpirationTimeFinalized              = time.Hour
+	DefaultExpirationNodeErrors                 = 250 * time.Millisecond
+	CacheNumCounters                            = 100000000 // expect 10M items
+	unixPrefix                                  = "unix:"
 )
 
 type CacheServer struct {
@@ -210,12 +221,18 @@ func (cs *CacheServer) ExpirationForChain(averageBlockTimeForChain time.Duration
 }
 
 // SharedStateTipExpiration returns the TTL for a pod's published chain tip in shared-state
-// mode: SharedStateTipBlockMultiplier * averageBlockTime, floored at ExpirationNonFinalized so
-// a chain whose spec omits the average block time still gets a sane, non-zero TTL (a zero TTL
-// would never expire).
+// mode: SharedStateTipBlockMultiplier * averageBlockTime, floored at ExpirationNonFinalized and
+// then at the hard MinSharedStateTipExpiration.
+//
+// The second floor is the load-bearing one. ExpirationNonFinalized is derived from two operator
+// flags (duration * multiplier), so it is neither guaranteed non-zero — a zero TTL means "never
+// expires" in ristretto, so a dead pod's tip would linger forever — nor guaranteed long enough:
+// at its 500ms default a tip for a spec that omits average_block_time (0 * multiplier = 0) would
+// evaporate between relays and silently disable shared state. MinSharedStateTipExpiration is a
+// constant precisely so no flag combination can defeat either guarantee.
 func (cs *CacheServer) SharedStateTipExpiration(averageBlockTimeForChain time.Duration) time.Duration {
 	ttl := averageBlockTimeForChain * SharedStateTipBlockMultiplier
-	return lavaslices.Max([]time.Duration{ttl, cs.ExpirationNonFinalized})
+	return lavaslices.Max([]time.Duration{ttl, cs.ExpirationNonFinalized, MinSharedStateTipExpiration})
 }
 
 func (cs *CacheServer) GetTotalCacheSize() int64 {
--- a/protocol/endpointtip/store.go
+++ b/protocol/endpointtip/store.go
@@ -1,23 +1,15 @@
 // Package endpointtip holds the single, process-wide source of truth for the
 // per-endpoint observed block tip.
 //
-// Why a standalone leaf package: the tip is written by the high layer
-// (endpointstate's poll + relay-harvest observers) but also read by the low layer
-// (lavasession's per-endpoint QoS sync-block). endpointstate already imports
-// lavasession, so lavasession cannot import endpointstate to reach the observation
-// store — that would be an import cycle. A leaf package that imports only the
-// standard library can be imported by BOTH layers, so they share ONE store instead
-// of each keeping its own copy (the divergence that this package exists to remove).
+// It is a standard-library-only leaf so both the high layer (endpointstate's poll +
+// relay observers, which write it) and the low layer (lavasession's QoS sync-block,
+// which reads it) can import it without a cycle — sharing ONE store rather than each
+// keeping a divergent copy. It stores the whole {Block, ObservedAt, Source} triple so
+// the guard and the relay-freshness gate see a consistent value.
 //
-// The store holds the whole atomic tip triple {Block, ObservedAt, Source} — never a
-// bare block — because the monotonic guard needs Block and ObservedAt together and
-// the relay-freshness gate reads Source. Splitting the triple across two stores would
-// reintroduce exactly the divergence this consolidation removes.
-//
-// Generation gating (rejecting writes from a removed/replaced tracker) is NOT done
-// here: this leaf cannot know about tracker lifecycles. Callers in endpointstate
-// gate on generation FIRST and only then call Set, which applies the time-monotonic
-// guard.
+// Generation gating (rejecting writes from a removed tracker) is done by the
+// endpointstate callers before Set; this leaf only applies the block-monotonic guard
+// with a staleness backstop (T4/C-D — see Set).
 package endpointtip
 
 import (
@@ -54,9 +46,13 @@ func (s Source) String() string {
 // either advances all three or none.
 type Tip struct {
 	Block      int64     // most recent observed block (always > 0 once set)
-	ObservedAt time.Time // wall-clock of the observation that set Block (monotonic)
+	ObservedAt time.Time // wall-clock of the observation that set Block (freshness stamp)
 	Source     Source    // origin of this observation
 }
+
+// The staleness horizon is NOT defined here — this leaf is pure mechanism. The caller
+// (endpointstate's monitor) derives it from chainstate.StalenessWindow, the one source of
+// truth for the fresh/alive horizon, and passes it to Set as staleAfter.
 
 // Store is the per-endpoint tip store, keyed by the composite Key. It is safe for
 // concurrent use. The exported instance type (rather than only package globals) lets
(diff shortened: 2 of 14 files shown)


avitenzer (commented):
## Review — verified as a faithful re-land

> **Update:** all four findings below have been fixed, and the fixes are now **in this PR** (`ee754cf`, `708b810`, `be4b2fc`) at the author's request. This PR therefore no longer carries "no new code" — the PR description has been updated to match. The re-land audit below still stands for commits `ac2165c`…`3e626e7`, which are unchanged.

### The audit that matters

Two facts, which only prove the claim together — measured at `3e626e7`, before the review-fix commits:

| Check | Result |
|---|---|
| `HEAD^{tree}` vs `origin/t4-block-monotonic-endpoint-tip^{tree}` | both `13a6d924…` — **identical** |
| `merge-base(HEAD, origin/main)` vs `origin/main` | both `548c5fa` — **merge-base is main's tip** |

The re-land produced exactly the tree of the branch that already held both #226 and #228 — nothing added, nothing dropped. That single check subsumes a diff read: git's tree object is a Merkle hash over the whole content, so one equality proves the resulting filesystem is bit-for-bit what was approved.

And because the merge-base *is* main's tip, the three-dot diff equals the true content difference — which is why this succeeds where #234 reported `CONFLICTING`. Worth naming the mechanism for the record: `git rebase` compares **patch-ids** and silently drops commits whose changes are already present, which a merge has no concept of. That is the whole reason the rebase route works here.

### Verified

- `go build ./...` — clean
- `go vet ./...` — clean (exit 0). This one earns its place: `Store.Set` gained a required third parameter, a breaking signature change, and `go build` skips `_test.go` files. `vet` compiles them, confirming no caller anywhere in the module still uses the 2-arg form.
- `gofmt` — clean on every touched file
- `go test` — green on `endpointtip`, `chainstate`, `endpointstate`, `probing`, `ecosystem/cache`, `rpcsmartrouter`, both before and after the review fixes

I also traced the T10 key alignment by hand rather than trusting the description: `ecosystem/cache/handlers.go:556` keys as `chainID + "_" + sharedStateId`, and both sides now pass `listenEndpoint.Key()`. The `peerTip <= localSeenBlock` guard is load-bearing and correct — `GetRelay` max-merges the caller's own `SeenBlock` into the reply (`handlers.go:229-231`), so without that guard the router would re-adopt its own echo on every request.

### Findings — all resolved

Every finding was inherited from the already-approved #226/#228; none was introduced by the re-land. Only the first carries a behavior change.

| # | Finding | Fixed in | Behavior change |
|---|---|---|---|
| 1 | T4 doc sweep missed `endpointTipPreferStore` | `708b810` | no |
| 2 | Shared-state tip TTL floor doesn't deliver its stated guarantee | `ee754cf` | **yes** (specs missing `average_block_time` only) |
| 3 | `adopted=false` conflates reject with downward-adopt | `708b810` | no (debug-log field split) |
| 4a | Doc credits `SetLatestBlock` with a `Recompute` it never calls | `708b810` | no |
| 4b | One write stores a hybrid tip triple, against `Tip`'s stated invariant | `be4b2fc` | no (docs only) |

**Finding 1 detail** (it was the only one not anchorable inline, since the line fell outside the original diff): `endpointTipPreferStore`'s doc at `rpcsmartrouter_server.go:2135` still read *"fed … under a **TIME-monotonic** guard, so it always reflects the newest observation — including a newer LOWER block after a reorg/rollback."* Under T4 that second clause is false — a lower block is held off while the stored higher tip is fresh, so the downward correction lags by up to the staleness window. This was precisely the correction commit `8fe9926` applied to `TestEndpointTipPreferStore`'s rationale **in this PR**: the test-side twin was fixed while the production doc was not, leaving the only surviving `time-monotonic` reference in the tree. The prefer-store conclusion is unchanged, and in fact sharper — the store eventually self-heals downward, the monotonic-max atomic never does.

### One open question

`ee754cf` commits to a **hard constant** (`MinSharedStateTipExpiration = 2s`) rather than an operator-tunable flag, deliberately: a tunable minimum reintroduces the same hole, since an operator could zero it and get an immortal cache entry. If the preference is for a dedicated flag instead, it would still need its own internal clamp — at which point the constant is doing the work regardless. Flagging it as the one design call in this series worth a second opinion.

### Out of scope

The forward-looking claims about #233 (9 commits auto-dropping, byte-identical diff after rebase) are not this PR's content. I neither verified nor endorse them — and that dry-run predates the three review-fix commits here, so it is worth re-running against the merged `main`.

ecosystem/cache/server.go
```diff
@@ -203,6 +209,15 @@ func (cs *CacheServer) ExpirationForChain(averageBlockTimeForChain time.Duration
 	return lavaslices.Max([]time.Duration{eighthBlock, cs.ExpirationNonFinalized})
 }
 
+// SharedStateTipExpiration returns the TTL for a pod's published chain tip in shared-state
+// mode: SharedStateTipBlockMultiplier * averageBlockTime, floored at ExpirationNonFinalized so
+// a chain whose spec omits the average block time still gets a sane, non-zero TTL (a zero TTL
+// would never expire).
+func (cs *CacheServer) SharedStateTipExpiration(averageBlockTimeForChain time.Duration) time.Duration {
+	ttl := averageBlockTimeForChain * SharedStateTipBlockMultiplier
+	return lavaslices.Max([]time.Duration{ttl, cs.ExpirationNonFinalized})
```
avitenzer: > **✅ Resolved in `ee754cf`** — the fix is now in this PR. Original finding retained below for the record.

---

**Finding 2 (robustness, inherited from #228): this floor did not deliver the guarantee its docstring claimed.**

The doc said the floor exists so a chain gets "a sane, non-zero TTL (a zero TTL would never expire)". `ExpirationNonFinalized` cannot underwrite that, because it is derived as `flag * multiplier` (`server.go:75`) and **both** are operator-settable — `--expiration-non-finalized=0` or a `0` multiplier makes it `0`. `SharedStateTipExpiration` then returned `0`, and ristretto treats a zero TTL as *never expires*: a dead pod's tip would be stranded in the shared cache forever. That is the exact failure the floor was written to prevent.

Separately, even at the 500ms default the floor is roughly three orders of magnitude short for a chain tip. It is reachable: `JsonRPCChainParser.ChainBlockStats` computes `time.Duration(spec.AverageBlockTime) * time.Millisecond`, so a spec omitting `average_block_time` yields `0 * 10 = 0` → 500ms, and the published tip evaporates between relays — shared state silently degrades back to off, which is the very condition T10 exists to fix.

**Fix as landed:** `MinSharedStateTipExpiration = 2 * time.Second` added as a third, independent `max()` term, mirroring the router-side floor (`chainstate.minStalenessWindow`) without `ecosystem/cache` taking a dependency on `protocol/`. `ExpirationNonFinalized` is left in the expression untouched — nothing is merged or removed. Deliberately a constant rather than a flag: a tunable minimum reintroduces the same hole.

Blast radius is one case — a spec with no `average_block_time` — since `abt * 10 >= 2s` for any `abt >= 200ms`, so ETH (12s) and Solana (400ms) are unchanged. `TestSharedStateTipExpiration_FloorSurvivesZeroedFlags` pins the zeroed-flag case.

protocol/rpcsmartrouter/rpcsmartrouter_server.go
```diff
@@ -2806,6 +2807,28 @@ func isFinalizedForCacheWrite(requestedBlock, replyLatestBlock, trackedLatestBlo
 	return spectypes.IsFinalizedBlock(requestedBlock, latest, finalizationDistance)
 }
 
+// adoptSharedStateTip feeds a peer pod's chain tip — read from the shared cache under the
+// chain-level key, so it is the fleet-max of every pod's guarded tip — into this pod's
+// ChainState, but only when it is ahead of our local view. SetLatestBlock applies the same
+// outlier/Recompute guards as a local observation, so a poisoned peer value is snapped down
```
avitenzer: > **✅ Resolved in `708b810`** — the fix is now in this PR. Original finding retained below for the record.

---

**Finding 4a (doc nit, inherited from #228):** `SetLatestBlock` never calls `Recompute` — it applies the outlier threshold against the clustered baseline (or against the fresh tip when no baseline exists yet, `chain_state.go:336-343`). The baseline arrives from a separate `Recompute` call site entirely, so naming it here sent a reader looking for something that is not in the call path.

**Fix as landed:** the doc now describes the guard actually in the call path — "outlier threshold against the clustered baseline, or against the fresh tip before a baseline exists".

protocol/rpcsmartrouter/rpcsmartrouter_server.go
```diff
@@ -2806,6 +2807,28 @@ func isFinalizedForCacheWrite(requestedBlock, replyLatestBlock, trackedLatestBlo
...
+		return
+	}
+	// SetLatestBlock re-guards: advanced=false means the anti-lie guard rejected/snapped down the
+	// peer tip (e.g. a lying-high value). Log the outcome so a rejected peer tip is visible rather
+	// than silent — a stream of adopted=false is the signature of a poisoned shared tip.
+	tip, _, advanced := rpcss.chainState.SetLatestBlock(peerTip)
+	utils.LavaFormatDebug("shared-state tip fed to chain state",
+		utils.LogAttr("peer_tip", peerTip),
+		utils.LogAttr("local_seen_block", localSeenBlock),
+		utils.LogAttr("adopted", advanced),
```
avitenzer: > **✅ Resolved in `708b810`** — the fix is now in this PR. Original finding retained below for the record.

---

**Finding 3 (observability, inherited from #228): `adopted=false` conflated "the guard rejected it" with "it was adopted downward" — which undercut this commit's own stated goal.**

`SetLatestBlock` returns `advanced=false` in the stale-tip downward re-adoption branch (`chain_state.go:324-328`) **after** having changed the tip to the peer's value. So `advanced` is false in two opposite situations, and commit `3e626e7` — whose whole purpose is making a rejected peer tip visible — could not distinguish them.

I confirmed this is reachable rather than theoretical, with a test against the real `ChainState`:

```go
now = now.Add(30 * time.Second)  // local tip goes stale
tipAfter, _, advanced := cs.SetLatestBlock(1500)
require.False(t, advanced)                    // passes
require.Equal(t, int64(1500), tipAfter)       // passes — the peer value WAS taken
```

It needs the local tip stale and `localSeenBlock < peerTip < cs.latestBlock`. The `chain_state_tip` field in the same line did still reveal the truth, so it was recoverable from the log — but the boolean read backwards.

**Fix as landed:** the log now carries `peer_tip_taken` (`tip == peerTip`) alongside `tip_advanced`. The resulting tip is the clean discriminator, and it maps correctly across all five `SetLatestBlock` outcomes (accepted upward / over-threshold reject / already equal / fresh-tip lower reject / stale downward self-heal), so `peer_tip_taken=false` is now the unambiguous signature of a poisoned shared tip. `TestAdoptSharedStateTip_StaleTipTakesLowerPeerValue` pins the distinction so a future change that collapses the two fields fails there.

protocol/endpointtip/store.go
```diff
@@ -78,27 +74,43 @@ func Key(chainID, apiInterface, url string) string {
...
 	}
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	if cur, ok := s.tips[key]; ok && t.ObservedAt.Before(cur.ObservedAt) {
-		return false
+	cur, ok := s.tips[key]
+	switch {
+	case !ok || t.Block > cur.Block:
+		if ok && cur.ObservedAt.After(t.ObservedAt) {
+			t.ObservedAt = cur.ObservedAt // freshness stamp only advances, never regresses
```
avitenzer: > **✅ Resolved in `be4b2fc`** — documented, as recommended. Original finding retained below for the record.

---

**Finding 4b (invariant note, inherited from #226): this line is the one place the tip triple is stored as a hybrid, and the `Tip` doc forbade that.**

`Tip` was documented as "the atomic per-endpoint tip triple. The three fields move together — a write either advances all three or none." This branch stores `Block` and `Source` from the incoming write but `ObservedAt` from the stored one, which is a genuine, deliberate exception to that sentence.

The anchoring itself is right, and I did not suggest changing it. The consequence worth documenting is `freshRelayTip`, the only reader that keys off `Source` *and* `ObservedAt` together: it can see `Source: Relay` stamped with a *poll's* timestamp, and suppress one poll cycle on a relay tip marginally older than `relayGateFreshness` — which is the one thing that function's own doc calls "load-bearing" (only served traffic earns a skipped poll).

The magnitude is small, and I checked both paths before raising it: `endpoint_poller.go:363` and `rpcsmartrouter_server.go:2128` both stamp `time.Now()` immediately after the round-trip, right before the store write, so the window is bounded by lock acquisition — sub-millisecond.

**Fix as landed:** no code change. `Tip`'s doc now names the exception, the affected reader, and the bound — so the next person to read "all three or none" is not misled into building on a guarantee the code does not make.

avitenzer (approved):

PR #236, opened 2026-07-28
fix(endpointstate): send REST POST block polls to the method path (MAG-2597)

AnnaR-prog:
# Description

Closes: [MAG-2597](https://magmadevs.atlassian.net/browse/MAG-2597)

`EndpointPoller.sendRawRequest` handled REST `GET` and REST `POST` asymmetrically. The GET branch joins the request data onto the base URL as a path; the POST branch handed the body to `directConnection.SendRequest`, which POSTs to `nodeUrl.Url` — the **bare base URL**. `parsing.ApiName` was passed into the function but used only as a gRPC header, never for HTTP.

Any spec whose `GET_BLOCKNUM` directive lives in a `rest` collection with `"type": "POST"` therefore polled the host root. Tron answers that with **405 on every poll**, which drives endpoint availability to zero until the router stops serving the interface entirely (`"No pairings available."`).

**Most critical file to review:** `protocol/endpointstate/endpoint_poller.go`.

### Why this went unnoticed

The relay path was never affected — `chainlib/rest.go:50-53` sets `urlPath = craftData.Path` for POST — so relays worked while block-polling failed. In the probe that surfaced this, 92/100 REST POST methods passed while the chain tracker was failing 100%. `smartrouter health` is built on `ChainFetcher`, which uses the correct path, so the diagnostic tool could not reproduce it either.

### The fix

REST non-GET now goes through the same path-joining helper as REST GET (`doRESTRequest`), which also removes the duplicated httpDoer cast / `JoinURLPath` / status-check block.

Gated on **`apiInterface == rest`**, not on `connectionType == POST` — jsonrpc, tendermintrpc and gRPC all reach that branch with POST and must keep targeting the base URL. A rest fix that silently redirected every jsonrpc chain would be a far worse bug than the one being fixed, so the jsonrpc path is verified live below, not just reasoned about.

## Testing

Three router runs, same config, **one variable each**, against TRX mainnet + TRXT Shasta + TRXN Nile. `TRXN` was grafted in from [lava-specs#96](https://github.com/Magma-Devs/lava-specs/pull/96) so the `TRX → TRXT → TRXN` inheritance chain is genuinely exercised.

| Run | Binary | `tron.json` | TRX | TRXT | TRXN | Chain-tracker hash reads |
|---|---|---|---|---|---|---|
| A | `main` | `main` (path in body slot) | 20× 405 | 20× 405 | 20× 405 | 0 |
| B | `main` | spec fix ([lava-specs#101](https://github.com/Magma-Devs/lava-specs/pull/101)) | 20× 405 | 20× 405 | 20× 405 | 0 |
| C | **this PR** | spec fix | ✅ | ✅ | ✅ | 51 |

Run B matters: the companion spec fix **alone** leaves failure counts byte-identical to the baseline. This change is what restores rest polling.

Run C, per chain, with real hashes off the chain tracker:

```
TRX  (mainnet)   latestBlock=84829439
TRXT (Shasta)    latestBlock=66980760   hash=0000000003fe0b9845e86c62c6558189466c2a7bee
TRXN (Nile)      latestBlock=69550153   hash=00000000042540494f6038570c4cadd7ddd1ae43c5
```

End-to-end relays through the router (`POST /wallet/getnowblock` per listener) vs. the same call direct to upstream:

| Chain | via router | direct upstream |
|---|---|---|
| TRX | 84830254 | 84830255 |
| TRXT | 66981568 | 66981569 |
| TRXN | 69550966 | 69550968 |

**jsonrpc no-regression** (this binary, TRX jsonrpc): `smartrouter_latest_block{apiInterface="jsonrpc",spec="TRX"}` = 84830298, `eth_blockNumber` through the router returned `0x50e686d` = 84830317, 0 poll failures.

`go build ./...` clean; `protocol/endpointstate` and `protocol/lavasession` tests pass.

## Notes for reviewers

- **Unit tests.** This is covered end-to-end above but has no new unit test. `sendRawRequest` is a method on `EndpointPoller` and needs a `directConnection` to exercise; the existing `direct_rpc_connection_test.go` covers `SendRequest`/`DoHTTPRequest` at the transport layer but nothing asserts *which* URL the poller targets per interface. A table test over (apiInterface, connectionType) → expected URL would lock in the jsonrpc-vs-rest split — happy to add it here if preferred over a follow-up.
- **Separate observation, not addressed here.** In run C the rest chain tracker was demonstrably working (51 hash reads, relays returning correct heights) yet **no `smartrouter_latest_block` series appeared for any rest spec**, while the jsonrpc run emits one. The gauge looks like it is not wired on the rest path. Not a regression from this change; worth its own ticket.
- **Environmental.** TronGrid's free tier throttles at `allowed_rps(3)`; back-to-back runs produce 429 cascades that can fail boot verification. Runs above were spaced ~2 min apart, and the remaining 429s are on the verification-relay path, never on the `GET_BLOCKNUM` poll, so they don't contaminate the 405 counts.

---

## Author Checklist

I have...

* [x] provided a link to the relevant issue or specification — MAG-2597, and the companion spec fix lava-specs#101
* [ ] included the necessary unit and integration tests — **not done**: verified end-to-end against three live chains instead; see the first reviewer note
* [x] updated the relevant documentation or specification — the doc comment on `sendRawRequest` now states the REST POST contract, and `doRESTRequest` is documented
* [ ] confirmed all CI checks have passed — pending on this PR

[MAG-2597]: https://magmadevs.atlassian.net/browse/MAG-2597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/protocol/endpointstate/endpoint_poller.go
+++ b/protocol/endpointstate/endpoint_poller.go
@@ -365,6 +365,7 @@ func (ecf *EndpointPoller) ObserveLatestBlockPoll(block int64, transportLatency 
 
 // sendRawRequest sends a raw request to the endpoint and returns the response.
 // For REST/GET requests, requestData is a URL path that must be appended to the base URL.
+// For REST/POST requests, requestData is the JSON body and apiName is the URL path.
 // For JSON-RPC/POST requests, requestData is the JSON body.
 func (ecf *EndpointPoller) sendRawRequest(ctx context.Context, requestData []byte, connectionType string, apiName string) ([]byte, error) {
 	if ecf.directConnection == nil {
@@ -374,31 +375,16 @@ func (ecf *EndpointPoller) sendRawRequest(ctx context.Context, requestData []byt
 	// REST GET: requestData is a URL path (e.g. "/cosmos/base/tendermint/v1beta1/blocks/latest")
 	// Must be appended to the base URL and sent as an HTTP GET.
 	if connectionType == "GET" {
-		httpDoer, ok := ecf.directConnection.(lavasession.HTTPDirectRPCDoer)
-		if !ok {
-			return nil, fmt.Errorf("connection does not support HTTP requests for endpoint %s", ecf.endpointURL)
-		}
+		return ecf.doRESTRequest(ctx, "GET", string(requestData), nil)
+	}
 
-		fullURL, err := common.JoinURLPath(ecf.directConnection.GetURL(), string(requestData))
-		if err != nil {
-			return nil, fmt.Errorf("failed to build REST URL: %w", err)
-		}
-
-		resp, err := httpDoer.DoHTTPRequest(ctx, lavasession.HTTPRequestParams{
-			Method: "GET",
-			URL:    fullURL,
-		})
-		if err != nil {
-			return nil, err
-		}
-		if resp.StatusCode >= 400 {
-			return nil, &lavasession.HTTPStatusError{
-				StatusCode: resp.StatusCode,
-				Status:     fmt.Sprintf("%d", resp.StatusCode),
-				Body:       resp.Body,
-			}
-		}
-		return resp.Body, nil
+	// REST with a non-GET method (Tron's rest collection is POST-typed): the method lives
+	// in the URL path, not the body. Unlike JSON-RPC — where every call targets the base
+	// URL and the method is named in the body — a REST POST must be sent to its own path,
+	// so append apiName the way the GET branch appends the template. Without this the poll
+	// lands on the bare base URL and the upstream answers 405 (MAG-2597).
+	if ecf.apiInterface == spectypes.APIInterfaceRest {
+		return ecf.doRESTRequest(ctx, connectionType, apiName, requestData)
 	}
 
 	// JSON-RPC / Tendermint RPC / gRPC / POST: send requestData as the body.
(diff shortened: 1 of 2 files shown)


AnnaR-prog:
## Merge order: this PR is first — and it is the one that actually fixes Tron rest

Completing the test matrix changed the picture. I had only run three of the four combinations; the missing one (**fixed binary + unfixed spec**) turns out to work.

| Run | Binary | `tron.json` | HTTP 405s | Chain-tracker hash reads |
|---|---|---|---|---|
| A2 | `main` | `main` (path in body slot) | 84 | 0 |
| B2 | `main` | spec fix (lava-specs#101) | 84 | 0 |
| C | **this PR** | spec fix | 0 | 51 |
| **D** | **this PR** | `main`, **unfixed** | **0** | **56** |

Run D, per chain: `TRX` 84856450, `TRXT` 67007519, `TRXN` 69577092 — all three tracking, zero 405s.

**The binary is the only variable that matters.** This PR restores Tron rest on its own; the companion spec fix has no runtime effect either way against TronGrid.

### Order

1. **This PR (#236)** — merge and **release**. A merged-but-unreleased router does not help a running one; the deployed image is what polls.
2. **[lava-specs#101](https://github.com/Magma-Devs/lava-specs/pull/101)** — any time, independent of this. Correctness, not function: with the path left in the body slot the poll only succeeds because TronGrid ignores a junk body on a zero-argument endpoint. That is an accident of this upstream, not a contract — a stricter node could reject it, and any other rest/POST directive that takes real arguments would send a malformed body.
3. **[lava-specs#96](https://github.com/Magma-Devs/lava-specs/pull/96)** (Tron Nile testnet) — after this is released, then re-run its pipeline for a clean verdict.

### Correction to my earlier note

I previously said Tron rest "only comes back when both are in." Run D disproves that: this PR alone is sufficient. lava-specs#101 is still worth merging, but as correctness hygiene rather than a functional dependency.

avitenzer (approved):

PR #238, opened 2026-07-29
docs(cbor): CBOR support feasibility & design for IC-based chains (MAG-2193)

AnnaR-prog:
Research deliverable for [MAG-2193](https://magmadevs.atlassian.net/browse/MAG-2193). **Doc only — no code changes.**

Concrete driver: [lava-specs PR #78](https://github.com/Magma-Devs/lava-specs/pull/78) (ORIGYN) is blocked. The Internet Computer HTTP interface is CBOR-encoded end to end, so the mandatory chain-id verification and every parse directive fail on boot.

## Headline

**Feasible — but "CBOR support" undercounts the work and overstates what's achievable.** Decoding CBOR is small and low-risk. What actually caps ORIGYN is the IC's architecture.

## Key findings

**1. Blocker reproduced live** against `icp-api.io`. The node is reachable and healthy — `root_key` and `replica_health_status` are readable right there in the CBOR — and the spec is valid. Rejection happens at the **pre-parser guard** (`rest.go:556`, `HandleJSONFormatError`), not in the parser. So there are **two** JSON surfaces to clear, and a parser-only fix never gets past the first.

**2. The gRPC precedent is the blueprint.** smart-router already transcodes a binary wire format to JSON so the generic parser works (`grpcMessage.go:133-148`, via `jsonpb`). CBOR follows the same seam and is *simpler* — CBOR is self-describing, so it needs no descriptor resolution.

**3. No JSON-gateway escape hatch.** Checked whether ORIGYN could be served as JSON with zero router changes (the Cardano/Blockfrost pattern). It can't: the IC agent API is CBOR-only by spec, and no self-hostable generic JSON gateway exists. CBOR work is genuinely required.

**4. Multi-subnet architecture caps QoS — the big one.** ORIGYN's canisters span **7 independent subnets**. Each is a separate blockchain with its own height and no shared clock, so any head we poll observes exactly one subnet while traffic goes to all of them, and the counters can't be combined. **No single number can represent "the chain's head."** Recommendation: run with **no block tracking**. QoS sync, consistency, fork detection, archive routing and caching go inert — none error; their preconditions don't exist on IC.

**5. Latent spec defect.** `root_key` is insufficient as `chain-id` — it's network-wide and identical on every IC subnet, so it passes even for a provider that cannot reach a single ORIGYN canister. Pair it with a canister-scoped constant (`icrc1_symbol()`).

## Recommended MVP — approx. 2-3 weeks

| Layer | Scope | Est. |
|---|---|---|
| L1 | CBOR response decode (both surfaces) + identity | 3-5 dd |
| L2a | relay pass-through — clients bring signed IC-agent envelopes | approx. 1 wk |
| thin L2b | one canister-scoped verification (hardcoded arg, no Candid library) | 1-2 dd |
| L3b | no block tracking | 0 |

Deferred/conditional: canister tip index (only if OGY-ledger traffic dominates), generic Candid codec, certificate verification.

## Reviewer notes

- Empirical claims are live-verified against IC mainnet and labelled as such; inferences are flagged separately.
- §12 carries file:line references for every code claim.
- One item is explicitly marked unverified: whether the legacy `3hhby` collections expose a constant-returning method for per-subnet verification.

[MAG-2193]: https://magmadevs.atlassian.net/browse/MAG-2193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/CBOR-SUPPORT-DESIGN.md
+++ b/docs/CBOR-SUPPORT-DESIGN.md
@@ -37,10 +37,29 @@ The Internet Computer (ICP) HTTP interface is **CBOR-encoded end to end**; `smar
 | **L1** — CBOR decode + identity | ✅ **Implemented & verified live** | `health` against `icp-api.io` reports `ok: true`; previously failed at `rest.go:556` |
 | **L2a** — relay pass-through | ✅ **Implemented & verified live** | canister queries relayed through the router: `icrc1_symbol` → `"OGY"`, `icrc1_total_supply` → `1041447811516099738`, `icrc1_decimals` → `8`, all HTTP 200 / `status: replied` |
 | **L3b** — no block tracking | ✅ **Verified** | `latestBlock: 0`, pod healthy, nothing errors |
-| **thin L2b** — canister-scoped verification | ⛔ **Blocked** — needs a Go-side hook (§5.2, Route A) | cannot be a spec directive: `ingress_expiry` is dynamic and the body is binary |
+| **thin L2b** — canister-scoped verification | ✅ **Implemented & verified live** (Route A) | `origyn-identity` verification passes: the router crafts the CBOR envelope in Go, queries `icrc1_symbol` on the OGY ledger, and matches `"OGY"`. Negative control with a wrong `expected_value` fails as it should. |
 | L3c / L3c-certified / full L2b | Not started — conditional or deferred | see §5.3.2, §5.3.5 |
 
-**What shipped:** an explicit `CollectionData.Encoding` field (empty = JSON, so every existing chain is untouched), a CBOR→JSON transcoder on the message seam, the pre-parser guard bypass, and a content-type fix on the direct-relay path.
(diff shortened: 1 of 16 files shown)


VicSheCodes:
Heads up on this PR. Its title and body describe it as docs-only (the body says doc-only, no code changes), but the diff is ~1858 lines across 10 runtime .go files, including new cbor.go and ic_request.go and edits to the REST request/response path. One change (RestMessage.NewParsableRPCInput) makes every REST response route through the new parser, not just CBOR chains. For non-CBOR responses it returns the default parser, intended to keep output identical, but that shared path now runs for every REST chain.

The CBOR/IC feature itself is dormant (gated behind encoding=cbor, which no deployed chain sets), so nothing runtime is at risk today.

Suggestion: review it as a code change rather than docs-only, and run the REST compare and smoke suites (lava-rest, aptos, stellar, tezos) as a before/after regression check on the shared parser change.

PR #239, opened 2026-07-29
fix(metrics): mirror the guarded tip into smartrouter_latest_block on every poll observation

avitenzer:
Fixes [MAG-2629](https://magmadevs.atlassian.net/browse/MAG-2629).

## Bug

`smartrouter_latest_block` was written **only in the relay-harvest path** (`SetRouterLatestBlock` had exactly one call site), while the ChainState tip it reports has **two feeders** — polls and relay harvest. On a router serving no relay traffic the gauge froze at the last relay's value as polls kept the real tip fresh:

- dashboards read "falling behind head" on an idle pod, indistinguishable from a genuinely stuck tip
- it contradicted the poll-fed `rpc_endpoint_latest_block` beside it ("polls fine, consensus broken" — the wrong conclusion)
- on a poll-only router the gauge was never exported at all — *no data*, quieter than a flat line

Archaeology: T3 rewired the gauge to the guarded tip in the relay path; Topic C later added the poll feeder to ChainState, and no matching metric write came with it.

## Fix

The `OnTipObservation` closure becomes a named method, `onTipObservation`: it feeds `ChainState.SetLatestBlock` exactly as before, then mirrors the **gated** `GetLatestBlock()` into the gauge.

- **Only-when-fresh** — a stale gated read leaves the gauge at its last fresh value (zeroing would turn routine TTL expiry into a false "chain reset" on the graph)
- **Guarded, never raw** — a lying-high endpoint cannot inflate the gauge through the poll path either
- The harvest-path write stays: redundant now (the hook fires first on relays too), kept as the explicit relay-side guarantee, comment updated accordingly

## Verification

**Regression pin** — `TestOnTipObservation_GaugeAdvancesUnderPollOnlyTraffic` drives the hook with poll-only traffic: gauge must seed, advance, and hold under a warped-stale read. Verified to **fail against the pre-fix wiring** with `smartrouter_latest_block not reported` — which *is* the bug. The gauge-reading helper is lifted to file scope, shared with the existing harvest-path test.

**Live, real Ethereum mainnet** (PublicNode + Tenderly, zero relays sent after boot):

| | 65 s pre-fix (yesterday, merged main) | 40 s post-fix (temp router, this branch) |
|---|---|---|
| gauge movement | **+0** (frozen 8 behind head) | **+3, 1 block behind the live head** |
| per-endpoint views | +6, tracking head | tracking head |

**Suites** — `rpcsmartrouter`, `metrics`, `provideroptimizer` all green.

[MAG-2629]: https://magmadevs.atlassian.net/browse/MAG-2629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/protocol/rpcsmartrouter/chainstate_glue_test.go
+++ b/protocol/rpcsmartrouter/chainstate_glue_test.go
@@ -368,31 +368,7 @@ func TestHarvest_ArchiveRoutingTipGuardedByChainState(t *testing.T) {
 	mm := metrics.NewSmartRouterMetricsManager(metrics.SmartRouterMetricsManagerOptions{})
 	require.NotNil(t, mm)
 
-	// routerLatestBlock reads the smartrouter_latest_block GAUGE for this chain/interface. Unlike the
-	// counters elsewhere in this package it is Set to an ABSOLUTE value, and these tests run
-	// sequentially (no t.Parallel), so the value read straight after a harvest is the one that
-	// harvest reported — no before/after delta needed.
-	routerLatestBlock := func(t *testing.T) float64 {
-		t.Helper()
-		mfs, err := prometheus.DefaultGatherer.Gather()
-		require.NoError(t, err)
-		for _, mf := range mfs {
-			if mf.GetName() != "smartrouter_latest_block" {
-				continue
-			}
-			for _, m := range mf.GetMetric() {
-				lm := map[string]string{}
-				for _, lp := range m.GetLabel() {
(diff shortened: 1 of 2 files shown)


Tomelia1999 (approved):

nimrod-teich (changes requested):
Why do we have merge commits?

PR #243, opened 2026-08-02
feat(optimizer)(MAG-2387): restart availability from a probation baseline on proven recovery [DO NOT MERGE]

Tomelia1999:
## The problem

A provider that recovered from an outage kept getting ~1% of traffic for minutes afterwards.

Availability is a decaying weighted average kept as `num / denom`. Crossing back over `MinAcceptableAvailability` (0.80) requires success weight `>= (min*denom - num) / (1 - min)` — **four units of success weight per unit of accumulated failure weight**, because `0.8 / (1 - 0.8) = 4`. The one-hour half-life removes under 2% over the few minutes an outage lasts, so nothing fades in time to help.

A three-minute outage (30 probe checks at weight 0.25) therefore needs ~120 healthy checks — about twelve minutes — to climb back over the line. Worse in production, where real relays fail during the outage and each carries **four times** a probe's weight.

Nothing corrected this. The prober reaches its own recovery verdict from K consecutive healthy polls and returns the provider to routing via `RestoreRecoveredProvider` — but that only clears the session-level block. The optimizer was never told, and went on scoring the provider off the `availability=0` samples the prober itself had fed throughout the outage. **The router spent minutes contradicting a verdict it had reached in seconds.**

Note the collapse itself is *intentional* (`weighted_selector.go`, the `availabilityDead` branch): a dead provider must not coast on stale latency/sync numbers. This PR does not change that. It changes how a provider gets back out.

## The change

On proven recovery, the availability average restarts at exactly `MinAcceptableAvailability` over a denominator of one relay's weight:

```
num = 0.8, denom = 1.0   ->   availability = exactly 0.80
```

**Why the minimum and not 1.0.** A provider with seconds of proven uptime hasn't earned parity with a peer that never failed. Landing exactly *at* the minimum clears the dead-provider collapse (the test is a strict `<`) and returns it to normal weighted selection in the **bottom tier**: its availability contribution is still zero, but its latency, sync and stake contributions stop being discarded. Most of the recovered composite comes from those three.

**Why denominator 1.0.** The oversized denominator *is* the pathology — the more accumulated weight an average carries, the less any new sample moves it. Starting small hands control back to current behaviour within a handful of cycles, and it cuts both ways: a still-broken provider re-collapses on its next failing sample. Probation, not amnesty.

It is exactly `1.0` rather than another pair that also divides to 0.8, because dividing by 1.0 is lossless in IEEE-754 — the rebased `Resolve()` returns *bit-for-bit* `MinAcceptableAvailability`. One ULP low and the collapse re-fires and the change has no observable effect. There's a `require.Equal` pinning it.

### Two paths report recovery

| Path | Covers |
|---|---|
| `Endpoint.RecordProbeVerdict` -> `RebaseAvailabilityOnRecovery` | the endpoint disable -> enable transition |
| `AppendProbeData` (sustained health) | the general case, where endpoints stayed nominally enabled and no transition ever occurs |

The second exists because the first turned out **not to fire in the reported scenario**. `RecordProbeVerdict` early-returns for enabled endpoints, and a provider whose endpoints never got disabled never crosses that edge — yet its score still collapsed. That is the shape a real outage most commonly takes, and it's the shape the reproductions take.

The bar in that path is **three consecutive fully-healthy probe cycles**, latched to fire once per recovery episode. Any imperfect sample resets the streak and re-arms the latch. The latch lives in the lock stripe rather than being inferred from the stored score, because `providersStorage` is ristretto and its `Set` is async — a caller can't reliably read back what it just wrote.

## Conformance against the product spec

The spec: *"A provider is confirmed recovered after K consecutive successful health checks at probe cadence. Defaults: K=3, cadence 5s (~15s total). K=3 is a starting assumption, not a measured optimum — it stays a config knob. On confirmation, the failure history stops counting against the availability score. Selection weight then ramps with continued successes and reaches full normal share within ~1 minute. Any failure during the ramp drops the weight immediately and restarts confirmation."*

| # | Requirement | Status | Detail |
|---|---|---|---|
| 1 | K consecutive successful health checks, K=3, ~15s | **Partial** | K=3 and ~15s hold (measured 18.1s on ETH). Two shortfalls — see 1a/1b below. |
| 1a | K stays a **config knob** | ❌ **Gap** | `probeRecoveryStreakBar` is a compile-time `const`. So is the prober's own `DefaultProbeReEnableHysteresis`. Neither has flag plumbing today; changing K means recompiling. |
| 1b | K *health checks* | ❌ **Gap** | The streak counts probe **cycles**, not distinct checks. The prober runs every ~5s but health data refreshes at `averageBlockTime / 2`. On ETH (6.5s) three cycles are three real checks; on BTC (60s poll) three cycles can read the *same* poll. `RecordProbeVerdict` already requires a strictly newer `LastSuccessfulPoll` — this path should adopt the same rule. |
| 2 | Failure history stops counting on confirmation | ✅ **Met** | The rebase discards accumulated `num`/`denom` and restarts at `0.8 / 1.0`. |
| 3 | Weight **ramps**, full share within ~1 min | ❌ **Gap** | It *jumps*. See below — the significant one. |
| 4 | Any failure during the ramp drops weight immediately and restarts confirmation | ✅ **Met** | Confirmation genuinely restarts (streak reset + latch re-arm on any sample < 1.0). A **relay** failure re-collapses the provider for the first ~16 checks (~96s), covering the whole ramp window. A **probe-only** failure stops being decisive after 4 checks (~24s). |
| 5 | Motivation: 2-provider pods keep failover headroom | ✅ **Addressed** | Survivor carries the full load for ~16s instead of 10-40 min. |

### Gap 3 in detail — it jumps rather than ramps

At confirmation the availability ingredient contributes zero, but latency, sync and stake all return at **full value at once**:

| Time from heal | Composite | First-pick share | Fraction of fair share |
|---|---|---|---|
| 0 – 12.6s | 0.01 | 0/40 | 0% |
| **15.8s** (confirmation) | **0.675** | **9/40** | **~68%** |
| 25s | ~0.75 | 13/40 | ~100% |

One probe cycle after confirming it already carries about two-thirds of a fair share, and full share by ~25s. The spec wants it starting low and reaching full at ~60s, precisely because *"a just-recovered provider is the most likely to fail again."*

There is a mild built-in ramp afterwards — composite drifts 0.67 -> ~0.87 over ~72s as raw availability climbs 0.80 -> 0.95 — but it rides on top of a jump that has already delivered most of the traffic.

**This cannot be fixed by rebasing to a lower value.** Anything below 0.80 re-triggers the collapse and pins the provider back at 0.01; the threshold is binary and there is no "slightly alive" value. A true ramp needs a separate probation multiplier on selection weight — roughly `weight = composite * min(1, cyclesSinceConfirmation / rampCycles)` applied at `weighted_selector.go` (`SelectionWeight: compositeScore`), fed from the per-provider state this PR already tracks in the stripe, and reset on any failure. That is a genuine addition touching the selection path, which is why it is **not** in this PR.

## Measurements

Three-provider simulated ETH chain, local k3d, ~3 minute outage.

| | Before outage | During | 90s after healing |
|---|---|---|---|
| **Without this change** | 0.915 | 0.010 | 0.010 — no recovery |
| **With this change** | 0.915 | 0.010 | **0.675, reached at 18.1s** |

First-pick share, sampled 40 requests at a time (fair share ≈ 13/40):

```
 0 – 12.6s    0/40     still collapsed
    15.8s     9/40     rebase fires
    19.0s    11/40     climbing
    25s+     12–14/40  steady at fair share
```

## Observability

The rebase emits one log line per event:

```
optimizer rebased availability after proven recovery
  providerAddress=... trigger="sustained probe health"
  availability_before=0.2504 availability_after=0.8
```

Two things reviewers should know about the metrics:

- **Only `rpc_endpoint_selection_score{score_type="composite"}` shows the recovery** as a step 0.01 -> ~0.67. The `availability` series does **not** move at the rebase: raw 0.80 rescales to exactly 0.0, the same value it already showed. It lifts off zero only once raw availability clears 0.80 under its own steam a few checks later.
- **Neither series is live.** Both are written in `SetProviderSelected`, which runs only on normal unpinned selection. With no traffic the gauges are stale and will not show the jump until the next real request.

There is currently **no metric** for the rebase event, only the log. A counter (`..._rebased_total{provider, trigger}`) would make repeated rebasing — the flapping-provider signature — alertable. Not included here; happy to add.

## Risk / trade-offs

1. **Failure history is erased, not faded.** After a rebase the router cannot tell a 30-second outage from a 30-minute one. That long memory was the defect, so removing it is the point — but long-run availability reputation is gone. Latency and sync history are untouched.
2. **The score is jumpy for ~24s after recovery.** A nearly-empty jar is what makes recovery fast and also what makes it twitchy. A marginal provider can sawtooth between the floor and ~85% with a period around 30s. Truthful, but noisier than the old flat line.
3. **It slightly softens the dead-provider protection.** Stale-but-flattering latency numbers count again after three good checks, until real relays refresh them.
4. **It also fires at startup.** Providers sit around 0.667 during the first cycles after boot and get rebased to 0.80. Probably a small improvement, but the log line appears on every restart.
5. **Partially degraded providers get no help, by design.** "Fully healthy" means availability exactly 1.0, so a provider with some endpoints permanently dead never qualifies.
6. **The streak map has no eviction.** One small struct per provider per stripe. The sibling SyncBlock floor map in the same struct behaves identically, so this follows the existing pattern.

## Testing

- `protocol/provideroptimizer/recovery_rebase_test.go` — new, 12 tests: the lift works; the bit-exact 0.80 guarantee; never lowers a healthy score; idempotent across multi-endpoint recovery; still-broken re-collapses; flapping never qualifies; partial degradation never qualifies; latency/sync/sync-block survive; score config carried forward; unknown provider is a no-op; the latch re-arms after a genuine dip.
- `protocol/rpcsmartrouter/probe_loop_test.go` — 2 tests: the prober reports recovery exactly once on the transition, and never when the provider has not recovered.
- `provideroptimizer`, `rpcsmartrouter`, `lavasession`, `utils/score`, `probing` all pass. `go vet ./...` clean, `gofmt` clean.

**Pre-existing flake, unrelated:** `lavasession/TestEndpointHealthRecovery` fails intermittently under parallel load (`"disabled endpoints should make GetSessions fail"`). Verified against unmodified `origin/main` in a clean worktree — 2 failures in 6 loaded runs there, versus 2 in 6 on this branch, and 0 in 14 when run unloaded on either. Same rate both sides; not caused by this change. Worth its own ticket.

## Follow-ups

In the order I'd suggest:

1. **Gap 1b** — require a strictly newer successful poll per streak increment, matching `RecordProbeVerdict`. Smallest, and a correctness issue on slow chains.
2. **Gap 1a** — make K a flag rather than a const, on both this path and the prober's.
3. **Gap 3** — the probation ramp. Worth deciding first where the ramp should *start*, since that is the risk dial and the spec doesn't pin it.

Relates to MAG-2387.

https://claude.ai/code/session_013BpPUfXZtNX2Sm9VS35W33

### DIFF
--- a/protocol/provideroptimizer/provider_optimizer.go
+++ b/protocol/provideroptimizer/provider_optimizer.go
@@ -122,9 +122,12 @@ type ProviderData struct {
 const providerLockStripes = 256
 
 // providerSyncFloor is one stripe: it serializes the score read-modify-write for every provider
-// hashing to it AND holds the authoritative monotonic SyncBlock floor for those providers.
+// hashing to it, and holds the per-provider state that must NOT live in the async cache — the
+// authoritative monotonic SyncBlock floor, and the sustained-health bookkeeping behind the recovery
+// rebase. Both exist for the same reason: they have to be readable and writable synchronously under
+// the stripe lock, which providersStorage cannot guarantee.
 //
-// Why a floor at all (Finding 5): the three writers (appendRelayData, AppendProbeData, the legacy
+// Why a floor at all: the three writers (appendRelayData, AppendProbeData, the legacy
 // AppendProbeRelayData) each do getProviderData → mutate → providersStorage.Set. ristretto's Set is
 // ASYNC — a subsequent Get can miss a just-written value (the package's own tests sleep to let it
 // settle) — so the cache cannot be the source of truth for "SyncBlock never decreases": a serialized
@@ -133,6 +136,58 @@ const providerLockStripes = 256
 type providerSyncFloor struct {
 	mu    sync.Mutex
 	floor map[string]uint64 // provider → highest SyncBlock ever accepted (lazily created)
+	// probeRecovery tracks each provider's consecutive fully-healthy probe samples and whether the
+	// current recovery episode has already been acted on. Guarded by the same mu as floor: every
+	// writer already holds the stripe lock. Lazily created.
+	probeRecovery map[string]*probeRecoveryState
+}
+
+// probeRecoveryState is one provider's sustained-health bookkeeping.
+type probeRecoveryState struct {
+	streak  uint64 // consecutive fully-healthy probe samples
+	rebased bool   // this recovery episode already triggered a rebase
+}
+
+// recordProbeHealthLocked advances a provider's consecutive fully-healthy probe streak and reports
+// whether it has JUST crossed into proven recovery — true at most once per recovery episode. Caller
+// MUST hold s.mu.
+//
+// "Fully healthy" is strict: AppendProbeData's availability is the FRACTION of the provider's
+// endpoints healthy this cycle, so anything below 1.0 means part of the provider is still down and
+// must not count toward proven recovery. Any imperfect sample resets the streak AND re-arms the
+// episode, which is what makes this safe against a flapping provider: alternating good/bad cycles
+// never reach the bar, and a provider that genuinely dips again can be rebased on its NEXT real
+// recovery.
+//
+// The once-per-episode latch lives HERE, in the stripe, rather than being inferred from the stored
+// availability score — because providersStorage is ristretto and its Set is async, so a caller
+// cannot reliably read back the value it just wrote (the same hazard the SyncBlock floor above
+// exists to solve). Deciding from cache state would make the rebase fire repeatedly under rapid
+// writes.
+func (s *providerSyncFloor) recordProbeHealthLocked(provider string, availability float64, bar uint64) bool {
+	if s.probeRecovery == nil {
+		s.probeRecovery = make(map[string]*probeRecoveryState)
+	}
+	st, ok := s.probeRecovery[provider]
+	if !ok {
+		st = &probeRecoveryState{}
+		s.probeRecovery[provider] = st
+	}
+
+	if availability < 1.0 {
+		st.streak = 0
+		st.rebased = false // re-arm: a later genuine recovery is allowed to rebase again
+		return false
+	}
+	if st.rebased {
+		return false // already acted on for this episode
+	}
+	st.streak++
+	if st.streak < bar {
+		return false
+	}
+	st.rebased = true
+	return true
 }
 
 // applyLocked stamps providerData.SyncBlock to the authoritative monotonic floor — the max of the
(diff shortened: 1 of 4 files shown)


Tomelia1999:
Decision about the proper fix in https://magmadevs.atlassian.net/jira/software/projects/MAG/boards/100?selectedIssue=MAG-2387

VicSheCodes:
@Tomelia1999 we will need to have black-box tests for optimizer probation-baseline behavior after provider recovery before this PR merged:
https://magmadevs.atlassian.net/browse/MAG-2651

PR #244, opened 2026-08-02
fix(grpc): forward auth-headers on gRPC dials (MAG-2218)

AnnaR-prog:
## Problem

`auth-headers` were honoured by every transport **except gRPC**:

| transport | where it applies them |
|---|---|
| HTTP | `direct_rpc_connection.go:319`, `:416` |
| WebSocket | `direct_rpc_connection.go:532`, `upstream_ws_pool.go:57` |
| REST | `rest.go:508` |
| TendermintRPC | `tendermintRPC.go:721` |
| JSON-RPC connector | `connector.go:95-100` |
| **gRPC** | **nowhere** |

So an authenticated gRPC upstream answered `Unauthenticated` for every call — including the boot-time spec verification, which is why the provider failed to *verify* rather than merely failing relays. Surfaced by Canton (MAG-2218), but it affects any authenticated gRPC upstream.

## Approach

Attach at **dial time** rather than per call site. Both gRPC stacks construct their connections through `chainproxy.GRPCConnector` — the chainlib `GrpcChainProxy` used by boot verification, and the lavasession `GRPCDirectRPCConnection` used by relays and per-endpoint polls — so one `PerRPCCredentials` on the connector's dial options covers unary calls, server reflection and streams at once.

`connectorNodeUrl` is already a value copy of `g.nodeUrl`, so `AuthConfig` was inside the connector all along — unused, not unplumbed.

Two notes on scope, both differing from the ticket's description:

- The ticket lists **three** paths to fix. The third (streaming) is dead code — `createUpstreamStream` is only reachable from `StartSubscription`, which has zero callers (see MAG-2643). It gets the credential here but stays untestable end to end.
- The connector itself had **three dial sites** (`:257`, `:437`, `:451`), not one. They now share `grpcDialOptions`, so the attachment cannot be present on one path and missing on another — the class of bug this was.

A `SendRequest`-only patch would not have fixed verification, since that path goes through `GrpcChainProxy`.

## Design decisions

**`RequireTransportSecurity()` reports false.** Returning true makes grpc-go refuse every plaintext dial, which is the local-node and private-network case operators actually run, and it matches HTTP/WS which already send auth headers over plaintext `http://`. A warning is logged once per connector when credentials are configured on an unsecured transport.

**Header names are lowercased and screened at dial time.** An HTTP/2-illegal name would otherwise be rejected by the transport at call time and surface as an opaque per-relay failure rather than a config error.

## Testing

`go test ./protocol/common/... ./protocol/chainlib/... ./protocol/lavasession/... ./protocol/rpcsmartrouter/...` — all pass.

New tests assert the credential actually reaches the wire, against a plaintext server, on **both** a unary call and the reflection stream. A dial-option count alone would not have caught the original defect.

## Not covered

Validation against a real authenticated upstream (step 1.7 of the MAG-2218 plan) — needs a live CNQS node and token, so this is unit-tested but **not** verified end to end.

### DIFF
--- a/protocol/chainlib/chainproxy/connector.go
+++ b/protocol/chainlib/chainproxy/connector.go
@@ -186,6 +186,10 @@ func NewGRPCConnector(ctx context.Context, nConns uint, nodeUrl common.NodeUrl) 
 		nodeUrl:     nodeUrl,
 	}
 
+	// MAG-2218: warn once per connector, not per dial. createConnection may still upgrade an
+	// unconfigured endpoint to TLS on retry, so this reflects the configured intent.
+	nodeUrl.TokenOverInsecureWarning(nodeUrl.AuthConfig.GetUseTls() || strings.HasPrefix(nodeUrl.Url, "grpcs://"))
+
 	rpcClient, err := connector.createConnection(ctx, nodeUrl, connector.numberOfFreeClients())
 	if err != nil {
 		return nil, utils.LavaFormatError("grpc failed to create the first connection", err, utils.Attribute{Key: "address", Value: nodeUrl.UrlStr()})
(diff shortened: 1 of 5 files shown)


avitenzer (approved):

VicSheCodes:
@AnnaR-prog  We will need to add automation test coverage for gRPC auth-header forwarding from router to the upstream provider before this PR is merged.
https://magmadevs.atlassian.net/browse/MAG-2650

PR #246, opened 2026-08-02
fix(relaycore): drop the finality distance from EndpointLagThreshold

Tomelia1999:
# Description

`EndpointLagThreshold` was derived as:

```
max(allowed_block_lag_for_qos_sync × gapFactor, block_distance_for_finalized_data, 10)
```

That last term means raising a chain's **finality distance** silently loosens the **endpoint-staleness gate** — two unrelated things. This PR removes it.

## Why the finality term can't do its stated job

The comment justified it as *"to ensure endpoints can serve finalized blocks"*. That cannot hold here:

- The gate it feeds (`ValidateEndpointCapability`, via `rpcsmartrouter_server.go:1924`) runs **only for `LATEST_BLOCK`**. `ShouldSkipConsistencyValidation` skips every concrete block number and the `finalized` / `safe` / `earliest` / `pending` tags — so the finalized-data requests it claimed to protect never reach it.
- For the one case that *does* reach it ("give me the head"), a deeper finality distance argued for tolerating a **staler** head. That's backwards: a chain with deep reorgs is not a chain where a lagging endpoint is more acceptable.

The two values answer unrelated questions — *how deep before the chain stops rewriting a block* vs *how out of date an endpoint may be before we stop routing to it* — and the spec already carries `allowed_block_lag_for_qos_sync` (plus `--consistency-block-gap-factor`) for the latter.

The coupling is inherited verbatim from `lavanet/lava` (`protocol/relaycore/consistency_config.go`), where it made sense alongside provider-side data reliability and finalization proofs. This fork has neither — nothing reads `blocks_in_finalization_proof` at all.

## Blast radius

`block_distance_for_finalized_data` is bound at exactly four non-test sites. This PR touches one:

| Site | Consumer | Changed? |
|---|---|---|
| `rpcsmartrouter_server.go:162` | consistency gate → probe verdicts | **yes** |
| `rpcsmartrouter_server.go:578` | cross-validation finality metric label | no |
| `rpcsmartrouter_server.go:3028` | cache write finalized/temp split | no |
| `chain_fetcher.go:625` | inert — all 5 production `NewChainFetcher` call sites pass `Cache: nil`, and `populateCache` is gated on `cache.CacheActive()` | no |

`EndpointLagThreshold` is the only path from the finality distance into consistency validation and the probe loop: `ValidateEndpointCapability`'s other inputs come from the endpointtip store and `chainstate` (`DefaultConfig` derives everything from block time), and `probeVerdictConfigFor` only copies `LagToleranceBlocks`. `Spec.GetBlockDistanceForFinalizedData()` has zero callers.

## No behaviour change today

Computed `max(2×lag, bdffd, 10)` vs `max(2×lag, 10)` across all 24 specs in `specs/` — **identical for every one**. The `2×lag` term or the 10 floor already dominated everywhere, including `ETH1` (bdffd=8, but 2×2=4 → floor 10 wins) and `BTC` (bdffd=6 → floor 10 wins).

It only takes effect if a chain later raises `block_distance_for_finalized_data` past that point — which is exactly the case this decouples. Context: [lavanet/lava#2316](https://github.com/lavanet/lava/pull/2316) proposes raising it on DOGE (1→40), BCH (1→15), LTC (1→13), TRX (4→20) et al. for cache-immutability reasons. Without this change, importing those values would also quadruple DOGE's tolerated endpoint lag (10→40 blocks, ~15 min → ~60 min) and double TRX's — an unrelated, unrequested loosening. With it, that spec change moves the gate by zero.

## Note on the signature

`blockDistanceToFinalization` is **removed from the parameter list** rather than left unused, so the coupling can't be reintroduced by accident and the signature honestly reflects the derivation. This is a small API break for callers (one production call site, two test files, all updated here). Say the word if you'd rather keep the parameter for signature parity with upstream.

## Tests

- `consistency_validation_test.go` — the three cases that asserted `max(…, finalization)` (expecting 64 / 32 / 100) now assert the QoS-lag derivation, and are re-labelled to say which term binds (floor vs QoS lag). The `block-gap-factor` case keeps its values; only its now-stale comment about the finalization term changed.
- `probe_loop_test.go` — this test used `finalization=64` purely as a *lever* to push the threshold above 10 so it could assert the probe sources its tolerance from the per-chain threshold. Lever switched to `blockLagForQosSync=32`; the assertion is unchanged.

`go build ./...`, `go vet`, and `go test ./protocol/... ./types/... ./utils/...` all pass. `gofumpt` reports nothing on the touched files.

---

## Author Checklist

* [ ] confirmed `!` in the type prefix if API or client breaking change — internal Go signature only, no client/API surface; flagged under "Note on the signature" above
* [ ] confirmed all CI checks have passed — pending

### DIFF
--- a/protocol/relaycore/consistency_config.go
+++ b/protocol/relaycore/consistency_config.go
@@ -50,29 +50,30 @@ func DefaultConsistencyValidationConfig() *ConsistencyValidationConfig {
 //
 // Parameters:
 //   - blockLagForQosSync: The block lag threshold used for QoS sync calculations
-//   - blockDistanceToFinalization: The number of blocks until finalization
 //   - averageBlockTime: The average time between blocks for this chain
 //   - blockGapFactor: polling-relief multiplier on blockLagForQosSync (0 = default 2)
 //
 // Derivation logic:
-//   - EndpointLagThreshold: max(blockLagForQosSync*blockGapFactor, blockDistanceToFinalization), minimum 10
+//   - EndpointLagThreshold: max(blockLagForQosSync*blockGapFactor, 10)
 //   - MaxWaitTime: averageBlockTime * 2 - wait up to 2 average block times
+//
+// EndpointLagThreshold bounds how stale an endpoint may be before we stop routing to it, so it
+// derives from blockLagForQosSync alone: this gate only ever sees LATEST_BLOCK requests, since
+// ShouldSkipConsistencyValidation skips concrete blocks and the finalized/safe/earliest/pending
+// tags. A chain's finality distance bounds something else — when data stops changing — and belongs
+// to cache-write finalization. Raise allowed_block_lag_for_qos_sync, or set
+// --consistency-block-gap-factor, when a chain needs more tolerance.
 func NewConsistencyValidationConfig(
 	blockLagForQosSync int64,
-	blockDistanceToFinalization uint32,
 	averageBlockTime time.Duration,
 	blockGapFactor int64,
 ) *ConsistencyValidationConfig {
-	// Calculate endpoint lag threshold: use the larger of double QoS sync lag
-	// or the finalization distance, to ensure endpoints can serve finalized blocks.
+	// Calculate endpoint lag threshold from the QoS sync lag.
 	gapFactor := int64(2)
 	if blockGapFactor != 0 {
 		gapFactor = blockGapFactor
 	}
 	endpointLagThreshold := blockLagForQosSync * gapFactor
-	if finalizationThreshold := int64(blockDistanceToFinalization); finalizationThreshold > endpointLagThreshold {
-		endpointLagThreshold = finalizationThreshold
-	}
 	// Apply a minimum of 10 blocks
 	if endpointLagThreshold < 10 {
 		endpointLagThreshold = 10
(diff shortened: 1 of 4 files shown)


Tomelia1999:
Parking this as a draft rather than merging.

Concern raised: removing the finality distance entirely could bite us later on **commitment-level** work. If the router ever starts gating `finalized` / `safe` (or Solana `finalized`-commitment) requests on endpoint capability — rather than skipping them as `ShouldSkipConsistencyValidation` does today — then "how far behind may an endpoint be and still hold every finalized block" becomes a real question, and `block_distance_for_finalized_data` is the right input to it.

That is a legitimate future case. Noting for whoever picks this up: it would want a **per-request** rule (for a finalized-commitment request, lag ≤ finality distance is acceptable), not the current global `max()`, which also loosens the threshold for plain `LATEST_BLOCK` requests that have nothing to do with finality.

Leaving the branch and analysis intact. The immediate consequence of keeping the coupling is captured in the description: importing raised finality values (e.g. lavanet/lava#2316) also widens the endpoint-staleness gate on DOGE 10→40, TRX 10→20, BCH 10→15, LTC 10→13 blocks. That needs to be a conscious call when those specs land.

Tomelia1999:
Un-parked — proceeding with the straight removal as originally built. No new field and no flag: the finality distance is simply no longer an input to `EndpointLagThreshold`.

If commitment-aware validation is built later, the value is still available from `ChainBlockStats()` at the call site; it just isn't threaded into this config. The note above about it wanting a per-request rule rather than a global `max()` still stands.

Ready for review.

PR #247, opened 2026-08-03
fix(smart-router): gate probe re-enable on replaying the failing relay (MAG-2550)

avitenzer:
## Summary

Fixes [MAG-2550](https://magmadevs.atlassian.net/browse/MAG-2550): a provider whose polls succeed but whose relays keep failing was re-enabled on poll evidence alone (~3 healthy `eth_blockNumber` polls), then failed a fresh 50-request budget of real client traffic before re-disabling — flapping indefinitely. The root cause is an evidence mismatch: the re-enable signal (cheap polls) and the disable trigger (real relay failures) measure different things, so no amount of poll hysteresis can prove the relay path recovered.

## What changed

**Record/replay recovery evidence** — the relay path records the failing request on the endpoint just before `MarkUnhealthy` (`Endpoint.RecordFailingRelay`), together with the relay path's **effective timeout** for that request, so the replay judges under the budget the request actually failed with (floored at 10s, extended by the endpoint's own `NodeUrl.Timeout` — exactly like the relay path). Eligibility is about the **shape** of the request, never the cause of the failure: single read-only JSON-RPC requests only — never writes (`Stateful == CONSISTENCY_SELECT_ALL_PROVIDERS` — those are broadcast to all providers anyway), never subscriptions, hanging APIs, batches (a batch reply can't be judged by single-response health classification), or oversized payloads. On a JSON-RPC router this makes recording the **common** disable path — 5xx storms and transport faults record too; the poll-only fallback covers the exceptions above plus non-JSON-RPC interfaces.

**Two-step re-enable with a three-way judge** — when a disable episode carries recorded evidence, `RecordProbeVerdict` holds the re-enable at the poll-hysteresis threshold. The probe loop replays the exact recorded request through the endpoint's own connection and judges by the same health-classification rules the relay path uses, with three outcomes:
- **recovered** — no health-affecting failure (user-level node errors count as answered) → completes the re-enable (`ConfirmRelayRecovery`);
- **still-failing** — a failure the relay path would disable on (5xx except 501, timeouts, transport faults, health-affecting node errors) → resets the streak and escalates the flap hysteresis (`RelayProbeFailed`) — proving poll-healthy/relay-broken **without exposing a single client request**;
- **inconclusive** — the replay proved nothing either way (429, unrecognized 4xx such as auth drift or an interfering proxy, unparseable 200 bodies) → neither confirms nor escalates (`RelayProbeInconclusive`); it only resets the streak to pace the next attempt.

**Bounded escape hatch** — every non-recovered replay consumes one unit of a per-evidence attempt budget (`maxRelayProbeAttempts = 3`). Exhausting it **drops the evidence** and the episode falls back to the poll-only re-enable path, so evidence whose replay can never pass (too expensive for that endpoint, pruned state, permanently unjudgeable replies) can never park an otherwise-healthy endpoint until the epoch reset. The trial budget still bounds client exposure after the fallback.

**Replays never block the prober** — the probe loop is a single serial ticker that also feeds every provider's QoS sample, so replays run on their own goroutines (`relayProbeRunner`, single-flight per endpoint) and the verdict is applied on completion through the endpoint's race-safe transitions.

**Trial refusal budget** — probe-granted re-enables carry a 3-consecutive-failure trial budget instead of a full `MaxConsecutiveConnectionAttempts` (50) reset, shrinking each residual flap cycle's client exposure ~16x. A successful real relay (`ResetHealth`) restores the full budget.

**Unsupported methods never poison health** — method-not-found / unimplemented / and the *retryable* `NODE_METHOD_NOT_SUPPORTED` (method disabled on this specific node, which previously marked unhealthy) no longer count toward the disable threshold and are never recorded as recovery evidence: a per-method capability gap is not an endpoint fault.

**Observability** — `/debug/endpoint-state` rows now expose `RelayProbeMethod` / `RelayProbeAttempts` / `ReenableProbeFlaps` (a held endpoint is distinguishable from one still earning its streak, and you can see *what* must replay); `/debug/probe-loop` exposes `EvidenceGatedCount` plus cumulative `ReplaysAttempted` / `ReplaysRecovered` / `ReplaysStillFailing` / `ReplaysInconclusive`.

## Testing

- New unit tests: `endpoint_relay_probe_test.go` (gate/confirm/fail/inconclusive/trial-budget state machine, evidence-drop escape hatch — a gated endpoint provably escapes without an epoch reset), `recovery_probe_test.go` (probe-cycle integration, three-way replay judge incl. fail-open cases, recorded-timeout plumbing and floor, async non-blocking cycle, single-flight guard, batch/whitespace-batch recording eligibility), plus `classifyEndpointHealth` unsupported-method cases.
- Updated existing hysteresis tests for the trial-budget semantics.
- Full suites pass under `-race`: `protocol/lavasession`, `protocol/rpcsmartrouter`, `protocol/probing`, `protocol/common`.

[MAG-2550]: https://magmadevs.atlassian.net/browse/MAG-2550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Jira ticket: MAG-2550

### DIFF
--- a/protocol/lavasession/consumer_types.go
+++ b/protocol/lavasession/consumer_types.go
@@ -207,9 +207,19 @@ type Endpoint struct {
 	// successful relay (ResetHealth) decays it back to 0. The cap keeps a flapping endpoint re-probed
 	// within a bounded window rather than parked until the epoch transition.
 	reenableProbeFlaps uint64
-	Addons             map[string]struct{}
-	Extensions         map[string]struct{}
-	mu                 sync.RWMutex // Protects Connections, ConnectionRefusals, Enabled, consecutiveHealthyProbes, disabledAt, lastRecoveryPoll, probeReenabled, reenableProbeFlaps
+	// relayProbeMethod / relayProbePayload capture the most recent health-affecting relay failure
+	// attributable to a READ-ONLY method (recorded by the relay path via RecordFailingRelay just
+	// before MarkUnhealthy — never for writes/subscriptions/hanging APIs, and never for
+	// unsupported-method responses, which don't reach MarkUnhealthy at all). While present on a
+	// DISABLED endpoint they GATE the probe re-enable: poll hysteresis alone is not enough, the
+	// prober must replay this exact request successfully first (MAG-2550 — the poll method and the
+	// failing relay method measure different things, so polls can never prove the relay path
+	// recovered). Cleared on every re-enable so each disable episode records fresh evidence.
+	relayProbeMethod  string
+	relayProbePayload []byte
+	Addons            map[string]struct{}
+	Extensions        map[string]struct{}
+	mu                sync.RWMutex // Protects Connections, ConnectionRefusals, Enabled, consecutiveHealthyProbes, disabledAt, lastRecoveryPoll, probeReenabled, reenableProbeFlaps, relayProbeMethod, relayProbePayload
 
 	// Per-endpoint observed tip lives in the shared endpointtip store (single source of
 	// truth), keyed by chain+apiInterface+NetworkAddress — not on the Endpoint — so the
@@ -240,6 +250,10 @@ type EndpointHealthSnapshot struct {
 	DisabledAt               time.Time
 	ConsecutiveHealthyProbes uint64 // distinct post-disable successful polls counted toward F1 re-enable
 	LastRecoveryPoll         time.Time
+	// RelayProbeMethod is the read-only method recorded as this disable episode's failing relay
+	// (MAG-2550) — the request the prober must replay successfully before re-enabling. Empty when
+	// no method-attributable evidence was recorded (transport-fault episodes, write methods).
+	RelayProbeMethod string
 }
 
 // HealthSnapshot returns the endpoint's enable/recovery state under e.mu. Read-only companion to
@@ -254,6 +268,7 @@ func (e *Endpoint) HealthSnapshot() EndpointHealthSnapshot {
 		DisabledAt:               e.disabledAt,
 		ConsecutiveHealthyProbes: e.consecutiveHealthyProbes,
 		LastRecoveryPoll:         e.lastRecoveryPoll,
+		RelayProbeMethod:         e.relayProbeMethod,
 	}
 }
 
@@ -295,10 +310,53 @@ func (e *Endpoint) clearRecoveryStreakLocked() {
 // streak AND the disable timestamp, so a re-enabled endpoint gets a clean slate and a future
 // disable starts a fresh streak. Every re-enable site funnels through this, making "an Enabled
 // transition resets the recovery fields" a property of the type rather than a block each caller
-// must remember to copy. Caller must hold e.mu.
+// must remember to copy. The recorded failing-relay evidence is cleared here too: it belongs to
+// the disable episode that just ended — if the endpoint is still broken, the trial traffic after
+// re-enable re-records fresh evidence on its way back down. Caller must hold e.mu.
 func (e *Endpoint) clearRecoveryTrackingLocked() {
 	e.clearRecoveryStreakLocked()
 	e.disabledAt = time.Time{}
+	e.relayProbeMethod = ""
+	e.relayProbePayload = nil
+}
+
+// RecordFailingRelay stores one health-affecting relay failure as replayable recovery evidence for
+// this endpoint's next disable episode (MAG-2550). The caller (the relay path, just before
+// MarkUnhealthy) is responsible for eligibility: READ-ONLY methods only — never writes,
+// subscriptions, hanging APIs, or unsupported-method responses — and a self-contained payload
+// (JSON-RPC POST body). Oversized or empty payloads are dropped rather than truncated: a replay of
+// half a request proves nothing. Rolling: the newest eligible failure wins, so the evidence always
+// reflects what was failing when the disable threshold was finally crossed.
+func (e *Endpoint) RecordFailingRelay(method string, payload []byte) {
+	if method == "" || len(payload) == 0 || len(payload) > maxRelayProbePayloadBytes {
+		return
+	}
+	// Copy: the caller's buffer is the live request data owned by the relay path.
+	stored := make([]byte, len(payload))
+	copy(stored, payload)
+	e.mu.Lock()
+	e.relayProbeMethod = method
+	e.relayProbePayload = stored
+	e.mu.Unlock()
+}
+
+// reenableFromProbeLocked applies a PROBE-GRANTED Enabled→true transition. Unlike ResetHealth
+// (real-relay evidence → full clean slate), a probe grant rests on cheap polls plus at most one
+// replayed relay, so the endpoint returns to rotation with only a TRIAL refusal budget
+// (probeReenableTrialBudget consecutive failures re-disable it) instead of a fresh
+// MaxConsecutiveConnectionAttempts — bounding each flap cycle's client exposure (MAG-2550).
+// Caller must hold e.mu and logs the transition itself.
+func (e *Endpoint) reenableFromProbeLocked() {
+	e.Enabled = true
+	e.ConnectionRefusals = 0
+	if MaxConsecutiveConnectionAttempts > probeReenableTrialBudget {
+		e.ConnectionRefusals = MaxConsecutiveConnectionAttempts - probeReenableTrialBudget
+	}
+	// Mark the re-enable as probe-granted: a subsequent disable with this still set is a flap,
+	// a successful relay clears it (ResetHealth). clearRecoveryTrackingLocked deliberately does NOT
+	// touch reenableProbeFlaps — the escalation must persist across re-enables to dampen the flap.
+	e.probeReenabled = true
+	e.clearRecoveryTrackingLocked()
 }
 
 // MarkUnhealthy increments connection refusals and disables endpoint if threshold exceeded.
@@ -405,13 +463,16 @@ func (e *Endpoint) RecordProbeVerdict(recoveryPoll time.Time, recoveryHealthy bo
 		return false
 	}
 
-	e.Enabled = true
-	e.ConnectionRefusals = 0
-	// Mark the re-enable as probe-granted: a subsequent disable with this still set is a flap (above),
-	// a successful relay clears it (ResetHealth). clearRecoveryTrackingLocked deliberately does NOT
-	// touch reenableProbeFlaps — the escalation must persist across re-enables to dampen the flap.
-	e.probeReenabled = true
-	e.clearRecoveryTrackingLocked()
+	// Poll hysteresis satisfied — but if this disable episode recorded a failing READ-ONLY relay,
+	// polls alone are categorically insufficient evidence (the poll method and the failing relay
+	// method measure different things — MAG-2550). Hold the re-enable; the endpoint is now a
+	// replay candidate (PendingRelayProbe) and the prober must confirm with ConfirmRelayRecovery
+	// after successfully replaying the recorded request, or report RelayProbeFailed.
+	if len(e.relayProbePayload) > 0 {
+		return false
+	}
+
+	e.reenableFromProbeLocked()
 	utils.LavaFormatInfo("probe re-enabled recovered endpoint",
 		utils.LogAttr("endpoint", e.NetworkAddress),
 		utils.LogAttr("is_direct_rpc", e.IsDirectRPC()),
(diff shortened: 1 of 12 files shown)


protocol/rpcsmartrouter/recovery_probe.go
```diff
@@ -0,0 +1,128 @@
...
+// REPLAYS it before re-enabling the disabled endpoint (replayFailingRelay). Together they close
+// the evidence gap that caused the disable/re-enable flap: the poll method (eth_blockNumber) and
+// the failing relay method (e.g. eth_call) measure different things, so healthy polls alone must
+// never re-enable an endpoint whose relay path was the thing that broke.
+
+// relayProbeTimeout bounds one replay of the recorded failing request. Deliberately generous
+// relative to the probe cadence: the replay runs at most once per earned poll-hysteresis streak
+// per disabled endpoint, off the data plane, and a slow-but-correct answer is still recovery
+// evidence.
+const relayProbeTimeout = 10 * time.Second
```
AnnaR-prog: **[High] A flat 10s replay timeout is far below the relay path's, so slow-but-healthy methods can never pass the replay.**

The relay path uses `nodeUrl.LowerContextTimeoutWithDuration(ctx, relayTimeout)` (`direct_rpc_relay.go:329`), where `relayTimeout = chainlib.GetRelayTimeout(...)` — the spec's per-API `TimeoutMs`, or a CU-derived budget, plus a hanging-API extra, plus the per-endpoint `nodeUrl.Timeout`. For heavy read methods that is routinely tens of seconds. This replay gets a flat 10s.

That matters because timeouts classify as `CategoryInternal` -> `shouldMarkUnhealthy`, so they are *exactly* the failures most likely to be recorded as evidence in the first place.

Concrete failure: a node is disabled by a wave of ~25s `eth_getLogs` / `debug_trace*` timeouts. It recovers to a normal 12s for that query. Every replay still times out at 10s, so `RelayProbeFailed` fires forever and the endpoint is held out of rotation until the epoch `ResetHealth` (`rpcsmartrouter.go:2801`) — the ~15 minutes this PR's own comment cites. Before this change it would have returned in ~15s.

Cleanest fix: the relay path already computed the effective timeout at record time, so carry it with the evidence — `RecordFailingRelay(method, payload, timeout)` — and use that for the replay instead of a constant.

protocol/lavasession/consumer_types.go
```diff
@@ -421,6 +482,84 @@ func (e *Endpoint) RecordProbeVerdict(recoveryPoll time.Time, recoveryHealthy bo
...
+
+// RelayProbeFailed is the prober reporting that the recorded failing relay STILL FAILS on replay:
+// the endpoint answers cheap polls but its relay path is demonstrably still broken. The streak
+// resets (the next replay attempt must re-earn the full poll hysteresis — this paces replays, one
+// per earned streak, instead of one per probe cycle) and the flap escalation advances: unlike the
+// legacy escalation, which needed a re-enable→re-disable cycle THROUGH REAL CLIENT TRAFFIC to
+// learn this, the replay proves poll-healthy/relay-broken without exposing a single client request.
+// lastRecoveryPoll is preserved (invalid-evidence convention) so an already-counted poll cannot be
+// re-counted toward the next streak.
+func (e *Endpoint) RelayProbeFailed() {
```
AnnaR-prog, line 586: **[High] Nothing ever gives up on poisoned evidence — a replay that can never pass parks the endpoint until the epoch.**

`RelayProbeFailed` resets the streak and escalates `reenableProbeFlaps`, but it never clears `relayProbePayload`, and the flap counter caps at `maxReenableProbeFlaps = 2`. So the same request is replayed forever at K<<2 = 12-poll intervals. Since `RecordProbeVerdict` now returns false unconditionally while evidence is present, there is no other way back: the only exit is the epoch `ResetHealth`.

If the recorded request can never succeed — the timeout mismatch in my other comment, a pruned-state `eth_call`, a range that always OOMs the node, or a call that is simply too expensive for that endpoint — the endpoint is a black hole for a full epoch even though it is healthy for everything else.

An escape hatch preserves the whole benefit at near-zero cost: after 2-3 consecutive failed replays, drop the evidence and fall back to the poll-only path. The trial budget (3 failures) still bounds client exposure, so you keep the anti-flap protection without the permanent park. Aging the evidence out would work too.

This is also the missing test — there is no case asserting that a gated endpoint eventually escapes.

protocol/rpcsmartrouter/rpcsmartrouter_server.go
```diff
@@ -2427,6 +2429,23 @@ func runProbeCycleCore(
 			if onRecover != nil {
 				onRecover(ep.ProviderAddress)
 			}
+		} else if method, payload, ok := ep.Endpoint.PendingRelayProbe(cfg.ReEnableHysteresis); ok {
```
AnnaR-prog: **[High/Medium] The replay runs synchronously inside the probe cycle, which is also the QoS sampler for every provider.**

`runProbeLoop` is a single serial ticker at `defaultProbeCadence = 5s` (`rpcsmartrouter_server.go:2233`), and this call is inline in the endpoint loop. One replay can burn 10s (more once the timeout is fixed to match the relay path); N pending endpoints serialize to N x timeout.

The bad case is exactly the incident this PR is for: a provider-wide degradation disables several endpoints together, their polls all recover at once, and every one of them becomes a replay candidate in the same cycle. While those replays run, the loop stops calling `AppendProbeData` for *all* providers and stops evaluating re-enables for every other endpoint. `time.Ticker` drops ticks rather than queueing, so cycles are silently skipped — poll evidence still advances by wall-clock poll time, but the optimizer's QoS input goes stale across the board.

Options: run replays in goroutines and apply the verdict on the next cycle, cap it to one replay per cycle, or bound total replay time per cycle.

protocol/rpcsmartrouter/rpcsmartrouter_server.go
```diff
@@ -3670,6 +3692,7 @@ func (rpcss *RPCSmartRouterServer) relayInnerDirect(
 		needsBackoff = needsBackoffHTTP
 
 		if shouldMarkUnhealthy && targetEndpoint != nil {
+			rpcss.recordRelayProbeEvidence(targetEndpoint, chainMessage, originalRequestData)
```
AnnaR-prog: **[Medium] The PR description understates the blast radius, and this call site is why.**

The description says:

> Episodes without method-attributable evidence (transport faults, 5xx storms) keep the existing poll-only path.

But `recordRelayProbeEvidence` is called at *both* `MarkUnhealthy` sites — including this one, the `statusCode >= 500` branch. And transport faults still carry a valid `chainMessage`, so they record too. On a JSON-RPC router, essentially every disable episode from read traffic records evidence.

So the poll-only fallback is the rare path (REST/gRPC/tendermint listeners, writes, hanging APIs, subscribe, >8KB payloads), not the common one. That is worth correcting in the description, because "5xx storms keep the old path" is the assumption that makes the two High comments above look survivable.

protocol/rpcsmartrouter/recovery_probe.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	if len(resp.Data) > 0 && json.Unmarshal(resp.Data, &nodeErr) == nil && nodeErr.Error != nil {
+		classified := common.ClassifyError(nil, chainFamily, common.TransportJsonRPC, nodeErr.Error.Code, nodeErr.Error.Message)
+		if classified.IsRateLimited() {
+			return false
+		}
+		unhealthy, _ := classifyEndpointHealth(classified, false)
+		return !unhealthy
+	}
+	return true
```
AnnaR-prog: **[Medium] Batch evidence makes the gate a near-no-op, silently.**

`originalRequestData` is the raw client body, and the JSON-RPC path explicitly uses it because it "supports batch requests" (`direct_rpc_relay.go:336`). A batch reply is a JSON array, so the `json.Unmarshal` above fails and control falls through to this `return true`.

Net effect: for any batch-derived evidence — very common on eth — the replay confirms recovery on transport success alone. That proves no more than the `eth_blockNumber` poll it was supposed to improve on, but it *looks* like relay-path evidence and grants the re-enable.

The comment frames this as an acceptable "all a multi-request payload can prove". I would argue the inverse is the safe default: evidence you cannot judge should not be *recorded*, so the episode falls back to the poll-only path plus the trial budget, with honest semantics. None of the eligibility cases in `TestRecordRelayProbeEvidence_Eligibility` covers a batch.

Minor, related: the recorded `method` for a batch is the synthesized `a + SEP + b + SEP + ...` name from `jsonRPC.go:224`, which is unbounded. Harmless as a log attribute today, but it is the same unbounded-batch-label shape 6fef8e0 just fixed for metrics — worth a note so it does not get wired into one later.

protocol/rpcsmartrouter/recovery_probe.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	if err != nil {
+		// HTTP-status errors mirror the relay path's status rule (5xx unhealthy except 501,
+		// which is a capability answer, not a fault).
+		if httpErr, ok := err.(*lavasession.HTTPStatusError); ok {
+			if httpErr.StatusCode == http.StatusTooManyRequests {
+				return false // busy: the failing handler was never exercised — no evidence
+			}
+			unhealthy, _ := classifyHTTPStatus(httpErr.StatusCode)
```
AnnaR-prog: **[Medium] The judge fails open on anything it cannot classify.**

`classifyHTTPStatus` returns `(false, false)` for 400/401/403/404, so `return !unhealthy` yields **true** — any non-429 4xx reads as "recovered". The same happens for an HTTP 200 carrying a non-JSON body (a proxy's HTML error page): the `Unmarshal` below fails and the function returns true.

Auth is fine today — both `SendRequest` and `DoHTTPRequest` apply `AuthConfig.AddAuthPath` and set the same `AuthHeaders` map, so the replay is authenticated and hits the same URL as the relay path. But if that ever drifts, the gate degrades to "always confirm" with no signal that it has stopped working.

Consider a third verdict — recovered / still-failing / **inconclusive** — where inconclusive neither confirms nor escalates. 429 already wants that shape (the comment says it "proves nothing either way" but the code routes it to the still-failing branch, which escalates the flap counter). It pairs naturally with the give-up counter I suggested on `RelayProbeFailed`.

protocol/lavasession/consumer_types.go
```diff
@@ -240,6 +250,10 @@ type EndpointHealthSnapshot struct {
 	DisabledAt               time.Time
 	ConsecutiveHealthyProbes uint64 // distinct post-disable successful polls counted toward F1 re-enable
 	LastRecoveryPoll         time.Time
+	// RelayProbeMethod is the read-only method recorded as this disable episode's failing relay
+	// (MAG-2550) — the request the prober must replay successfully before re-enabling. Empty when
+	// no method-attributable evidence was recorded (transport-fault episodes, write methods).
+	RelayProbeMethod string
```
AnnaR-prog, line 269: **[Medium] No production observability for a feature that can hold an endpoint out of rotation.**

`RelayProbeMethod` is added to the snapshot, but the only production consumer — the `/debug/endpoint-state` row builder at `rpcsmartrouter.go:1190-1205` — does not include it. Outside tests this field is write-only, so there is no way to see *why* an endpoint is being held.

Also missing:
- `probeStats.recordCycle` counts `reEnabled` but not replays attempted / passed / failed.
- No gauge for "endpoints currently held pending replay" (the gate returns false from `RecordProbeVerdict`, so a held endpoint is indistinguishable from one that simply has not earned its streak).

Right now on-call gets one `LavaFormatInfo` line per failed replay and nothing else. Given the permanent-park risk in the High comments, adding the field to the debug row plus two counters would make the difference between diagnosing this in minutes and not knowing to look.

AnnaR-prog (commented):
Review of the record/replay design. The diagnosis is right and the important insight — that the disable trigger and the re-enable signal measure different things, so no amount of poll hysteresis can close the loop — is correct. The state machine is carefully built: the replay runs outside the endpoint mutex, payloads are copied in both directions, `ConfirmRelayRecovery`/`RelayProbeFailed` are inert against a racing re-enable, and the write-safety claim genuinely holds for batches because `SpecCategory.Combine` takes `max(Stateful)` and `CONSISTENCY_SELECT_ALL_PROVIDERS = 1` is the max.

My concern is the **failure mode of the new gate**: it removes the poll-only escape hatch entirely, and nothing bounds how long a bad replay can hold an endpoint out of rotation. Comments 1 and 2 below are the ones I'd want resolved before merge — together they can turn today's 15-second flap into a 15-minute outage for an endpoint that is actually healthy.

Verification: built and tested `ce84f37` in a clean worktree. `go build ./...` clean; `protocol/rpcsmartrouter` and `protocol/common` pass; `protocol/lavasession` failed once on `Failed locking selection after MaximumNumberOfSelectionLockAttempts` (unrelated to this diff) and passed on re-run — flaky, not broken. The new tests are race-clean under `-race`.

AnnaR-prog (approved):

PR #248, opened 2026-08-03
feat: read-only secondary cache tier (MAG-2596)

avitenzer:
## Summary

Implements the **Secondary Cache Backend** PRD: an optional, independently configured, **read-only** secondary cache consulted when the primary produces no hit (miss, error, timeout, or primary down/unconfigured), before provider fall-through. A secondary hit is sanitized, served as `Cached`, and backfilled into the router's own primary under the existing populator's eligibility rules — keyed by the exact block that hit, with the validity floor lifted so the backfill is visible to its own follow-up lookup.

Operator-facing overview, including the PRD traceability table and a runnable two-zone walkthrough: `docs/SECONDARY-CACHE.md`.

Jira ticket: MAG-2596

**Companion PR — must land together:** [smart-router-automation#872](https://github.com/Magma-Devs/smart-router-automation/pull/872). The dangling-config change below moved a contract that was pinned in that repo.

## Highlights

- **Structural read-only**: the secondary is held behind a narrow `performance.CacheReader` (`CacheActive`/`GetEntry` only) — `SetEntry`/`Flush` on it are compile-time errors; the debug flush and all writers keep targeting the primary only.
- **Trust boundary**: every secondary reply is deep-copied and sanitized before serving or backfill — signatures zeroed, `Metadata` reduced to a transport allowlist (`Content-Type`/`Content-Encoding`), the foreign `LatestBlock` zeroed and re-stamped from the local tip, and the foreign block-hash→height mappings never requested. No foreign scalar reaches local chain-scoped state.
- **Entry-kind contract**: additive `IsNodeError`/`StatusCode` wire fields (zero = legacy unknown, both version-skew directions safe); cached node errors serve truthfully but are rejected from backfill by the populator's own checks — not a pre-filter.
- **Independence**: bypass rules evaluated once; each tier attempted iff itself active. Secondary-only topology boots (advisory warning) and keeps serving while the primary is down.
- **Config**: `secondary-cache-be` (required), `secondary-cache-timeout` (50ms default), `secondary-cache-mode` (read-only, the only v1 value); flags + YAML via the existing viper binding, fail-fast validation on anything that would make an *enabled* secondary misbehave, and a startup config log.
- **Observability**: `cache_tier` label on all four `smartrouter_cache_*` series, `outcome=miss|error|timeout` on failed, latency observed on every attempt (semantics change — migration note in `docs/METRICS.md`), per-tier tracing spans with `cache.tier`/`cache.outcome`.

## Changes since review

Two rounds. The first (`17cb29e`..`4b70b2c`) is described in [this comment](https://github.com/Magma-Devs/smart-router/pull/248#issuecomment-5510900339) and was verified fixed in @AnnaR-prog's re-review. The second (`02d338e`..`f6a92f0`) addresses her ship check and adversarial pass.

### Round 2 — trust boundary

- **Foreign `BlocksHashesToHeights` steered local routing.** The field sits on the *outer* `CacheRelayReply`, structurally beyond `SanitizeForeignCacheReply`'s `*RelayReply` parameter, and was used raw: the miss path returned it and the caller folded it into `latestBlockHashRequested`/`earliestBlockHashRequested`. Those steer `resolveRequestedBlock` (which gates endpoint sync and optimizer selection) and `UpdateEarliestAndValidateExtensionRules` (archive routing) — and the fold took max-for-latest / min-for-earliest, so the more extreme value always won and the foreign tier beat the local primary **by construction**. Reachable on the common path: `getBlockHeightsFromHashes` runs on hits and misses alike. The lookup now neither asks for the mappings nor returns them, pinned on the *request* so the fold cannot be re-plumbed without the test failing first. `mergeBlockHashHeights` is gone. Cost: hash-keyed archive detection falls back to the primary's mappings alone — where a router with no cache already is.
- **Dropping all `Metadata` was a serving regression, not only a privacy win.** `Reply.Metadata` is the *only* source of client response headers, so nulling it dropped `Content-Type` and `Content-Encoding` too. The sanitizer's comment claimed `Content-Type` was minted locally; the fiber handlers only pre-set `application/json` as a default that upstream metadata overrides — exactly the case a non-JSON REST body needs. Now an **allowlist**: unknown header names are still dropped by default (the open-set property the denylist could never have), without the regression. `Content-Length` stays out — `outputFormatter` re-stamps the payload.
- **`Provider-Latest-Block` was omitted on secondary hits.** Zeroing `LatestBlock` was not quite inert: its one remaining serving-side reader is `appendHeadersToRelayResult`. Taking option (b) from the re-review — the value is re-stamped from this router's own *gated* tip after sanitization. Locally sourced and anti-lie-guarded, so none of the three backfill problems the zeroing closed can reopen (`isFinalizedForCacheWrite` sees `max(localTip, localTip)`; `SetRelay` stores `max(localTip, local SeenBlock)`).

### Round 2 — contract, docs, tests

- **Dangling-config change broke a cross-repo test contract.** Two cases in `smart-router-automation`'s `TestFailFastValidation` asserted a non-zero exit for exactly the configurations `ec6148d` made advisory. Both fail on both assertions against this branch. Fixed in [automation#872](https://github.com/Magma-Devs/smart-router-automation/pull/872) by moving them to `TestBootWithWarnings`; MAG-2596's configuration contract is [updated in the ticket](https://magmadevs.atlassian.net/browse/MAG-2596) so the spec no longer says "startup error".
- **Mixed cache engines: reverted to open.** The doc answered "Yes"; MAG-2596 records "don't support mixed cache engines", and the reporter is not the PR author. The doc now says the ticket's resolution stands, and keeps only the mechanism claim (the router never learns which engine is behind either address). Sign-off requested on the ticket.
- **LATEST-tagged cross-zone lookups disclosed under Scope.** The block is resolved against this router's own tip *before* the GET, so the cache server's tag resolution never runs and the key is `hash ‖ <this zone's tip>`. Two zones a block apart miss every time, and re-query per request rather than once per entry — which the Backfill bullet claimed outright. Explicit-block queries, which this tier is for, are unaffected.
- **Metrics identity restored.** `TestSmartRouterRecordCacheResult_TotalEqualsSuccessPlusFailedPerTier` asserts `requests_total == success_total + Σ failed_total{outcome}` per tier over a mixed run, both sides required non-zero. `METRICS.md` states the `sum without (outcome)` aggregation, since that is what changed for query authors.
- **Tier-symmetry doc claims corrected.** "byte-identical" is replaced with what is actually true: both tiers emit the same locally minted header set and the same body; a secondary-sourced entry replays only its `Content-Type`/`Content-Encoding`.

**Not addressed, deliberately:** the generated `smartrouter_eth_zone_{internal,external}.yml` are already covered by `.gitignore:36` (`config/smartrouter_examples/smartrouter_*.yml`) — `git check-ignore -v` confirms both. The untested `sendRelayToEndpoint` wiring is pre-existing (`chainstate_glue_test.go:317`) and out of scope here.

## Verification

Run from a `git worktree` checkout of the pushed commit `f6a92f0`, not the working directory.

| Check | Result |
| --- | --- |
| `go build ./...` | clean |
| `go vet ./...` (whole repo, incl. test files) | clean |
| `gofmt -l ./protocol ./ecosystem` | clean |
| `go test -race ./ecosystem/cache/... ./protocol/performance/... ./types/relay/...` | ok |
| `go test ./protocol/rpcsmartrouter/... ./protocol/metrics/... ./protocol/tracing/...` | ok |

**Known pre-existing failure, not from this branch:** `protocol/metrics` `TestHappyFlow` fails under `-race` (16s, timing-sensitive). It fails identically on `4b70b2c`, the commit reviewed before this round, so it predates these commits. It passes without `-race`, which is how the row above runs it — same as the previous rounds' tables.

Each of the three code fixes is pinned by a test verified to **fail without it**, checked by reverting the fix rather than trusting a green run:

| Revert | Fails |
| --- | --- |
| allowlist → `Metadata = nil` | `TestSanitizeForeignCacheReplyKeepsTransportHeaders`, `…KeepsNonJSONContentType` |
| `BlocksHashesToHeights: nil` → populated | `TestSecondaryLookupNeverRequestsForeignBlockHashHeights` ("Expected nil, but got: []*relay.BlockHashToHeight{…}") |
| drop the `LatestBlock` re-stamp | `TestSecondaryHitRestampsLatestBlockFromLocalTip` |

The hash test uses ETH1/`trace_transaction`, the bundled spec's hash-parsed method — `eth_getBlockByHash` carries no `BLOCK_HASH` parser and yields no hashes, which made the first version of that assertion vacuously true.

Cross-repo, against a binary built from this branch: **12/12** automation startup tests pass with [automation#872](https://github.com/Magma-Devs/smart-router-automation/pull/872); without it, exactly the two dangling cases fail and the other four fail-fast cases pass.

Two-zone integration lane: `scripts/pre_setups/init_smartrouter_eth_secondary_cache.sh` (`RUN_DEMO=1` proves secondary-hit → backfill → primary-hit with per-tier metric deltas).

## Notes for reviewers

- Commits are ordered by dependency; reviewable commit-by-commit, squash-merge when landing per repo convention.
- Deploy note: the *secondary* backend should run a build including the wire-contract commit to report entry kind explicitly; older backends still work via the legacy `CACHED_ERROR` placeholder fallback.
- Operator-visible even without a secondary: the metrics label additions and latency-histogram semantics (see the METRICS.md migration note). **No ticket yet names who updates the Grafana boards** — flagged on MAG-2596.


AnnaR-prog:
## Review — read-only secondary cache tier (MAG-2596)

Reviewed at `3faa14d`, checked out into a worktree and built/tested locally.

The shape of this is good. Holding the secondary behind a narrow `CacheReader` so `SetEntry`/`Flush` are compile-time errors is the right way to make "read-only" a property rather than a promise; routing backfill eligibility through the existing populator instead of re-deriving it means the two tiers can't drift; and `TestSecondaryHitBackfillsPrimaryWithExactKeyAndValidSeenBlock` running against a *real* cache server is exactly the test that a fake `SetEntry` recorder could not have written. The exact-key + lifted-`SeenBlock` reasoning in `tryCacheWriteResolved` is subtle and, as far as I can tell, correct.

One blocker, one correctness gap I'd fix before merge, and some notes.

---

### 🔴 Blocker — the `ecosystem/cache` test package does not compile

`3faa14d` removed `newCacheServerForTest` from `set_relay_negative_block_test.go` and left a pointer to a file that isn't in the branch:

```
ecosystem/cache/set_relay_negative_block_test.go:14
// newCacheServerForTest lives in helpers_test.go.
```

There is no `helpers_test.go`, and `setRelayForTest` / `getRelayForTest` — called by both new test files — don't exist anywhere either:

```
$ go vet ./ecosystem/cache/...
vet: ecosystem/cache/is_node_error_roundtrip_test.go:19:9: undefined: newCacheServerForTest

$ go test -run zzzNONE ./ecosystem/cache/...
ecosystem/cache/is_node_error_roundtrip_test.go:19:9:  undefined: newCacheServerForTest
ecosystem/cache/is_node_error_roundtrip_test.go:21:2:  undefined: setRelayForTest
ecosystem/cache/is_node_error_roundtrip_test.go:23:16: undefined: getRelayForTest
...
FAIL  github.com/magma-Devs/smart-router/ecosystem/cache [build failed]
```

This takes the whole package's tests down, including ones that already pass on `main` (`set_relay_negative_block_test.go`, `flush_test.go`, `metrics_test.go`, `shared_state_tip_test.go`, `endpoint_observations_test.go`). Two consequences worth stating plainly:

- The PR body's "Unit/behavioral suites green, including `-race` on … `ecosystem/cache`" can't be true at this HEAD, and `is_node_error_roundtrip_test.go` / `seen_block_validation_test.go` — the tests that prove the wire-contract commit — have never run in this state.
- CI won't catch it. No workflow under `.github/workflows/` runs the Go unit tests (I grepped for `go test`, `make test`, `gotestsum` — nothing; `pr-gate-build-artifact.yml` is a build + k8s/Helm deploy validation). So this would land on `main` silently and break `go test ./...` for everyone.

**I think there's one root cause, and it explains a second oddity in the same commit.** `3faa14d` looks partially staged: it extracted helpers into a file that was never `git add`ed, and in the same commit added a shell function that is defined and never called —

```
.github/workflows/require-jira-ticket.yml:54
annotate() { ... printf '::%s title=Jira ticket gate::%s\n' "$level" "$detail"; }
```

`grep -n annotate` on that file returns only line 54, and no `::error`/`::notice` is emitted anywhere else in it. The function's own comment argues why the annotations matter, so the intended `annotate 'error' …` calls inside `fail_gate` etc. are presumably sitting in the same unstaged working tree. Worth a `git status` before the fix-up push — one action should recover both.

---

### 🟠 The foreign `Reply.LatestBlock` survives the sanitizer and reaches your own cache server's chain tip

`SanitizeForeignCacheReply` (`protocol/performance/sanitize.go:21`) drops `Sig`, `SigBlocks` and `Metadata`, and `sanitize_test.go:44` deliberately asserts `LatestBlock` survives. But on the backfill path `LatestBlock` isn't inert — the sanitized clone becomes `relayResult.Reply` and goes straight into `tryCacheWriteResolved`, where the foreign value is read three times:

**1. It picks the TTL store.** `rpcsmartrouter_server.go:3457-3460`:

```go
latestBlock := relayResult.Reply.LatestBlock   // foreign
finalized := isFinalizedForCacheWrite(requestedBlock, latestBlock, int64(rpcss.getLatestBlock()), ...)
```

and `isFinalizedForCacheWrite` (`:3299`) takes `max(replyLatestBlock, trackedLatestBlock)`. The comment two lines up says finalization "uses the GATED `getLatestBlock` … never `getLatestBlockAllowStale`: a stale or too-high head here would falsely finalize a mutable block into the long-TTL store" — the foreign value walks around that gate.

**2. It becomes the entry's staleness floor.** `SetRelay` stores `latestKnownBlock = max(Response.LatestBlock, SeenBlock)` (`ecosystem/cache/handlers.go:359`), and `GetRelay` only rejects an entry whose stored value is *too low*, never too high (`handlers.go:230`). An inflated value makes the backfilled entry pass every future staleness check.

**3. It is published as the cache server's chain-level latest block.** This is the one I'd actually fix. Last lines of `SetRelay`:

```go
s.setSeenBlockOnSharedStateMode(chainId, sharedStateId, latestKnownBlock, ...)
s.setLatestBlock(latestBlockKey(chainId, ""), latestKnownBlock)   // handlers.go:388
```

`setLatestBlock` is a monotonic max write (`performInt64WriteWithValidationAndRetry`, `handlers.go:309-330`) — it only ever raises, and can't be lowered until the entry expires. That key is what `GetRelay` uses at `handlers.go:188-189` to resolve negative block tags (`replaceRequestedBlock`: LATEST / SAFE / FINALIZED / PENDING all map to it). A single over-high foreign `LatestBlock` therefore shifts tag resolution for the *entire chain* on your own primary until expiry, so those lanes look up keys nobody wrote and miss permanently. In shared-state mode the same value also goes out as the fleet tip via `setSeenBlockOnSharedStateMode` (re-guarded at the router by `adoptSharedStateTip`, so that half is defended).

Bounding this honestly:

- `IsFinalizedBlock` returns `false` for a negative `requestedBlock` (`types/spec/constants.go:32`), so `LATEST`-tagged queries can't be pushed into the finalized store. The TTL exposure in (1) is explicit near-head block queries only.
- The entry was already served to the caller once, so (1) and (2) are about **persistence** — a secondary-sourced entry outliving what a locally-written one would — not about first-response exposure.
- A live provider can already lie about `Reply.LatestBlock` on the normal `tryCacheWrite` path, so (3) isn't a brand-new class of exposure — but it is that exposure extended across the trust boundary this PR itself declares, and the backfill is the only path where a *cache-sourced* `LatestBlock` reaches `setLatestBlock`.

The cheap fix, and the one consistent with the drop-everything policy already argued for `Metadata`: zero `LatestBlock` in `SanitizeForeignCacheReply` (or clamp it to the local gated tip). Zeroing costs nothing on the serving side — nothing downstream of `SetResponse` reads it, since the MAG-2160 rule already keeps it out of tip state — and `isFinalizedForCacheWrite` then falls back to the tracked tip alone, which is what its own comment says it wants.

---

### Smaller items

- **`outcome="hit"` is recorded before the hit can still fall through.** `secondary_cache.go:78-82` classifies and records the outcome, then `:99-104` returns `served=false` when `DeepCopyProtoObject` fails, and the request goes to a provider anyway. Rare, but it means `smartrouter_cache_success_total{cache_tier="secondary"}` isn't quite "requests served from the secondary" as METRICS.md describes it. Recording after the copy (or downgrading to `outcome="error"` on copy failure) closes it.

- **Dangling-config is a hard startup failure.** `SecondaryCacheConfig.Validate` (`secondary_config.go:31`) refuses to start when `secondary-cache-timeout`/`-mode` are set with no address. Defensible fail-fast, but it will take down a router if a shared or templated YAML carries the tuning keys across a fleet where only some routers have a secondary. A warning would still catch the typo it's aimed at.

- **The metrics break is operator-visible even with no secondary configured** — `cache_tier` on all four `smartrouter_cache_*` series, and the latency histogram now observing non-hits. The METRICS.md migration note is good and accurate; worth landing the Grafana/alert updates in the same window, since `sum by (spec, method)` survives but exact `{spec=…,apiInterface=…,method=…}` matchers do not.

- **The PR description points at a deleted file:** "Design + full PRD traceability: `docs/SECONDARY-CACHE-DESIGN.md`" — removed in `00a1e43`. No stale in-tree references remain (`6e78dbf` cleaned those up properly), it's just the description that still sends reviewers there.

- **The required `jira-ticket/validated` check is red:** "Missing Jira ticket line". The gate wants a literal line matching `^\s*jira\s+ticket\s*:` — i.e. `Jira ticket: MAG-2596` on its own line. The `[MAG-2596](…)` link in the summary doesn't satisfy it.

---

### Checked and cleared

Listing these so they don't get re-litigated. All verified against the code, not assumed:

Typed-nil `CacheReader` from the wiring — `InitCache` returns a non-nil `*Cache` even when the dial fails (`performance/cache.go:150-164`) and `CacheActive()` is nil-receiver-safe, so `secondaryCacheActive()` can neither panic nor report a dead client as active (`TestSecondaryCacheActiveNilSafety` covers the typed-nil case). `outputFormatter` being applied *before* the backfill, so the stored body carries this request's JSON-RPC id — harmless, because the formatter is `sjson.SetBytes(data, "id", …)`, an unconditional overwrite (`chainlib/cacheformat/formatter.go:97`), so a later read re-stamps it. Newly setting `IsNodeError` on **primary** cache hits — does not change retry or QoS behaviour: `ResultsManagerInst.setValidResponse` classifies via `protocolMessage.CheckResponseError`, not the flag (`relaycore/results_manager.go:130`), and `shouldFailSessionForResult` is only reached from the live-relay path; the sole new effect is the `lava-identified-node-error` header at `rpcsmartrouter_server.go:4842`, which is the intent. The two new `CacheValue` fields — stored as an in-memory `interface{}` in ristretto (`handlers.go:533`), so no serialization or existing-entry compatibility concern. `viper.IsSet` for the dangling-config check — `find(key, false)` ignores pflag defaults, so an untouched flag correctly doesn't count as set. GET/SET key agreement — both derive from `protocolMessage.HashCacheRequest`, and the backfill runs before `updateProtocolMessageIfNeededWithNewEarliestData` can mutate the message. Race on the async SET — `tryCacheWriteResolved` deep-copies synchronously and the backfill precedes `SetResponse`, so the response path's header appends can't race it.

### Local verification

```
go build ./...                                              clean
go vet $(go list ./... | grep -v 'ecosystem/cache$')        clean  (exit 0, no output)
go vet ./ecosystem/cache/...                                FAILS  (see blocker)
go test ./protocol/performance/... ./protocol/metrics/... \
        ./types/relay/... ./protocol/tracing/...            ok
go test ./protocol/rpcsmartrouter/...                       ok (90.7s)
go test ./ecosystem/cache/...                               build failed
```

So the compile breakage is exactly one package — every other package in the repo type-checks including its test files, which means the `RecordCacheResult` signature change was propagated completely.


[MAG-2596]: https://magmadevs.atlassian.net/browse/MAG-2596?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

avitenzer:
Thanks — this was a genuinely useful review. All items are addressed at `4b70b2c`, in six commits on top of what you read, and the branch is rebased onto current `main`.

Your single-root-cause call was right: the preceding commit was made against a partially staged tree, and that explains both symptoms.

---

### 🔴 Blocker — `ecosystem/cache` did not compile

Exactly as diagnosed. `helpers_test.go` was extracted but never `git add`ed. Committed in `17cb29e`, unchanged from what the commit intended.

Both of your consequences were correct and worth stating plainly rather than glossing:

- The previous PR body's claim that `ecosystem/cache` was green **could not have been true** at that HEAD, and `is_node_error_roundtrip_test.go` / `seen_block_validation_test.go` — the tests that prove the wire-contract commit — had never run. They run now.
- Your CI observation holds: the gate is builds, lint, govulncheck and the Jira check, and nothing runs `go test`. A green PR said nothing about this. I've stopped verifying from the working directory for exactly this reason — see the verification note at the bottom.

### 🟠 Foreign `Reply.LatestBlock` — fixed, and confirmed empirically

Fixed in `7de7d70`: `LatestBlock` is now zeroed in `SanitizeForeignCacheReply` alongside `Sig`/`SigBlocks`/`Metadata`, which is the option you recommended and the one consistent with the drop-everything policy already argued for `Metadata`.

Your third mechanism was not just theoretically reachable — it fires, end to end. With the fix reverted, one poisoned entry carrying `LatestBlock: 987654321` produces:

```
DBG setting latest block key=LAVA_ latestBlock=987654321
DBG Got Cache Get requested_block=-2 requested_block_parsed=987654321 seen_block=0
    Error: Expected value not to be nil.
```

That is the whole chain of events you described: the foreign head is published as the chain-level tip, the `LATEST` tag then resolves *to it*, and the lookup misses — a key nobody ever wrote. With the fix, the same lookup logs `requested_block_parsed=100` and hits. So a single foreign entry really did retarget negative-tag resolution chain-wide on the router's own primary, and the resulting misses would have persisted until expiry.

`TestSecondaryPoisonedEntrySanitizedForCallerAndBackfill` now carries a foreign head of `987654321` and asserts the **consequence** rather than the mechanism: a `LATEST`-tagged lookup must still resolve to this router's own tip and hit. That keeps the test load-bearing if the zeroing ever moves elsewhere. I verified it fails on the un-fixed code (`Should be zero, but was 987654321`) rather than trusting that it passes on the fixed code.

Your bounding was accurate and I've kept it in the rationale: `IsFinalizedBlock` returning `false` for negative blocks does limit the TTL exposure to near-head queries, and (1) and (2) are about persistence rather than first-response exposure. The reason I still treated it as blocker-grade is your last point — the backfill is the only path on which a *cache-sourced* `LatestBlock` reaches `setLatestBlock`, so it is precisely the trust boundary this PR declares.

### The partially-staged `annotate()`

You were right that it was the same commit and the same cause. I dropped it (`6aa9c76`) rather than completing it. The annotations are a real improvement to that gate, but they are unrelated to this tier and belong in a change that is about the gate, where they can be reviewed and tested as such. The PR no longer touches `.github/` at all.

### `outcome="hit"` recorded before the hit could still fall through

Fixed in `655cbfd`. The deep copy now runs before classification. I took the second of your two suggestions as well as the first: a reply the secondary genuinely returned but that this router could not handle is recorded as `outcome="error"`, not `"miss"` — telling those apart is the point of the label, and a miss would file a local defect as ordinary cache behaviour. Latency is still measured across the `GetEntry` call only.

### Dangling config is now a warning

Fixed in `ec6148d`. Your fleet argument won: the shape is usually a typo, but it is also what one templated YAML looks like across a fleet where only some routers run a secondary, and a hard failure there converts an unused key into an outage on the routers that don't have one. It now warns, names the dangling options, and leaves the tier disabled. Anything that would make an *enabled* secondary misbehave — zero timeout, `read-write` mode — stays fatal. `docs/SECONDARY-CACHE.md` is updated, including the walkthrough that previously claimed all three cases exit non-zero.

### Metrics break is operator-visible

Agreed, and no code change — flagging that I've read it rather than silently passing over it. The `cache_tier` label and the latency-histogram semantics change land whether or not a secondary is configured, and your note about `sum by (spec, method)` surviving while exact `{spec=…,apiInterface=…,method=…}` matchers do not is the right framing for the dashboard/alert update. Worth landing those in the same window.

### Stale PR description

Fixed. The `docs/SECONDARY-CACHE-DESIGN.md` pointer is gone; traceability now lives in `docs/SECONDARY-CACHE.md`, which also gains a "no foreign chain state adopted" row covering both channels (`SeenBlock` and `LatestBlock`) and naming their tests.

### Jira gate

Fixed — `Validated MAG-2596`. It wanted the bare `^\s*jira\s+ticket\s*:` line, as you said; the `[MAG-2596](…)` link did not satisfy it.

---

### Verification

Run from a `git worktree` checkout of the pushed commit rather than the working directory — the blocker above is exactly what working-directory verification hides, and since no workflow runs `go test`, a local run is the only test evidence this PR has.

| Check | Result |
| --- | --- |
| `go build ./...` | clean |
| `go vet ./...` (whole repo, incl. test files) | clean |
| `gofmt -l ./protocol ./ecosystem` | clean |
| `go test -race ./ecosystem/cache/... ./protocol/performance/... ./types/relay/...` | ok |
| `go test ./protocol/rpcsmartrouter/... ./protocol/metrics/... ./protocol/tracing/...` | ok (91s) |

Also rebased onto current `main` to clear `BEHIND`, which pulled in `bd64f59` (subscription keying) — same package as this change, so the `rpcsmartrouter` suite was re-run after the rebase, not just before it.

Thanks also for the "checked and cleared" section — the typed-nil wiring, the `outputFormatter`-before-backfill ordering, the `IsNodeError`-on-primary-hits behaviour and the async-SET race were the four I most wanted a second pair of eyes on, and listing them explicitly means they won't get re-litigated.


[MAG-2596]: https://magmadevs.atlassian.net/browse/MAG-2596?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

AnnaR-prog:
## Re-review at `4b70b2c`

All six items from the first pass are addressed, and I verified each one rather than taking the commit messages for it. The `LatestBlock` commit in particular went further than I asked — asserting the *consequence* (a `LATEST`-tagged lookup still resolving to this router's own tip) instead of just the field being zero is the right test to have written, and it's the one that would actually catch a regression.

### Verified fixed

| First-pass item | Fix | How I checked |
| --- | --- | --- |
| 🔴 `ecosystem/cache` doesn't compile | `17cb29e` — `helpers_test.go` committed | `go vet ./...` now clean repo-wide; `go test ./ecosystem/cache/...` **ok** |
| Dead `annotate()` in the Jira gate | `6aa9c76` — removed | workflow is back to base; no longer in the PR diff |
| 🟠 Foreign `Reply.LatestBlock` | `7de7d70` — zeroed in the sanitizer | `TestSanitizeForeignCacheReplyDropsLatestBlock` + the rewritten poisoned-entry test; `isFinalizedForCacheWrite` now falls back to the tracked tip, and `SetRelay`'s stored floor to the locally derived `SeenBlock` |
| `outcome="hit"` before the copy could fail | `655cbfd` — copy moved ahead of classification | copy failure now records `outcome="error"`, not `"miss"` — right call, a miss would have filed a local defect as ordinary cache behaviour |
| Dangling config aborted startup | `ec6148d` — warning, tier left disabled | zero timeout and `read-write` mode correctly stay fatal |
| PR description | Jira line added, deleted design doc no longer referenced | `jira-ticket/validated` → **pass** ("Validated MAG-2596") |

I also re-ran everything on `4b70b2c` (worktree, merge base `bd64f59`):

```
go build ./...                                     clean
go vet ./...                                       clean          ← was the blocker
go test ./ecosystem/cache/...                      ok  (1.2s)     ← was build-failed
go test ./protocol/performance/...                 ok
go test ./protocol/metrics/...                     ok
go test ./types/relay/...                          ok
go test ./protocol/tracing/...                     ok
go test ./protocol/rpcsmartrouter/...              ok  (92.6s)
```

and confirmed the PR's own diff against the current merge base is 32 files, all secondary-cache scoped — the force-push didn't drag anything unrelated in.

---

### One consequence of the `LatestBlock` fix, and it's mine to own

Zeroing `LatestBlock` isn't quite inert on the serving path. `appendHeadersToRelayResult` emits `Provider-Latest-Block` (`protocol/common/endpoints.go:24`) whenever the value is positive:

```go
// protocol/rpcsmartrouter/rpcsmartrouter_server.go:4812
if relayResult.Reply.LatestBlock > 0 {
    metadataReply = append(metadataReply, pairingtypes.Metadata{
        Name:  common.PROVIDER_LATEST_BLOCK_HEADER_NAME,
        Value: strconv.FormatInt(relayResult.Reply.LatestBlock, 10),
    })
}
```

A **primary** hit still carries that header (its stored `LatestBlock` is untouched); a **secondary** hit now omits it. It's the only serving-side reader of the field — the consistency feed that used to read it is gone (`relaycore/relay_processor.go:663`, deliberately removed) — so nothing breaks. But it does make two doc lines slightly untrue:

- `docs/SECONDARY-CACHE.md:55` — "It is inert when serving"
- `docs/SECONDARY-CACHE.md:51` and `:469` — a secondary hit is "byte-identical to a primary-cache hit"

Two ways to square it, either fine:

- **(a)** Accept the omission, soften those three lines to say the two tiers are indistinguishable apart from `Provider-Latest-Block`. Smallest diff.
- **(b)** Stamp the *local* gated tip into `copyReply.LatestBlock` after `SanitizeForeignCacheReply`, instead of leaving zero. The header stays truthful and locally sourced, the tiers stay symmetric, and the backfill is unaffected — `isFinalizedForCacheWrite` would see `max(localTip, localTip)` and `SetRelay` `max(localTip, seenBlock)`, both local values, so none of the three problems the sanitizer fix closed can reopen. Two lines, and it keeps the doc claim as written.

I'd lean (b) since the doc's symmetry claim is load-bearing for operators reading `cache_tier` dashboards, but (a) is perfectly defensible.

---

### Also worth a look while you're in there

`docs/SECONDARY-CACHE.md:293` says steps 2 and 3 of the walkthrough print `Lava-Provider-Address: Cached` and are "indistinguishable" — same caveat as above once `Provider-Latest-Block` differs between the tiers.

Nothing else outstanding from my side. With (a) or (b) applied this looks ready to me.

AnnaR-prog:
## Ship check at `4b70b2c` — one blocker, outside this repo

I ran a full readiness pass rather than just re-checking the fixes, and it turned up something neither of my earlier reviews would have: **this PR breaks two tests in `Magma-Devs/smart-router-automation`.**

### 🔴 Blocking — the dangling-config change invalidates a cross-repo test contract

`tests/infrastructure/binary_startup/test_secondary_cache_config_validation.py` →
`TestFailFastValidation::test_bad_config_stops_startup_with_error` is parametrized with six cases. Two of them:

```python
pytest.param(["--secondary-cache-timeout", "100ms"], ERR_DANGLING,
             id="timeout_without_address_is_dangling"),
pytest.param(["--secondary-cache-mode", "read-only"], ERR_DANGLING,
             id="mode_without_address_is_dangling"),
```

and the test body asserts:

```python
assert run.exit_code != 0
assert ERR_WRAPPER in run.output      # "invalid secondary cache configuration"
```

After `ec6148d` both configurations boot with a warning — exit code 0, and the wrapper string (emitted only on the error path, `rpcsmartrouter.go:3235`) never appears. Both cases fail on both assertions. The `"dangling configuration"` detail string does survive in the new warning text, so `expected_detail` still matches; it's the exit code and the wrapper that go.

The other four cases — zero timeout, negative timeout, `read-write` mode, unknown mode — still pass, because those stayed fatal. That's the split working as intended; it's only the two dangling cases that moved category.

No companion PR exists in the automation repo (that file was last touched 2026-08-03 by #464).

**This one is on me as much as anyone.** My first-pass note suggested a warning would be enough, and I didn't know the contract was pinned in another repo. **Clears when** a companion automation PR moves those two cases from `TestFailFastValidation` to `TestBootWithWarnings` — asserting boot plus the advisory line instead of a non-zero exit — and the two land together.

Related, same root: **MAG-2596's configuration contract still specifies the error.** The ticket comment says *"`secondary-cache-timeout` or `secondary-cache-mode` set while `secondary-cache-be` is empty → **startup error** ("dangling secondary cache configuration")"*. The PR body documents the deviation; the ticket doesn't. Worth updating it so the next person doesn't re-derive the old behaviour from the spec — as the automation suite did.

### 🟡 Worth reconciling before this ships

**Mixed cache engines.** MAG-2596's description is a single line — *"Open question → don't support mixed cache engines."* `docs/SECONDARY-CACHE.md:592-597` answers it the other way:

> *Should mixed cache engine configurations be supported?*
> Yes. The two tiers are independent connections that speak the Smart Router cache protocol; the storage engine behind each address is that cache service's own concern and is invisible to the router.

The reasoning is sound and I'd probably agree with it. But it reverses the ticket's recorded resolution, the reporter (omer) isn't the PR author, and this is customer-facing text — so it wants their sign-off rather than landing as a doc assertion.

**`Provider-Latest-Block`** — still open from my previous comment; either of the two options there is fine.

### 🟢 Non-blocking

- `TestSmartRouterRecordCacheResult_TotalEqualsSuccessPlusFailedAfterMixed` was removed and not replaced. Nothing now asserts `requests_total == success_total + Σ failed_total[outcome]` — an invariant that matters *more* than it did, since `failed` gained a label and recovering the identity now requires summing across `outcome`.
- Grafana dashboards and alerts live outside this repo, so the `cache_tier` / `outcome` / latency-semantics migration can't be proven here. METRICS.md documents it well; there's no ticket naming who updates the boards.
- `init_smartrouter_eth_secondary_cache.sh` generates `config/smartrouter_examples/smartrouter_eth_zone_{internal,external}.yml` into a tracked directory, and neither is gitignored — running the lane leaves untracked files behind.

### Not proven (not a defect — just unmeasured)

The tier's wiring inside `sendRelayToEndpoint` — bypass rules widened to `primaryActive || secondaryActive`, the `if primaryCacheActive` wrap, `served → return nil`, and the `mergeBlockHashHeights` fold — has no test driving it. Nothing in the repo drives that function at all, and `chainstate_glue_test.go:317` already records that limitation, so this is pre-existing rather than something the PR introduced. The helpers around it are covered thoroughly; the ~20 lines of call-site wiring are covered only by the two-zone script, which I did not run (needs a live ETH endpoint). `RUN_DEMO=1 scripts/pre_setups/init_smartrouter_eth_secondary_cache.sh` with the two metric deltas captured would settle it.

### Verified myself, at this commit, in a worktree

| Check | Result |
| --- | --- |
| `go build ./...` | clean |
| `go vet ./...` (whole repo, incl. test files) | clean |
| `gofmt -l ./protocol ./ecosystem` | clean |
| `go test -race ./ecosystem/cache/... ./protocol/performance/... ./types/relay/...` | ok |
| `go test ./protocol/rpcsmartrouter/... ./protocol/metrics/... ./protocol/tracing/...` | ok |

**5/5 of the PR body's verification table confirmed**, including the falsifiability claim: deleting `reply.LatestBlock = 0` fails `TestSanitizeForeignCacheReplyDropsLatestBlock` ("Should be zero, but was 999999999") *and* `TestSecondaryPoisonedEntrySanitizedForCallerAndBackfill` ("was 987654321"), with the cache server logging `setting latest block key=LAVA_ latestBlock=987654321` exactly as described. That claim is worth having written down — it made the check a two-minute confirmation instead of an argument.

One thing I checked and cleared, so nobody re-opens it: the automation repo's Prometheus suite (`tests/simulator/simulator_prometheus_tests/test_prometheus_metrics.py`) reads `cache_success_total` with `label_filter={"method": …}` under `agg="single"`. Adding `cache_tier` keeps that at exactly one matching sample on a single-tier router, so the assertion holds — only the explanatory comment beside it goes stale.

AnnaR-prog:
## Follow-up: three more findings, two of which my earlier reviews missed

I ran a dedicated adversarial pass over `4b70b2c` after the ship check. It found things my first two reviews did not, and I've verified each one myself before repeating it here — every line reference below is one I opened.

### 🔴 The trust boundary has a second hole, in the same reply we just patched

`SanitizeForeignCacheReply` takes a `*RelayReply` (`sanitize.go:43`), so it can only reach `Sig`, `SigBlocks`, `Metadata`, `LatestBlock`. **`BlocksHashesToHeights` sits on the outer `CacheRelayReply`, structurally out of its reach — and it is used unsanitized.**

The miss path returns the raw foreign reply (`secondary_cache.go:121`), the caller folds it in at `rpcsmartrouter_server.go:3847-3848`, and `mergeBlockHashHeights` takes **max for latest, min for earliest** — so the more extreme value always wins and the foreign tier beats the local primary by construction. Those two scalars then steer two decisions:

```go
// rpcsmartrouter_server.go:1364-1368  — resolveRequestedBlock
if latestBlockHashRequested >= 0 &&
    ((reqBlock == spectypes.LATEST_BLOCK && protocolMessage.GetUsedDefaultValue()) ||
        reqBlock >= 0 && reqBlock < latestBlockHashRequested) {
    reqBlock = latestBlockHashRequested        // → endpoint sync gating, optimizer selection
}
```

```go
// rpcsmartrouter_server.go:5020-5023
extensionAdded := newProtocolMessage.UpdateEarliestAndValidateExtensionRules(
    rpcss.chainParser.ExtensionsParser(), earliestBlockHashRequested, addon, relayRequestData.SeenBlock)
if extensionAdded && relayState.CheckIsArchive(...) { relayState.SetIsArchive(true) }
```

It is reachable on the miss path, which is the common one: `getBlockHeightsFromHashes` runs on hits *and* misses and is assigned unconditionally (`ecosystem/cache/handlers.go:246`, `:249`), and the hashes are populated by every parser for hash-bearing calls like `eth_getBlockByHash`.

What makes this worth blocking on rather than filing: `sanitize.go:18-38` is a 20-line argument that a foreign scalar must never reach local chain-scoped state. It was written for `LatestBlock`. It applies verbatim to these two heights and was never made. We diagnosed the bug class last round and fixed one of its two instances in the same message.

Worst case is the secondary-only topology, where the primary block is skipped entirely, `:3835` never runs, and both scalars come *solely* from the foreign tier.

### 🟠 The two tiers do not serve the same response headers, and the docs say they do

`sanitize.go:50` sets `reply.Metadata = nil`. `Reply.Metadata` is the **only** source of client response headers:

```go
// protocol/chainlib/common.go:314-320
func addHeadersAndSendBytes(c *fiber.Ctx, metaData []pairingtypes.Metadata, data []byte) error {
	for _, value := range metaData { c.Set(value.Name, value.Value) }
	return c.Send(data)
}
```

`SetResponseFromRelayResult` does the same into `http.Header` (`base_chain_parser.go:612-614`), and for REST the entire upstream `res.Header` is what lands in `Metadata` in the first place (`rest.go:549`).

So on a secondary hit the caller loses every **upstream** header the primary tier passes through — `Content-Type` included. To be precise about what is and isn't affected, because the sanitizer's comment is half right: locally-minted headers *do* survive, because `appendHeadersToRelayResult` appends them after sanitization (`rpcsmartrouter_server.go:4942`), and `Lava-Provider-Address: Cached` is one of them (`:4679-4684`). But `sanitize.go:14-16` also names `Content-Type` as locally minted, and it is not — nothing mints it.

That leaves two artifacts in this repo asserting opposite policies for the same field: `sanitize_test.go:32-33` pins dropping `Content-Type` as correct ("nothing in Metadata is load-bearing"), while `rest_integration_test.go:348-354` asserts an arbitrary upstream header must reach the caller. And `docs/SECONDARY-CACHE.md:51` claims a secondary hit is "byte-identical to a primary-cache hit".

It also doesn't stay a first-request problem: the sanitized clone is what backfills (`secondary_cache.go:124` → `:160`), so the primary's own copy is stripped from then on.

The drop-everything policy was the right instinct for identity headers, and I argued for it. The inverse of the denylist you (correctly) rejected is an **allowlist** — keep `Content-Type`/`Content-Encoding`, drop the rest — which gets the security property without the serving regression. This supersedes my `Provider-Latest-Block` note from the last comment: same class, and I had found one instance of a much broader effect.

### 🟡 LATEST-tagged requests can't hit across zones — and the docs promise they can

For a LATEST request the block is resolved **client-side, before the GET**: `requestedBlockForCache = rpcss.getLatestBlock()` (`rpcsmartrouter_server.go:3706-3722`), passed unchanged to the secondary (`:3843` → `secondary_cache.go:66-75`). Because it arrives non-negative, the cache server's own tag resolution (`handlers.go:186-190`) never runs, and the key is `hash ‖ <this router's tip>` while the foreign zone wrote `hash ‖ <its tip>`. Two independently-tracking zones one block apart miss every time.

Explicit-block queries work fine, and a cross-zone cache is plausibly *for* those — so this is a scope question, not a defect. But it isn't disclosed, and one doc line states the opposite:

> `docs/SECONDARY-CACHE.md:38` — "the secondary is consulted once per entry, not per request"

For LATEST the key advances with the tip, so every request re-queries the secondary and misses, paying the cross-zone timeout each time. Nothing tests it either: `fakeCacheReader.GetEntry` (`secondary_cache_test.go:49-64`) returns its canned reply regardless of `RequestedBlock`, so no test can observe the rejection, and the demo pins a historical block (`init_smartrouter_eth_secondary_cache.sh:173`, `params: ["0x112a880", false]`) — the one shape immune to it. Worth either supporting LATEST explicitly or naming it in "Scope of this release" (`:534`).

### Confirmed, not new

The pass independently reached the same conclusion as my ship check on the untested `sendRelayToEndpoint` wiring, and ranked the bundled latency-histogram semantics change last, as taste. It also cleared several things I'd checked: the JSON (not protobuf) wire making `IsNodeError`/`StatusCode` skew-safe both ways, `CacheActive` nil-safety, lookup/backfill key agreement via `GetChainIdAndApiInterface`, and `FinalizedBlocksHashes` being inert (consumed nowhere under `protocol/` outside tests). It called the backfill tests genuinely good, and I agree.

nimrod-teich:
## Review at `4b70b2c` — independent pass

Checked out into a worktree (merge base `bd64f59`), built, vetted, and ran the suites myself. I read the three earlier reviews and deliberately re-verified every open item against the code rather than inheriting the conclusions — two of them I'd re-grade, one I'd close, and I found nothing new that blocks. The design holds up: `CacheReader` makes read-only structural, the backfill reuses the populator's eligibility, and the exact-key + lifted-`SeenBlock` write is correct against the real server.

**Verdict: not mergeable yet, for two reasons — one outside this repo, one a small deletion in it.** Everything else is docs.

---

### 🔴 Must land together: the automation companion PR

Confirmed at `origin/main` of `smart-router-automation`: `tests/infrastructure/binary_startup/test_secondary_cache_config_validation.py` (from #464) has `timeout_without_address_is_dangling` and `mode_without_address_is_dangling` under `TestFailFastValidation`, asserting `exit_code != 0` and the `"invalid secondary cache configuration"` wrapper. After `ec6148d` both boot with a warning, so both fail on both assertions. The other four fail-fast cases still hold.

Needs a companion PR moving those two cases to `TestBootWithWarnings` (assert boot + the advisory line). MAG-2596's comment also still specifies "startup error" for this case — update it so the spec matches what ships.

### 🟠 Foreign `BlocksHashesToHeights` — real, but latent here; the fix is a deletion

Verified: `SanitizeForeignCacheReply` works on `*RelayReply`, `BlocksHashesToHeights` is on the outer `CacheRelayReply`, and the miss path returns the raw foreign reply into `mergeBlockHashHeights` → `resolveRequestedBlock` / `updateProtocolMessageIfNeededWithNewEarliestData`.

One thing the earlier review didn't weigh: **no smart-router writer ever populates that server-side map.** The SET sends `BlocksHashesToHeights: nil` (`rpcsmartrouter_server.go:3560`, "Not available in direct RPC mode"), and `setBlocksHashesToHeights` (`handlers.go:346`) is the only writer to `blocksHashesToHeightsCache`. So a secondary fed by smart-routers always returns `NOT_APPLICABLE` heights and this path is dead — it goes live only when the secondary was populated by another lineage, which is exactly the case the sanitizer's own comment argues we must defend against.

Recommended fix is smaller than an allowlist: send `BlocksHashesToHeights: nil` on the secondary GET, drop the merge, delete `mergeBlockHashHeights` and its test. Fewer lines, and both resolution scalars stay local-only by construction.

### 🟡 Metadata / `Content-Type` — the earlier finding is overstated; this is a docs fix

The claim "nothing mints `Content-Type`" is wrong: `jsonRPC.go:455` and `rest.go:343` set `application/json; charset=utf-8` at the top of every HTTP handler, before `addHeadersAndSendBytes` overlays the reply's `Metadata`. So the sanitizer comment is right for `Content-Type`, and a secondary hit serves a valid JSON response.

What *is* true: a primary hit replays the upstream headers stored in the entry, a secondary hit (and the primary entry it backfills) does not, and neither carries `Provider-Latest-Block`. That's the safe direction and I would **not** add an allowlist for it. Fix the three doc lines instead — `docs/SECONDARY-CACHE.md:51`, `:293`, `:470` ("byte-identical" / "indistinguishable") → same body and `Lava-Provider-Address: Cached`; upstream pass-through headers and `Provider-Latest-Block` are absent on a secondary hit and its backfill.

### 🟡 LATEST-tagged requests can't hit across zones — agree, disclose it

Confirmed: for LATEST the key block is resolved client-side (`rpcsmartrouter_server.go:3706-3722`) and arrives non-negative, so `GetRelay`'s tag resolution never runs and the key is `hash ‖ <this router's tip>` vs the foreign zone's `hash ‖ <its tip>`. Two zones one block apart miss every time, and each such miss costs the full `secondary-cache-timeout` on the request path. Scope, not a defect — add it to "Scope of this release" and fix `docs/SECONDARY-CACHE.md:38` ("once per entry, not per request"), which is false for LATEST.

### 🟡 Mixed cache engines — needs the reporter's sign-off

Confirmed on the ticket: MAG-2596's description is literally *"Open question → don't support mixed cache engines."* `docs/SECONDARY-CACHE.md:592` answers "Yes". The reasoning is fine, but it reverses a recorded decision in customer-facing text — one line from omer on the ticket settles it either way.

### 🟢 Non-blocking

- Re-add the `requests_total == success_total + Σ failed_total{outcome}` identity test in `cache_metrics_test.go`. The invariant matters more now that recovering it requires a sum over `outcome`.
- Consider a sentence in the docs on the serial cost: every primary miss now waits up to `secondary-cache-timeout` before reaching a provider. By design per the PRD, but it's the number an operator tuning the timeout needs.

---

### Checked and cleared (so nobody re-opens them)

- **Generated zone configs are gitignored.** `.gitignore:36` (`config/smartrouter_examples/smartrouter_*.yml`) matches them — the earlier "leaves untracked files behind" note doesn't hold; the tracked examples are force-added.
- **Secondary timeouts don't thrash the gRPC client.** `resetOnConnectionError` (`performance/cache.go:129`) only resets on `codes.Unavailable`; a `DeadlineExceeded` from the 50 ms budget leaves the connection alone.
- **StatusCode asymmetry is harmless.** A primary hit hardcodes 200; a secondary hit serves the stored status (only zero → 200). Writers only store 2xx or 0, so the difference is unobservable unless a foreign lineage stores something else — and then serving its body already implies trusting it.
- **`IsNodeError` on primary hits** — same conclusion as before: the results manager classifies via `CheckResponseError`, so the only new effect is the truthful header.
- `resolveCachedEntryKind` decides kind before the placeholder substitution erases the marker, and a legacy-placeholder node error is correctly rejected from backfill via `relayResult.IsNodeError`.

### Verification (worktree at `4b70b2c`, go 1.26.6)

| Check | Result |
| --- | --- |
| `go build ./...` | clean |
| `go vet` on all touched packages | clean |
| `gofmt -l ./protocol ./ecosystem ./types` | clean |
| `go test -race ./ecosystem/cache/... ./protocol/performance/... ./types/relay/...` | ok |
| `go test ./protocol/metrics/... ./protocol/tracing/...` | ok |
| `go test ./protocol/rpcsmartrouter/...` | ok (95 s) |
| `go test -race ./protocol/metrics/...` | **FAIL — pre-existing.** `TestHappyFlow` in `relays_monitor_test.go` races identically on `origin/main` (`bd64f59`). Not this PR; worth its own ticket. |

I also confirmed the PR body's falsifiability claim holds at this HEAD (the `LatestBlock` tests are the ones that would catch a regression) and that the diff against the merge base is the same 32 secondary-cache-scoped files.

avitenzer:
Both passes addressed at `f6a92f0`, in four commits on top of `4b70b2c`. The adversarial pass earned its keep — the `BlocksHashesToHeights` finding is the same bug you diagnosed last round, and I fixed one of its two instances while writing a 20-line comment arguing why the other one mattered.

---

### 🔴 `BlocksHashesToHeights` — fixed in `c376d1e`

Correct on every point, including that the miss path is the common one. The lookup now neither asks for the mappings nor returns them; `mergeBlockHashHeights` is gone with them.

I pinned it on the **request** rather than the reply. The reply side is already structural — the function no longer returns a `*CacheRelayReply`, so a fold is a compile error — but a future change could re-add the return and re-plumb it. A request that never asks can't. `TestSecondaryLookupNeverRequestsForeignBlockHashHeights` fails on the reverted code with `Expected nil, but got: []*relay.BlockHashToHeight{…}`.

**One thing worth flagging, because my first version of that test was worthless.** I wrote it against the existing REST fixture and it passed — then passed *again* with the fix reverted. `eth_getBlockByHash` carries no `BLOCK_HASH` parser in the bundled ETH1 spec, so `GetRequestedBlocksHashes()` returns empty and `newBlocksHashesToHeightsSliceFromRequestedBlockHashes` returns nil either way. The assertion was vacuously true. The test now uses `trace_transaction` (`parse_type: BLOCK_HASH` on `.params.[0]`) and asserts the fixture genuinely carries a hash before asserting anything about the request. Same reason your `LatestBlock` note asked for the consequence rather than the mechanism.

I took the cost you'd expect: hash-keyed archive detection now uses the primary's mappings alone, or none in a secondary-only topology. That is where a router with no cache configured already is, so it's a lost optimization rather than a regression.

### 🟠 Metadata → allowlist — fixed in `02d338e`

You were right, and specifically right about `Content-Type` not being minted. The fiber handlers *do* call `fiberCtx.Set(HeaderContentType, application/json)` — but at handler entry, as a default that `addHeadersAndSendBytes` overrides from `Metadata`. So the sanitizer's comment was true for JSON-RPC by accident and false for exactly the case it matters, a non-JSON REST body. That is now the test (`TestSanitizeForeignCacheReplyKeepsNonJSONContentType`) rather than a claim.

Allowlisted `Content-Type` and `Content-Encoding`, as you proposed. `Content-Length` stays out: `outputFormatter` re-stamps the payload before it's served, so a replayed length can disagree with the body — the primary tier has the same exposure, but there's no reason to import it here.

Your two-conflicting-artifacts point is what settled it for me. `sanitize_test.go` pinning "nothing in Metadata is load-bearing" against `rest_integration_test.go` pinning an arbitrary upstream header reaching the caller isn't a disagreement about the secondary tier — it's the repo asserting two policies for one field, and only one of them was reasoned about.

### `Provider-Latest-Block` — option (b), in `c376d1e`

Took (b). Your reasoning for it was the deciding factor: the doc's symmetry claim is load-bearing for operators reading `cache_tier` dashboards, and (a) would have made the docs true by lowering what they promise. Re-stamped from the gated tip after sanitization, so `isFinalizedForCacheWrite` sees `max(localTip, localTip)` and `SetRelay` stores `max(localTip, local SeenBlock)` — both local, exactly as you predicted. No local tip yet → still zero, header still omitted, which is what every other path does.

### 🟡 LATEST cross-zone — documented, not fixed

Confirmed by reading the same lines. Disclosed under *Scope of this release* as a limitation of the release rather than a rejected configuration, since the config is valid and only its benefit is bounded. Your sharper observation is the one I put in the docs: it isn't just "misses", it's that the key advances with the tip, so it re-queries **per request** and pays the cross-zone timeout each time — which flatly contradicted the Backfill bullet's "once per entry, not per request". That line now scopes itself to explicit-block queries.

Not fixed because making two zones agree on a block for `latest` is a protocol change, not a configuration one. Operators whose workload is mostly `latest`-tagged now get told to keep the timeout tight.

### 🔴 The cross-repo blocker — [automation#872](https://github.com/Magma-Devs/smart-router-automation/pull/872)

You called this exactly right, down to which assertions fail and which four cases survive. I built a binary from this branch and ran the suite both ways:

```
with automation#872:      12 passed
with origin/main's file:  timeout_without_address_is_dangling  FAILED
                          mode_without_address_is_dangling     FAILED
                          zero_timeout_rejected                PASSED
                          negative_timeout_rejected            PASSED
                          read_write_mode_rejected             PASSED
                          unknown_mode_rejected                PASSED
```

Both cases move to `TestBootWithWarnings`, asserting boot + the advisory + **no secondary configured** — the third one because a router that warned and then enabled an address-less secondary would pass the first two. I also wrote the error/warning split into the docstrings, since that boundary is the thing that moved: fatal iff an address is present *and* a setting would make that secondary run wrong.

I don't think this was yours to own. Your first-pass note argued the fleet case correctly; nothing in the smart-router repo points at that file, and the traceability table's pointer to it runs the wrong direction. The automation PR's docstring now says a case that looks like it belongs in the other class needs a companion smart-router PR — which is the missing signal, in the place someone would actually read it.

MAG-2596's configuration contract is updated too, so the spec no longer says "startup error" for a case the code warns about.

### 🟡 Mixed cache engines — reverted to open

Agreed, and I shouldn't have landed it as a doc assertion. Reverted: the doc now says MAG-2596's resolution stands and that this release does not overturn it, and keeps only the mechanism claim (the router never learns which engine is behind either address, so a mixed pairing would be undetected rather than supported). Sign-off asked for on the ticket rather than assumed here.

### 🟢 Non-blocking

- **Metrics identity** — restored as `TestSmartRouterRecordCacheResult_TotalEqualsSuccessPlusFailedPerTier`, per tier over a mixed run, with both sides required non-zero so it can't pass vacuously. Your framing that it matters *more* now is in the test comment and in METRICS.md, which states the `sum without (outcome)` aggregation.
- **Grafana** — can't be proven here, agreed. Flagged on MAG-2596 that no ticket names an owner; that's the actionable part.
- **Generated configs** — I think this one's a false alarm: `.gitignore:36` is `config/smartrouter_examples/smartrouter_*.yml`, which both zone files match. `git check-ignore -v` confirms. Nothing to change unless you're seeing them untracked on your side, in which case I'd want to know why.
- **Untested `sendRelayToEndpoint` wiring** — no change, and I agree with your ranking. It's pre-existing (`chainstate_glue_test.go:317`) and widening it is a bigger change than this PR should carry.

---

### Verification

Worktree checkout of `f6a92f0`, not the working directory.

| Check | Result |
| --- | --- |
| `go build ./...` | clean |
| `go vet ./...` (whole repo, incl. test files) | clean |
| `gofmt -l ./protocol ./ecosystem` | clean |
| `go test -race ./ecosystem/cache/... ./protocol/performance/... ./types/relay/...` | ok |
| `go test ./protocol/rpcsmartrouter/... ./protocol/metrics/... ./protocol/tracing/...` | ok |

**One thing that table hides, so I'd rather say it than have you find it:** `protocol/metrics` `TestHappyFlow` fails under `-race`. I checked it against `4b70b2c` in a separate worktree and it fails there identically, so it predates this round — but it means the metrics row above is a non-`-race` run, same as your earlier tables, and neither of us has actually run that package under `-race` green.

Each fix was verified by reverting it and watching the test fail, not by trusting a green run on the fixed code — the practice paid for itself on the hash test above, which was green for the wrong reason.

avitenzer:
One more finding after `f6a92f0`, fixed at `ff97ae2` in one commit. It is the same bug class rounds 1 and 2 closed — a foreign scalar reaching local chain-scoped state — on a fourth channel that all of the earlier passes missed, including both of mine. I also ran the two-zone lane, which closes the one item the ship check left unmeasured.

---

### 🔴 The fourth channel: the backfill **key** itself — fixed in `ff97ae2`

For an explicit-block request, `requestedBlockForCache` is the **raw user-requested height** — only `LATEST` is resolved against the local tip before the GET (`rpcsmartrouter_server.go:3706-3722`). The exact-key backfill then lifted the SET's `SeenBlock` to that height unconditionally, and its own comment justified it with a claim that is true only on the `LATEST` path: *"the resolved block came from the LOCAL guarded tip at lookup time."* On the explicit-block path it is the user's number, vouched for by nothing but the foreign tier's willingness to answer that key.

Downstream, `SetRelay` publishes `max(Response.LatestBlock, SeenBlock)` as the cache server's **chain-level tip** (`ecosystem/cache/handlers.go:359`, `:388`) through the monotonic-max write nothing can lower until expiry, and that key resolves every negative block tag (`:187-189`). So: one request at a user-chosen far-future key, answered by a foreign secondary whose reply is *fully sanitize-clean* — zero `LatestBlock`, no metadata, no signatures; **the poison is the key, not the payload, so it is structurally beyond `SanitizeForeignCacheReply`'s reach** — retargets `LATEST`/`SAFE`/`FINALIZED`/`PENDING` resolution chain-wide on this router's own primary until expiry. Demonstrated before fixing:

```
DBG setting latest block key=LAVA_ latestBlock=987654321
POISON CONFIRMED: LATEST-tagged lookup resolved to the far-future key and served it
```

**The fix** clamps the lift to the local gated tip. The legitimate cases never exceed it, so nothing is lost: a `LATEST`-resolved block *came from* `getLatestBlock()`, and a requester's `SeenBlock` is a local parse-time value — the backfill still lands on its exact key and stays visible to its own follow-up GET. Worth stating honestly: with round 2's re-stamp in place, `Response.LatestBlock` already carries the local tip, so the unclamped lift was only load-bearing at **exactly** the values that needed clamping. I kept the clamp rather than deleting the lift because the re-stamp exists for a serving concern (header symmetry) — a future change could touch it without thinking about validity floors, and the floor should not silently depend on it. The exact-key test's docstring now records that the floor has two local sources and regresses only if both go.

**Falsifiability, checked by reverting as usual:** `TestSecondaryFutureKeyBackfillNeverRaisesPrimaryChainTip` is a consequence test in the shape this thread has converged on — an honest entry at the tip first (so a nil result cannot be told apart from "no tip was ever published"), then the far-key backfill, then: stored floor ≤ tip, and `LATEST` still serves the *honest* entry. On the unclamped code it fails with `"987654321" is not less than or equal to "100"`, with the cache server logging the poison line above.

Two consequential corrections ride along: the traceability row now says **four** channels (`docs/SECONDARY-CACHE.md`), and my round-2 line "no foreign scalar reaches local chain-scoped state" was false between `2da5a7b` and `ff97ae2` — the count in that comment and the old doc row was wrong. The exact-key test also now sets the guarded tip its scenario always described; previously the fixture had no tip and the lift was silently doing the re-stamp's work.

### 🟢 The unmeasured wiring, now measured

@AnnaR-prog's ship check left the `sendRelayToEndpoint` wiring "not proven — just unmeasured" and named the run that would settle it. Ran it: `RUN_DEMO=1 scripts/pre_setups/init_smartrouter_eth_secondary_cache.sh` at `f6a92f0`, binary provenance checked the paranoid way (`strings $(which smartrouter) | grep -c f6a92f0` → 6, version `v1.4.1-25-gf6a92f0`). Steps 2 and 3 both served `Lava-Provider-Address: Cached`, and the tiers split exactly as documented:

```
smartrouter_cache_success_total{...,cache_tier="secondary",method="eth_getBlockByNumber"} 1   ← the cross-zone rescue
smartrouter_cache_success_total{...,cache_tier="primary",method="eth_getBlockByNumber"}   1   ← the repeat, after backfill
```

The gate-0 table earned its keep on the way there: the first attempt ran a **stale May-7 binary** — `make install` failed on an IDE-injected `GOFLAGS` env var (make re-exports its `GOFLAGS` with embedded quoted `-ldflags`, which the env-var form can't carry), and `$PATH` quietly served a 6.1.0 build that rejected `--secondary-cache-be`. The `strings`/version check caught it; `env -u GOFLAGS` re-ran clean. Local quirk, not a repo change — but if anyone else's lane "fails to start", check that first. (`ff97ae2` doesn't touch the demo path — the clamp is a no-op for in-range keys — but strictly, the lane evidence is from `f6a92f0`.)

### 🟢 Two observations from the same pass, not addressed

- **Status asymmetry, now observable:** a secondary hit serves the writer-recorded status (a stored 203 reaches the caller — pinned by `TestSecondaryHitPreservesWriterRecordedStatusThroughBackfill`), but the primary-hit path hardcodes 200 — so the same entry answers 203 on first serve and 200 on the repeat. nimrod cleared the asymmetry as unobservable last round; the backfill is what makes it observable. Two-line fix on the primary path or one doc sentence — happy to do either if you want it in this PR.
- **`SAFE`/`FINALIZED`/`PENDING` cross zones where `latest` can't:** they reach the secondary as negative tags and resolve against the *foreign* server's tip (`handlers.go:187-189`) — no chain-state adoption (a negative `resolvedBlock` skips both the key override and the lift; traced it), but the Scope section currently singles out `latest` and says nothing about the other tags behaving differently. One sentence would close it.

### Verification

Run against the exact content of `ff97ae2` — from a working tree whose tracked files were byte-identical to the pushed commit, not from a separate worktree checkout:

| Check | Result |
| --- | --- |
| `go build ./...` | clean |
| `go vet ./...` (whole repo, incl. test files) | clean |
| `gofmt -l ./protocol ./ecosystem` | clean |
| `go test -count=1 ./...` — all 36 packages | ok |
| `go test -race ./ecosystem/cache/... ./protocol/performance/... ./types/relay/...` | ok |
| clamp reverted → new test | **fails** ("987654321 is not less than or equal to 100") |

`TestHappyFlow` under `-race` still fails and is still pre-existing — re-confirmed this round by running it at the merge base `bd64f59` and at HEAD back-to-back: identical failure, 16s both.

`ff97ae2` is unreviewed, and it edits the exact area rounds 2 and 3 reworked — it needs the same eyes the last four commits never formally got. Requesting review from @AnnaR-prog / @nimrod-teich; automation#872 is unchanged by this round and still lands together with this PR.

AnnaR-prog (approved):

PR #252, opened 2026-08-04
fix(rpcsmartrouter): stop relay-race cancellations penalising endpoint health and QoS

Tomelia1999:
# Description

A stateful method (`stateful: 1` in the spec — `eth_sendRawTransaction`, Solana
`sendTransaction`) is broadcast to **every** endpoint. The first response wins and the
router cancels the rest. Those cancelled relays were being recorded as **endpoint
failures** at three independent sites.

On a broadcast the fastest node wins and every *other* healthy node takes the hit, so the
penalty is structural rather than correlated with node quality. Reported by a customer on
Polygon: healthy internal Bor endpoints flipping available/unavailable 30–110×/3h with
availability scores at 0.0–0.4, while block lag was 0 and sync ~0.99.

**Nothing was broken for clients** — the winning relay always succeeded. The damage was to
the router's own endpoint scoring and to every dashboard built on the error metrics.

## Root cause

A carve-out for exactly this case already existed at one of the three sites and had never
worked. `classifyDirectRPCError` rebuilt the error from `err.Error()` — a plain string —
so `LavaWrappedError` carried no cause:

```go
return classified, common.NewLavaError(classified, err.Error())
//                                                 ^^^^^^^^^^^ flattened
```

`errors.Is(err, context.Canceled)` was therefore always false, `IsClientCancellation`
always returned false, and `classifyEndpointHealth` fell through to the
`CategoryInternal` arm → `MarkUnhealthy()` + backoff.

The classification itself was correct (`PROTOCOL_CONTEXT_CANCELED`) — only the sentinel
needed by the carve-out was lost.

## Approach

Restore the signal at the root, then honour it at all three sites.

**1. Keep the cause** (`error_registry.go`, `error_mapper.go`)

`LavaWrappedError` gains a `cause`, `Unwrap` returns it, and a new `As` keeps the
`*LavaError` reachable. One call site switches to `NewLavaErrorWrapping`.

The singular `Unwrap() error` is deliberate. `Unwrap() []error` is the tidier-looking form
and is **wrong here**: `errors.Unwrap` is documented to return nil for multi-unwrap types,
which breaks three existing tests and the chain walker in `utils/score/score_errors.go`.
The comment in the code explains this so a future refactor does not silently reintroduce
the bug.

This change alone revives the health carve-out with no edit to `classifyEndpointHealth`.

**2. Release the session without blaming it** (`consumer_session_manager.go`)

New `OnSessionCancelled`, sharing a private helper with the existing `OnSessionDiscarded`
— which already had exactly this accounting, on exactly this reasoning ("no upstream
request was made, so availability was not actually tested"). A race loser is the same
category: dispatched, but no verdict about the endpoint was obtained.

**3. A third relay outcome** (`smartrouter_metrics_manager.go`)

`RecordDirectRelayEnd` takes a `RelayOutcome` (success / error / cancelled) instead of a
success bool. Two words of vocabulary were not enough for three real outcomes.

### The trap that shapes 2 and 3

Neither call could simply be skipped for a cancelled relay, because **both do cleanup as
well as blame**:

- `RecordDirectRelayEnd` also decrements the in-flight gauge that `RecordDirectRelayStart`
  incremented. Skipping it leaks `rpc_endpoint_requests_in_flight` upward forever, one per
  race loser.
- `OnSessionFailure` also frees the session and returns its compute units. Skipping it
  leaks sessions until the provider stops being granted new ones.

Each needed a third path that keeps the cleanup and drops only the penalty.
`TestCancelledRelay_InFlightGaugeReturnsToZero` pins the naive fix out of existence.

## New metric

| Metric | Type | Labels | Description |
| --- | --- | --- | --- |
| `rpc_endpoint_total_cancelled` | Counter | `spec`, `apiInterface`, `endpoint_id`, `function` | Relays the router aborted before completion — race losers on stateful broadcasts, and client disconnects. Not an endpoint fault. |

A separate counter rather than silence: dropping these entirely would hide the broadcast
cancel rate, which is the number you need to tune fan-out. Documented in `docs/METRICS.md`
with the expected `(N-1)/N` rate on write traffic.

## Behaviour changes reviewers should know about

- **`rpc_endpoint_total_errored` will drop sharply on write-heavy specs.** That volume
  moves to `rpc_endpoint_total_cancelled`. This is the fix working, but it will look like
  a measurement break on a dashboard.
- **`smartrouter_requests_*` no longer counts cancelled relays.** For a stateful broadcast
  to N endpoints, `requests_total` now counts completed relays rather than all dispatched
  ones. This preserves `requests_total == requests_success + requests_failed` exactly, at
  the cost of lower write-method request counts.
- **Client disconnects are fixed as a side effect.** They ride the same
  `IsClientCancellation` rule, which is correct — a user hanging up is not an endpoint
  fault. It means the error-rate drop is not limited to broadcast traffic.
- **Cancelled relays no longer record latency.** A cancel timing measures "time until the
  winner returned", not that endpoint's performance, and would have dragged the slowest
  endpoints' distributions toward zero.

### Live A/B on a 3-endpoint SOLANAT router

Same server, same config, same 90 stateful broadcasts. Only the binary changed. All three
`direct-rpc` entries pointed at the **same upstream URL**, so latency and sync were
identical and the only variable was which endpoint answered first.

| Measurement | Before | After |
| --- | --- | --- |
| `rpc_endpoint_total_errored` | **160** | **0** |
| `rpc_endpoint_total_cancelled` | *(did not exist)* | 167 |
| `selection_score{availability}` | 0.00 / 0.00 / 0.00 | 0.93 / 0.90 / 0.94 |
| `selection_score{sync}` | 1.00 / 1.00 / 1.00 | 1.00 / 1.00 / 1.00 |
| `rpc_endpoint_overall_health` flips | 28 | 0 (723 samples) |
| `rpc_endpoint_requests_in_flight` | 0 | 0 |
| Client-visible failures | 0 / 91 | 0 / 90 |

Accounting balances on both sides: `serviced + (errored | cancelled) == 90 × 3 == 270`.

### Genuine faults are still penalised

The obvious risk in a fix like this is trading a false-positive for a much worse
false-negative. Verified two ways:

- **Live:** a deliberately broken 4th endpoint (a proxy killed mid-run) took **17 errors**
  and went unhealthy, while the three healthy endpoints took **0** in the same run.
- **Unit:** `TestGenuineFaults_StillPenalised` asserts connection refused / reset / host
  unreachable / deadline / EOF all still mark the endpoint unhealthy **while the context
  is cancelled** — the exact state a race loser is in.

Also pinned: rate limiting keeps its "healthy but busy" treatment, `OnSessionFailure`
still records QoS penalties, and cross-validation stragglers on `WithoutCancel` contexts
are correctly *not* treated as cancelled.

### Test plan

- 24 new/updated tests across `common`, `metrics`, `lavasession`, `rpcsmartrouter`
- `go build ./...`, `go vet ./...`, `gofmt`: clean
- Full suite: **32 packages, 0 failures**
- `-race` on the changed packages: no races in this change. One pre-existing flake
  (`TestHappyFlow`, `relays_monitor_test.go`) reproduces identically on clean `main` —
  confirmed by running `main` 10× (2 failures) vs this branch 7× (2 failures)
- Review note (avitenzer, 2026-08-05): two more pre-existing flakes reproduce on clean
  `main` in full-package `lavasession` runs — `TestCheckAndUnblock_BackupUnblockedWhenHealthy`
  and `TestEndpointHealthRecovery` (both pass in isolation; confirmed flaky in a clean `main`
  worktree, so not introduced by this branch)

### Why the previous tests missed this

`TestClassifyEndpointHealth_ClientCancellationCarvesOut` passes
`isClientCancellation` as a literal `true`. It verifies the branch but never whether the
flag is ever computed true on the real path, so it stayed green for the entire life of the
bug. `IsClientCancellation` itself had no test at all. Both are now covered from a real
transport error forward.

## Known gap — this fix is HTTP-only

**The gRPC transport still has this bug, with an inverted and worse failure mode.**

`handleGRPCError` rebuilds the error as `GRPCStatusError{Code, Message}` with no cause,
and grpc-go's `status.Error(codes.Canceled, ...)` does not wrap `context.Canceled` either.
Because `codes.Canceled` is 1 (below the `>= 13` node-error threshold) the consumption arm
returns a nil error, so a cancelled gRPC relay is recorded as a **success** — including a
fabricated latency sample and an active `ResetHealth()`.

Confirmed by driving the real code path. Affects the seven specs with a `grpc`
apiInterface (Cosmos family + Lava). Not fixed here because it needs its own change and
its own verification; tracked with a full plan in the ticket.

## Follow-up

From code review, tracked separately:

- **gRPC transport** (above) — highest priority
- Race losers still appear in the client's errored-providers metadata
- `rpc_endpoint_total_cancelled` merges race losers with client disconnects; needs a
  `reason` label to separate them
- Client-disconnected requests are now invisible in `smartrouter_requests_*`
- Nil guards on `LavaWrappedError`
- Adjacent pre-existing bug: gRPC `DeadlineExceeded` is also recorded as a success

## Files most worth reviewing

1. `protocol/common/error_registry.go` — the root fix; read the `Unwrap` comment
2. `protocol/rpcsmartrouter/rpcsmartrouter_server.go` — the two call sites; note that the
   `isClientCancel` branch must precede the `shouldFailSession` test
3. `protocol/metrics/smartrouter_metrics_manager.go` — the in-flight decrement must run
   for every outcome

---

## Jira ticket

Jira ticket: MAG-2648

---

### DIFF
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -73,7 +73,7 @@ They split into **endpoint-scoped** (`rpc_endpoint_*`) and **router-scoped**
 | Metric | Type | Labels | Description |
 | --- | --- | --- | --- |
 | `rpc_endpoint_total_relays_serviced` | Counter | `spec`, `apiInterface`, `endpoint_id`, `function` | Relays successfully served by this endpoint. |
-| `rpc_endpoint_total_errored` | Counter | `spec`, `apiInterface`, `endpoint_id`, `function` | Errored relays for this endpoint. Excludes relays the router cancelled — see `rpc_endpoint_total_cancelled`. |
+| `rpc_endpoint_total_errored` | Counter | `spec`, `apiInterface`, `endpoint_id`, `function` | Errored relays for this endpoint. Excludes relays the router itself cancelled — see `rpc_endpoint_total_cancelled`. |
 | `rpc_endpoint_total_cancelled` | Counter | `spec`, `apiInterface`, `endpoint_id`, `function` | Relays the router aborted before completion: relay-race losers on stateful broadcasts, and client disconnects. **Not an endpoint fault** — excluded from `rpc_endpoint_total_errored` and from QoS/availability scoring. |
 | `rpc_endpoint_requests_in_flight` | Gauge | `spec`, `apiInterface`, `endpoint_id`, `function` | Relays currently in flight to this endpoint. |
 | `rpc_endpoint_end_to_end_latency_milliseconds` | Histogram | `spec`, `apiInterface`, `endpoint_id`, `function` | End-to-end latency per function for this endpoint. |
@@ -81,18 +81,25 @@ They split into **endpoint-scoped** (`rpc_endpoint_*`) and **router-scoped**
 | `rpc_endpoint_overall_health_breakdown` | Gauge | `spec`, `apiInterface` | Aggregate health per chain/interface. |
 | `rpc_endpoint_selection_score` | Gauge | `spec`, `apiInterface`, `endpoint_id`, `score_type` | Selection scores by `score_type` (availability / latency / sync / stake / composite). |
 | `rpc_endpoint_latest_block` | Gauge | `spec`, `apiInterface`, `endpoint_id` | Latest block reported by the endpoint. |
-
-> **Reading cancelled relays.** A write (`stateful: 1`, e.g. `eth_sendRawTransaction`) is
-> broadcast to every endpoint; the first response wins and the rest are cancelled. So a
-> healthy endpoint on write-heavy traffic will show a high `rpc_endpoint_total_cancelled`
-> rate by design — roughly `(N-1)/N` of broadcasts, for N endpoints. That is the number to
-> watch when tuning broadcast fan-out, not a fault signal. Cancelled relays still decrement
-> `rpc_endpoint_requests_in_flight`, and never move the `smartrouter_requests_*` group, so
-> `requests_total == requests_success + requests_failed` remains exact.
 | `rpc_endpoint_fetch_latest_fails` | Counter | `spec`, `apiInterface`, `endpoint_id` | Failed latest-block fetches. |
 | `rpc_endpoint_fetch_block_fails` | Counter | `spec`, `apiInterface`, `endpoint_id` | Failed specific-block fetches. |
 | `rpc_endpoint_fetch_latest_success` | Counter | `spec`, `apiInterface`, `endpoint_id` | Successful latest-block fetches. |
 | `rpc_endpoint_fetch_block_success` | Counter | `spec`, `apiInterface`, `endpoint_id` | Successful specific-block fetches. |
+
+> **Reading cancelled relays.** A stateful method (`stateful: 1` in the spec — e.g.
+> `eth_sendRawTransaction`, Solana `sendTransaction`) is broadcast to *every* endpoint; the
+> first response wins and the rest are cancelled. A healthy endpoint under write traffic will
+> therefore show a high `rpc_endpoint_total_cancelled` rate **by design** — with N endpoints,
+> roughly `(N-1)/N` of every broadcast. That is the number to watch when tuning broadcast
+> fan-out; it is not a fault signal.
+>
+> The two counters partition the non-success outcomes: `total_errored` is the endpoint's
+> fault, `total_cancelled` is ours. Cancelled relays still decrement
+> `rpc_endpoint_requests_in_flight`, and never move the `smartrouter_requests_*` family, so
+> `requests_total == requests_success + requests_failed` remains exact.
+>
+> Measured on a 3-endpoint SOLANAT router (MAG-2648): 90 stateful broadcasts produced 103
+> serviced + 167 cancelled = 270 relays = 90 × 3, with zero errored.
 
 ### Optimizer
 
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -73,13 +73,22 @@ They split into **endpoint-scoped** (`rpc_endpoint_*`) and **router-scoped**
 | Metric | Type | Labels | Description |
 | --- | --- | --- | --- |
 | `rpc_endpoint_total_relays_serviced` | Counter | `spec`, `apiInterface`, `endpoint_id`, `function` | Relays successfully served by this endpoint. |
-| `rpc_endpoint_total_errored` | Counter | `spec`, `apiInterface`, `endpoint_id`, `function` | Errored relays for this endpoint. |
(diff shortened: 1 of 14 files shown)


avitenzer:
**`protocol/common/error_registry.go` — nil guards in `Is` and the new `As`**

Both methods dereference `e.LavaErr` unguarded. `Is` is pre-existing, but `As` is new in this PR, so it would ship with the same latent nil-deref the follow-up list already tracks ("Nil guards on `LavaWrappedError`") — and with a worse failure mode than a missed match: if `LavaErr` is ever nil, `As` returns `true` while writing a nil `*LavaError` into the target, so `extractLavaError(err).Code`-style callers panic.

```go
func (e *LavaWrappedError) As(target any) bool {
	if e == nil || e.LavaErr == nil {
		return false
	}
	if t, ok := target.(**LavaError); ok {
		*t = e.LavaErr
		return true
	}
	return false
}
```

Same two-line guard at the top of `Is`. Constructors always set `LavaErr` today, so this is latent — but it's two lines now vs. a tracked follow-up, and it closes the new surface this PR adds rather than deferring it.

*(Noted during review of the MAG-2648 changes — the rest of the error-registry change looks correct, including the singular-`Unwrap` reasoning.)*

Tomelia1999:
Agreed, and taken in 7f446b2 — with the guard extended past the two methods you named.

`As` was the one worth catching, for exactly the reason you give: returning `true` with a nil `*LavaError` is strictly worse than never matching, because `errors.As` stops walking. There's now a test that builds a `LavaErr`-less wrapper around a resolvable inner error and asserts the walk still reaches it.

Two more methods had the same unguarded dereference:

- **`Error()`** — pre-existing like `Is`, and the worst failure mode of the set: it *panics*, inside a logging call. Now returns `Context`.
- **`Unwrap()`** — this one is new surface from this PR. It returned `e.LavaErr` directly, so a nil there produces a non-nil `error` interface holding a nil `*LavaError`; the next `errors.Is` calls `LavaError.Is` on a nil receiver and panics. Only reachable because this PR gave `Unwrap` a branch. Now returns a plain `nil`.

Nil-receiver checks on all four as well, since `errors.As` will call these on a nil `*LavaWrappedError` held in a non-nil interface.

One small correction on the example: `extractLavaError(err).Code` doesn't actually go through the new `As`. It targets `**LavaWrappedError`, which matches the concrete type directly, and its main call site already checks `if classified == nil`. The panicking shape you're describing is the `errors.As(err, &le)` → `le.Code` caller, which is real. The finding stands; it's just attached to a different mechanism.

The tests panic without the guards rather than merely failing, so they'd catch a regression here rather than quietly passing. Full suite green: 32 packages.

This also clears "Nil guards on `LavaWrappedError`" off the follow-up list.

avitenzer (approved):

PR #257, opened 2026-08-06
refactor(smart-router): clean up leftovers from the dead-code sweep

nimrod-teich:
# Description

Follow-up to #254 / #255, from a review of those diffs after merge.

`deadcode` and `staticcheck -checks=U1000` report **functions and methods**. Neither reports **types**, and golangci-lint's `unused` skips exported identifiers. So the sweep removed the methods but left behind the declarations that existed only to carry them — each now sitting there with a doc comment describing behaviour that is gone. This PR closes that gap and fixes the comments that outlived their code.

**Orphaned by their own removed methods:**
- `TestModeChainRouter` — both of its `ChainRouter` methods were removed, so it no longer implements the interface it exists for; its comment still described routing behaviour.
- `DummyChainFetcher` — all three methods plus `NewVerificationsOnlyChainFetcher` removed, leaving an embed-only shell.
- `MONTHS_IN_YEAR` and `DebugPaymentE2E` — the latter's only reader was `NextMonth`, and nothing sets it via `-ldflags -X` either.

**Files left holding nothing but a package clause** (deleted): `utils/convert.go`, `utils/time.go`, `utils/debug.go`, `utils/specfetcher/expand.go`, `protocol/lavasession/testing_helpers.go`, and the `rpcclient` `endpoints.go` / `inproc.go` stubs (license header + `package rpcclient`, nothing else).

**`protocol/relaycore/consistency.go` — please sanity-check this one.** The file was reduced to a retirement note ending *"Only the key derivation survives, because the shared-state cache path still identifies a caller by dapp+IP … tracked as follow-up task T10."* But #254 deleted `UserDataKey`, which is the key derivation that sentence is about. Reading the current code, the shared-state cache path keys on the listen endpoint (`sharedStateId = rpcss.listenEndpoint.Key()`), not dapp+IP — so T10's premise looks already satisfied and the note is simply stale. I deleted the file rather than keep a code-free file carrying a false claim; the same rationale (including the F14 production incident) is already recorded in full on `ValidateEndpointCapability`. **If T10 is still open, say so and I'll restore a corrected note instead.**

**`ipcListen` on windows and js.** #254 removed `ipcListen` from `ipc_unix.go` along with the server half, but `ipc_windows.go` and `ipc_js.go` still defined their own. Unused funcs aren't a compile error so nothing broke — and nothing would have flagged it either, since CI only builds linux and darwin. I cross-compiled `GOOS=windows` to confirm both before and after.

**Comments that outlived their code:**
- `utils/sigs/sigs.go` had a dangling `//nolint:staticcheck` sitting inside the import block after its `ripemd160` import was removed — it now suppresses nothing and reads like an import went missing. Its package header also still described a `Signable` interface and `Sign()` / `ExtractSignerAddress()` / `RecoverPubKey()`; the package is down to `EncodeUint64` and `HashMsg`.
- The `rpcclient` package doc still documented `NewServer`, `ServeCodec`, `ServeListener` and `ClientFromContext`, all removed. Rewritten to state plainly that this is now the client half only, and why the method-registration machinery is still there (the client uses it to dispatch notifications).

**Not addressed, worth tickets rather than code:** two of the removals look like half-wired features rather than pure cruft, and deleting them erased the evidence. `handleCodeErrors` was the only gRPC `codes.DeadlineExceeded` → `common.ContextDeadlineExceededError` mapper, and `chaintracker` still branches on that error at two call sites. And `--test-mode` is a live flag (`common.IsTestMode`, checked in `rpcsmartrouter.go` and `jsonRPC.go`) whose dedicated router was never hooked up. Both were genuinely unreachable, so nothing regressed — flagging them so the gap is tracked deliberately.

Jira ticket: MAG-2690

## Verification

`go build ./...` for linux **and windows**, `go vet ./...`, and `go test ./...` all pass (33 packages, 0 failures).

`deadcode -test ./...` reports **zero findings**. Getting it to run took sorting out a toolchain trap worth recording: `go run golang.org/x/tools/cmd/deadcode@latest` picks its toolchain from *x/tools'* own `go` directive (`1.25.0`) and silently **downgrades to go1.25.12**, which then cannot type-check this module's `go 1.26.5` packages — it fails with a wall of `package requires newer Go version go1.26 (application built with go1.25)` that looks like a broken checkout rather than a toolchain mismatch. The fix is to pin the toolchain explicitly:

```sh
GOTOOLCHAIN=go1.26.5 go run golang.org/x/tools/cmd/deadcode@latest -test ./...
```

`GOTOOLCHAIN=local` does **not** work as a substitute unless your PATH `go` is itself >= the `go.mod` directive — mine reports 1.26.5 only because `auto` upgrades it, and `local` fell back to 1.23.3. To confirm the zero-findings result was real analysis and not a silent no-op, I dropped an unreachable `ProbeCanary` function into `protocol/common`, re-ran, and checked it was reported before removing it.

## The `:3000` test hazard, fixed here

The earlier draft of this PR flagged this as out of scope; it is now fixed in the second commit, because it is the same "silently wrong rather than loudly failing" defect the rest of the PR is about.

Four tests bound hardcoded ports (`:3000`, `:3010`, `:3011` in `chainlib`, `:3000` in `metrics`) and swallowed the bind error, so on a host already using that port they did not fail — they raced, then dialled whatever else was listening. `TestExtractDappIDFromWebsocketConnection` is the worst case: it hangs to the 10-minute package timeout, presenting as a stuck CI job rather than a test failure. They all now bind `127.0.0.1:0` and dial the address the kernel handed back, which also removes the `time.Sleep`-and-hope delays since the socket is bound before the dial.

`lavasession`'s `createGRPCServer` now returns the address it actually bound, so `TestEndpointSortingFlow` can use port 0 too — retiring `getDelayedAddress` and its "if it collides use 3336 instead" workaround. Its `require.NoError` also moved up to the call site: it sat after two backoff-free connect loops, so a bind failure busy-waited to the timeout instead of reporting itself.

Verified the way it should be: on a host where `:3000` was already occupied, the suite previously hung in `chainlib` and failed in `metrics`; it now passes clean.

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change — n/a, every removed identifier has zero references
* [x] provided a link to the relevant issue or specification — MAG-2690
* [x] included the necessary unit and integration tests — n/a, removals and comment fixes only; suite passes unchanged
* [x] updated the relevant documentation or specification — the doc-comment fixes above are the documentation change

## Reviewers Checklist

I have...


avitenzer:
Sanity-checked the `consistency.go` deletion as requested — **deleting the file was the right call, and no corrected note is needed: T10 is closed.**

Traced the history:

- T10 was implemented by name in 901ac07 (*fix(shared-state): rebuild cross-pod consistency on the guarded chain tip (T10)*, July 22). That commit switched the shared-state **read** key from `relaycore.UserDataKey(userData)` (`dappId__consumerIp`) to `rpcss.listenEndpoint.Key()`, matching the write key — its message confirms the read had missed every time before that. The tests in `ecosystem/cache/shared_state_tip_test.go` were added there and lock the contract, including a case asserting the old per-user key now misses. What that commit didn't do was update the note in `consistency.go` — that's when it went stale.
- #254 then removed the caller-less `UserDataKey`, which made the note's closing claim ("only the key derivation survives … still identifies a caller by dapp+IP") false on both counts.
- No history is lost: the retirement rationale including the F14 incident lives on `ValidateEndpointCapability`'s doc block (`protocol/relaycore/consistency_validation.go`), cross-referenced from `rpcsmartrouter_server.go`, with the F14 regression gate in `consistency_liar_repro_test.go`.

If anything, the PR description undersells it — T10 wasn't just "premise satisfied", it was deliberately shipped as T10 in a dedicated commit with tests.

avitenzer (approved):

PR #258, opened 2026-08-09
feat(smart-router/debug): add POST /debug/poll-now (MAG-2649)

AnnaR-prog:
# Description

Adds `POST /debug/poll-now`: force ONE endpoint's ChainTracker to run its dedicated poll right now, and answer only once that poll's result has been recorded.

The automation suite can only observe what a ChainTracker records after a poll by waiting out real clock time — ~6.5 s for the per-endpoint cadence on Ethereum, ~30 s for a warm-up that waits on a forced poll, and the poll-settle half of the ~130 s staleness check. Nothing could make a poll happen sooner. With this, a test becomes: set up state → trigger → read.

## Jira ticket

Jira ticket: MAG-2649

---

## How it works

Fidelity is the whole point, so the trigger is thin. `ChainTracker.PollNow` hands a request to the tracker's **own poll goroutine** over an unbuffered channel; that goroutine runs `fetchAllPreviousBlocksIfNecessary` — the same function the cadence timer calls, with the same fetch timeout — and replies when it returns. Nothing is reimplemented: the parse, the observation write (`LatestBlock`/`Source`, `ConsecutivePollFailures`, `LastSuccessfulPoll`), the endpoint tip and the per-chain tip are the production ones, and all complete before the reply — which is what makes the caller's subsequent read race-free.

**Running it on the poll goroutine is not incidental.** The cycle mutates `relaySkipsSinceRealPoll` and `blockCheckpoint` unlocked, on a single-writer assumption, so calling it from the HTTP goroutine would be a real data race. For the same reason `EndpointMonitor.PollNow` copies the tracker pointer and releases `m.mu` **before** waiting: the cycle takes `m.mu` itself when it observes a new block (`newLatestCallback` → `setTrackerState`), so holding it across the wait deadlocks.

## Two deliberate behaviours

- **Forces past the relay traffic gate** — via a `force` parameter on the production function rather than a second copy of it. Otherwise a poll-now issued right after a set-up relay would silently no-op inside the freshness window (`relayGateFreshness` = `averageBlockTime`, 12 s on ETH) and hand the caller a stale record. It still resets the gate's skip budget, because a real poll did happen.
- **Leaves cadence alone** — `fetchFails` is untouched and the timer is never rescheduled, so a forced poll cannot move the schedule that the cadence, idle-ceiling, gate and backoff tests measure in real time. The mirror of `/debug/reset-probe-backoff` ("cadence only, data untouched").

Note the **two distinct failure counters**: the tracker-local `fetchFails` (cadence backoff — untouched here) and the observation's `ConsecutivePollFailures`, which is what the ticket means by "the running count of failed polls" and **is** updated, by the shared record path.

## `Polled` is a claim about the record, not about the goroutine

The endpoint's entire value is that a harness can trust what it reads back, so the one field everything else hangs off is `Polled`: **this row's observation is this poll's**. It is false whenever the response cannot vouch for the record — and the record handed back is then the prior one, explicitly stale.

That phrasing matters because "a poll ran" and "I can tell you what it recorded" are not the same claim. A poll whose result outlives the caller's budget *does* run — it is bounded by the tracker's context, not the requester's, so it finishes and writes regardless — but this response read the record before it landed. Reporting that as `Polled: true` would hand back a pre-poll block **and a pre-poll failure streak** while labelling both fresh, which is exactly the confusion the endpoint exists to remove.

Three sentinels make the causes distinguishable to `errors.Is`, and all three map to `Polled: false`:

| Sentinel | Meaning |
| --- | --- |
| `ErrorPollNowNotDelivered` | the poll goroutine never took the request — tracker still starting, retrying its init, or stopped. Nothing ran and nothing was written. |
| `ErrorPollNowUnsupported` | this tracker cannot poll at all (`DummyChainTracker`). |
| `ErrorPollNowResultNotAwaited` | the request **was** taken and the cycle is running, but it outlived the caller's budget. The poll completes and records on its own; this response simply did not witness it. |

## Deadlines: delivery and result are bounded separately

A tracker that is not yet `polling` has no goroutine to receive the trigger, so waiting out the full 25 s handler budget only delays a verdict that will not change. `EndpointMonitor.PollNow` therefore caps that wait at 2 s (`pollNowUnstartedGrace`) — a shorter deadline, not a rejection, so a state that merely lags the goroutine can never refuse a poll that would have worked.

**That grace bounds delivery only.** `ChainTracker.PollNowWithDeliveryDeadline` takes the two contexts separately, because the two waits have unrelated natural lengths: delivery either happens in microseconds (the goroutine is at its select) or never (there is no goroutine), while the cycle behind it is bounded by the tracker's own fetch timeout, `max(10s, MinimumTimePerRelayDelay)` — several times the grace. Capping both with the grace would abandon healthy polls mid-flight and leave the caller holding a pre-poll record.

## API

`POST /debug/poll-now` with `{"network_address": "<url>"}` (required; optional `chain_id` / `api_interface` narrow the match when the same URL is registered on more than one chain or interface). Response is a row per matched endpoint, sharing `/debug/endpoint-state`'s field vocabulary plus `Polled`, `PollError`, `TriggerError`, `DurationMs`.

| Status | Meaning |
| --- | --- |
| 200 | at least one matched endpoint returned a record this response can vouch for — including when the poll reached upstream and **failed** (`PollError` set), which is exactly what a test asserting on the failure streak needs |
| 400 | unusable request (bad JSON, missing `network_address`) |
| 404 | no ChainTracker serves that address |
| 504 | no matched endpoint produced a trustworthy record. **`TriggerError`, not the status code, says why** — either no poll could be started (the tracker's lifecycle state is named, since `retrying_start` is a diagnosis and a bare timeout is not), or a poll was started and outlived the budget. The second is not an endpoint fault. |

`PollError` is only ever set alongside `Polled: true`, so the pair can never describe a failure streak that was not actually recorded.

The 25 s budget is **shared across all matched targets** rather than granted per target, so a multi-target address cannot multiply the worst-case hang by the number of matches. The cost is that a late target can inherit a nearly-spent budget and answer `Polled: false` — honest, and the reason the not-awaited case is reported that way instead of claiming a poll it did not witness. Narrow with `chain_id` / `api_interface` when a test needs every matched endpoint to get a full budget.

Debug-only needs no new flag: the whole `/debug` mux exists only under `--debug-address`. SVM is covered by construction — the poll goes through `iChainFetcherWrapper` and the `PollObserver` hook.

- `protocol/chaintracker/chain_tracker.go` — the new `pollNowCh` case in the poll loop, `PollNow` / `PollNowWithDeliveryDeadline`, and the `force` parameter.
- `protocol/endpointstate/endpoint_monitor.go` — `PollNow`, the delivery-only grace, and the lock release that avoids the deadlock described above.
- `protocol/rpcsmartrouter/rpcsmartrouter.go` — the handler and `resolvePollNowTargets` (resolves under `router.mu`, releases it before polling).

## Testing

`go test -race` across `chaintracker`, `endpointstate` and `rpcsmartrouter`, including an end-to-end handler test over a real HTTP upstream.

Four tests are load-bearing beyond their assertions:
- the endpointstate test drives an **advancing** block, which is what makes the cycle fire `newLatestCallback` → `m.mu.Lock()`. A constant-block mock would pass against a version that deadlocks.
- a **relay-then-poll-now** test — the pytest suite's actual sequence — proves the gate bypass end to end through the channel, not just by calling the cycle directly.
- `TestEndpointMonitor_PollNow_LaggingTrackerStateDoesNotTruncateThePoll` forces the lagging-state window deterministically (the real race is microseconds wide) and drives a poll longer than the grace, asserting the **fresh** block comes back.
- `TestEndpointMonitor_PollNow_ResultNotAwaited_ReportsNotPolled` asserts `Polled: false` and an unchanged pre-poll failure streak, then waits for the abandoned poll to land — proving `Polled: false` is honest rather than an under-report.

Both of the last two were verified against the previous behaviour: with the fix reverted they fail, with it in place they pass.

**Pre-existing, unrelated:** three `chaintracker` tests (`TestChainTrackerCallbacks`, `TestChainTrackerFetchSpreadAcrossPollingTime`, `TestChainTrackerPollingTimeUpdate`) fail under `-race` on a `smallestBlockGap` / `blockEventsGap` race in the legacy adaptive cadence. Reproduced on a clean `origin/main` and again with this branch's later commits stashed. Untouched here; worth its own ticket.

## Notes for whoever writes the pytest side

All documented in the handler's doc comment:
- **Assert `Polled` before reading any other field in a row.** A false `Polled` with a "not awaited" `TriggerError` is not a broken tracker — the poll is running and will record, the response just could not wait for it. Retry, or narrow with `chain_id` / `api_interface` so fewer endpoints share the budget.
- The per-chain consensus baseline is recomputed on its own tick — assert on `ObservedTip`, not `/debug/chain-state`'s `ConsensusBaseline`, right after a poll-now.
- A head moved **backward** will not be visible: the endpoint tip is block-monotonic and holds a lower block as a late straggler until the stored tip goes stale (`StalenessWindow`, ~120 s on ETH), and the tracker does not walk its own head down. That is the production anti-flap rule faithfully applied, not a failed trigger.
- Do **not** use this endpoint for tests that measure poll cadence, the idle-poll ceiling, the gate's skip behaviour, or backoff growth — per the ticket's Caution, those must measure real elapsed time.

---

## Author Checklist

* [x] confirmed `!` in the type prefix if API or client breaking change — n/a, additive debug-only endpoint
* [x] updated the relevant documentation or specification, including comments for documenting Go code — Go doc comments on every new symbol; CHANGELOG is generated at release time from commit subjects, so not hand-edited here

## Reviewers Checklist

### DIFF
--- a/protocol/endpointstate/endpoint_monitor.go
+++ b/protocol/endpointstate/endpoint_monitor.go
@@ -30,11 +30,16 @@ const (
 	defaultTrackerStartRetryMax = 30 * time.Second
 	trackerStartRetryJitterDiv  = 5
 
-	// pollNowUnstartedGrace bounds how long PollNow waits on a tracker that is not yet Polling
-	// (MAG-2649). Such a tracker has no poll goroutine to receive the trigger, so waiting out the
-	// caller's full budget only delays a verdict that will not change. Comfortably longer than the
-	// microseconds a live goroutine needs to take the send, so it never cuts short a tracker whose
-	// state merely lagged.
+	// pollNowUnstartedGrace bounds how long PollNow waits for a tracker that is not yet Polling to
+	// TAKE the trigger (MAG-2649). Such a tracker has no poll goroutine to receive it, so waiting out
+	// the caller's full budget only delays a verdict that will not change. Comfortably longer than
+	// the microseconds a live goroutine needs to take the send, so it never cuts short a tracker
+	// whose state merely lagged.
+	//
+	// Delivery only. It deliberately does NOT bound the poll cycle that follows: that is bounded by
+	// the tracker's own fetch timeout, max(10s, MinimumTimePerRelayDelay), which is several times
+	// this grace. Capping both with this value would abandon healthy polls mid-flight and leave the
+	// caller reading a pre-poll record.
 	pollNowUnstartedGrace = 2 * time.Second
 )
 
@@ -591,14 +596,21 @@ func (m *EndpointMonitor) BackoffSnapshot() map[string]time.Duration {
 // /debug/poll-now). It exists so a test can set a provider's state, trigger the poll, and read the
 // result — instead of waiting out the per-endpoint cadence (avgBlockTime/2).
 //
-// polled reports whether the poll CYCLE ACTUALLY RAN. It separates the two outcomes a caller must
-// never conflate:
+// polled reports whether THE RETURNED OBSERVATION IS THIS POLL'S. It is the caller's licence to
+// trust the record, and it separates outcomes that must never be conflated:
 //   - polled=true with a non-nil err — a real poll reached (or tried to reach) upstream and
 //     failed. The observation records that failure, exactly as a timer-driven failure would:
 //     ConsecutivePollFailures incremented, LastSuccessfulPoll untouched.
-//   - polled=false — no poll happened at all (no tracker for this URL, the tracker is still
-//     starting/retrying its init, or it is a non-polling dummy). The returned observation is
-//     whatever was already on record and says nothing about now.
+//   - polled=false — the observation predates this call and says nothing about now. Either no poll
+//     happened at all (no tracker for this URL, the tracker is still starting/retrying its init, or
+//     it is a non-polling dummy), or one was started and outlived the caller's budget. err
+//     distinguishes them; both mean the same thing about the record.
+//
+// That second case is why polled is phrased around the RECORD rather than around the cycle. A poll
+// whose result was not awaited did run — it finishes and writes on its own — but this call cannot
+// say what it wrote, and the record available now is the one from before it. Reporting that as
+// polled=true would hand a harness a pre-poll block and a pre-poll failure streak while telling it
+// both were fresh, which is precisely the confusion this endpoint exists to remove.
 //
 // The tracker's poll goroutine takes m.mu during a block-advancing cycle (newLatestCallback →
 // setTrackerState), so the lock is released BEFORE the trigger — holding it across the wait would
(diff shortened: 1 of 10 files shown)


protocol/endpointstate/endpoint_monitor.go
```diff
@@ -578,6 +586,70 @@ func (m *EndpointMonitor) BackoffSnapshot() map[string]time.Duration {
...
+	// send is taken instantly anyway, so nothing that could have succeeded is refused. A Polling
+	// tracker keeps the full budget — it may legitimately be mid-cycle on a slow upstream.
+	attemptCtx := ctx
+	if stateBefore != EndpointChainTrackerPolling {
+		var cancelAttempt context.CancelFunc
+		attemptCtx, cancelAttempt = context.WithTimeout(ctx, pollNowUnstartedGrace)
+		defer cancelAttempt()
+	}
+
+	pollErr := tracker.PollNow(attemptCtx)
```
avitenzer: `pollNowUnstartedGrace` bounds the **poll**, not just the delivery — so a truncated poll can be reported as `Polled: true` with a stale record.

The doc comment above states the intent clearly: *"a shorter DEADLINE and not a rejection … in that window the send is taken instantly anyway, so nothing that could have succeeded is refused."* That holds for the send, but `attemptCtx` is handed to `ChainTracker.PollNow`, which uses the same context for **both** phases (`chain_tracker.go:131-142`):

```go
select { case cs.pollNowCh <- req:            // phase 1: delivery
         case <-ctx.Done():  return ErrorPollNowNotDelivered… }
select { case err := <-req.resp: return err   // phase 2: the poll itself
         case <-ctx.Done():  return fmt.Errorf("poll started but its result was not awaited: %w", ctx.Err()) }
```

So the 2 s also has to cover the whole upstream fetch. Concretely:

1. `stateBefore` reads as `retrying_start` because `trackerStates` lags the goroutine by a hair → `attemptCtx` = 2 s.
2. The send is taken in microseconds; the cycle then runs `FetchLatestBlockNum` + `forkChanged` + `fetchAllPreviousBlocks` against a real upstream — comfortably >2 s on a slow node.
3. `PollNow` returns `"poll started but its result was not awaited"`, which is **neither sentinel**, so line 650 returns `polled=true`.
4. The handler emits `Polled: true`, `PollError: "poll started but its result was not awaited…"`, and a `LatestBlock` that `GetObservation` read *before* the poll finished — the stale pre-poll value.

That contradicts the handler's own contract (`rpcsmartrouter.go:1137`): *"False means nothing was polled … and the record below is stale."* A harness trusting `Polled == true` reads a stale block as fresh — the exact failure mode the rest of the PR is built to prevent. And it's reachable in the use case the description calls out ("a warm-up that waits on a forced poll"), since that is precisely when tracker state is most likely to be lagging.

The tests can't catch it as written: `newPollNowMonitor` waits via `require.Eventually` until the state is `Polling`, so every success test takes the full-budget path, and `TestEndpointMonitor_PollNow_TrackerNotPolling_NamesTheState` uses a tracker that never starts, so phase 2 is never reached under the grace window.

### Suggested fix

Split the two deadlines rather than shortening both — in `chaintracker`:

```go
func (cs *ChainTracker) PollNow(ctx context.Context) error {
	return cs.pollNow(ctx, ctx)
}

// PollNowWithDeliveryDeadline bounds only the wait for the poll goroutine to TAKE the
// request; once taken, ctx governs the wait for the result.
func (cs *ChainTracker) PollNowWithDeliveryDeadline(ctx, deliver context.Context) error {
	return cs.pollNow(deliver, ctx)
}
```

and have `EndpointMonitor.PollNow` pass the 2 s context as `deliver` only.

Independently — and worth doing either way — add a third sentinel (`ErrorPollNowResultNotAwaited`) for the phase-2 timeout and map it to `polled=false`. "The cycle started but I can't tell you what it recorded" is a different claim from "a poll ran and I'm reporting it," and only a sentinel makes that distinguishable to `errors.Is`. The PR already draws this distinction at the error level with two sentinels; the gap is that the *deadlines* aren't split the same way.

A regression test would be: register a tracker, force `trackerStates` to a non-`Polling` value, point it at an upstream that sleeps ~3 s, and assert the result is **not** `polled=true` with the pre-poll block.

AnnaR-prog: Good catch — fixed in `14b842e`.

Confirmed the mechanism as you describe it: one context bounded both phases, the phase-2 timeout returned a bare `fmt.Errorf` matching neither sentinel, so the guard fell through to `polled=true` over an observation read before the poll wrote. Your read of the test gap was right too — `newPollNowMonitor` waits for `Polling`, so every success test took the full-budget path.

Two things came out differently from the suggestion, both pushing toward the sentinel:

**Reachability is broader than the grace window.** The handler builds one 25 s context *outside* the target loop (`rpcsmartrouter.go:1456`) and passes it to every `PollNow`. So with `stateBefore == Polling` and no race at all, a late target inherits a nearly-spent budget and hits the same false `Polled: true`. Also worth noting `MinimumTimePerRelayDelay` is a tunable `var` capped only against `DefaultTimeout` — above ~12.5 s even the single-target path can exceed 25 s.

**That inverts the fix priority.** Splitting the deadlines doesn't help the multi-target path — a split creates no new budget on a shared context. So `ErrorPollNowResultNotAwaited` → `polled=false` is the load-bearing fix, and the split is the follow-on that keeps the grace from truncating healthy polls. Both are in; `PollNowWithDeliveryDeadline` bounds delivery only.

One sharper consequence worth recording: on the phase-2 timeout the row carried `Polled: true` **and** `PollError`, which per the handler's own vocabulary means "reached upstream and failed, `ConsecutivePollFailures` incremented." It hadn't — so it forged a failure streak, breaking the exact case the PR headlines. `PollError` is now only ever set alongside `Polled: true`.

Your regression test is in, plus one for the inverse. Both verified differentially: reverted to the single-context behaviour, both fail (`Received unexpected error` / `Should be false`); with the fix, both pass.

Follow-up in `51bf1a8`: making the not-awaited case `polled=false` routes it through the same 504 as "nothing could start", so the 504 docs had become wrong in the other direction. `TriggerError` now carries the distinction. PR description updated to match.

avitenzer (approved):

PR #259, opened 2026-08-09
feat(smart-router/debug): add GET /debug/provider-scores (MAG-2707)

AnnaR-prog:
# Description

Adds `GET /debug/provider-scores`: read the per-provider quality scores the router is holding right now — the values it uses to decide where the next request goes. Read-only.

The router could already **reset** these scores (`/debug/reset-scores`) but never **read** them. Their only exposure was Prometheus gauges on the metrics port, which is not published in the cluster, so reading them meant a hand-held tunnel: impossible from an automated nightly, and worse — when the tunnel was down the read came back empty and the test passed having measured nothing. Every question about how the router scores providers was untestable from outside.

## Jira ticket

Jira ticket: MAG-2707

---

## The error design is the point

That silent-empty failure mode is what this endpoint exists to end, so it deliberately breaks the convention of the other `/debug` state endpoints (which answer `200 []` when unwired):

| Status | Meaning |
| --- | --- |
| 200 | scores follow. A provider that exists but has not been sampled yet **is** included, with zero scores — a real answer, not an omission |
| 404 | `chain_id` was given and no optimizer is registered for it |
| 503 | the scores could not be obtained: no QoS sampler wired, no optimizer registered, or every matched chain has no providers known yet. The body **names the chains**, so a failing test says *why* rather than just "empty" |

Never 200 with an empty array.

## Design decisions

**Computed on demand, not served from the sampler's cache.** Not for freshness (the sampler runs at 1 s), but because that cache is empty until the first tick and after a score reset — so "not sampled yet" would be indistinguishable from "no providers", the exact ambiguity this endpoint removes. On-demand also lets one test reset scores, send traffic, and read the result back in sequence.

**Reading publishes nothing.** `appendOptimizerQoSReport` is split into a pure `buildOptimizerQoSReport` plus the usage-sink emit; `SnapshotReports` calls the builder only, and the periodic sampler keeps emitting exactly as before. Without that split a debug *read* would fire OTel usage events and double-count every dashboard fed by the sink.

**Everything routes through the shared builder,** which keeps `sanitizeFloat` on every float. `encoding/json` **errors** on NaN, and `writeDebugRows` only logs a warning after the status line is written — so a raw-float path would answer 200 with a truncated body: the same silent pass in a new costume.

**`SnapshotReports` refuses to drop a registered chain whose provider set is still empty.** The periodic sampler skips that case silently; a reader must not, or an unpopulated router is indistinguishable from a healthy idle one.

**Row shape.** Both score families, because they answer different questions: the raw EWMA values (`AvailabilityScore`, `LatencyScore` and `SyncScore` in seconds) are what a "was this provider marked down?" test reads, while the `Selection*` values are those same signals normalised, with `SelectionComposite` the number selection ranks on. Plus stake, node-error rate and selection counters for context.

`debugRowKey` gained `ProviderAddress` as a tiebreaker — `sort.Slice` is **not** stable, so same-chain rows would otherwise come back in arbitrary order. Absent on every other row type, where comma-ok yields `""` and the key is unchanged.

Read-only throughout, and behind `--debug-address` like the rest of `/debug`, so no new flag.

## One addition beyond the ticket — easy to strike

`NetworkAddresses` on each row is **not** in the ticket's four acceptance points. I added it because scores are keyed by **provider address** while endpoint health is keyed by **network address**, and neither identity can be derived from the other — so without it the automation cannot ask "this provider's score moved, was its endpoint healthy?" without a third lookup. It is also the only part of the diff that pulls `router.mu` into a read path that otherwise touches just the QoS client.

If a reviewer would rather keep the endpoint minimal, `providerEndpointURLs` + that one field can be removed without unpicking anything else.

## Related finding — likely a separate bug, not fixed here

The optimizer is being keyed by **two different identities**. Every path feeds it the provider address (`PublicLavaAddress`): the probe loop (`rpcsmartrouter_server.go:2574`), the relay path (`AppendRelayDataConsensus`), and stake (`CalcWeightsByStake` → `weights[cswp.PublicLavaAddress]`). But the **WS subscription path feeds it the endpoint URL** — `direct_ws_subscription_manager.go:579, 591, 601` pass `selectedEndpoint.Url`.

Reports enumerate only the stake map's keys (provider addresses), so WS-sourced QoS entries are written under keys nothing ever reads: invisible in scores, in metrics, and — if selection also enumerates addresses — in selection itself. That would mean WS relay outcomes are silently discarded as QoS feedback in production.

Out of scope here and worth its own ticket. Worth noting that this endpoint is exactly the tool that would have surfaced it.

- `protocol/metrics/consumer_optimizer_qos_client.go` — the build/emit split and `SnapshotReports`.
- `protocol/rpcsmartrouter/rpcsmartrouter.go` — the handler (status-code logic above all), `providerEndpointURLs`, and the `debugRowKey` tiebreaker.

## Testing

`go test -race -count=1` on both packages.

The strongest test is `TestSnapshotReports_DoesNotEmitToUsageSink`: it asserts the read path publishes nothing **and** carries a positive control proving the periodic path still does — so it cannot pass against a sink that never fires. The join test wires two endpoints on one provider, declared in reverse order, so a pass proves the URLs are collected *and* sorted rather than echoed back in declaration order.

Lock order checked by hand: every `router.mu` acquisition is a debug handler, startup, or the epoch tick, and `updateEpoch` already goes `router.mu` → `csm.lock` — the same direction as the new code, so no inversion is introduced.

**Pre-existing, unrelated:** `metrics/TestHappyFlow` (a race in `relays_monitor`) and the `chaintracker` adaptive-cadence tests fail under `-race` on a clean `origin/main` — verified by stashing this branch's changes and re-running at the base commit. Untouched here.

---

## Author Checklist

* [x] confirmed `!` in the type prefix if API or client breaking change — n/a, additive debug-only endpoint
* [x] updated the relevant documentation or specification, including comments for documenting Go code — Go doc comments on every new symbol; CHANGELOG is generated at release time from commit subjects, so not hand-edited here

## Reviewers Checklist

### DIFF
--- a/protocol/metrics/consumer_optimizer_qos_client.go
+++ b/protocol/metrics/consumer_optimizer_qos_client.go
@@ -5,6 +5,7 @@ import (
 	"math"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -220,7 +221,16 @@ func (coqc *ConsumerOptimizerQoSClient) getLastRNGValue(chainId, providerAddress
 	return 0
 }
 
-func (coqc *ConsumerOptimizerQoSClient) appendOptimizerQoSReport(report *OptimizerQoSReport, chainId string, epoch uint64) OptimizerQoSReportToSend {
+// buildOptimizerQoSReport assembles the wire form of one provider's report. It is PURE: it reads
+// the counter maps (caller must hold at least the read lock) and writes nothing, emits nothing.
+//
+// Split out from appendOptimizerQoSReport (MAG-2707) so the read-only snapshot path can reuse the
+// exact same assembly — including sanitizeFloat on every float — without the usage-sink emit that
+// belongs to the periodic sampler alone. Two properties depend on going through here rather than
+// marshalling a raw OptimizerQoSReport: NaN/Inf are neutralised (encoding/json ERRORS on NaN, and a
+// debug handler that has already written its status line cannot take that back), and the enrichment
+// (stake, selection counters, node-error rate) is applied identically to both paths.
+func (coqc *ConsumerOptimizerQoSClient) buildOptimizerQoSReport(report *OptimizerQoSReport, chainId string, epoch uint64) OptimizerQoSReportToSend {
 	// must be called under read lock
 	// Use sanitizeFloat to prevent NaN/Inf values in JSON output
 	optimizerQoSReportToSend := OptimizerQoSReportToSend{
@@ -255,9 +265,19 @@ func (coqc *ConsumerOptimizerQoSClient) appendOptimizerQoSReport(report *Optimiz
 		StakeContribution:        sanitizeFloat(report.StakeContribution),
 	}
 
-	// Fire the report at the usage sink (OTel). Non-blocking; NoopUsageSink
-	// when OTel emission is disabled. The in-memory reportsToSend cache
-	// (built by the caller) still feeds the Prometheus /metrics endpoint.
+	return optimizerQoSReportToSend
+}
+
+// appendOptimizerQoSReport builds a report AND fires it at the usage sink (OTel). Non-blocking;
+// NoopUsageSink when OTel emission is disabled. The in-memory reportsToSend cache (built by the
+// caller) still feeds the Prometheus /metrics endpoint.
+//
+// The emit is why this wrapper exists separately from buildOptimizerQoSReport: it is the periodic
+// sampler's step, and a read-only caller (SnapshotReports) must not perform it — a debug read that
+// published usage events would double-count every dashboard fed by the sink.
+func (coqc *ConsumerOptimizerQoSClient) appendOptimizerQoSReport(report *OptimizerQoSReport, chainId string, epoch uint64) OptimizerQoSReportToSend {
+	// must be called under read lock
+	optimizerQoSReportToSend := coqc.buildOptimizerQoSReport(report, chainId, epoch)
 	coqc.usageSink.EmitOptimizerQoS(optimizerQoSReportToSend)
 	return optimizerQoSReportToSend
 }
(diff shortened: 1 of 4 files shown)


VicSheCodes:
Reviewed against MAG-2707's acceptance points and against how the automation will consume this. All five are met: it reads the live scores, needs no manual setup, publishes nothing when read, sits behind `--debug-address`, and fails loudly instead of returning an empty success. Good to merge from our side.

The build/emit split with `TestSnapshotReports_DoesNotEmitToUsageSink` and its positive control is the right shape, and computing on demand rather than serving the sampler cache is what makes reset-then-read work for us. That sequence is the thing we could not do before.

### One ask: please keep `NetworkAddresses`

You offered to strike `NetworkAddresses` and `providerEndpointURLs` as beyond the ticket's four acceptance points. We need both.

Scores are keyed by provider address. Endpoint health in `/debug/endpoint-state` is keyed by network address. Neither identity is derivable from the other on the test side, so without this field every assertion of the form "this provider's score moved, was its endpoint healthy at the time?" needs a third lookup and a provider-to-URL mapping we would have to build and keep in sync ourselves. With the field it is two reads and a join on data the router already owns.

`TestDebugProviderScores_CarriesEndpointURLsForJoin`, with the two endpoints declared in reverse order, is the right test for it.

### Two correctness notes, neither blocking

Both need a router process serving more than one chain. Every router we run serves exactly one, so neither reaches us. Raising them because they are real, not because they hold up the merge.

1. A partly-empty router answers 200 and drops the empty chains without saying so. The guard at line 1416 fires only when `Reports` is empty across the whole router, so if one chain has providers and another does not, the caller gets 200, the rows of the populated chain, and no indication that the other produced nothing. `ChainsUnavailable` is computed correctly but is only read inside the 503 message, so on the 200 path it is discarded. What we would want in that situation is for the answer to state which chains produced nothing, in the body rather than a header, since a header is easy to ignore and would recreate the silent pass this endpoint exists to end. Shape is your call.

2. `TestDebugProviderScores_UnavailableIsNot200` covers both cases where the router produced nothing at all. The mixed case is the one that would lock the above: two chains registered, one with providers and one without, asserting the empty chain is still visible in the answer.

AnnaR-prog:
Thanks — both notes were right, and note 1 reproduced exactly as you described. Fixed in `354e5df`.

**Heads-up first, since it affects anything you have already drafted:** the 200 body changed shape. It was a bare array; it is now an envelope.

```json
{"rows": [ ... ], "chains_unavailable": ["SOLANA"]}
```

If you have a parser against the old shape it needs `resp["rows"]`. Nothing else consumes this endpoint yet, so now is the cheap moment to change it.

### Note 1 — partly-populated router

You read the code correctly: the guard at what was line 1416 fired only when `Reports` was empty across the whole router, and `ChainsUnavailable` was computed correctly but then read only inside the 503 message — so the 200 path discarded it. One chain populated and another not gave you a clean-looking 200 with no mention of the empty chain.

`chains_unavailable` now names every matched chain that produced no rows, and is `[]` (never `null`) when the answer is complete. The 503 is unchanged and still fires when *nothing* could be obtained, including when you narrow to an empty chain with `?chain_id=`.

Two shapes I considered and rejected, in case they come up in review:

- **Sentinel rows** for empty chains keep the array shape, but hand you a row with an empty `ProviderAddress` that a naive loop reads as a provider — a footgun pointing at the exact bug being fixed.
- **Failing the whole request** when any chain is empty is maximally loud, but makes a healthy chain unreadable while another is still warming up. Too strict for a diagnostic.

The envelope deviates from the flat-array style of the other `/debug` state endpoints. That is deliberate, and it is the same deviation the status codes already make on this endpoint.

### Note 2 — the test that locks it

`TestDebugProviderScores_PartlyPopulatedRouterNamesTheEmptyChain`: two chains registered, only ETH1 populated. It asserts the populated chain is still readable at 200, that `chains_unavailable` is `["SOLANA"]`, and that `?chain_id=SOLANA` is still an outright 503.

### `NetworkAddresses`

Staying, per your ask — the offer to strike it is withdrawn. Your reasoning matches why I added it: neither identity derives from the other on your side, so removing it would push a provider-to-URL mapping onto the test suite to build and keep in sync.

### One thing neither note raised

`SnapshotReports` builds both chain lists by ranging a map, so `chains_unavailable` would have varied in order between identical reads. Now sorted — it matters more than it did when the list only ever appeared in an error string. `writeDebugRows`' ordering is factored into `sortDebugRows` so the enveloped response keeps the same deterministic row order.

Full `-race` run green in both packages. The two pre-existing races called out in the description (`metrics/TestHappyFlow`, the `chaintracker` adaptive-cadence tests) are still there and still reproduce on a clean `origin/main`.

avitenzer (approved):

PR #260, opened 2026-08-09
chore(lavasession)!: remove the legacy synthetic prober

Tomelia1999:
Jira ticket: MAG-2733

# ⚠️ Testing status — please read before approving

**Only unit tests have been run. Nothing else.** This change deletes code across the session manager, the provider optimizer, the metrics interface and the CLI surface, and it has not been exercised against a running router.

What *was* verified:

- `go build ./...` — clean
- `go vet ./protocol/...` — clean
- `go test ./protocol/... ./types/... ./utils/...` — all green
- `go test -race -count=2` on the migrated concurrency test — green
- `gofumpt` — no new findings on touched files

What was **not** done, and should be before this merges:

- **No integration or e2e run.** None of `scripts/pre_setups/*` (`test_uc1_smartrouter_lava.sh`, `test_uc2_smartrouter_lava.sh`, `test_uc4_smartrouter_sim.sh`, `test_smartrouter_sol.sh`, `test_chaintracker_chainstate_sim.sh`) or `scripts/test_mag2062_e2e.sh` was executed.
- **No manual run** against real upstream endpoints — no router was started with this build.
- **No deployment / sim validation.** The dev-sim PR gate has not been run.
- **Note: CI does not run `go test` at all.** No workflow in `.github/workflows/` invokes it — the green checks here are build, lint, govulncheck and security scanning only. The unit-test result above came from a local run and is not reproduced by CI.

Areas most worth a real-traffic sanity check, since deletions touched them:

1. **Startup** — two CLI flags were removed. Any deployment passing `--enable-periodic-probe-providers` or `--periodic-probe-providers-interval` will now **fail to start** with an unknown-flag error. This repo is clean (grepped across Go, shell, yaml, md and json, including `config/smartrouter_examples/` and `docker/`), but **deployment manifests and Helm values live outside this repo and still need checking — that check is not done and is the one real pre-merge gate.**
2. **Epoch / pairing update** — `UpdateAllProviders`' defer no longer calls `probeProviders`; confirm the re-blocked-provider unblock path still behaves.
3. **Provider selection** — the optimizer interface lost a method; confirm scores still populate and selection is sane under traffic.

---

# Description

Removes the legacy synthetic prober. It was unreachable in direct-RPC mode, and it was one of five writers into the provider optimizer's QoS scores — this is step 1 of consolidating those.

## Why it was dead

Two independent reasons:

1. **Every provider is static.** `convertProvidersToSessions` (`rpcsmartrouter.go:1686`) sets `providerEntry.StaticProvider = true`, and it builds all three provider sets — primary (`:2072`), backup (`:2076`) and recovered (`:3050`). `probeProviders` returned early for static providers, so the goroutine body never executed for any provider.
2. **The loop was off by default.** `PeriodicProbeProviders = false`; the ticker only started behind `--enable-periodic-probe-providers`. Enabling it would have changed nothing, because of (1).

The other caller — the `UpdateAllProviders` defer — did run, but hit the same early return for every provider.

## What was removed, bottom-up

| Layer | Symbol |
|---|---|
| Optimizer | `ProviderOptimizer.AppendProbeRelayData` |
| Interface | `ProviderOptimizer.AppendProbeRelayData` entry (`consumer_types.go`) |
| CSM | `csm.probeProviders`, `csm.PeriodicProbeProviders` |
| Vars | `PeriodicProbeProviders`, `PeriodicProbeProvidersInterval` |
| Flags | `--enable-periodic-probe-providers`, `--periodic-probe-providers-interval` (+ name constants) |
| Metrics | `ConsumerMetricsManagerInf.SetProviderLiveness` — only caller was the dead path, and **both** implementations were already no-ops, so no metric is lost |

`runProbeLoop` (`--probe-loop-interval`) is now the only probe loop and the only probe-fed QoS writer. `AppendProbeData` remains the single non-test entry point.

**`csm.probeProvider` (singular) is untouched** — it still backs `GenerateReconnectCallback` (`:2334`) and `checkAndUnblockHealthyReBlockedProviders` (`:2447`), as do `--debug-probes` / `DebugProbesEnabled`.

## One behaviour change, worth reading

`UpdateAllProviders` no longer probes the pairing list on epoch update, so **endpoints are no longer latency-sorted there**.

This never happened in this fork: `probeProviders` skipped static providers, and every provider is static. So nothing that ran in production stops running. But it does mean per-provider endpoint latency sorting on pairing update is absent — and was already absent. If we want it for static providers, that is a **new feature**, not a regression this PR introduces. Sorting still happens inside `probeProvider` (`:696`) on the reconnect and unblock paths.

`TestEndpointSortingFlow` covered exactly that path, using non-static providers (`createPairingList` leaves `StaticProvider` false) — a configuration the smart router cannot produce. It is removed with the behaviour it tested.

## Tests changed

- **`TestPeriodicProbeProvidersTickerCleanup`** — removed with the function it tested.
- **`TestEndpointSortingFlow`** — removed, see above.
- **`provider_optimizer_test.go`** — 8 call sites migrated to `AppendProbeData(addr, 1, latency, true, 0, false, SyncReference{})`: availability 1, latency fed, no sync evidence. Same dimensions and same `ProbeUpdateWeight` as the call they replace, so the scores under test are unchanged.
- **`TestProviderOptimizer_SyncBlockNeverDecreasesUnderConcurrency`** keeps its three racing writers — the third is now a sync-less `AppendProbeData`, which has the same read-modify-write-the-whole-record shape that made it a floor-regression risk.

Net: **-351 / +48** across 10 files.

## Review follow-ups (applied)

Two items from @avitenzer's review, both fixed in the second commit:

- **`rawPairing` was left write-only.** `probeProviders` was its only reader; with that deleted the field was declared (`:58`), assigned on every `UpdateAllProviders` (`:333`) and never read, with a comment naming the deleted function. Removed the field, the assignment, and the mention in `ResetTransientFailureState`'s doc comment — same reasoning as the write-only `EndpointMetrics` sweep in MAG-2691.
- **The replacement comment in `checkAndUnblockHealthyReBlockedProviders` was wrong.** It claimed a backup can never appear in `reportedProviders`. It can: `blockProvider`'s `AddressIndexWasNotFoundError` branch marks `blockedBackupProviders` (`:1832`) and then *falls through* to the `reportProvider` block (`:1842`), so a backup that was selected and failed a relay does get reported. The guard is correct for a simpler reason — `!isBackup` short-circuits, so a backup's report state is never consulted. Restated that way in the code, the test doc comment, and the test's assertion message.

One correction to the review, for the record: the second counterexample it offered — `probeProvider`'s `AllProviderEndpointsDisabledError` path calling `blockProvider(reportProvider=true)` at `:628` — is unreachable in this fork, since `probeProvider` returns at `:622` for static providers and every provider here is static. The load-bearing counterexample is the relay-failure one, which is what the new comment now cites.

## Follow-up

This leaves four QoS writers. The remaining consolidation is the relay paths — `AppendRelayDataConsensus` and `AppendRelayFailure` in the CSM, plus the websocket manager's pair — which is a design change rather than dead-code removal and is not attempted here.

**`probeProvider`'s entire gRPC branch may be dead by this PR's own argument.** Removing `TestEndpointSortingFlow` leaves `sortEndpointsByLatency` and `EndpointInfoList` with zero test coverage — but the sharper point is that if every provider is static, `probeProvider` returns at `:622` and lines `:625`–`:698` (the `fetchEndpointConnectionFromConsumerSessionWithProvider` call, the gRPC probe loop, and the latency sort) are unreachable for the same reason `probeProviders` was. That is ~75 lines, more than this PR deletes. Either it is reachable and now untested, or it is the next dead-code sweep. Deliberately **not** resolved here — it needs its own reachability audit rather than being bolted onto this one.

---

## Author Checklist

* [x] confirmed `!` in the type prefix if API or client breaking change — title is now `chore(lavasession)!:`. Two CLI flags are removed; they were no-ops, but a config passing them will fail to start, and release notes are generated from commit subjects so the `!` is the only signal operators get
* [ ] included the necessary unit and integration tests — **unit only; see Testing status above**
* [ ] confirmed all CI checks have passed — pending; note CI does not run unit tests

### DIFF
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -56,7 +56,6 @@ type ConsumerSessionManager struct {
 	rpcEndpoint    *RPCEndpoint // used to filter out endpoints
 	lock           sync.RWMutex
 	pairing        map[string]*ConsumerSessionsWithProvider // key == provider address
-	rawPairing     map[uint64]*ConsumerSessionsWithProvider // key == provider index in pairing. Used for periodic probing of providers
 	stickySessions *StickySessionStore
 	currentEpoch   uint64
 	numberOfResets uint64
@@ -330,8 +329,6 @@ func (csm *ConsumerSessionManager) UpdateAllProviders(epoch uint64, pairingList 
 
 	csm.lock.Lock()         // start by locking the class lock.
 	defer csm.lock.Unlock() // we defer here so in case we return an error it will unlock automatically.
-
-	csm.rawPairing = pairingList
 
 	if epoch < previousEpoch { // sentry shouldn't update an old epoch
 		return utils.LavaFormatError("trying to update provider list for older epoch", nil, utils.Attribute{Key: "epoch", Value: epoch}, utils.Attribute{Key: "currentEpoch", Value: csm.atomicReadCurrentEpoch()})
@@ -2395,11 +2392,16 @@ func (csm *ConsumerSessionManager) checkAndUnblockHealthyReBlockedProviders(ctx 
 			continue // Provider not in current pairing or backup list
 		}
 
-		// reportedProviders is populated by relay-failure reporting (ReportProvider), which only
-		// ever fires for providers that actually served traffic. A backup provider that was never
-		// selected therefore never appears there, so using !IsReported for a backup would signal
-		// health it has not demonstrated and unblock it without any real check. Always run an
-		// explicit probe for backup providers.
+		// A backup ALWAYS takes the comprehensive-probe branch: !isBackup short-circuits, so a
+		// backup's report state is never consulted at all. That is deliberate — report state is
+		// only meaningful as a health signal for a provider we actually have evidence about, and
+		// a backup that was never selected has none. Falling through to !IsReported for it would
+		// read "absent from reportedProviders" as "healthy" and unblock it with no real check.
+		//
+		// Note a backup CAN legitimately appear in reportedProviders: blockProvider's
+		// AddressIndexWasNotFoundError branch marks blockedBackupProviders and then falls through
+		// to the reportProvider block below it. That is irrelevant here precisely because the
+		// short-circuit means we never look.
 		if !isBackup && !csm.reportedProviders.IsReported(blockedAddr) {
 			// Non-backup provider whose probe succeeded (it wasn't added to reportedProviders).
 			// Unblock immediately.
@@ -2412,7 +2414,7 @@ func (csm *ConsumerSessionManager) checkAndUnblockHealthyReBlockedProviders(ctx 
 			csm.validateAndReturnBlockedProviderToValidAddressesListLocked(blockedAddr)
 			csm.reportedProviders.RemoveReport(blockedAddr)
 		} else {
-			// Either a backup provider (needs an actual probe since it has no relay history),
+			// Either a backup provider (always routed here by the short-circuit above),
 			// or a non-backup that was reported for relay failures.
 			// In both cases, run a comprehensive probe with tryReconnect=true.
 			providersNeedingComprehensiveProbe[blockedAddr] = reBlockedProviderInfo{cswp: cswp, isBackup: isBackup}
@@ -2494,7 +2496,7 @@ func (csm *ConsumerSessionManager) SetLavaBlockHeightCallback(getLavaBlockHeight
 }
 
 // ResetTransientFailureState clears every cross-epoch failure-tracking store
-// without forcing an epoch transition. The "live pairing" (pairing, rawPairing,
+// without forcing an epoch transition. The "live pairing" (pairing,
 // pairingAddresses, validAddresses, currentlyBlockedProviderAddresses,
 // backupProviders) is left intact so the very next relay can route normally.
 //
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -30,7 +30,7 @@ const (
 	BlockedProviderSessionUnusedStatus = uint32(0)
 )
 
-// debugProbes is an atomic toggle: it is read by probe goroutines (probeProviders) while the CLI
+// debugProbes is an atomic toggle: it is read by probe goroutines (probeProvider) while the CLI
 // layer sets it from the parsed --debug-probes flag. A plain bool here raced the flag write against
 // those reads under -race (a leaked probe goroutine from one test vs another test building the cobra
 // command). Use SetDebugProbes/DebugProbesEnabled rather than touching it directly.
@@ -43,13 +43,11 @@ func SetDebugProbes(enabled bool) { debugProbes.Store(enabled) }
 func DebugProbesEnabled() bool { return debugProbes.Load() }
 
 var (
-	retrySecondChanceAfter         = time.Minute * 3
-	PeriodicProbeProviders         = false
-	PeriodicProbeProvidersInterval = 5 * time.Second
-	// ProbeLoopInterval is the configurable cadence (MAG-2161 D5) of the real proactive health prober
-	// (rpcsmartrouter.runProbeLoop). Default 5s; validated at startup (a non-positive value is rejected
-	// back to the default). Distinct from PeriodicProbeProvidersInterval, which clocks the legacy
-	// synthetic probe loop — the prober does NOT use that knob.
+	retrySecondChanceAfter = time.Minute * 3
+	// ProbeLoopInterval is the configurable cadence (MAG-2161 D5) of the proactive health prober
+	// (rpcsmartrouter.runProbeLoop) — the single source of truth for direct-RPC endpoint health and
+	// probe-fed QoS. Default 5s; validated at startup (a non-positive value is rejected back to the
+	// default).
 	ProbeLoopInterval = 5 * time.Second
 )
 
@@ -322,9 +320,6 @@ func (csm *ConsumerSessionManager) UpdateAllProviders(epoch uint64, pairingList 
 		time.Sleep(time.Duration(rand.Intn(500)) * time.Millisecond) // sleep up to 500ms in order to scatter different chains probe triggers
 		go func() {
 			ctx := context.Background()
-			// Probe all providers to eliminate offline ones from affecting relays
-			csm.probeProviders(ctx, pairingList, epoch) // pairingList is thread safe it's members are not (accessed through csm.pairing)
-
 			// Check re-blocked providers from previous epoch and unblock if healthy
 			csm.checkAndUnblockHealthyReBlockedProviders(ctx, epoch)
 		}()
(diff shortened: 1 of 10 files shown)


Tomelia1999:
⚠️ **Testing caveat — flagging explicitly rather than leaving it implied.**

Beyond unit tests, **no testing has been done on this change**. No integration run, no e2e, no manual run against real endpoints, no sim/deployment validation. The change deletes code across the session manager, the provider optimizer, the metrics interface and the CLI surface, so someone should confirm nothing is broken before merging.

Two things that make this easier to under-weigh than it should be:

1. **CI does not run `go test`.** No workflow in `.github/workflows/` invokes it. The green checks on this PR are build, lint, govulncheck and security scanning — the unit-test pass reported in the description came from a local run and is not reproduced here.
2. **Two CLI flags were removed.** Any deployment still passing `--enable-periodic-probe-providers` or `--periodic-probe-providers-interval` will fail to start with an unknown-flag error. Worth grepping deployment manifests and Helm values before rollout, even though both flags were no-ops.

Suggested minimum before merge: one `scripts/pre_setups/` run (e.g. `test_uc1_smartrouter_lava.sh` or `test_uc4_smartrouter_sim.sh`), plus a router started against real endpoints long enough to confirm probe-fed QoS still populates and provider selection is sane.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -321,9 +319,6 @@ func (csm *ConsumerSessionManager) UpdateAllProviders(epoch uint64, pairingList
 		time.Sleep(time.Duration(rand.Intn(500)) * time.Millisecond) // sleep up to 500ms in order to scatter different chains probe triggers
 		go func() {
 			ctx := context.Background()
-			// Probe all providers to eliminate offline ones from affecting relays
-			csm.probeProviders(ctx, pairingList, epoch) // pairingList is thread safe it's members are not (accessed through csm.pairing)
-
 			// Check re-blocked providers from previous epoch and unblock if healthy
 			csm.checkAndUnblockHealthyReBlockedProviders(ctx, epoch)
```
avitenzer, line 323: Dropping the `probeProviders` call here orphans `csm.rawPairing` — that function was its only reader (`PeriodicProbeProviders`, pre-change `:626`). The field is still declared at `:58` and still assigned on every `UpdateAllProviders` at `:333`, so it becomes write-only state that pins a reference to the pairing map for the process lifetime. Its comment also still reads "Used for periodic probing of providers", naming the function this PR deletes.

Since this is a dead-code sweep, and MAG-2691 removed write-only `EndpointMetrics` state for exactly this reason, I'd drop the field, the assignment, and the mention in `ResetTransientFailureState`'s doc comment (`:2436`) here rather than leave a stale pointer for the next pass.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -2404,11 +2334,11 @@ func (csm *ConsumerSessionManager) checkAndUnblockHealthyReBlockedProviders(ctx
 			continue // Provider not in current pairing or backup list
 		}
 
-		// reportedProviders reflects probe outcomes only for pairingList providers —
-		// probeProviders only probes pairingList, so backup providers are never probed
-		// and will never appear in reportedProviders. Using !IsReported for a backup
-		// provider would incorrectly signal a successful probe and unblock it without
-		// any real health check. Always run an explicit probe for backup providers.
+		// reportedProviders is populated by relay-failure reporting (ReportProvider), which only
+		// ever fires for providers that actually served traffic. A backup provider that was never
```
avitenzer: This rationale isn't accurate — and it's the same "never" claim the old comment made, just via a different route.

`blockProvider` does reach `ReportProvider` for a backup: `removeAddressFromValidAddresses` returns `AddressIndexWasNotFoundError`, the branch at `:1831` marks `blockedBackupProviders`, and control then falls through to `if reportProvider` at `:1842` → `:1846`/`:1864`. So a backup that was selected and failed a relay does land in `reportedProviders`. `ReportProvider` also isn't relay-only: `probeProvider`'s `AllProviderEndpointsDisabledError` path calls `blockProvider(..., reportProvider=true, ...)`.

The guard itself is fine, because `!isBackup` short-circuits before `IsReported` is ever consulted. So the accurate version is simpler: backups always take the comprehensive-probe branch because `!isBackup` fails — their report state is never read, so whether they *can* appear in `reportedProviders` doesn't matter. Same applies to the assertion message in `TestCheckAndUnblock_BackupRoutedToComprehensiveProbe` ("backup provider should never appear in reportedProviders"), which holds for that fixture but not in general.

avitenzer (commented):
Reviewed against the code. Build, vet, and the affected packages' tests pass locally on `137e276`.

## The core claim holds

All four non-test `UpdateAllProviders` callers (`rpcsmartrouter.go:2095,2908,3083,3180`) pass maps built by `convertProvidersToSessions` (`:1774`, `StaticProvider = true`) or copies that preserve the bit (`:2822`, `:2851`). So `probeProviders`' early return did fire for every provider and the goroutine body never ran — the dead-code argument is sound. No reference to any removed symbol or flag survives anywhere in the repo (Go, shell, yaml, md, json). No orphaned Prometheus metric either: `SetProviderLiveness` really was a no-op in both implementations.

The optimizer test migration is exactly equivalent. `AppendProbeData(a, 1, lat, true, 0, false, SyncReference{})` takes the same stripe lock, floors with `observed = 0` (pure preserve), uses the same `ProbeUpdateWeight`, feeds availability=1 + latency, and skips sync. All 8 migrated sites passed `success=true`, so the failure branch never diverges.

Local verification on this head:

- `go build ./...` — clean
- `go vet` over lavasession / provideroptimizer / metrics / rpcsmartrouter — clean
- `go test ./protocol/lavasession/... ./protocol/provideroptimizer/... ./protocol/metrics/...` — all pass
- `gofumpt` — the two findings on touched files are pre-existing import grouping, present on `HEAD~1` too, so "no new findings" checks out

Also confirming the note in the description: none of the nine workflows in `.github/workflows/` invokes `go test`, so the local run really is the only functional evidence here.

## Blocking

**The Jira line is still the `MAG-XXXX` placeholder.** `jira-ticket/validated` is failing with "Malformed Jira ticket line" — it is the only red check.

## Worth doing before merge

**Add the `!`: `chore(lavasession)!: ...`.** Two CLI flags are removed, so any deployment passing them fails to start. Release notes are generated from commit subjects (see CHANGELOG.md), so without the `!` operators get no signal — the explanation in the PR body doesn't reach them. You already identified this and left the box unchecked; I'd just mark it.

Two inline comments below: `rawPairing` is left write-only, and the replacement comment in `checkAndUnblockHealthyReBlockedProviders` carries forward a claim that isn't true.

## Non-blocking

Removing `TestEndpointSortingFlow` leaves zero coverage of `sortEndpointsByLatency`, `EndpointInfoList`, and the gRPC branch of `probeProvider` — code this PR deliberately keeps. By the PR's own argument (every provider is static → `probeProvider` returns early into `probeDirectRPCEndpoints` at `:620`), that sort is equally unreachable. So either it is reachable and now untested, or it belongs in the follow-up list alongside the four remaining QoS writers. Worth naming in the Follow-up section either way.

Of the three risk areas flagged in the description, #1 (startup with the removed flags) is the only one I would treat as load-bearing. Nothing inside this repo references either flag, so a grep of the deployment manifests should settle it.

protocol/lavasession/consumer_session_manager.go
Tomelia1999: Fixed in 0185a79 — dropped the field (`:58`), the assignment in `UpdateAllProviders`, and the `rawPairing` mention in `ResetTransientFailureState`'s doc comment. Confirmed zero references left in the repo.

Agreed on the MAG-2691 precedent; that was the right frame. One note: "pins a reference to the pairing map for the process lifetime" slightly overstates the cost — it was one map, replaced on each `UpdateAllProviders`, and in direct-RPC mode there are no epoch transitions anyway. So this was hygiene rather than a leak, which is why it is in this PR and not a hotfix.

protocol/lavasession/consumer_session_manager.go
Tomelia1999: You are right, and I took your suggested framing — fixed in 0185a79.

Verified the mechanism: `blockProvider`'s `AddressIndexWasNotFoundError` branch marks `blockedBackupProviders` at `:1832` and then falls through with no return to `if reportProvider` at `:1842` → `ReportProvider` at `:1846`/`:1864`. So a backup that was selected and failed a relay does land in `reportedProviders`, and my "never appears there" was just the old absolute claim wearing a new hat.

The comment now leads with the short-circuit — `!isBackup` fails, so a backup's report state is never consulted — and explicitly notes that a backup *can* be reported and that this is irrelevant precisely because we never look. Same fix applied to the test's doc comment and its assertion message, which now says "this backup never served traffic, so it must not be in reportedProviders" — true of the fixture, and not claiming to be true in general.

One correction on the supporting argument: the second counterexample — `probeProvider`'s `AllProviderEndpointsDisabledError` path calling `blockProvider(..., reportProvider=true, ...)` at `:628` — is unreachable in this fork. `probeProvider` returns at `:622` for static providers and every provider here is static, which is this PR's own central premise. So that half does not carry weight; the relay-failure path is the one that does, and it is what the new comment cites.

That same observation is why I added the gRPC-branch reachability question to Follow-up rather than resolving it here — see the reply on your non-blocking point.

Tomelia1999:
All four review points addressed — 0185a79 for the code, plus PR title/body.

**1. Jira placeholder** — now `MAG-2733` ([MAG-2733](https://magmadevs.atlassian.net/browse/MAG-2733)).

**2. `!` in the type prefix** — retitled `chore(lavasession)!: remove the legacy synthetic prober`, checklist box now ticked with the reasoning (release notes come from commit subjects, so the `!` is the only signal that reaches operators).

**3 & 4. The two inline comments** — both fixed, replies on the threads. `rawPairing` is gone; the unblock rationale now leads with the `!isBackup` short-circuit instead of a claim about `reportedProviders` that is not true in general.

**5. `sortEndpointsByLatency` coverage (non-blocking)** — added to Follow-up, and I would push your point further. Confirmed zero test references to `sortEndpointsByLatency` or `EndpointInfoList` after `TestEndpointSortingFlow` goes. But by this PR's own argument the exposure is bigger than the sort: if every provider is static, `probeProvider` returns at `:622` and lines `:625`–`:698` — the `fetchEndpointConnectionFromConsumerSessionWithProvider` call, the whole gRPC probe loop, and the sort — are unreachable for exactly the reason `probeProviders` was. That is ~75 lines, more than this PR deletes. Deliberately not resolved here: it deserves its own reachability audit, not a bolt-on to this one.

**6. Deployment manifests** — I grepped this repo across Go, shell, yaml, md and json (including `config/smartrouter_examples/` and `docker/`): clean, no reference to either flag. **The manifests and Helm values live outside this repo and have not been checked.** Agreed with your read that this is the only load-bearing risk of the three I listed — since the deleted code provably never executed, an integration run mostly proves the build works, not that the removal is safe. So this grep is the real pre-merge gate and it needs an owner outside this PR.

**Verification on `0185a79`:**

- `go build ./...` — clean
- `go vet ./protocol/...` — clean
- `go test ./protocol/... ./types/... ./utils/...` — all green
- `gofumpt` — the one finding on `consumer_session_manager.go` is byte-identical to the finding on `137e276`, i.e. pre-existing import grouping, not introduced here

Testing status is otherwise unchanged from the original caveat: unit only, and CI still does not run `go test`.

[MAG-2733]: https://magmadevs.atlassian.net/browse/MAG-2733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

avitenzer (approved):

PR #261, opened 2026-08-10
feat(cache): RESP-compatible (Redis/Valkey) cache backend

avitenzer:
Jira ticket: MAG-2647

Adds a RESP-compatible (Redis/Valkey) cache backend as an alternative to the gRPC cache
sidecar. The router runs the same cache engine in-process and stores entries in a remote,
persistent, shareable backend.

`cache-be` is preserved as the rollback path: with both configured, resp-cache wins and
`cache-be` is never dialled; removing the `resp-cache:` block reverts to today's behaviour on
the next start. With no resp configuration at all, behaviour is identical to `main` — the
metrics register lazily, so not even an empty series appears.

Operator documentation: `docs/RESP-CACHE.md`, metrics in `docs/METRICS.md`.

## What's in it

- **Storage-agnostic cache engine** extracted from the gRPC server, with the router seamed
  onto a `CacheBackend` interface (lifecycle `Close`).
- **`redisstore`** — standalone / sentinel / cluster topologies, TLS and mTLS, streaming
  credentials with live rotation, and an optional read/write endpoint split.
- **Selection and precedence** in `SelectCacheBackend`, plus the config surface (`resp-cache:`
  YAML block, `--resp-cache-addresses`, `--resp-cache-topology`) with flags outranking YAML.
- **Observability** — `smartrouter_resp_cache_*` series for connection errors, per-op
  failures split by `op`/`kind`, connectivity, and pool stats.
- **Degraded-mode guarantees** — a dead or unauthenticated backend degrades relays to cache
  misses rather than failing them.

## Verification

- **Feature parity.** Every behavioural case runs against *both* backends — a real
  `ecosystem/cache` gRPC server on a loopback port and `RespCache` — via
  `go test ./protocol/performance -run '^(TestParity|TestSelectBackend)'` → 15 tests, 35 PASS.
  Covers finalized-vs-temp variant coexistence, request-hash and seen-block validation,
  compression, node-error TTL expiry, `"latest"` resolution and tip monotonicity, flush, and
  backend selection including precedence and rollback.
- **Real servers in docker** (opt-in lanes, all PASS): sentinel failover with the primary
  killed mid-run and the same Store serving through the promotion; mTLS against a Valkey
  terminating TLS with certificate-less clients rejected; a three-master cluster reached
  through a single discovery endpoint with slot-spread writes and cross-slot lookups.
- **Credential rotation** against a real server: live connections re-authenticate *in place* —
  the same server-side `CLIENT ID` reports the new ACL user, and the pool converges to
  dial-free serving. Proven twice: once driving the provider directly, and once through
  `redisstore.New()` so the password-file watcher the router actually owns is what notices the
  rewritten secret. No restart, no reconnect.
- **Failure behaviour**: backend down → relays keep serving from upstream,
  `smartrouter_resp_cache_connected` goes to 0, and transition-only logs report recovery.
  Authentication failures are classified distinctly from network faults and the log detail is a
  fixed phrase, so no password, username, or server `WRONGPASS` reply is ever emitted.
- **Shared-backend safety**: purge is prefix-scoped via SCAN + UNLINK, so foreign keys survive
  a flush. The key prefix is restricted to `[A-Za-z0-9._-]+` precisely because it is fed into a
  SCAN MATCH glob.

## Notes for reviewers

- The docker-backed lanes are opt-in through environment variables and **do not run in CI**.
  They now fail hard rather than skipping when opted into without a reachable daemon — a
  skipped test still leaves the package `ok` and the exit code 0, which is how an unverified
  lane gets mistaken for a passing one.
- No CI job runs `go test` on this repo, so local `make test` is the effective test gate.
- `golangci-lint` reports a few advisory findings in the new code (`forcetypeassert` in a test,
  `nilnil`, and `ST1012` error-var naming). The naming one matches the existing convention in
  `protocol/performance/errors.go`; changing it would make the new code inconsistent with the
  old. Happy to address any of them on request.


AnnaR-prog:
## Review

Reviewed the ~2.5k lines of production code (`ecosystem/cache/core`, `ecosystem/cache/redisstore`, `ristretto_store.go`, `protocol/performance/resp_*`, and the router seams); the docs, shell lanes, and tests I read for corroboration rather than line-by-line.

`go build ./...` and `go test ./ecosystem/cache/... ./protocol/performance/...` both pass on `7e88e8d`.

**Verdict: the design holds and the extraction is faithful. One verified defect I'd want fixed before merge (a 2-line change), plus five fix-or-file items.**

I checked the two things a refactor of this shape most often gets wrong, and both are clean:

- **Engine extraction fidelity.** Diffed `core.Engine` against `main`'s `handlers.go` statement by statement — `getRelayInner` precedence, the 3-goroutine fan-out, the seen-block rejection, `replaceRequestedBlock`, TTL selection, and `FlushCache` (identical guards, same per-store `Clear`/`Wait`, same `endpointObservations.clear()`) all survive unchanged. The `cacheSource` reconstruction via `(i == 0) == Finalized` is correct in all four cases. `formatHashKey`'s in-place `binary.AppendUint64` on the caller's slice is gone, and the old `chainId + "_" + ""` chain-tip key no longer shares a namespace with shared-state tips — two latent hazards removed on the way past.
- **"No resp config → identical to `main`."** This is the PR's load-bearing claim, and it rests on `v.IsSet("resp-cache")` staying false once `BindPFlags` has bound the flat flags. `TestRealCommand_NoRespConfigKeepsLegacyPath` proves it through the *real* cobra command's viper rather than a synthetic flag set. That test is the right one to have written.

Also worth saying: the both-backends parity matrix, the degraded-mode tests (particularly `TestRespCacheCallerBudgetBoundsEstablishedConnections`), and the deliberate "every key stays volatile so `volatile-*` eviction sees the whole keyspace" design are the parts of this PR that make the rest reviewable.

---

### 1. A corrupt shared-tip key permanently wedges tip publishing — `redisstore/store.go:355`

`setInt64GEScript` compares against `tonumber(cur)` with no nil guard. When the stored value isn't numeric, `tonumber` yields nil and Lua raises on the comparison — so the write never lands, and because the key is never overwritten it **cannot self-heal**. Every subsequent `SetSharedTip` for that chain/pod fails the same way until someone deletes the key by hand.

Verified against real Redis 7 (not just miniredis):

```
127.0.0.1:6379> SET sr:tip:ETH1:pod-a "not-a-number"
OK
127.0.0.1:6379> EVAL "<setInt64GEScript>" 1 sr:tip:ETH1:pod-a 12345 30000
(error) ERR user_script:3: attempt to compare number with nil
127.0.0.1:6379> GET sr:tip:ETH1:pod-a
"not-a-number"          <- still corrupt, forever
```

This contradicts the invariant the PR states two functions up, in `GetInt64` (`store.go:372`):

> The in-memory adapter fatals on a corrupt shared-tip entry; a router-embedded backend on a **SHARED** store must not be crashable by a foreign writer, so corruption reads as a miss instead.

The read side honours that. The write side doesn't. And the sibling script already has the guard — `setChainTipGEScript` (`store.go:401`) does `local curb = tonumber(...)` then `if curb and ...`, which I confirmed returns `1` and self-heals on the same corrupt input. The fix is to mirror it:

```lua
local cur = redis.call('GET', KEYS[1])
local curn = cur and tonumber(cur)
if curn and tonumber(ARGV[1]) < curn then
	return 0
end
```

A `TestSetInt64OverCorruptValue` alongside the existing `TestCorruptEntryReadsAsMiss` would lock it in.

> **Testing-infra caveat, worth one line in a comment somewhere:** under miniredis the *chain-tip* script also fails on a corrupt value (`bad argument #1 to tonumber (value expected)`), because gopher-lua's `tonumber` rejects nil where PUC Lua 5.1 returns nil. That script is **correct** — I verified it self-heals on real Redis. Without this noted, whoever adds the corrupt-value tests will see the chain-tip case fail under miniredis and "fix" working code. It also explains why this whole class of bug wasn't caught: the corrupt-value tests that exist only cover envelopes, which decode in Go rather than in Lua.

### 2. The sentinel endpoint tracker is non-deterministic — `redisstore/config.go:256`

`trackingDialer` is installed as `FailoverOptions.Dialer`, and go-redis's `sentinelOptions()` (v9.22.0, `sentinel.go`) copies `opt.Dialer` verbatim for the **sentinel control-plane** connections. So `endpointTracker` records sentinel addresses as well as the master, last-write-wins, and sentinel connections re-dial at arbitrary times (discovery, the `+switch-master` pub/sub subscription). The comment at `config.go:254` — "Records the DATA-node address, which under sentinel is whichever node currently holds the master role" — isn't what the code does.

The header isn't *wrong* so much as *unreliable*, which matters here specifically: the PR's sentinel-failover verification leans on watching `Lava-Cache-Backend` change across a promotion, and that signal can be a sentinel address. `FailoverOptions` exposes no separate sentinel dialer, so the practical fix is to filter noted addresses against `SentinelAddrs` before calling `tracker.note`.

### 3. Block-hash→height lookups are N sequential round trips inside a 50 ms budget — `core/engine.go:93`

`getBlockHeightsFromHashes` loops `e.Store.GetHeight` one key at a time. Free over ristretto; over RESP it's one network round trip per hash, running concurrently with the entry lookup but sharing `common.CacheTimeout = 50ms`. A request carrying several block hashes can spend the whole budget here and turn a warm cache into a miss — and the multi-region lane this PR ships (`init_smartrouter_eth_redis_multiregion.sh`) is exactly where RTT makes that bite.

`KVStore` already documents batching as the adapter's job for `GetEntries` ("Adapters should batch where the transport allows"); heights need the same surface, because an adapter cannot batch across separate `GetHeight` calls. A `GetHeights(ctx, keys []string) ([]int64, []bool, error)` would collapse it to one pipeline, matching what `GetEntries` already does.

The write side (`setBlocksHashesToHeights`, `engine.go:105`) has the same shape but is currently unreachable from the router — `tryCacheWrite` passes `BlocksHashesToHeights: nil`, and it's async on a 5s budget regardless. Read path is the one that matters.

### 4. `TTL == 0` creates a permanently non-volatile key — `core/policy.go:81` (latent)

`ForRelayEntry(finalized=true, isNodeError=true, averageBlockTime=0, ...)` returns `min(0, 250ms) = 0`, and `Store.SetEntry` maps a non-positive TTL to a plain `SET`. Confirmed: the key comes back with `TTL == -1`, i.e. it never expires. That defeats the "every key stays volatile so `volatile-*` maxmemory policies see the whole keyspace" property `chainTipRetention` was deliberately bounded to preserve — a single such write leaves a key no `volatile-*` policy can ever evict.

**Not reachable today**, and I want to be clear about that: `tryCacheWrite` hardcodes `IsNodeError: false`, `chain_fetcher.go`'s `populateCache` never sets the field, and every router-side `ChainFetcherOptions` literal passes `Cache: nil`. So this is hardening against a future caller, not a live bug. Cheapest guard is a floor in `ForRelayEntry` (`isNodeError` with a zero `averageBlockTime` should fall back to `p.NodeErrors`, not to zero) rather than a special case in the adapter, since the ristretto adapter has the identical "0 means never expires" semantics.

### 5. Sentinel + `password-file` runs a watcher that can't do anything — `redisstore/store.go:158`

`New` starts `watchCredentials` whenever `PasswordFile != ""`, but `buildClient` doesn't hand the `StreamingProvider` to the sentinel path at all (it uses `CredentialsProviderContext`, per the go-redis v9.22 gap documented at `config.go:246`). So under sentinel the provider has zero subscribers, and every rotation logs:

> `resp-cache credentials rotated; re-authenticating live connections  connections=0`

An operator reading that reasonably concludes rotation took effect in place; it didn't — it applies on the next reconnect. Either skip the watcher for sentinel, or make the log say which of the two happened. (`RESP-CACHE.md` documents the sentinel restriction for the *sentinel* password; the data-node one under sentinel deserves the same note.)

### 6. A password containing `:` is silently reinterpreted — `redisstore/credentials.go:40`

`FileCredentials.Credentials` does an unconditional `strings.Cut(raw, ":")`, so a password file holding `p@ss:word` authenticates as user `p@ss` / password `word`. Fails closed rather than open, but the failure mode is a puzzling `WRONGPASS` with no hint at the cause — and the auth-error log path deliberately suppresses the server's reply, so there's nothing to go on. Worth either requiring an explicit opt-in for the combined form, or documenting the constraint where operators will hit it.

---

### Housekeeping

- **GitGuardian's 2 findings are false positives** — `placeholder-sentinel-credential`, `placeholder-not-a-real-credential`, and the `datauser`/`datapass` fixture in `TestFailoverOptionsMapping`. Worth dismissing in the dashboard so the check goes green rather than leaving a red mark future reviewers have to re-triage.
- **The branch is `CONFLICTING` with `main`** (base has moved ~20 commits, including v1.4.1) and needs a rebase.
- **On the "no CI job runs `go test`" note in the PR description** — agreed it's pre-existing, but it's also a reviewer-level concern for a 9k-line change that touches the relay hot path: the only gate is whoever remembers to run `make test` locally, and the docker lanes that cover sentinel/cluster/mTLS are opt-in on top of that. Making them fail-hard instead of skip was the right call; it just doesn't help while nothing invokes them.
- Findings 3–6 are all fix-or-file as far as I'm concerned. Only #1 I'd hold the merge for, and it's two lines.

AnnaR-prog:
## Re-review — `cf653b9` + merge `7d930b8`

Re-reviewed against my [previous review](https://github.com/Magma-Devs/smart-router/pull/261#issuecomment-5509982223). All six findings are addressed, each with a test. `go build ./...`, `go vet`, and `go test ./ecosystem/cache/... ./protocol/performance/... ./protocol/chainlib/... ./protocol/rpcsmartrouter/...` are green on the merged tree (`7d930b8`).

**Verdict: #1 — the one I'd have held the merge for — is correctly fixed and the fix is genuinely test-covered. No remaining blockers. One residual on #2 that I'd still want closed, because the fix does not hold in the deployment shape this PR's own docs recommend.**

I verified the two claims that carry the most risk rather than taking them on faith:

- **Removing the sentinel watcher (#5) did not silently delete rotation.** The new startup log promises rotation "applies on the next reconnect or failover," which only holds if the sentinel path re-resolves per dial. It does: `failoverOptions` sets `CredentialsProviderContext` to a closure over `source` (`config.go:285`), and `FileCredentials.Credentials` re-reads the file on every call (`credentials.go:49`). So the watcher was genuinely redundant there, not load-bearing. `Close()` already guarded a nil `stopWatcher`, so skipping the goroutine doesn't panic on shutdown.
- **"Each with a regression test" is true, not just asserted.** I reverted the two Lua lines and the `averageBlockTime <= 0` return in a worktree and re-ran: `TestSetInt64OverCorruptValue` fails with `attempt to compare number with nil`, and `TestPolicyForRelayEntry` fails `expected: 250ms / actual: 0s`. Both tests fail against the old code for the right reason, which is what makes them regression tests. The miniredis caveat I raised was taken seriously rather than worked around: binding `match` before `tonumber` made the chain-tip path testable without touching its semantics, and `TestSetChainTipOverCorruptValue` now covers the half that was previously unreachable under the fake.

Also checked, since the merge landed on top: main's new `RedactSecrets` now rewrites every log attribute value (`lavalog.go:601`), and the auth-log suite asserts fixed phrases. `protocol/performance` still passes, so the redaction pass doesn't reshape those assertions.

On #4 specifically — the floor is complete for the surface that can reach it. `ForRelayEntry`'s other branches still return 0 if `Finalized`/`NonFinalized` are flag-set to 0, but the router-embedded backend always builds its policy from `core.DefaultPolicy()` (`resp_config.go:110`), whose constants are non-zero. The flag-driven zero is reachable only on the standalone gRPC server over ristretto, where it is pre-existing `main` behaviour. Nothing to do here.

---

### The one residual: sentinel exclusion is static, but the sentinel list is not (#2)

`trackingDialerExcluding` builds `excluded` **once**, from the configured `addrs` (`config.go:234`). go-redis grows its sentinel list at runtime:

- `discoverSentinels` runs `SENTINEL sentinels <master>` and **appends** every address not already in the slice (`sentinel.go:1212-1234`).
- Every entry in `c.sentinelAddrs` — seeded *and* discovered — is then dialed via `NewSentinelClient(c.opt.sentinelOptions(addr))` (`sentinel.go:1017`, and again at `1105` for replica lookups), and `sentinelOptions` copies `Dialer: opt.Dialer` verbatim (`sentinel.go:265`).

So a discovered sentinel is dialed through the tracking dialer and is not in the exclude set — `tracker.note` fires and `Lava-Cache-Backend` names a sentinel again.

This isn't a corner case, because of an address-form mismatch: `SENTINEL sentinels` reports the peers' **IP** addresses, while the configuration example in `RESP-CACHE.md:105` seeds **hostnames**:

```yaml
addresses: ["sentinel-1:26379", "sentinel-2:26379", "sentinel-3:26379"]
```

`slices.Contains` compares raw strings, so `10.0.0.5:26379` never matches `sentinel-2:26379`. It gets appended, it gets dialed, and it is not excluded. Configuring all three sentinels doesn't help — discovery returns the *other* sentinels, always in IP form. In a Docker or Kubernetes deployment, which is what the sentinel lane exercises, the pollution is close to guaranteed rather than rare.

`TestSentinelAddressesAreNotRecordedAsTheDataNode` passes because it calls `trackingDialerExcluding` directly with the address already in the exclude list. It pins the mechanism, not the wiring — which is exactly how this survived the fix.

The constraint any fix has to meet is that it must discriminate by *which client is dialling*, not by how the address is spelled — and it can't probe the connection, because the dialer hands back a raw `net.Conn` before go-redis's handshake, with no AUTH run and different credentials on the two paths anyway.

One mechanism in v9.22 does meet it structurally. `NewFailoverClient` replaces the data-path dialer with `masterReplicaDialer` (`sentinel.go:559`), builds the data pool through the returned client's own hook chain (`sentinel.go:582`), and `masterReplicaDialer` threads the caller's context into `failover.opt.Dialer` (`sentinel.go:647`). The internal sentinel clients are separate `*Client`s built from `sentinelOptions`, so they never traverse that hook chain. A `DialHook` added to the returned failover client therefore fires on data-path dials *only* — by construction, not by heuristic. It can't carry the address itself (the pool dials the `"FailoverClient"` placeholder and `masterReplicaDialer` resolves the real master downstream), but it can mark the context, and the tracking dialer can record only when that mark is present.

If that's more machinery than the header is worth, the honest alternative is to stop claiming the data node under sentinel and describe the value as "an endpoint in the sentinel set" — the comment at `config.go:275` currently states a guarantee the code doesn't provide.

### Nits

- `RESP-CACHE.md:170` says the colon case "logs a **warning**"; `credentials.go:61` logs at Info. The startup timing in the doc is right — `New` resolves credentials eagerly at `store.go:126` to fail fast, which trips the `sync.Once` there.
- Because `cfg.credentialsSource()` returns a fresh `FileCredentials` per call, the "once" is once per instance — the fail-fast probe, the provider, and (under sentinel) each `buildClient` each own one, so the line can appear 2–3 times. Bounded and not per-connection, which was the point, so this is only worth knowing when reading the logs.

### Housekeeping

- **GitGuardian is still failing** and `mergeStateStatus` is `BLOCKED`. Both findings are the placeholder fixtures I confirmed last time; they need dismissing in the dashboard.
- **The rebase is done** — the branch is `MERGEABLE` and the merge note's conflict resolution matches what's in the tree. I re-checked both facts the fix commit depends on directly in the merged tree rather than from the note: `tryCacheWrite` still early-returns on node errors (`rpcsmartrouter_server.go:3358`) and still hardcodes `IsNodeError: false` (`:3502`), so #4 stays hardening; and the cache-get path still populates `BlocksHashesToHeights` from the protocol message (`:3685`), so #3 stays on a reachable path.

nimrod-teich:
## Review at `7d930b8` — independent pass

Worktree at the merged head (merge base `bd64f59`), built, vetted, and ran the suites myself — including `-race` on the cache packages. I read the two earlier reviews and re-checked their open residual against go-redis v9.22.0's source rather than inheriting it; the rest of this is things those passes did not cover.

The extraction is faithful and the seam is the right one: `core.Engine` over `KVStore` is the same code both backends run, the parity matrix proves it against a real gRPC server, and the corrupt-value / TTL-floor fixes from the first round are real regression tests. `CacheBackend` with a typed-nil convention keeps the no-config path byte-identical to `main`, and `TestRealCommand_NoRespConfigKeepsLegacyPath` pins that through the real cobra command.

**Verdict: not mergeable yet.** One degraded-mode gap the docs promise away, one integration hazard with #248 that will silently break that PR's contract if the merge is done mechanically, plus the sentinel residual already on the record. Everything else is fix-or-file.

---

### 🟠 A black-holed **TLS** backend costs cold lookups the 5 s dial timeout, not the 50 ms budget — `redisstore/config.go:231`

`trackingDialer` delegates to `redis.NewDialer`, whose TLS branch is `tls.DialWithDialer(netDialer, …)` (`options.go:626`) — a ctx-less API. The pool wraps each attempt in `attemptCtx` bounded by the caller's deadline (`pool.go` dial loop), but that context never reaches the TLS handshake, so a fresh dial against a black-holed (SYN-dropped, not refused) TLS endpoint runs the full `DialTimeout`. And `Config.DialTimeout` is passed through unset, so go-redis's default of **5 s** applies (`options.go:485`). Plaintext is fine — `DialContext(ctx)` honours the budget — which is why `TestRespCacheSlowBackendTimesOutWithinBudget` passes: it sets `DialTimeout: 200ms` explicitly and runs without TLS.

Bounded, not unbounded: go-redis fails fast once `dialErrorsNum ≥ PoolSize` and probes in the background. But `PoolSize` defaults to `10 × GOMAXPROCS`, so on a 16-core router that's ~160 relays paying up to 5 s each at the start of every partition, plus the async `SetEntry` goroutines holding pool slots for their 5 s write budget. `RESP-CACHE.md:192` ("lookups degrade to cache misses within the relay's budget") is stated without the TLS caveat that `:79` half-admits.

Two fixes, do both: (1) build the TLS path with `(&tls.Dialer{NetDialer: …, Config: tlsCfg}).DialContext(ctx, …)` so the caller's deadline reaches the handshake — `trackingDialer` already owns the dial closure, so it costs no coupling; (2) default `dial-timeout` to something in the hundreds of milliseconds when unset, rather than inheriting a client default sized for batch jobs. A test with a TLS listener that accepts and never handshakes, under a 150 ms context, would pin it the way the existing plaintext test does.

### 🟠 Integration with #248: the RESP envelope must carry `IsNodeError` / `StatusCode`, or the secondary-cache contract silently breaks

I trial-merged #248's head (`4b70b2c`) onto this branch: conflicts in `handlers.go`, `rpcsmartrouter.go`, `rpcsmartrouter_server.go`, `LOCAL-COMPOSE.md`. Textual conflicts are routine; the semantic one is not. #248 adds `IsNodeError` and `StatusCode` to `CacheValue` and `formatCacheValue` — the fields its secondary tier reads to decide whether a foreign entry may be backfilled. Here `CacheValue` becomes an alias of `core.Envelope`, and the RESP wire form is a separate `storedEnvelope` (`store.go:283`) with its own field list. Whoever resolves the conflict has to thread both fields through `core.Envelope`, `NewEnvelope`, `ToCacheReply`, **and** `storedEnvelope` (with the `v` bump the version field exists for), or a RESP-backed cache reports every entry as `IsNodeError=false, StatusCode=0` and #248's node-error rejection quietly degrades to the legacy placeholder check. Nothing in either PR's test set would catch that combination. Worth agreeing the merge order now and adding a parity case for those two fields to whichever lands second.

### 🟠 The sentinel `Lava-Cache-Backend` residual stands — confirmed in the vendored source

The previous review's residual is accurate: `sentinelOptions` copies `Dialer: opt.Dialer` (`sentinel.go:265`), `discoverSentinels` appends learned addresses (`:1232`), and those are dialled through the tracking dialer outside the static exclude set. Since discovery reports peers as IPs and the docs seed hostnames, this fires in the exact topology the sentinel lane exercises. Either the `DialHook` approach the earlier review sketched, or stop claiming the data node under sentinel and reword `config.go:275` and the doc line. Not a correctness issue — the header is debug-only — but the comment states a guarantee the code doesn't provide.

---

### 🟡 Fix-or-file

- **Backend failures are invisible to the shared cache metrics.** `RespCache.GetEntry` swallows `core.StoreError` and returns `(reply, nil)` (`resp_cache.go:228-232`), so the call site sees `cacheError == nil` and a nil reply — a clean miss. Today that only mislabels the debug log; once #248 lands, `smartrouter_cache_failed_total{outcome}` will report a RESP outage as `miss` rather than `error`/`timeout`, contradicting the METRICS.md that PR ships. The gRPC *client* returns transport errors, and the router already degrades any error to a miss (`cacheHit := cacheError == nil && …`, `GetReply()` is nil-safe), so returning the joined error is behaviour-preserving. The comment says it mirrors the gRPC *server's* contract; the seam it replaces is the client's. `degradedGet`'s `require.NoError` would need to become a `require.ErrorIs(err, core.StoreError)`.
- **`CacheActive()` is unconditionally true for RESP** (`resp_cache.go:215`). The gRPC client nils itself on `Unavailable` and skips lookups until the reconnect loop succeeds — an implicit breaker. RESP has a health loop that already knows `connected`, but nothing consults it, so every relay pays the lookup budget against a dead backend for the whole outage. Gating `CacheActive` on the probe (with the 10 s cadence tightened during an outage) would make "degrades to misses" cost nothing after the first failure instead of 50 ms per relay.
- **Write-side bookkeeping is 2–3 sequential round trips per relay** (`engine.go:292-303`: `SetEntry`, then the chain-tip `EVAL`, plus the shared-tip `EVAL` when shared state is on). Over the gRPC path this was one RPC with in-memory bookkeeping on the server. Async, so not on the request path, but each write goroutine holds a pool connection for the sum, and at cross-region RTTs that is what saturates `pool-size` first. A `KVStore` batch-write, or letting the adapter pipeline `SET` + `EVAL`s, collapses it to one.
- **Cluster: the two lookup variants land on different shards.** `rel:f:` and `rel:t:` differ before the identity portion, so `GetEntries` in cluster mode is two shard round trips (concurrent, but two). A hash tag around `chainId:hash:block` co-locates them and would also allow `MGET`. Cheap to do now, since the key scheme becomes a compatibility surface the moment this ships.
- **Dangling-config policy diverges from #248.** Here `resp-cache-topology` without addresses is fatal (`resp_config.go:327`); #248 was changed during its review to *warn* on the equivalent shape for fleet-templated YAML. Both are defensible; the two cache features landing in the same release should agree.
- **Chain-tip freshness is now judged on the writer's clock** (`store.go:477` compares `time.Now()` against a deadline another router stamped). The in-memory store used one clock. With a 500 ms freshness window, skew between routers sharing a backend is the whole window — NTP-synced hosts are fine, but a note in the docs' shared-state section is warranted.

### 🟢 Non-blocking

- GitGuardian is still red and `mergeStateStatus` is `BLOCKED`; both findings are the placeholder fixtures the earlier review cleared. Needs a dismissal in the dashboard, not a code change.
- `GetSharedTip` logs at Info on every miss (`engine.go:75`). Pre-existing on the sidecar; embedded in the router it is now a per-relay Info line whenever shared state is on and the tip key is absent.
- The four new lanes honour the ownership-safety rule (no `killall`, labelled containers only); `init_smartrouter_eth_resp_cache.sh` is the only one that `docker rm -f`s by name, and it label-checks first.

### Checked and cleared

- Every `rpcss.cache` / `cf.cache` / `deps.cache` call site nil-checks the interface before probing, and the typed-nil `*Cache` methods are nil-receiver safe — the "no config → identical to `main`" claim holds at the call sites, not just in `SelectCacheBackend`.
- `Close()` ordering in `Stop` is after the client-facing drain, `Cache.Close` races an in-flight reconnect correctly (`closed` re-checked under the lock before installing), and `RespCache.Close` is `sync.Once`-guarded.
- `Purge` is prefix-scoped SCAN + single-key UNLINK, per-master under cluster; the prefix regex makes the glob safe.
- Auth-failure logging never carries the credential or the server reply; `RedactSecrets` on main does not reshape the fixed-phrase assertions.
- No customer names, no real credentials in the diff; the `datapass`/`sentinelpass` defaults in the sentinel lane are local-docker placeholders overridable by env.

### Verification (worktree at `7d930b8`, go 1.26.6)

| Check | Result |
| --- | --- |
| `go build ./...` | clean |
| `go vet ./...` (whole repo) | clean |
| `gofmt -l ./ecosystem ./protocol ./types` | clean |
| `go test -race ./ecosystem/cache/... ./protocol/performance/...` | ok (core, redisstore, performance incl. the degraded suite) |
| `go test ./protocol/rpcsmartrouter/...` | ok (91 s) |
| `go test ./protocol/metrics/... ./protocol/chainlib/` | ok |
| `golangci-lint run --new-from-rev origin/main` | 3 advisory: `forcetypeassert` (config_test.go:158), `nilnil` (config.go:101), and a `gofumpt` complaint at rpcsmartrouter_server_test.go:3452 — the first two match the PR body; the third is new and is a one-line format fix |
| Trial merge of #248 (`4b70b2c`) onto this head | 4 conflicting files (see the integration item) |

The docker-backed lanes (sentinel failover, cluster, mTLS, rotation) I did not run — they need a daemon and are opt-in by design. The claims they back are the ones I could not independently confirm here.

avitenzer:
Both passes addressed at `51800e2`, in five commits: the main merge (#248 landed, so the integration you both flagged is now done, not planned), the two dialer findings, the docs/UX corrections, and one fix-or-file item that the merge upgraded to a fix. Every code fix is pinned by a test verified to fail against the un-fixed code.

---

### The #248 integration — merged, with the semantic port (`be7c824`)

@nimrod-teich called the shape exactly: the mechanical merge produces 4 textual conflicts plus two traps a clean resolution would ship. Both are closed in the merge commit itself:

- **`IsNodeError`/`StatusCode` are ported through `core.Envelope`, `NewEnvelope`, `ToCacheReply`, `Engine.SetRelay`, and the RESP `storedEnvelope`** (additive JSON fields, no `v` bump — absent decodes to the legacy-unknown zero values, which is the cache wire convention `types/relay/cache_test.go` pins). Without the port, a RESP-backed cache reports every entry as `IsNodeError=false, StatusCode=0` and the secondary tier's backfill-eligibility silently degrades to the legacy `CACHED_ERROR` placeholder check. main's own `is_node_error_roundtrip` / `seen_block_validation` suites now run **through the engine** and fail with the threading removed (verified by removing it); your requested parity case exists as `TestParityEntryKindAndStatusRoundTrip`, green against both backends.
- **A trap the conflict list doesn't show:** main's `helpers_test.go` and this branch's `engine_roundtrip_test.go` both declared `setRelayForTest` with different signatures — git merges them *cleanly* into a package-level redeclaration, so `ecosystem/cache` compiles only after someone notices. Renamed to `roundTripSet` (main's helpers file is the declared one-home-per-helper spot).
- One combination the merge newly creates: `resp-cache` primary + `secondary-cache-be`. The secondary's backfill goes through the `CacheBackend` seam, so it works — and its "secondary without a primary: nothing backfills" advisory now receives the *effective* primary rather than keying on `cache-be` alone (it would have warned falsely there). No dedicated combo test yet; the interface seam plus the parity suite is the argument. Happy to add one if you want it in this PR.

### 🟠 TLS dial budget — fixed (`cea1807`), with a correction to the framing

The finding is real and the fix is what you proposed: `baseDialer` dials TLS through `tls.Dialer.DialContext` (context reaches the handshake; sooner of deadline and dial-timeout wins) and `dial-timeout` defaults to **500ms** instead of inheriting go-redis's 5s.

One layer of the impact statement didn't survive measurement, so correcting it here: **the caller of a command was already bounded.** Probed empirically against a black-holed TLS listener with `DialTimeout: 5s` and a 150ms caller deadline: `store.Ping` returned `context deadline exceeded` at **151ms in both states**; the *direct dialer call* took **151ms fixed vs 5.001s un-fixed**. So "~160 relays paying up to 5s each" wasn't the caller-visible shape — what the un-fixed path really cost was the attempt burning its full 5s underneath (`dialConn` runs the dialer synchronously): the pool slot stays held, retries serialize behind it, and the health probe can spend seconds per tick. That's why the regression test (`TestTLSHandshakeBoundedByCallerContext`) pins the dialer layer — the store-level latency doesn't discriminate (I watched a store-level version pass against the broken dialer). Fails at `5.001s is not less than 2s` on the reverted dialer.

### 🟠 Sentinel tracker — took the `DialHook` approach (`cea1807`), both halves of the residual

@AnnaR-prog's structural sketch is what shipped: `markDataDialsHook`, installed on the failover client's own hook chain, stamps a context mark; the sentinel tracker records **marked dials only**. Discrimination is by which *client* dials, never by address spelling, so runtime-discovered sentinels (the IP-vs-hostname mismatch that defeated the exclude list) can't pollute the tracker. `trackingDialerExcluding` and the exclude machinery are gone. `config.go`'s comment now describes a guarantee the code provides.

And because "it pins the mechanism, not the wiring" is how the last fix survived review: there are **two** tests (`51800e2`). The mechanism test covers the discovered-sentinel shape directly; `TestNewSentinelWiringDoesNotRecordControlPlaneDials` goes through the real `New()` — a fake sentinel accepts the TCP dial and never speaks RESP, and `ReadEndpoint` must stay empty. Verified failing both with the mark check dropped and (by construction) if `buildClient` stops installing the hook. The *positive* wiring half — data dials recorded through real construction — needs a live master and remains with the docker sentinel drill and the demo lane, which is one more reason those lanes are the standing run-me item.

### Fix-or-file: `StoreError` swallowing — upgraded to fixed (`286e2f1`)

You filed this as "matters once #248 lands." It landed, so it's fixed: `RespCache.GetEntry` now returns the joined store failure (semantic misses stay error-free), matching the gRPC *client* seam it replaces. The call site already degrades any error to a miss, and `ClassifyCacheLookupOutcome` reads through the `errors.Join` to `DeadlineExceeded` — so a RESP outage now labels `outcome="error"/"timeout"` instead of masquerading as `"miss"`. `TestRespCacheStoreFailureClassifiesAsErrorNotMiss` pins both halves; fails with the error swallowed again.

### Docs/UX (`2e5c45f`)

- `--shared-state`'s help text was wrong in both directions after this branch: "requires --cache-be" (the consistency seen-block now travels through either backend) while unconditionally promising the poll observations (the half RESP does *not* carry). It now says which half needs what.
- The example config's rollback path pointed at `127.0.0.1:20100` — the router container's own loopback, with no sidecar in the documented stack. It now names the cache overlay's `cache:20100` and the comment spells out the actual rollback move (drop the block, add `-f docker/docker-compose.cache.yml`); LOCAL-COMPOSE.md says the same.

### Dispositions on the rest

- **`CacheActive()` breaker, batched write-side bookkeeping, cluster hash-tag co-location, chain-tip clock-skew note, `GetSharedTip` per-miss Info log** — filing as follow-ups rather than growing this PR further; ticket keys to follow on MAG-2647.
- **Dangling-config policy divergence** (secondary-cache warns per its review; resp-cache fails fast): genuine policy question, not something I want to settle unilaterally inside a fix commit. My lean is that resp keeps fail-fast — an enabled-but-half-configured *primary* is a worse surprise than an unused tuning key — but if either of you wants the fleet-template argument applied uniformly, it's a small change and I'll make it. Say which.
- **gofumpt nit** at `rpcsmartrouter_server_test.go` — will fold into whichever commit lands next.
- **GitGuardian stays red by design of the check**: both findings live in historical commits (`sup3r-s3cret-value`, `sentinel-secret` — fixture literals against miniredis/temp files, confirmed no real credential material anywhere in the branch's 18 commits), so no push can clear it. It needs the two incidents dismissed in the dashboard (35807866, 35909064); squash-merge keeps the flagged commits off main regardless.

### Verification

Worktree at `51800e2` (full gate also run at `286e2f1`; `51800e2` adds only the wiring test):

| Check | Result |
| --- | --- |
| `go build ./...` / `go vet ./...` (whole repo, incl. tests) | clean |
| `gofmt -l ./ecosystem ./protocol ./config` | clean |
| `go test -count=1 ./...` | ok, all packages |
| `go test -race ./ecosystem/cache/... ./protocol/performance/...` | ok |
| `go mod tidy -diff` | no changes |
| Parity + selection (`-run '^(TestParity|TestSelectBackend)'`) | **38/38** (was 35 — the entry-kind case and its subtests) |

Every fix revert-verified: the port (three main-suite tests fail without the threading), the TLS dialer (`5.001s`), the mark check (mechanism *and* wiring tests), the StoreError return. The docker lanes (sentinel failover, cluster, mTLS, rotation) remain the run-me item — doubly so now that the sentinel tracker's positive half lives there.

avitenzer:
Docker lanes run at `51800e2` (docker 28.0.1, macOS host) — the standing run-me item from the reviews, plus the live proof of the sentinel tracker's positive half.

### The four opt-in drills — 6/6 PASS

| Drill | Result |
| --- | --- |
| `RESP_CACHE_TEST_TLS_DOCKER=1 … -run TestTLSDockerValkey` | **PASS** (10.2s) — real Valkey terminating TLS with `--tls-auth-clients yes`; mTLS round-trips, the certificate-less client is rejected |
| `RESP_CACHE_TEST_SENTINEL_DOCKER=1 … -run TestSentinelFailover` | **PASS** (9.2s) — primary killed mid-run, sentinels promote, the SAME `Store` resumes writes with no re-construction |
| `RESP_CACHE_TEST_CLUSTER_DOCKER=1 … -run TestClusterDocker` | **PASS** (4.2s) — three masters via one discovery endpoint, writes spread across slots, cross-slot pipelined `GetEntries`, prefix-scoped purge empties `sr:*` on every master while the foreign prefix survives in full |
| `RESP_CACHE_TEST_VALKEY_ADDR=… -run TestLiveRotation` | **PASS ×3** — the miniredis smoke, the real-Valkey acceptance (server-side `CLIENT LIST` shows the ORIGINAL connection id re-authenticated as the new ACL user — in-place, no reconnect), and the production-wiring variant where only the `password-file` watcher `New()` owns notices the rewritten secret |

### Sentinel demo lane — the tracker's positive half, live

My previous comment noted the wiring test covers only the negative half (control-plane dials never recorded) and that the positive half — data dials recorded through the real construction — lives here. Ran it:

```
sentinels agree: master = 10.0.0.10:63811        (setup SMOKE PASS; replication OK)

# cached hit BEFORE failover — the header names the DATA node, not a sentinel (:26390-92):
Lava-Provider-Address: Cached
Lava-Cache-Backend: 10.0.0.10:63811

# --failover: promoted 10.0.0.10:63811 -> 10.0.0.10:63812; the router was never restarted
# and kept serving. The header follows the promotion:
Lava-Provider-Address: Cached
Lava-Cache-Backend: 10.0.0.10:63812

smartrouter_resp_cache_connected: dipped to 0 inside the kill window, self-healed to 1
within one probe tick. --recover: the old primary rejoined as a REPLICA of the new
master — no split brain, no manual step.
```

So the full `markDataDialsHook` pair is now evidenced at every level: mechanism (unit), wiring-negative (real `New()` against a fake sentinel), and wiring-positive (real router, real quorum, header naming the master and flipping across a failover — with three sentinels dialling throughout and never appearing in the header).

With this, every verification row in the PR is either CI-checked or has a named run on record. Remaining before merge is unchanged: the GitGuardian dashboard dismissal (incidents 35807866 / 35909064), your re-review of the five round-2 commits, and the dangling-config policy answer.

PR #262, opened 2026-08-10
feat(provideroptimizer): configurable upstream selection policy (mode + priority presets)

Tomelia1999:
Jira ticket: MAG-2780

Single consolidated PR for the former stack — #263 and #264 have been merged up into this branch, so this now carries the whole feature set. Implements FAILOVER-TASKS section 1: let the operator decide **what** the router optimises for when choosing who serves a request, and **how** the winner is picked.

Four commits, reviewable independently and in order.

---

## 1. `--qos-selection-mode` — how the winner is picked

```
--qos-selection-mode weighted_random|best   (default: weighted_random)
```

`best` always routes to the highest-scoring upstream. The default is unchanged, so **production behaviour is identical unless the flag is set**.

### Why the split lands where it does

Only one step in the selection flow is mode-dependent. Scoring, the stats payload, the metrics emit, and the `wantedProviders` sampling loop are all shared:

| layer | mode-dependent? |
|---|---|
| CSM branch + `wantedProviders` loop | no |
| `ChooseUpstreamWithStats` | no |
| `CalculateUpstreamScores` | no |
| **turning `[]UpstreamScore` into one address** | **yes — the whole delta** |

So `SelectUpstreamWithStats` is split along that seam rather than forked wholesale:

- `pickWeightedIndex` — today's cumulative-probability draw, verbatim
- `pickBestIndex` — the new argmax
- `buildSelectionResult` — the debug breakdown and `SelectionStats`, previously duplicated across **four** return paths

Hoisting the reporting is what makes the second mode cost ~12 lines. `SelectUpstreamWithStats` went from 108 lines to ~50 across four focused functions.

The guards stay mode-independent. The all-zero-scores path stays *uniform* for both modes — though note (thanks @avitenzer) that it is **unreachable at any sane config**: `CalculateScore` floors every candidate at `minSelectionChance`, so `totalScore` is zero only under `--qos-min-selection-chance 0`, which `NewUpstreamSelector` does not reject. Degraded-chain behaviour is decided by the tie-break below, not by that guard. The code comment has been corrected to say so.

### Tie-breaking is deterministic, and that is the point

**Changed since the first review.** `pickBestIndex` originally broke ties uniformly at random via reservoir sampling. It now breaks them **first-wins, with no RNG draw at all**.

`CalculateScore` collapses every upstream below `MinAcceptableAvailability` to *exactly* `minSelectionChance` (MAG-2237), so an all-tied candidate list is the routine state once a chain degrades — not a corner case. Randomising there meant every pick on a bad chain was a coin toss, which is `best` mode quietly reverting to a lottery in precisely the situation where the routing most needs to be explainable. Section 1 of FAILOVER-TASKS asks for the opposite in bold: *"the answer does not change between two identical requests."*

The starvation objection does not survive contact with this mode. `minSelectionChance` is a **lottery** mechanism — it buys a floored upstream a slice of the weighted draw. Under `best` there is no draw, so a floored upstream beside any higher-scoring peer already receives exactly zero traffic. The floor stopped protecting anyone the moment the operator asked for "always the best"; randomising the all-tied case never gave that back, it only removed the determinism the mode exists to provide.

Two consequences, both accepted deliberately:

- When everything is floored the candidates are equally dead, and the `wantedProviders` loop still walks the whole list within a request — only the *first* attempt concentrates.
- An upstream with no organic traffic recovers via the prober, not via routing. Already true of any non-leader under `best`.

"Earliest" carries no ranking meaning: `validAddresses` comes from iterating a Go map, so the order is arbitrary but **stable within an epoch** (which is what determinism needs) and differs per pod, so a fleet still spreads the all-tied case across routers.

Dropping the draw also removes the cost @avitenzer measured. Against the crypto-backed global RNG (183ns/draw, 3 allocs, behind a process-wide mutex shared with per-request GUID generation):

| candidates | all-tied, before | after |
|---|---|---|
| 2 | ~140ns, 3 allocs | 2.2ns, 0 allocs |
| 5 | ~630ns, 12 allocs | 4.2ns, 0 allocs |
| 10 | ~1440ns, 27 allocs | 9.1ns, 0 allocs |
| 20 | ~3040ns, 57 allocs | 14.4ns, 0 allocs |

Under `RunParallel` at 10 tied candidates the gap widens from ~160x to **~224x** — that part is lock contention, not CPU.

### Observability

`SelectionStats` gains `Mode`, surfaced in the selection-stats header ahead of RNG:

```
[provider1: 0.900, 0.800, 0.700, 0.100, 0.750] | Mode: best | RNG: 0.000000 | Selected: provider1
```

An `RNGValue` of `0` is otherwise ambiguous — it means *no draw happened* under `best`, but *single candidate* or *all scores zero* under `weighted_random`. `/debug/runtime-config` reports the mode per chain alongside the weights.

### Drive-by: removed `ChooseBestProvider` / `ChooseBestProviderWithStats`

Despite the name, these were semantically identical to `ChooseUpstream` / `ChooseUpstreamWithStats` — same scoring, same weighted draw, same return shape, differing only in two trace strings. "Best" never meant highest-scoring; sticky sessions were getting an ordinary weighted draw.

The sticky-session path now calls `ChooseUpstreamWithStats` directly with no behaviour change, and the non-stats variant had no callers at all.

---

## 2 + 4. `WeightedSelector` → `UpstreamSelector`

Once commit 1 lands, `WeightedSelector` names a type that does weighted random selection *and* highest-score selection. Scoring and reporting are shared; only the final pick varies by mode. The name promises one of the two.

Kept out of the feature commits deliberately — a repo-wide mechanical rename mixed into a feature diff makes the feature impossible to review.

### Why `Upstream` and not `Endpoint`

Commit 2 renamed to `ProviderSelector`; @nimrod-teich requested changes — *"We want to move away from the 'provider' terminology."*

`Endpoint` was the obvious candidate and is **wrong**, because `provider` and `endpoint` are not synonyms in this codebase — they're two levels of a real hierarchy:

```
ConsumerSessionsWithProvider     identity: address + stake
  └── Endpoints []*Endpoint      network address (host:port)
        └── Connections []*...   the actual connections
```

The selector ranks the **identity** level and returns one address; the session manager then picks an endpoint within it. So `EndpointSelector` would name the wrong level, and `ConsumerSessionsWithEndpoint{ Endpoints []*Endpoint }` is nonsense.

`Upstream` names that level without colliding with either `Endpoint` or `NodeUrl`, and is the ordinary vocabulary for the thing a router forwards to.

### The type's own API is renamed with it

A type called `UpstreamSelector` whose method is `SelectProviderWithStats` returning `[]UpstreamScore` just relocates the inconsistency:

| before | after |
|---|---|
| `WeightedSelector` / `ProviderSelector` | `UpstreamSelector` |
| `WeightedSelectorConfig` | `UpstreamSelectorConfig` |
| `NewWeightedSelector` | `NewUpstreamSelector` |
| `DefaultWeightedSelectorConfig` | `DefaultUpstreamSelectorConfig` |
| `ConfigureWeightedSelector` | `ConfigureUpstreamSelector` |
| `GetWeightedSelectorConfig` | `GetUpstreamSelectorConfig` |
| `UpdateWeightedSelectorStrategy` | `UpdateUpstreamSelectorStrategy` |
| `SelectProvider` / `SelectProviderWithStats` | `SelectUpstream` / `SelectUpstreamWithStats` |
| `ChooseProvider` / `ChooseProviderWithStats` | `ChooseUpstream` / `ChooseUpstreamWithStats` |
| `ProviderScore` / `ProviderScoreDetails` | `UpstreamScore` / `UpstreamScoreDetails` |
| `SelectionStats.ProviderScores` | `SelectionStats.UpstreamScores` |
| `CalculateProviderScores` | `CalculateUpstreamScores` |
| `weighted_selector.go` | `upstream_selector.go` (and `_test`) |

**309 insertions / 309 deletions** — perfectly balanced, which is what a pure rename looks like.

### Scope: the selector API surface only

`ProviderOptimizer`, `ProviderStakeCache`, the `provideroptimizer` package name, and the provider vocabulary throughout `lavasession`, `rpcsmartrouter` and `metrics` are **deliberately untouched**. That migration is ~280 identifiers across 19 packages (~6,900 mentions) and needs its own ticket — and a decision on the external contracts below.

### No external contract moves

All **101** externally-observable surfaces were snapshotted before and after the rename and diffed **byte-identical**:

| surface | count | examples |
|---|---|---|
| HTTP headers | 12 | `Lava-Provider-Address`, `lava-select-provider`, `Lava-Errored-Providers` |
| Prometheus metric names | 8 | `smartrouter_csm_blocked_providers`, `provider_stake` |
| debug URL paths | — | `/debug/provider-scores` |
| json/yaml/mapstructure tags | 5 | `provider-name`, `provider_stake`, `unresponsive_providers` |
| cobra flag / viper keys | ~6 | `static-providers-list`, `max-sessions-per-provider` |

Three packages are excluded outright to keep it that way — verified untouched via `git status`:

- **`protocol/metrics`** — owns `ProviderSelectionScores`, `SetProviderSelected` and the metric-name literals. Note `allProviderScores` there *does* contain the substring `ProviderScore`, so a repo-wide substitution would have renamed the variable and left its type behind.
- **`protocol/relaycore`** — its only `SelectProvider` mention is `TestArchiveAddDoesNotPreserveSelectProvider`, a guard on the `lava-select-provider` header resetting between retries. Unrelated to this API, and its own comment warns that regressing it breaks `test_2_1_one_provider_down_retry_to_next` in automation.
- **`debug_provider_scores_test.go`** — pins the `/debug/provider-scores` contract the automation suite reads.

### Deliberately NOT renamed

These genuinely describe the weighted draw and stay correct:

- `pickWeightedIndex` — this one really is the weighted selection
- `SelectionModeWeightedRandom` / the `weighted_random` flag value
- `updateDecayingWeightedAverage` — unrelated EWMA helper
- `"weighted composite score"` / `"weighted contributions"` comments — the composite genuinely is a weighted sum of the QoS axes

**Note for reviewers:** three log-message strings changed (the weight-validation warnings, `"invalid weighted selector weight"` → `"invalid upstream selector weight"`). If anything greps those, it needs updating — though they only fire on invalid startup config.

## 3. `--qos-selection-priority` — what to optimise for

```
```

|---|---|---|---|---|

**Dominant, not exclusive.** The chosen axis takes 0.70, but availability keeps real weight in every preset so a fast-but-flaky upstream cannot outrank a fast-and-solid one on latency alone. `balanced` is byte-identical to today's defaults — it's the flag's default value, so anything else would silently move every existing deployment.

### Why the axis presets zero stake

`CalcWeightsByStake` hands every static upstream an identical weight (`maxWeight * 10`, and `maxWeight` stays 1 in an all-static deployment). So the stake term is **the same constant for every candidate**:

- under `weighted_random` it actively flattens the lottery — everyone sits on the same pedestal, compressing the probability ratio between a great upstream and a mediocre one

### Composes with `--qos-selection-mode`, doesn't duplicate it

| flag | decides |
|---|---|

### Precedence

### Drive-by fix the new tests surfaced

`ParseSelectionMode` (from commit 1) rejected the empty string, which tied startup correctness to the flag being viper-bound: any call site that hadn't bound it reads `""` and aborts with a confusing `invalid selection mode:` instead of using the default. **Empty now means "not specified"** and resolves to the default in both parsers. Genuinely wrong values (`bset`, `quickest`) still hard-fail — that's the property that actually matters.

### `--strategy` deliberately untouched

Measured: its only scoring effect is an exponent nudge worth **at most +0.015 on the composite** (typically +0.006), and it does not move the weights at all. `StrategyBalanced` — the default, and the only reachable value for YAML users since the flag has no `BindPFlag` — is a genuine no-op in the switch. So it cannot interfere with this flag.

---

## Verification

- `go build ./...` — clean
- `go vet ./...` — clean repo-wide
- `make test` — **34 packages, zero failures** (~2 min wall)
- `--help` output verified for both flags

Rebased onto current `main` (`498c298`) with zero conflicts. **Correction to an earlier version of this section:** it claimed vet and the full suite were clean when they were not. `protocol/rpcsmartrouter/node_error_demotion_mag2156_test.go` still called `po.ChooseProvider` after the rename, so that test package did not compile. `go build ./...` was genuinely clean — the only caller was in a `_test.go`, which `go build` does not compile — and repo CI has no `go test` job and a `lint` job that cannot fail, so nothing contradicted the claim. Fixed in `ec0f0b6`; the CI gap is being tracked separately.

Tests include the end-to-end proof that the preset is more than a table of numbers: given the same two upstreams, `fastest` and `most-reliable` pick **different** ones. Plus the sum-to-1.0 invariant (a drifted preset would be silently renormalised by `NewUpstreamSelector`), the `balanced == today's defaults` backward-compat pin, and both halves of the precedence rule.

For commit 1: best-scorer-placed-last, deterministic tie-break across three floor-collapsed upstreams (one selector deliberately left unseeded, so a reintroduced draw cannot hide behind `math/rand`), earliest-maximum-wins on a `[0.5, 0.9, 0.5, 0.9]` list (a `>=` argmax would walk the pick to the last tied element), leader tracking, mode stamping across all six mode × short-circuit combinations, header formatting, parse/reject coverage, and a regression pin that the default mode still spreads 80/20. `TestDebugRuntimeConfig_SmartRouter_PerChainOptimizer` gives its two chains *different* modes, so the field is proven to travel per-chain rather than from a global default.

For the rename, verified mechanical beyond the test suite: the same transform was applied to the pre-rename tree and diffed against this one. Only the six prose edits above remain, and the test-file diff is empty — so no logic moved.

This branch is behind current `main` but rebases onto it with **zero conflicts**, including `protocol/rpcsmartrouter/rpcsmartrouter.go`, which moved on `main` since the stack was cut (MAG-2707, MAG-2649). Verified by rebasing and running the full suite green on top of current `main`.

## Notes for review

- The selection-stats header gained a `| Mode: <mode> |` segment. No test parses it, but external tooling reading it positionally would need a look.
- The mode and priority are global across chains (one flag → every chain's optimizer), consistent with the existing `qos-*` weight flags. Going per-chain later is config plumbing only; the selection seam does not move.

Consolidates and replaces #263 and #264 (both merged into this branch).

### DIFF
--- a/protocol/provideroptimizer/upstream_selector.go
+++ b/protocol/provideroptimizer/upstream_selector.go
@@ -183,6 +183,28 @@ func NewUpstreamSelector(config UpstreamSelectorConfig) *UpstreamSelector {
 		totalWeight = 1.0 // default weights sum to 1.0
 	}
 
+	// MinSelectionChance is a floor applied to every composite, so it only makes sense as
+	// a score in [0, 1]. Nothing validated it before, and out-of-range values fail in ways
+	// that look like a scoring bug rather than a config error:
+	//
+	//   > 1  — CalculateScore floors every candidate above 1.0 and then clamps back to
+	//          1.0, so EVERY upstream scores exactly 1.0. Ranking is switched off entirely
+	//          and silently: a 10ms upstream and a 25s upstream come out identical.
+	//   < 0  — the floor stops floor-ing; a negative composite is meaningless and would
+	//          drag totalScore below zero, diverting live traffic into the all-zero guard.
+	//
+	// 0 stays legal — it is the deliberate "no floor at all" choice — and is safe now that
+	// the all-zero guard keeps SelectionModeBest deterministic.
+	if math.IsNaN(config.MinSelectionChance) || math.IsInf(config.MinSelectionChance, 0) ||
+		config.MinSelectionChance < 0 || config.MinSelectionChance > 1 {
+		defaultChance := DefaultUpstreamSelectorConfig().MinSelectionChance
+		utils.LavaFormatWarning("invalid min selection chance, must be finite and within [0, 1], using default", nil,
+			utils.LogAttr("minSelectionChance", config.MinSelectionChance),
+			utils.LogAttr("usingInstead", defaultChance),
+		)
+		config.MinSelectionChance = defaultChance
+	}
+
 	if math.Abs(totalWeight-1.0) > 0.001 {
 		utils.LavaFormatWarning("endpoint selector weights do not sum to 1.0, normalizing",
 			nil,
@@ -672,20 +694,32 @@ func (ws *UpstreamSelector) SelectUpstreamWithStats(
 	}
 
 	if totalScore <= 0 {
-		// Defensive only — unreachable at any sane configuration, in BOTH modes.
+		// Reachable only when minSelectionChance is 0. CalculateScore floors every
+		// candidate at ws.minSelectionChance — including the availability-dead collapse —
+		// so with the default floor a score arriving here is never below it and totalScore
+		// is never zero. Set --qos-min-selection-chance 0 and that stops being true: on a
+		// chain where every upstream has gone unhealthy, every composite collapses to 0.
 		//
-		// CalculateScore floors every candidate at ws.minSelectionChance (default 0.01),
-		// including the availability-dead collapse, so a score reaching this function is
-		// never below that floor and totalScore is never zero. The single way in is
-		// --qos-min-selection-chance 0, which NewUpstreamSelector does not reject: it
-		// validates the four weights but leaves MinSelectionChance untouched.
+		// An all-zero list is an N-way tie with no signal to rank on, so each mode answers
+		// it the way it answers every other tie. SelectionModeBest must NOT draw here: it
+		// is sold on "the answer does not change between two identical requests", and a
+		// draw would break that silently — the log line still says selection_mode=best.
+		// Reported by @avitenzer, who reproduced it as 300 identical requests splitting
+		// roughly evenly across three candidates.
 		//
-		// So this guard is NOT what keeps a degraded chain sane. Once availability
-		// collapses, every candidate ties at exactly minSelectionChance and the pick is
-		// resolved by pickBestIndex (first-wins) or pickWeightedIndex (a flat draw) —
-		// not here. Do not reason about the dead-chain case from this branch.
-		utils.LavaFormatWarning("all provider scores are zero, using uniform selection", nil)
-		return ws.buildSelectionResult(ctx, providerScores, scoreDetails, ws.rng.Intn(len(providerScores)), 0.0, totalScore)
+		// This guard is still not what keeps a degraded chain sane at the DEFAULT floor.
+		// There every candidate ties at exactly minSelectionChance, totalScore is positive,
+		// and the pick is resolved by pickBestIndex or pickWeightedIndex — not here.
+		selectedIndex := 0
+		if ws.selectionMode != SelectionModeBest {
+			selectedIndex = ws.rng.Intn(len(providerScores))
+		}
+		utils.LavaFormatWarning("all provider scores are zero, falling back to a flat pick", nil,
+			utils.LogAttr("selectionMode", ws.selectionMode.String()),
+			utils.LogAttr("numCandidates", len(providerScores)),
+			utils.LogAttr("hint", "every upstream scored zero — reachable only with --qos-min-selection-chance 0"),
+		)
+		return ws.buildSelectionResult(ctx, providerScores, scoreDetails, selectedIndex, 0.0, totalScore)
 	}
 
 	var (
@@ -739,10 +773,18 @@ func (ws *UpstreamSelector) pickWeightedIndex(providerScores []UpstreamScore, to
 // address. That objection does not survive contact with this mode. minSelectionChance is a
 // *lottery* mechanism — it buys a floored provider a slice of the weighted draw. Under
 // SelectionModeBest there is no draw, so a floored provider sitting beside any
-// higher-scoring peer already receives exactly zero traffic. The floor stopped protecting
-// anyone the moment the operator asked for "always the best"; randomising the all-tied case
+// higher-scoring peer already receives exactly zero traffic. Randomising the all-tied case
 // does not give that protection back, it only removes the determinism the mode exists to
 // provide.
+//
+// That reasoning is about the TIE-BREAK, and it is not an argument for turning the floor
+// off. Do not read it as one, and do not set --qos-min-selection-chance 0 on the strength
+// of it: the floor is what keeps every composite positive, and at 0 a fully degraded chain
+// scores zero across the board and lands in the all-zero guard in SelectUpstreamWithStats.
+// An earlier revision of this comment did read as such an argument, and @avitenzer pointed
+// out it walks the operator into the one branch that used to reintroduce a draw. That guard
+// is deterministic under Best now, so the trap is closed — but the floor still earns its
+// keep under weighted_random, where it is the whole anti-starvation mechanism.
 //
 // Two consequences worth knowing, neither of which is a regression from this choice:
 //
--- a/protocol/provideroptimizer/upstream_selector.go
+++ b/protocol/provideroptimizer/upstream_selector.go
@@ -672,9 +672,18 @@ func (ws *UpstreamSelector) SelectUpstreamWithStats(
 	}
 
 	if totalScore <= 0 {
-		// Fallback to uniform random selection if all scores are zero. SelectionModeBest
-		// lands here too: with no signal to rank on, an argmax would be an arbitrary
-		// N-way tie, so uniform is the honest answer for both modes.
+		// Defensive only — unreachable at any sane configuration, in BOTH modes.
+		//
+		// CalculateScore floors every candidate at ws.minSelectionChance (default 0.01),
+		// including the availability-dead collapse, so a score reaching this function is
+		// never below that floor and totalScore is never zero. The single way in is
+		// --qos-min-selection-chance 0, which NewUpstreamSelector does not reject: it
+		// validates the four weights but leaves MinSelectionChance untouched.
+		//
+		// So this guard is NOT what keeps a degraded chain sane. Once availability
+		// collapses, every candidate ties at exactly minSelectionChance and the pick is
+		// resolved by pickBestIndex (first-wins) or pickWeightedIndex (a flat draw) —
+		// not here. Do not reason about the dead-chain case from this branch.
 		utils.LavaFormatWarning("all provider scores are zero, using uniform selection", nil)
 		return ws.buildSelectionResult(ctx, providerScores, scoreDetails, ws.rng.Intn(len(providerScores)), 0.0, totalScore)
 	}
@@ -716,26 +725,42 @@ func (ws *UpstreamSelector) pickWeightedIndex(providerScores []UpstreamScore, to
 	return len(providerScores) - 1, randomValue
 }
 
-// pickBestIndex returns the index of the highest-scoring provider.
+// pickBestIndex returns the index of the highest-scoring provider. Ties go to the earliest
+// candidate — first-wins, with no randomness and no RNG draw at all.
 //
-// Ties are broken uniformly at random via one-pass reservoir sampling rather than
-// first-wins. This is not defensive polish: CalculateScore collapses every provider whose
-// availability has fallen below score.MinAcceptableAvailability to exactly
-// minSelectionChance, so exact N-way ties are routine on a degraded chain — a first-wins
-// scan would pin all traffic onto whichever address happens to sort first in the pairing
-// list, which is precisely the starvation the floor exists to prevent.
+// That is the whole point of the mode: FAILOVER-TASKS section 1 asks for "the order is
+// fixed, not random ... the answer does not change between two identical requests". A
+// random tie-break would break exactly that, and would break it on a degraded chain, which
+// is precisely when an operator most needs the routing to be explainable.
+//
+// The tempting objection is MAG-2237 starvation: CalculateScore collapses every provider
+// below score.MinAcceptableAvailability to exactly minSelectionChance, so exact N-way ties
+// are routine once a chain degrades, and first-wins then pins every first attempt onto one
+// address. That objection does not survive contact with this mode. minSelectionChance is a
+// *lottery* mechanism — it buys a floored provider a slice of the weighted draw. Under
+// SelectionModeBest there is no draw, so a floored provider sitting beside any
+// higher-scoring peer already receives exactly zero traffic. The floor stopped protecting
+// anyone the moment the operator asked for "always the best"; randomising the all-tied case
+// does not give that protection back, it only removes the determinism the mode exists to
+// provide.
+//
+// Two consequences worth knowing, neither of which is a regression from this choice:
+//
+//   - When every provider is floored they are all equally dead, and the wantedProviders
+//     loop in getValidProviderAddresses walks the whole list anyway, so each one is still
+//     tried within a single request. Only the *first* attempt concentrates.
+//   - A provider that receives no organic traffic recovers via the proactive prober, not
+//     via routing. That is already true of any non-leader under this mode.
+//
+// "Earliest" carries no ranking meaning: validAddresses is built by iterating
+// csm.pairingAddresses, a Go map, so the order is an arbitrary permutation fixed at reset
+// time. It is stable within an epoch — which is what determinism requires — and differs
+// between pods, so a fleet still spreads the all-tied case across routers.
 func (ws *UpstreamSelector) pickBestIndex(providerScores []UpstreamScore) int {
-	best, tied := 0, 1
+	best := 0
 	for i := 1; i < len(providerScores); i++ {
-		switch {
-		case providerScores[i].SelectionWeight > providerScores[best].SelectionWeight:
-			best, tied = i, 1
-		case providerScores[i].SelectionWeight == providerScores[best].SelectionWeight:
-			tied++
-			// Replace with probability 1/tied → uniform across all maxima
-			if ws.rng.Intn(tied) == 0 {
-				best = i
-			}
+		if providerScores[i].SelectionWeight > providerScores[best].SelectionWeight {
+			best = i
 		}
 	}
 
--- a/protocol/provideroptimizer/upstream_selector.go
+++ b/protocol/provideroptimizer/upstream_selector.go
@@ -51,14 +51,14 @@ func (m *mutexRandomizer) Intn(n int) int {
 	return m.rng.Intn(n)
 }
 
-// ProviderSelector scores providers on composite QoS and picks one of them. It replaces
+// UpstreamSelector scores providers on composite QoS and picks one of them. It replaces
 // the tier-based selection system with a continuous, score-based approach where providers
 // are ranked by overall quality without artificial tier boundaries.
 //
 // Scoring is shared by every selection policy; only the final pick varies, governed by
 // selectionMode — a weighted random draw (the default) or the highest scorer. The type was
 // named WeightedSelector while the weighted draw was the only policy.
-type ProviderSelector struct {
+type UpstreamSelector struct {
 	// Configuration weights for different QoS metrics (should sum to 1.0)
 	availabilityWeight float64 // Default: 0.3 (30% weight)
 	latencyWeight      float64 // Default: 0.3 (30% weight)
@@ -85,8 +85,8 @@ type ProviderSelector struct {
 	adaptiveSyncGetter    func() (p10, p90 float64) // Function to get adaptive P10-P90 bounds for sync
 }
 
-// ProviderScore represents a provider's calculated scores for selection
-type ProviderScore struct {
+// UpstreamScore represents a provider's calculated scores for selection
+type UpstreamScore struct {
 	Address         string  // Provider address
 	CompositeScore  float64 // Normalized 0-1 composite score
 	SelectionWeight float64 // Score adjusted by strategy and stake
@@ -94,7 +94,7 @@ type ProviderScore struct {
 
 // SelectionStats contains detailed information about provider selection for debugging
 type SelectionStats struct {
-	ProviderScores   []ProviderScoreDetails // Scores for all candidates
+	UpstreamScores   []UpstreamScoreDetails // Scores for all candidates
 	RNGValue         float64                // Random number used for selection
 	SelectedProvider string                 // The provider that was selected
 	// Mode is the policy that produced this pick. It disambiguates RNGValue, which is
@@ -104,8 +104,8 @@ type SelectionStats struct {
 	Mode SelectionMode
 }
 
-// ProviderScoreDetails contains detailed scoring information for a single provider
-type ProviderScoreDetails struct {
+// UpstreamScoreDetails contains detailed scoring information for a single provider
+type UpstreamScoreDetails struct {
 	Address      string  // Provider address
 	Availability float64 // Availability score (0-1)
 	Latency      float64 // Latency score (0-1)
@@ -114,8 +114,8 @@ type ProviderScoreDetails struct {
 	Composite    float64 // Combined QoS score (0-1)
 }
 
-// ProviderSelectorConfig holds configuration options for creating a ProviderSelector
-type ProviderSelectorConfig struct {
+// UpstreamSelectorConfig holds configuration options for creating a UpstreamSelector
+type UpstreamSelectorConfig struct {
 	AvailabilityWeight    float64
 	LatencyWeight         float64
 	SyncWeight            float64
@@ -129,9 +129,9 @@ type ProviderSelectorConfig struct {
 	AdaptiveSyncGetter    func() (p10, p90 float64) // Phase 2: Function to get adaptive P10-P90 bounds for sync
 }
 
-// DefaultProviderSelectorConfig returns a configuration with balanced default weights
-func DefaultProviderSelectorConfig() ProviderSelectorConfig {
-	return ProviderSelectorConfig{
+// DefaultUpstreamSelectorConfig returns a configuration with balanced default weights
+func DefaultUpstreamSelectorConfig() UpstreamSelectorConfig {
+	return UpstreamSelectorConfig{
 		AvailabilityWeight: 0.3,  // Availability remains critical (30%)
 		LatencyWeight:      0.3,  // Latency shares equal emphasis (30%)
 		SyncWeight:         0.2,  // Sync is third priority (20%)
@@ -142,15 +142,15 @@ func DefaultProviderSelectorConfig() ProviderSelectorConfig {
 	}
 }
 
-// NewProviderSelector creates a new ProviderSelector with the given configuration
-func NewProviderSelector(config ProviderSelectorConfig) *ProviderSelector {
+// NewUpstreamSelector creates a new UpstreamSelector with the given configuration
+func NewUpstreamSelector(config UpstreamSelectorConfig) *UpstreamSelector {
 	// Validate and normalize weights
 	//
 	// Important: we only want to fallback the *weights* if they are invalid, but keep
 	// other user choices (Strategy / SelectionMode / MinSelectionChance) intact.
 	validateWeight := func(name string, w float64) bool {
 		if math.IsNaN(w) || math.IsInf(w, 0) || w < 0 {
-			utils.LavaFormatWarning("invalid provider selector weight, must be finite and >= 0",
+			utils.LavaFormatWarning("invalid endpoint selector weight, must be finite and >= 0",
 				nil,
 				utils.LogAttr("weightName", name),
 				utils.LogAttr("weight", w),
@@ -167,7 +167,7 @@ func NewProviderSelector(config ProviderSelectorConfig) *ProviderSelector {
 
 	totalWeight := config.AvailabilityWeight + config.LatencyWeight + config.SyncWeight + config.StakeWeight
 	if !weightsValid || math.IsNaN(totalWeight) || math.IsInf(totalWeight, 0) || totalWeight <= 0 {
-		utils.LavaFormatWarning("provider selector weights sum to zero/negative or contain invalid values, using default weights", nil,
+		utils.LavaFormatWarning("endpoint selector weights sum to zero/negative or contain invalid values, using default weights", nil,
 			utils.LogAttr("totalWeight", totalWeight),
 			utils.LogAttr("availabilityWeight", config.AvailabilityWeight),
 			utils.LogAttr("latencyWeight", config.LatencyWeight),
@@ -175,7 +175,7 @@ func NewProviderSelector(config ProviderSelectorConfig) *ProviderSelector {
 			utils.LogAttr("stakeWeight", config.StakeWeight),
 		)
 
-		defaultCfg := DefaultProviderSelectorConfig()
+		defaultCfg := DefaultUpstreamSelectorConfig()
 		config.AvailabilityWeight = defaultCfg.AvailabilityWeight
 		config.LatencyWeight = defaultCfg.LatencyWeight
 		config.SyncWeight = defaultCfg.SyncWeight
@@ -184,7 +184,7 @@ func NewProviderSelector(config ProviderSelectorConfig) *ProviderSelector {
 	}
 
 	if math.Abs(totalWeight-1.0) > 0.001 {
-		utils.LavaFormatWarning("provider selector weights do not sum to 1.0, normalizing",
+		utils.LavaFormatWarning("endpoint selector weights do not sum to 1.0, normalizing",
 			nil,
 			utils.LogAttr("totalWeight", totalWeight),
 			utils.LogAttr("availabilityWeight", config.AvailabilityWeight),
@@ -199,7 +199,7 @@ func NewProviderSelector(config ProviderSelectorConfig) *ProviderSelector {
 		config.StakeWeight /= totalWeight
 	}
 
-	return &ProviderSelector{
+	return &UpstreamSelector{
 		availabilityWeight:    config.AvailabilityWeight,
 		latencyWeight:         config.LatencyWeight,
 		syncWeight:            config.SyncWeight,
@@ -215,9 +215,9 @@ func NewProviderSelector(config ProviderSelectorConfig) *ProviderSelector {
 	}
 }
 
-// SetDeterministicSeed sets a specific seed for the provider selector's RNG
+// SetDeterministicSeed sets a specific seed for the endpoint selector's RNG
 // This should ONLY be used for testing to ensure deterministic selection
-func (ws *ProviderSelector) SetDeterministicSeed(seed int64) {
+func (ws *UpstreamSelector) SetDeterministicSeed(seed int64) {
 	ws.rng = &mutexRandomizer{
 		rng: stdrand.New(stdrand.NewSource(seed)),
 	}
@@ -226,7 +226,7 @@ func (ws *ProviderSelector) SetDeterministicSeed(seed int64) {
 // CalculateScore computes a composite score for a provider based on QoS metrics and stake.
 // stake and totalStake are raw stake amounts (e.g. in ulava).
 // Returns a normalized score between 0 and 1, where higher is better
-func (ws *ProviderSelector) CalculateScore(
+func (ws *UpstreamSelector) CalculateScore(
 	qos *pairingtypes.QualityOfServiceReport,
 	stake float64,
 	totalStake float64,
@@ -336,7 +336,7 @@ func (ws *ProviderSelector) CalculateScore(
 //   - ✅ Zero complexity, no state, no memory overhead
 //   - ✅ Clear business rule: < 80% availability = unacceptable
 //   - ✅ Stable and predictable
-func (ws *ProviderSelector) normalizeAvailability(availability float64) float64 {
+func (ws *UpstreamSelector) normalizeAvailability(availability float64) float64 {
 	// Phase 1: Simple Rescaling
 	const minAcceptable = score.MinAcceptableAvailability // currently 0.80
 	const maxAvailability = 1.0
@@ -388,7 +388,7 @@ func (ws *ProviderSelector) normalizeAvailability(availability float64) float64 
 // Phase 1 (Fallback):
 //   - Uses fixed maximum expected latency (score.WorstLatencyScore = 30s)
 //   - Formula: normalized = 1 - (latency / maxLatency)
-func (ws *ProviderSelector) normalizeLatency(latency float64) float64 {
+func (ws *UpstreamSelector) normalizeLatency(latency float64) float64 {
 	// Phase 2: Adaptive P10-P90 normalization (if enabled)
 	if ws.useAdaptiveLatencyMax && ws.adaptiveLatencyGetter != nil {
 		p10, p90 := ws.adaptiveLatencyGetter()
@@ -483,7 +483,7 @@ func (ws *ProviderSelector) normalizeLatency(latency float64) float64 {
 // Phase 1 (Fallback):
 //   - Uses fixed maximum expected sync lag (score.WorstSyncScore = 1200s)
 //   - Formula: normalized = 1 - (syncLag / maxSyncLag)
-func (ws *ProviderSelector) normalizeSync(syncLag float64) float64 {
+func (ws *UpstreamSelector) normalizeSync(syncLag float64) float64 {
 	// Phase 2: Adaptive P10-P90 normalization (if enabled)
 	if ws.useAdaptiveSyncMax && ws.adaptiveSyncGetter != nil {
 		p10, p90 := ws.adaptiveSyncGetter()
@@ -568,7 +568,7 @@ func (ws *ProviderSelector) normalizeSync(syncLag float64) float64 {
 // normalizeStake converts stake to 0-1 range relative to total stake.
 // stake and totalStake are raw stake amounts.
 // Uses square root scaling to reduce whale dominance while maintaining incentives
-func (ws *ProviderSelector) normalizeStake(stake float64, totalStake float64) float64 {
+func (ws *UpstreamSelector) normalizeStake(stake float64, totalStake float64) float64 {
 	if totalStake == 0 || stake == 0 {
 		return 0.0
 	}
@@ -606,7 +606,7 @@ func (ws *ProviderSelector) normalizeStake(stake float64, totalStake float64) fl
 
 // applyStrategyAdjustments modifies scores based on the configured strategy
 // This allows different strategies to emphasize different metrics
-func (ws *ProviderSelector) applyStrategyAdjustments(latency, sync float64) (float64, float64) {
+func (ws *UpstreamSelector) applyStrategyAdjustments(latency, sync float64) (float64, float64) {
 	switch ws.strategy {
 	case StrategyLatency:
 		// Boost latency importance by squaring good latency scores
@@ -634,27 +634,27 @@ func (ws *ProviderSelector) applyStrategyAdjustments(latency, sync float64) (flo
 	return latency, sync
 }
 
-// SelectProvider selects a provider from the scored candidates according to the
-// configured SelectionMode (weighted random by default, see SelectProviderWithStats)
-func (ws *ProviderSelector) SelectProvider(
+// SelectUpstream selects a provider from the scored candidates according to the
+// configured SelectionMode (weighted random by default, see SelectUpstreamWithStats)
+func (ws *UpstreamSelector) SelectUpstream(
 	ctx context.Context,
-	providerScores []ProviderScore,
+	providerScores []UpstreamScore,
 ) string {
-	selected, _ := ws.SelectProviderWithStats(ctx, providerScores, nil)
+	selected, _ := ws.SelectUpstreamWithStats(ctx, providerScores, nil)
 	return selected
 }
 
-// SelectProviderWithStats selects a provider and returns detailed selection statistics
+// SelectUpstreamWithStats selects a provider and returns detailed selection statistics
 // if scoreDetails is provided.
 //
 // Only the winner-picking step depends on ws.selectionMode: SelectionModeWeightedRandom
 // draws proportionally to SelectionWeight, SelectionModeBest takes the highest score.
 // The guards below, the stats payload and the debug breakdown are shared by both modes so
 // the reporting contract cannot drift between them.
-func (ws *ProviderSelector) SelectProviderWithStats(
+func (ws *UpstreamSelector) SelectUpstreamWithStats(
 	ctx context.Context,
-	providerScores []ProviderScore,
-	scoreDetails []ProviderScoreDetails,
+	providerScores []UpstreamScore,
+	scoreDetails []UpstreamScoreDetails,
 ) (string, *SelectionStats) {
 	if len(providerScores) == 0 {
 		return "", nil
@@ -695,7 +695,7 @@ func (ws *ProviderSelector) SelectProviderWithStats(
 // pickWeightedIndex draws a provider with probability proportional to its SelectionWeight
 // and returns the winning index alongside the random value that produced it.
 // Callers must ensure totalScore > 0.
-func (ws *ProviderSelector) pickWeightedIndex(providerScores []ProviderScore, totalScore float64) (int, float64) {
+func (ws *UpstreamSelector) pickWeightedIndex(providerScores []UpstreamScore, totalScore float64) (int, float64) {
 	// Generate random value in [0, totalScore)
 	randomValue := ws.rng.Float64() * totalScore
 
@@ -724,7 +724,7 @@ func (ws *ProviderSelector) pickWeightedIndex(providerScores []ProviderScore, to
 // minSelectionChance, so exact N-way ties are routine on a degraded chain — a first-wins
 // scan would pin all traffic onto whichever address happens to sort first in the pairing
 // list, which is precisely the starvation the floor exists to prevent.
-func (ws *ProviderSelector) pickBestIndex(providerScores []ProviderScore) int {
+func (ws *UpstreamSelector) pickBestIndex(providerScores []UpstreamScore) int {
 	best, tied := 0, 1
 	for i := 1; i < len(providerScores); i++ {
 		switch {
@@ -744,10 +744,10 @@ func (ws *ProviderSelector) pickBestIndex(providerScores []ProviderScore) int {
 
 // buildSelectionResult emits the debug breakdown and assembles the SelectionStats payload
 // for the winning candidate. Shared by every selection path, including the short-circuits.
-func (ws *ProviderSelector) buildSelectionResult(
+func (ws *UpstreamSelector) buildSelectionResult(
 	ctx context.Context,
-	providerScores []ProviderScore,
-	scoreDetails []ProviderScoreDetails,
+	providerScores []UpstreamScore,
+	scoreDetails []UpstreamScoreDetails,
 	selectedIndex int,
 	randomValue float64,
 	totalScore float64,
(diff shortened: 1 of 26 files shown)


avitenzer (commented):
Reviewed at `f04b290`. Verified locally: `go build ./...` and `go vet` clean on the three touched packages, `go test ./protocol/provideroptimizer/...` and `TestDebugRuntimeConfig*` both pass.

**Verdict: approve with changes.** The design is sound and the refactor is a genuine improvement. Two things to address before merge — one CI gate, one avoidable cost on the exact path the new tie-break exists to serve.

---

## What I verified

- **The seam holds.** `CalculateProviderScores` sets `SelectionWeight: compositeScore` verbatim (`weighted_selector.go:860`), so argmax-on-`SelectionWeight` and argmax-on-`CompositeScore` are the same operation. The mode genuinely cannot perturb scoring.
- **`pickBestIndex` reservoir sampling is correct.** `tied` starts at 1, resets to `(i, 1)` on a strict improvement, increments then replaces with probability `1/tied` on equality. `Intn` is only ever reached with `tied >= 2`, so the `n <= 0` panic in `threadSafeRand.Intn` is unreachable. Uniform over all maxima.
- **The `wantedProviders` loop is safe in Best mode.** This was my main concern — Best is deterministic, so an N-iteration loop could plausibly return the same address N times. It doesn't: `consumer_session_manager.go:1425` adds each pick to `ignoredProvidersListCopy` before the next call, so Best yields the top-N by score.
- **The `ChooseBestProvider` deletion is clean.** Zero remaining references repo-wide, and none in `smart-router-enterprise-internal` either.
- **`buildSelectionResult` guards the zero-`totalScore` division**, so the all-zero path can't emit `NaN` percentages.

The `ChooseBestProvider` removal is the highest-value part of this diff, independent of the feature. A method named "Best" that performed an ordinary weighted draw is an active trap — anyone reading `getValidProviderAddresses` would have concluded sticky sessions were pinned to the leader.

---

## 1. Blocking: Jira gate is failing

```
Validate Jira ticket     fail   Missing Jira ticket line
jira-ticket/validated    fail
```

Everything else is green (lint, govulncheck, all five build targets, GitGuardian, Docker). The gate parses the PR body for a bare `Jira ticket: MAG-XXXX` line. One line, but it blocks merge.

## 2. Should fix: `pickBestIndex` burns O(N) crypto/rand calls on ties

`ws.rng` defaults to `globalRandomizer` → `utils/rand.Intn` → `threadSafeRand.Intn`, which is **`crypto/rand.Int` plus a `big.Int` allocation behind a single process-wide mutex**. Measured:

```
crypto/rand Intn: 146ns/op
math/rand   Intn:   3ns/op   (41x)
```

`pickBestIndex` calls it once per tied element, so an N-way tie costs N-1 serialized crypto draws per selection — against exactly one `Float64()` for the entire weighted path.

This matters more than it looks, because of one thing the PR description gets slightly wrong. The description says `SelectionModeBest` "lands here too" for the `totalScore <= 0` uniform fallback. With default config it never does: `CalculateScore` floors *every* provider at `minSelectionChance` (default `0.01`), including the `availabilityDead` collapse, so `totalScore` is zero only if someone explicitly passes `--qos-min-selection-chance 0`. That guard is effectively dead code in production.

So the degraded-chain case never takes the cheap uniform path — it falls into `pickBestIndex` with an N-way tie at exactly `0.01`, i.e. the *maximum* number of crypto draws, on the *worst* chain, under retry-amplified load. The outcome is still uniform so behaviour is correct; the cost profile is just inverted from what the description implies.

Two-pass collapses it to one draw. The extra pass over a small `float64` slice is free next to a single 146ns crypto call:

```go
func (ws *WeightedSelector) pickBestIndex(providerScores []ProviderScore) int {
	best, tied := 0, 1
	for i := 1; i < len(providerScores); i++ {
		switch {
		case providerScores[i].SelectionWeight > providerScores[best].SelectionWeight:
			best, tied = i, 1
		case providerScores[i].SelectionWeight == providerScores[best].SelectionWeight:
			tied++
		}
	}
	if tied == 1 {
		return best
	}

	// One draw regardless of tie width: on a degraded chain CalculateScore collapses
	// every provider to exactly minSelectionChance, so `tied` is the whole pairing list.
	nth := ws.rng.Intn(tied)
	for i, ps := range providerScores {
		if ps.SelectionWeight == providerScores[best].SelectionWeight {
			if nth == 0 {
				return i
			}
			nth--
		}
	}
	return best
}
```

`TestSelectProviderBestModeTieBreak` passes unchanged against this — it asserts the distribution rather than the algorithm, which is the right way to have written it.

Worth amending the `totalScore <= 0` comment (and the matching PR-description paragraph) to say the guard is unreachable at default `minSelectionChance`, and that degraded-chain uniformity actually comes from the tie-break. Otherwise the next reader will assume that guard covers the dead-chain case.

---

## Minor — fine as follow-ups

**3. The mode doesn't reach metrics.** `convertSelectionStatsToMetrics` (`consumer_session_manager.go:1261`) extracts only `ProviderScores` and `RNGValue`, so `SetProviderSelected` can't distinguish policies. The `RNGValue == 0` ambiguity that `Mode` was added to resolve still exists in Prometheus/OTel — it's fixed only in the debug header and the debug log. If a chain gets flipped to `best`, dashboards won't show it.

**4. Asymmetric `/debug/runtime-config`.** `PerChainOptimizer` carries `SelectionMode`; the flat top-level block (`rpcsmartrouter.go:1413-1452`) doesn't. Defensible, since that block is documented as `DefaultWeightedSelectorConfig()` defaults rather than live values, but every other optimizer field appears in both.

**5. Debug-log behaviour change worth knowing about.** `buildSelectionResult` now fires the per-candidate breakdown on the single-provider and all-zero paths, which previously returned before reaching it. It's a consistency fix, but a chain with one provider now allocates a full `logAttrs` slice per relay under `--log-level debug` where it previously allocated none. Guarded by `IsDebugEnabled()`, so production is unaffected.

---

## Tests

Strong, and written to fail for the right reasons. `TestSelectProviderBestMode` places the winner **last** so a first-wins scan fails it; `TestSelectionStatsCarriesMode` covering all six mode x short-circuit combinations is what actually pins the "stats contract can't drift between modes" claim; `TestSelectProviderWeightedModeUnchanged` is a real regression pin on the default path. Giving the two chains *different* modes in `TestDebugRuntimeConfig_SmartRouter_PerChainOptimizer` is the right instinct — a shared value would pass against a global default.

Two low-priority gaps: nothing pins that Best mode plus the `wantedProviders` loop returns N *distinct* providers in descending score order (I verified by reading the code, but it isn't asserted), and nothing exercises `--qos-selection-mode` through the actual cobra parse into a running optimizer.

## Security

Nothing of concern — no new input surface beyond a validated enum flag, no changes to auth, sessions or network paths.

The one operational risk is inherent to the feature rather than the implementation: Best concentrates 100% of organic traffic on the leader, so unselected providers stop producing relay data and their QoS goes stale. The description acknowledges this and points at the prober. That's the correct call given MAG-2237 — ranking on the unfloored composite would let a dead node's frozen latency/sync make it look fast — but it does mean Best mode's correctness depends on the prober being healthy in a way `weighted_random` does not. Worth a line in whatever runbook ships with the flag.

---

**Requested before merge:** the `Jira ticket:` line, the single-draw `pickBestIndex`, and the corrected "all-zero guard covers Best mode" claim in both the code comment and the description. Items 3-5 can be follow-ups.

avitenzer:
## Follow-up: Best-mode traffic dynamics under a rate-limited leader

Separate from the code review above. This came out of tracing what actually happens when the top provider starts returning 429 — the failover path is fine, but Best changes the *dynamics* around it in a way the description doesn't cover. Nothing here is a defect in the diff; it's the operational envelope the flag ships with, and I think it belongs in the description and/or the runbook before anyone enables it on a chain with rate-limited upstreams.

### Failover within a request works, and is mode-independent

Worth stating explicitly since it was my first concern with a deterministic argmax — Best cannot get stuck retrying the same address. The exclusion happens a layer above the selection mode:

1. `relayInnerDirect` → `classifyHTTPStatus(429)` → returns `fmt.Errorf("HTTP 429")` (`rpcsmartrouter_server.go:3797-3817`)
2. `shouldFailSession = err != nil || statusCode >= 500 || statusCode == 429` (`:1826`) → `OnSessionFailure` (`:1915`)
3. → `go AppendRelayFailure(addr)` **and** `consumerSession.Free(err)`
4. `Free` → `usedProviders.RemoveUsed(addr, routerKey, err)` (`single_consumer_session.go:58`) → `setUnwanted` (`used_providers.go:276`)
5. Retry's `GetSessions` → `GetUnwantedProvidersToSend(routerKey)` → `tempIgnoredProviders` (`consumer_session_manager.go:1080`)
6. `CalculateProviderScores` skips ignored addresses outright (`weighted_selector.go:830-832`) — they never enter `providerScores`
7. `pickBestIndex` argmaxes over the survivors → #2

So the retry walks down the ranking, and with `RelayRetryLimit` defaulting to 2 it can reach ~#3. Identical in both modes. No change needed.

### Across requests, Best turns a slope into a threshold

A fresh request resets `UsedProviders`, so only the score carries over. For a 429 the score is the *only* demotion signal available:

- 429 deliberately does **not** mark the endpoint unhealthy — `classifyHTTPStatus` returns `shouldMarkUnhealthy=false, needsBackoff=true` (`:4573-4581`), on the reasoning that a rate-limited node is healthy, just busy. It stays in `validAddresses`.
- The `needsBackoff` return is **discarded at the only call site**: `relayLatency, err, _ := rpcss.relayInnerDirect(...)` (`:1751`). No cooloff is applied.
- So the sole effect is `AppendRelayFailure` → one availability sample of `0`.

Both of those are pre-existing main-branch behaviour, not introduced here. But under `weighted_random` a dinged availability score bleeds traffic away *continuously* — that proportional falloff is itself a form of backpressure. Under Best the rate-limited provider keeps **100%** of traffic until its composite drops strictly below #2. All-or-nothing, and every request in that window burns a wasted round-trip before failing over on retry.

The gap isn't necessarily small: `normalizeAvailability` rescales `[0.8, 1.0] → [0.0, 1.0]`, so at the default `AvailabilityWeight: 0.3`, moving availability 1.00 → 0.95 costs only 0.075 of composite.

### Recovery is probe-only, with a ~1 hour time constant

This is the part I'd most want a second opinion on. `ScoreStore` is `Resolve() = Num/Denom` with decay applied *only inside* `Update()`:

```go
newNum   = Num*decayFactor   + sample*Weight     // score_store.go:242
newDenom = Denom*decayFactor + Weight            // score_store.go:260
```

Both sides decay by the same factor, so **the ratio is frozen between samples** — waiting does nothing. A demoted provider sits at exactly the value it collapsed to until something feeds it a sample.

There are two sample sources: organic relays (weight `RelayUpdateWeight = 1`) and the probe loop (`AppendProbeData`, weight `ProbeUpdateWeight = 0.25`, every 5s over `GetAllDirectRPCEndpoints()` — all endpoints, so the demoted provider *is* covered). Under Best a non-leader gets exactly zero organic relays, so probes are the sole channel.

Once probes return healthy, let `gap = Denom - Num`. Each update multiplies the gap by `decayFactor` and adds nothing to it, so over wall-clock `T` with samples every `dt`:

```
gap(T) = gap(0) * d^(T/dt) = gap(0) * exp(-ln2 * T / HalfLife)
```

`dt` cancels. **Recovery halves every `HalfLife` regardless of probe cadence**, and `calculateHalfTime` returns `max(DefaultHalfLifeTime = 1h, median relay gap)` capped at `MaxHalfTime = 3h`.

Net: demotion takes seconds (all traffic, weight 1.0, at full request rate); recovery takes on the order of an hour (probes only, weight 0.25, ~1h time constant). And re-entry is all-or-nothing — the recovered provider takes 100% back at the instant it crosses.

The algebra above is derived from the two update formulas rather than measured, so worth a sanity check, but I believe the cadence-independence result is solid and it's the counter-intuitive bit: lowering `--probe-loop-interval` will *not* speed recovery up.

### The bit that connects back to MAG-2237

The description defends `minSelectionChance` as the anti-starvation floor, and under `weighted_random` it genuinely delivers: a demoted provider keeps ~1% of traffic, which at 100 RPS is ~1 weight-1.0 organic sample per second — about **20x** the probe feed of 0.05 weight/sec.

Under Best that floor grants **no traffic at all**. Argmax ignores it except in the exact-tie case. So Best doesn't only concentrate traffic; it removes the organic feedback channel the floor was designed to keep open, and hands recovery entirely to a 4x-lighter probe signal.

That seems worth calling out directly, because the tie-break rationale in the description leans on the floor being load-bearing for MAG-2237 — which is true in the all-tied case, but the floor stops delivering its recovery samples the moment there's a clear leader.

### What I'd suggest

Not asking for any of this in this PR — the flag is opt-in and defaults to the existing behaviour, which is the right call. But:

1. **Document the envelope.** The description's "recovery is driven by the prober" line is accurate; what it omits is the timescale (~1h half-life) and that Best is precisely what makes the prober the *only* driver. Worth stating in the description and wherever the flag gets operator docs.
2. **Flapping is a real risk.** Re-entry is all-or-nothing, so a provider that recovers just past the leader instantly absorbs the full request rate — exactly what rate-limited it before. `weighted_random` ramps back in gradually; Best slams it. Best + rate-limited upstreams can oscillate.
3. **Consider whether 429 deserves different treatment under Best**, as a follow-up — either honouring the currently-discarded `needsBackoff` with a short cooloff, or a rate-limit-specific penalty. The argument is that there's no longer any proportional bleed-off to absorb it.
4. **Worth knowing:** the obvious tuning lever doesn't work. `--probe-loop-interval` won't change recovery speed, and the actual constant (`DefaultHalfLifeTime`) has no flag.

Happy to be wrong on any of the numbers here — flagging the shape of it more than the exact magnitudes.

avitenzer:
Superseded by #282, which consolidates this stack (#262 → #263 → #264) into a single PR rebased onto current `main` (zero conflicts).

One change versus this stack: the rename in #263 now targets `EndpointSelector` rather than `ProviderSelector`, addressing @nimrod-teich's request to move away from "provider" terminology. "Endpoint" is the term the codebase is already converging on (`protocol/endpointstate`, `EndpointMetrics`, `GetAllDirectRPCEndpoints`, `EndpointLagThreshold`).

Tracked under [MAG-2780](https://magmadevs.atlassian.net/browse/MAG-2780) (new subtask of MAG-2599 Failover v2), which also clears the Jira gate that was failing here.

Closing in favour of #282.

[MAG-2780]: https://magmadevs.atlassian.net/browse/MAG-2780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

avitenzer:
Correction to my previous comment — please disregard it.

I closed this PR based on state I'd read at the start of the session, not noticing that #264 and #263 had since been merged **up the stack** (into `refactor/rename-provider-selector` at 11:23Z and `feat/provider-selection-mode` at 12:14Z). That already made this PR the single consolidated one: `feat/provider-selection-mode` now carries all three commits linearly against `main`.

I've reopened it. Nothing was lost.

There are now two candidates for the consolidated PR and it's worth a deliberate choice rather than letting both sit open:

**This PR (#262)** — consolidated by merging up the stack. Retains all three original review threads. Still on the original base point (not rebased onto current `main`), still names the type `ProviderSelector`, still carries @nimrod-teich's CHANGES_REQUESTED, and still fails the required Jira gate (no `Jira ticket:` line in the description).

**#282** — the same three commits rebased onto current `main` (zero conflicts), with the rename retargeted to `EndpointSelector` per @nimrod-teich's terminology request, a `Jira ticket: MAG-2780` line so the gate passes, and `make test` green across all 33 packages.

If #282 is the one to keep, this can be closed. If this one is preferred, #282 should be closed and the rename/Jira/rebase points folded in here instead.

avitenzer:
Folded the rename into this PR as `bd598f0` (fast-forward on top of Tomelia1999's rebase — no force-push).

@nimrod-teich — this addresses your request to move away from "provider" terminology. The selector is now `EndpointSelector`, matching the term the codebase is already converging on (`protocol/endpointstate`, `EndpointMetrics`, `GetAllDirectRPCEndpoints`, `EndpointLagThreshold`).

It lands as a follow-up commit rather than a rewrite of `9d6fd31`, so nobody with this branch checked out has to reset. **The Files-changed view shows only the net result** — `WeightedSelector` → `EndpointSelector`, with `ProviderSelector` appearing nowhere in the diff.

Scope is the selector type only; the surrounding `provideroptimizer` package and the wider `provider*` vocabulary are untouched. A full terminology migration is a much larger change and would want its own ticket.

Also added `Jira ticket: MAG-2780` to the description ([new subtask](https://magmadevs.atlassian.net/browse/MAG-2780) under MAG-2599 Failover v2), which clears the required Jira gate that was failing here.

Verification on this exact tree: `go build ./...` and `go vet ./...` clean, `make test` green across all 33 packages, zero failures.

#282 — the alternative consolidation I opened before noticing this stack had already been merged up — is now closed in favour of this one.

avitenzer:
Reworked the rename per the terminology discussion. Force-pushed with `--force-with-lease` — Tomelia1999's three commits are untouched; only my rename commit was replaced, and the branch was verified unmoved first.

@nimrod-teich — `Endpoint` turned out to be the wrong answer to your request, for a reason worth recording. `provider` and `endpoint` are not synonyms here, they're two levels of a real hierarchy:

```
ConsumerSessionsWithProvider     identity: address + stake
  └── Endpoints []*Endpoint      network address (host:port)
        └── Connections []*...   the actual connections
```

The selector ranks the **identity** level and returns one address; the session manager then picks an endpoint within it. So `EndpointSelector` named the wrong level, and `ConsumerSessionsWithEndpoint{ Endpoints []*Endpoint }` would have been nonsense.

It's now **`UpstreamSelector`** — that level, without colliding with `Endpoint` or `NodeUrl`. The type's whole API moved with it (`SelectUpstreamWithStats`, `ChooseUpstreamWithStats`, `UpstreamScore`, `CalculateUpstreamScores`), because a `UpstreamSelector` returning `[]ProviderScore` would just relocate the inconsistency. 309 insertions / 309 deletions — a pure rename.

**No external contract moves.** I snapshotted all 101 externally-observable surfaces before and after and diffed them byte-identical: the twelve `Lava-*` provider headers, `lava-select-provider`, the eight `smartrouter_*` provider metric names, `/debug/provider-scores`, every json/yaml/mapstructure tag, every cobra flag key. Three packages are excluded outright to keep it that way — `protocol/metrics`, `protocol/relaycore` and `debug_provider_scores_test.go` — all verified unmodified.

Two traps a repo-wide sed would have hit, for the record: `allProviderScores` in metrics contains the substring `ProviderScore` but is typed `ProviderSelectionScores`, so a blind substitution renames the variable and orphans the type; and `TestArchiveAddDoesNotPreserveSelectProvider` in relaycore guards the `lava-select-provider` **header**, not this API.

Scope is the selector's API surface only. `ProviderOptimizer`, `ProviderStakeCache`, the package name, and the provider vocabulary across lavasession/rpcsmartrouter/metrics are untouched — that's ~280 identifiers across 19 packages and wants its own ticket, plus a decision on whether the external contracts ever migrate.

`go build ./...`, `go vet ./...` and `make test` clean: 33 packages, zero failures.

Tomelia1999:
@avitenzer thanks — this was a genuinely useful review, and one of your findings changed the design rather than just the code. All of it is addressed. Branch is rebased onto current `main` (`498c298`) and force-pushed; new work is in `ec0f0b6`, `2b634c8`, `695468e`.

**Heads-up before the details:** you reviewed at `f04b290`, which was the branch when it was a *single* commit. It has since been force-pushed to a six-commit stack. So your pass covers commit 1 of 6 — `--qos-selection-priority`, both renames, and `resolveSelectionWeights` have had no review from anyone. Details at the bottom.

---

## 1. Jira gate — already green

Fixed before your review landed in my inbox. `jira-ticket/validated: pass — Validated MAG-2780`, and all 20 checks are green.

## 2. `pickBestIndex` — we went further than you suggested

You were right about the cost, and I reproduced your numbers: **183.7 ns/op, 3 allocs** for the crypto path vs 4.5 ns for `math/rand` — the same 41x you measured.

But digging into *why* the tie-break existed, I think the tie-break itself was wrong, not just its implementation. **It now breaks ties first-wins, with no draw at all.**

The reasoning: FAILOVER-TASKS section 1 asks in bold for *"the order is fixed, not random ... the answer does not change between two identical requests."* A random tie-break breaks exactly that — and, because `CalculateScore` collapses everything below `MinAcceptableAvailability` to the same floor, it breaks it on **every pick on a degraded chain**. That is `best` mode reverting to a lottery in the one situation where an operator most needs the routing to be explainable.

The starvation justification I originally wrote does not survive contact with the mode. `minSelectionChance` is a **lottery** mechanism: it buys a floored upstream a slice of the weighted draw. Under `best` there is no draw, so a floored upstream sitting beside any higher-scoring peer already gets exactly zero traffic. The floor stopped protecting anyone the moment the operator asked for "always the best" — randomising the all-tied case never gave that back, it only cost the determinism.

Two consequences, accepted deliberately and now documented in the code: when everything is floored the candidates are equally dead and the `wantedProviders` loop still walks the whole list within a request, so only the *first* attempt concentrates; and a non-leader recovers via the prober, which was already true under `best`.

Worth noting on the "first-wins pins one address" concern: `validAddresses` comes from iterating a Go map, so the order is arbitrary but **stable within an epoch** — which is what determinism needs — and **differs per pod**, so a fleet still spreads the all-tied case across routers.

Your two-pass rewrite was correct (I traced it against `[0.5,0.9,0.5,0.9]`, `[0.5,0.9,0.9,0.9]`, `[0.5,0.5,0.9]`), it just became unnecessary once the draw disappeared. Measured before/after, all-tied, against the real global RNG:

| candidates | before | after |
|---|---|---|
| 2 | ~140ns, 3 allocs | 2.2ns, 0 allocs |
| 5 | ~630ns, 12 allocs | 4.2ns, 0 allocs |
| 10 | ~1440ns, 27 allocs | 9.1ns, 0 allocs |
| 20 | ~3040ns, 57 allocs | 14.4ns, 0 allocs |

Under `RunParallel` at 10 tied candidates it goes from ~160x to **~224x** — which is the lock contention you were pointing at, and it is worse than the serial number suggests: that mutex is shared with per-request GUID generation (`rpcconsumer_logs.go:149`) and every WebSocket message (`consumer_websocket_manager.go:239`).

## 3. The `totalScore <= 0` claim — you were right, and this was the best catch

Confirmed and corrected in both the code comment and the PR description. `CalculateScore` floors every candidate at `minSelectionChance`, so the branch is unreachable unless someone passes `--qos-min-selection-chance 0` — and `NewUpstreamSelector` does not reject that, because it validates the four weights and leaves `MinSelectionChance` alone. The comment now says explicitly that degraded-chain behaviour is decided by the tie-break, not by that guard, so the next reader does not draw the same wrong conclusion I did.

The unvalidated `MinSelectionChance` is pre-existing (setting it to 0 silently removes the MAG-2237 floor entirely) — filing separately rather than widening this PR.

## 4. Mode doesn't reach metrics — agreed, follow-up ticket

Correct, and it is a real gap: the `RNGValue == 0` ambiguity `Mode` exists to resolve still exists in Prometheus/OTel. Not fixing it here because it changes the `SetProviderSelected` signature on `consumer_metrics_manager_inf.go:99`, which means touching `protocol/metrics` — the package this PR deliberately excludes to keep the 101 external surfaces byte-identical. Separate ticket.

## 5. `/debug/runtime-config` asymmetry — leaving as-is

Agreed it is asymmetric, but that block is documented as `DefaultUpstreamSelectorConfig()` defaults rather than live values. Adding a "default" mode beside per-chain live modes reads as *more* confusing, not less. Noting the wider version of this in the follow-up: under a non-`balanced` priority the top-level weights already diverge from every per-chain entry.

## 6. Debug-log change — acknowledged

Right on both counts, including that `IsDebugEnabled()` keeps production unaffected. Called out in the commit message so nobody debugs it later as a regression.

## Tests — both your gaps

**The queue gap is closed.** New `protocol/lavasession/best_mode_queue_test.go` covers the `wantedProviders` loop end to end: top-3 in descending order, the *full* queue (so a loop that ranks the head right but degenerates later cannot pass), stability across 20 identical calls, and the caller's ignore list on the retry path. Latencies are seeded worst-first, so a loop that ignored scores and walked the list would return the exact reverse.

One thing worth flagging for anyone writing tests in this area. My first version was **flaky, ~1 in 15 runs**, and the cause was not obvious: `UpdateAllProviders` starts async probing, every probe result feeds the optimizer, and a late probe sample shifts the adaptive latency P10–P90 band and reorders the ranking *between* the expectation and the assertion. Measured directly — 3 of 80 attempts saw a composite change between two consecutive reads. A longer sleep would only have hidden it, so the test now uses an optimizer whose `Append*` methods are no-ops. Selection is **not** stubbed: real `UpstreamSelector`, real `pickBestIndex`, real ignored-provider handling. 45 consecutive clean runs since.

Both suites are mutation-checked, so they are not decorative: deleting the loop's ignore-list update fails 3 of the 4 queue tests, and flipping the argmax to `>=` fails both new selector tests.

**Your second gap is still open** — nothing drives `--qos-selection-mode` through the cobra parse into a running optimizer. Fair hit, and it is the same asymmetry as `resolveSelectionWeights` being extracted *specifically* so it could be tested while the mode parse stayed inline in `RunE`. Tracking it.

## The compile break you could not have seen

The rename missed `protocol/rpcsmartrouter/node_error_demotion_mag2156_test.go:81`, so that test package did not compile at `277f546`. Not your miss — that file did not exist at `f04b290`, so your "`go vet` clean on the three touched packages" was accurate for what you had.

Worth everyone knowing *why* it survived: `go build ./...` does not compile `_test.go` files, and **nothing in CI ever does** — there is no `go test` job in any workflow, and the `lint` job carries `continue-on-error: true` on all three attempts, so it cannot fail. Every check was green on a tree where `go vet` fails. Fixed in `ec0f0b6`; the CI gap is written up and being filed separately, because it means a genuinely *failing* test would also merge green today.

The PR description's verification section has been corrected — it claimed vet and the full suite were clean when they were not. It now reads: 34 packages, zero failures, and states what went wrong.

---

## Ask

Could you take a second pass over commits 2–6? Your review predates the priority presets, both renames, and `resolveSelectionWeights` — in particular the precedence rule there decides overrides from `flags.Changed()` but reads values from `viper`, and a partially-overridden preset (`--qos-selection-priority fastest --qos-latency-weight 0.5`) sums to 0.80, which trips the normaliser warning and silently rescales the operator's number. I would value your read on that one specifically.

avitenzer:
### `best` mode has a config-reachable determinism hole

`upstream_selector.go` — the `totalScore <= 0` guard draws from the RNG **regardless of mode**:

```go
utils.LavaFormatWarning("all provider scores are zero, using uniform selection", nil)
return ws.buildSelectionResult(ctx, providerScores, scoreDetails, ws.rng.Intn(len(providerScores)), 0.0, totalScore)
```

The comment above it argues the branch is unreachable because `CalculateScore` floors every candidate at `minSelectionChance`. That's true — unless `minSelectionChance` is itself 0. The comment says as much ("the single way in is `--qos-min-selection-chance 0`, which `NewUpstreamSelector` does not reject"), but documents the hole rather than closing it, and it lands on the one invariant this mode is sold on.

#### Reproduced

`SelectionModeBest`, `MinSelectionChance = 0`, three candidates below `MinAcceptableAvailability` (so every composite floors to 0), 300 identical requests:

```
WRN all provider scores are zero, using uniform selection
DBG Provider selection completed  selection_mode=best  selected_provider=a  total_score=0
DBG Provider selection completed  selection_mode=best  selected_provider=b  total_score=0
DBG Provider selection completed  selection_mode=best  selected_provider=a  total_score=0
DBG Provider selection completed  selection_mode=best  selected_provider=b  total_score=0
DBG Provider selection completed  selection_mode=best  selected_provider=c  total_score=0
```

Identical input, three different answers — with `selection_mode=best` in the log line. This is exactly the property FAILOVER-TASKS section 1 asks for and that `pickBestIndex` spends thirty lines defending: *"the answer does not change between two identical requests."*

#### Why this config is likelier than the comment implies

The branch's own reasoning invites it. `pickBestIndex` argues:

> `minSelectionChance` is a **lottery** mechanism — it buys a floored upstream a slice of the weighted draw. Under `best` there is no draw, so the floor stopped protecting anyone the moment the operator asked for "always the best".

An operator who reads that and concludes "then I don't need the floor" sets `--qos-min-selection-chance 0` — and lands in the one branch that puts the draw back. The two flags are argued to be independent, but they aren't.

#### Suggested fix

Keep the guard shared, but let it answer the tie the way the mode answers every other tie:

```go
if totalScore <= 0 {
	utils.LavaFormatWarning("all provider scores are zero, using uniform selection", nil)
	selectedIndex := 0 // Best stays deterministic: an all-zero list is an N-way tie, and
	if ws.selectionMode != SelectionModeBest { // first-wins is this mode's answer to a tie.
		selectedIndex = ws.rng.Intn(len(providerScores))
	}
	return ws.buildSelectionResult(ctx, providerScores, scoreDetails, selectedIndex, 0.0, totalScore)
}
```

Worth considering alongside it: `NewUpstreamSelector` validates the four weights but never `MinSelectionChance`, so the same gap also accepts `--qos-min-selection-chance 2`, which floors every composite to 1.0 and flattens scoring entirely. That one is pre-existing, but this PR is well placed to close it.

#### Test gap

`TestSelectionStatsCarriesMode/best/all_zero` passes through this branch but only asserts `stats.Mode`, never the selected address — so it can't catch this. A determinism assertion on the returned address with `MinSelectionChance: 0` would pin it.

Tomelia1999:
@avitenzer confirmed and fixed in `b696dd9`. All three points were right, and I reproduced the first two rather than taking them on the code read.

**The determinism hole.** Your repro is exact. `best`, `MinSelectionChance = 0`, three candidates below `MinAcceptableAvailability`, 300 identical requests:

```
300 identical requests under selection_mode=best -> map[a:98 b:103 c:99]
```

Three answers to one question, in the mode whose whole contract is that there is only one.

**The floor above 1.0.** Also exact, and worse than "flattens scoring" suggests — it is not a compression, it is a total collapse. Driving the real scorer at `MinSelectionChance = 2`:

```
composite(availability 1.00, latency 10ms,  sync 0.1s) = 1
composite(availability 0.99, latency 25s,   sync 500s) = 1
```

Every composite is pushed above 1.0 and then clamped back to 1.0. A fast upstream and a hopeless one are indistinguishable, with nothing logged.

---

### What changed

**1. The guard is mode-aware.** Took your shape — first-wins under `best`, unbiased draw otherwise. One addition: the warning now carries the mode, the candidate count, and a hint that the branch is only reachable with `--qos-min-selection-chance 0`. If anyone ever does hit this in production, the log should say *why*, not just *what*.

**2. `MinSelectionChance` is validated.** NaN, Inf, `< 0` and `> 1` fall back to the default with a warning, matching how the four weights are already handled. Negative matters for the reason you would expect but is worth stating: it does not merely stop flooring, it can drag `totalScore` below zero and divert live traffic into the all-zero guard.

`0` stays legal. It is the deliberate "no floor at all" choice, and it is safe now that (1) landed. One judgment call worth flagging: I allow **exactly 1.0**, which still flattens everything to 1.0 — same end state as `2`, reached legitimately. Rejecting `1.0` while accepting `0.999` felt like an arbitrary line, and `> 1` is the genuinely incoherent case since this is a score floor and scores cap at 1. Happy to tighten it to `[0, 1)` if you disagree; it is one character.

**3. The comment that caused it — this was the sharpest part of your review.**

You are right that the reasoning invites the config. `pickBestIndex` argued the floor "stopped protecting anyone" under `best`, an operator reads that, concludes they do not need it, and lands in the one branch that put the draw back. My own comment walked them into the trap.

It now scopes that claim to the **tie-break** specifically, says plainly not to read it as a reason to disable the floor, and notes the floor still does the entire anti-starvation job under `weighted_random`. Credited to you in the source.

### Tests

Your suggested assertion, plus three more:

- Determinism on the **returned address** at `MinSelectionChance = 0`, 300 iterations. Deliberately **unseeded** — `SetDeterministicSeed` swaps the crypto RNG for `math/rand`, so a seeded selector would still be deterministic per seed and would hide a reintroduced draw completely.
- The mirror: `weighted_random` must still spread an all-zero list across every candidate, so the fix cannot quietly pin a degraded chain onto one address.
- A validation table covering the `1.0` boundary and the legal `0`.
- One that drives the real scorer to show the flattening when the constructor is bypassed, and its absence when it is not — so the *reason* for the validation is pinned, not just the validation.

Mutation-checked, since your point was that the existing coverage could not catch this: reverting the guard to an unconditional draw fails the determinism test, and removing the validation fails both validation tests. You were right about `TestSelectionStatsCarriesMode/best/all_zero` — it walks the branch and asserts only `stats.Mode`, never the address.

`go vet ./...` clean, full suite 34 packages / 0 failures.

---

Two rounds now where you have found something real in this mode, and both times the fix went beyond the line you pointed at. Thank you — genuinely.

Standing ask, unchanged: **commits 2–6 still have no review from anyone.** Your original pass was at `f04b290`, when this was a single commit. `--qos-selection-priority`, both renames, and `resolveSelectionWeights` have not been looked at. The two spots I would most value your eye on:

- `resolveSelectionWeights` decides *whether* a weight was overridden from `flags.Changed()` but reads the *value* from `viper` — two sources that agree only because every weight flag happens to be `BindPFlag`'d today.
- A partially-overridden preset (`--qos-selection-priority fastest --qos-latency-weight 0.5`) sums to 0.80, trips the normaliser warning at startup, and silently rescales the operator's `0.5` to `0.625`. That is a documented, supported combination.

protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -814,6 +815,57 @@ func resetEndpointHealthAndGauge(deps debugMuxDeps) int {
...
+	for _, w := range []struct {
+		flagName string
+		target   *float64
+	}{
+		{common.ProviderOptimizerAvailabilityWeight, &config.AvailabilityWeight},
+		{common.ProviderOptimizerLatencyWeight, &config.LatencyWeight},
+		{common.ProviderOptimizerSyncWeight, &config.SyncWeight},
+		{common.ProviderOptimizerStakeWeight, &config.StakeWeight},
+	} {
+		if (flags != nil && flags.Changed(w.flagName)) || viper.InConfig(w.flagName) {
```
avitenzer, line 867: **The gate and the value come from two different sources.**

`flags.Changed(w.flagName)` decides *whether* the weight was overridden; `viper.GetFloat64(w.flagName)` decides *what it is*. Those agree only because all four weight flags happen to be `BindPFlag`'d at L3222-3233 — 2,400 lines from the code that depends on it, with nothing asserting the coupling.

I ran the unbound case (`fs.Set` on the CLI, no `viper.Set`):

```
operator typed --qos-latency-weight 0.5 -> selector gets 0.0000
INF Working with provider selection priority: fastest availabilityWeight=0.2
    latencyWeight=0 manuallyOverridden=qos-latency-weight ...
```

Not a fallback to the preset's 0.70 — a hard **zero**, while the Info line reports `manuallyOverridden=qos-latency-weight`. Combined with `fastest` that normalises to 0.667 availability / **0.0 latency** / 0.333 sync: a latency-priority router with latency switched off, no error, and a log line that reads like success.

Read from the source you tested:

```go
if flags != nil && flags.Changed(w.flagName) {
    v, err := flags.GetFloat64(w.flagName)
    if err != nil {
        return config, err
    }
    *w.target = v
} else if viper.InConfig(w.flagName) {
    *w.target = viper.GetFloat64(w.flagName)
} else {
    continue
}
overridden = append(overridden, w.flagName)
```

The `viper.InConfig` half of the design is right and worth keeping — I confirmed `ReadInConfig()` runs before this, so a weight set only in `config.yml` really is invisible to `Changed()`. It's just the value read that needs to follow the branch.

protocol/rpcsmartrouter/selection_weights_test.go
```diff
@@ -0,0 +1,125 @@
...
+// TestResolveSelectionWeights_CLIFlagBeatsPreset is the headline precedence rule: a weight
+// typed on the command line wins over the preset, and the untouched axes keep the
+// preset's values.
+func TestResolveSelectionWeights_CLIFlagBeatsPreset(t *testing.T) {
+	withCleanViper(t)
+	viper.Set(common.ProviderOptimizerSelectionPriority, "fastest")
+
+	fs := newWeightFlagSet(t)
+	require.NoError(t, fs.Set(common.ProviderOptimizerLatencyWeight, "0.5"))
+	viper.Set(common.ProviderOptimizerLatencyWeight, 0.5)
```
avitenzer, line 60: **This line is the finding above, written down as test setup.**

`fs.Set(...)` on the previous line already expresses what the operator did. The `viper.Set` exists only because the production code reads the gate from one source and the value from the other — so the test has to keep them in sync by hand.

That means this test passes in exactly the world where production breaks: it can never fail on a missing `BindPFlag`, which is the only failure mode the split source has.

After the fix in `resolveSelectionWeights`, this line should be deletable and the test should still pass. That is the assertion worth having here.

protocol/provideroptimizer/upstream_selector.go
```diff
@@ -153,24 +167,46 @@ func NewWeightedSelector(config WeightedSelectorConfig) *WeightedSelector {
...
+		utils.LavaFormatWarning("invalid min selection chance, must be finite and within [0, 1], using default", nil,
+			utils.LogAttr("minSelectionChance", config.MinSelectionChance),
+			utils.LogAttr("usingInstead", defaultChance),
+		)
+		config.MinSelectionChance = defaultChance
+	}
+
 	if math.Abs(totalWeight-1.0) > 0.001 {
-		utils.LavaFormatWarning("weighted selector weights do not sum to 1.0, normalizing",
+		utils.LavaFormatWarning("endpoint selector weights do not sum to 1.0, normalizing",
```
avitenzer, line 209: **This fires on a documented, supported combination.**

`--qos-selection-priority fastest --qos-latency-weight 0.5` — the precedence rule the PR description advertises as "the preset is a starting point, never a cage" — produces:

```
preset fastest:  avail=0.2000 lat=0.7000 sync=0.1000 stake=0.0000 sum=1.0000
after --qos-latency-weight 0.5:                                  sum=0.8000
WRN endpoint selector weights do not sum to 1.0, normalizing ... totalWeight=0.7999999999999999
selector actual: avail=0.2500 lat=0.6250 sync=0.1250 stake=0.0000
```

The operator gets a startup WARNING that reads as their mistake, and their `0.5` runs as `0.625`. Any partial override of a preset hits this unless the typed value happens to equal the preset's.

Two things to be precise about before anyone "fixes" this:

- **The rescale is not distorting the scoring.** Normalisation preserves ratios — latency's share is `0.5/0.8 = 0.625` whether you divide or not. The division keeps the composite on `[0,1]` so `minSelectionChance` and the 1.0 clamp still mean what they say. This is a semantics/UX defect, not a scoring-math bug.
- **It is not a regression.** Pre-PR, `--qos-latency-weight 0.5` against the 0.3/0.3/0.2/0.2 defaults summed to 1.2 and warned identically. But this PR *promotes* the path — and it falsifies the description's note that these strings "only fire on invalid startup config". The normaliser warning now fires on the feature's headline use case.

**Two ways out — this one is your call:**

**(a) Honour the typed number.** Rescale only the *non-overridden* axes to fill `1 - Σ(overridden)`. `fastest` + `lat 0.5` → `0.333 / 0.500 / 0.167 / 0`; latency is genuinely 0.5. Needs a guard for `Σ(overridden) > 1`.

**(b) Keep the math, fix the telling.** Normalise inside `resolveSelectionWeights` so what reaches `NewUpstreamSelector` already sums to 1.0 and no warning fires, and log the *effective* weights in the existing Info line.

I'd take **(b)** — smaller, doesn't invent a second normalisation semantics, and it fixes the actual complaint (an operator warned at for a supported config and then never told the real number). (a) is more honest to the flag name but changes what a partial override means.

**Test gap either way:** `TestResolveSelectionWeights_CLIFlagBeatsPreset` asserts `0.5` on the struct *returned by* `resolveSelectionWeights` and never feeds it to `NewUpstreamSelector`. The suite pins the number the operator typed while production runs a different one, and nothing crosses that boundary. Whichever option you pick needs a test that ends at the selector's `GetConfig()`.

avitenzer (commented):
## Review of commits 2–8

Commit 1 got a pass at `f04b290`; everything after it — both renames, `--qos-selection-priority`, and `resolveSelectionWeights` — had no review from anyone until now. This covers that gap.

Re-verified rather than taken on trust: `go build ./...` clean, `go vet` clean on the three touched packages, and `provideroptimizer` / `rpcsmartrouter` / `lavasession` all green locally (50s / 91s / 37s). Branch is rebased on current `main`, 0 behind.

**Claims I checked and confirmed accurate**, since nobody had: `balanced` (0.30/0.30/0.20/0.20) is byte-identical to `DefaultUpstreamSelectorConfig()`, so backward-compat holds; all four presets sum to 1.0; `b6ef6e3` really is 309/309; and `ChooseBestProviderWithStats` on `main` did call the same `SelectProviderWithStats` weighted draw, so removing it is safe. Also cleared two things I went looking for: there is no `AutomaticEnv` anywhere in the tree, so no env-var precedence hole, and `viper.ReadInConfig()` (L2877) genuinely does run before `resolveSelectionWeights` (L3098) in the same `RunE`, so the `viper.InConfig` half is reachable as designed.

Two findings, both inline. Summary:

1. **`resolveSelectionWeights` gates on `flags.Changed()` but reads from viper.** Latent today — `BindPFlag` is the only thing holding it — but I ran the unbound case and the operator's `0.5` silently becomes `0.0`. One-line fix. The new test hand-mirrors the value into both sources, so it cannot catch this.
2. **A partial override trips the normaliser warning and rescales the typed value.** `--qos-selection-priority fastest --qos-latency-weight 0.5` → sum 0.80 → WARN at startup → runs as 0.625. Not a regression, but this PR promotes the path that triggers it. **Needs a product call — see the inline comment for the two options.**

Nothing here blocks the design. Commit 1's seam is the right one and commits 6/8 closed the determinism holes cleanly. #1 is a one-line fix and should land before merge; #2 needs a decision first.

Tomelia1999:
@avitenzer thanks for taking commits 2–8 — that closed the gap, and the re-verification of my own claims (the `balanced` byte-identity, the preset sums, the `ChooseBestProviderWithStats` equivalence) is the part I most needed someone else to do. Also useful that you cleared `AutomaticEnv` and confirmed `ReadInConfig()` ordering; both were open questions on my side.

One correction to me, not you: I flagged your `b6ef6e3` as a SHA that didn't exist. **You were reading origin and my local was stale** — the branch had been rebased onto newer `main` since my last push. `b6ef6e3` is the rename commit. My mistake.

---

## Finding 2 — fixed, option (b), in `03ec70b`

Went with your recommendation. `resolveSelectionWeights` now normalises before returning, so the selector receives a set already summing to 1.0 and its warning cannot fire. The Info line reports the **effective** weights plus a `rescaledToSumToOne` flag.

Your two clarifications both went into the code comment, because they are the things a future reader will otherwise re-litigate: that the rescale **preserves ratios** so this is a semantics/UX defect rather than a scoring bug, and that it is **not a regression** — this PR just promotes the path that triggers it.

The 0.001 tolerance deliberately mirrors `NewUpstreamSelector`'s own check. Matching it exactly is what makes the warning unreachable from this path rather than merely unlikely.

Genuinely broken input still passes through unscaled — negative, NaN, Inf, all-zero — so the existing validation rejects it loudly and falls back to defaults. Only the legitimate case goes quiet.

**One argument for (b) you didn't raise, which I think is the strongest one.** (a) would also move deployments that set a single weight by hand *today, with no preset involved*. `--qos-latency-weight 0.5` against the defaults sums to 1.20 and currently runs **0.4167**; under (a) it becomes **0.5000** the day they upgrade. Silently moving existing deployments is the one thing this PR has avoided everywhere else, so (a) would have broken that promise in the course of fixing a UX complaint. `TestResolveSelectionWeights_NoPresetOverrideIsUnchanged` now pins it.

**On your test-gap point — you were right, and it was worse than a gap.** The two precedence tests asserted the typed number on the struct `resolveSelectionWeights` returns, so the suite pinned `0.5` while production ran `0.625` with nothing crossing that boundary. They now assert effective weights *and the ratio*, since the ratio is what actually proves the typed value won rather than the preset's. Six tests total, all ending at `NewUpstreamSelector(...).GetConfig()`. Mutation-checked: making the normaliser a no-op fails five of them.

`go vet ./...` clean, full suite 34 packages / 0 failures.

---

## Finding 1 — accepted, deliberately deferred

Your diagnosis is correct and I reproduced it exactly:

```
operator typed --qos-latency-weight 0.5 -> selector gets 0.0000
selector actual: avail=0.6667 lat=0.0000 sync=0.3333
```

A `fastest` router with latency switched off, reported as a successful override. And your point about the test is the sharp one: the `viper.Set` line exists only to mirror what the split source separates, so the test passes in exactly the world where production breaks.

**Not fixing it in this PR.** The reasoning, so you can push back if you disagree:

- It cannot happen today. I checked: one production caller (`cmd.Flags()`), and all four weight flags are `BindPFlag`'d.
- It cannot happen tomorrow without a second, independent mistake — and forgetting `BindPFlag` on a new flag breaks that flag *everywhere* it is read through viper, not only here, so it is unlikely to reach production silently.
- This branch is already carrying two review-driven behaviour fixes plus the normalisation above. Adding a refactor with no user-visible effect extends the review surface for something you yourself characterised as latent.

It goes to the follow-up ticket alongside the `Mode`-to-metrics gap and the missing cobra-parse test for `--qos-selection-mode`. If you think it belongs before merge rather than after, say so and I will take it — it is a small change and I am not attached to the call, only to not growing this PR further without a reason a user would notice.

Worth recording for whoever picks it up: the fix is to read the value from whichever source answered the gate — `flags.GetFloat64` when `Changed()` is true, viper when `InConfig()` is true — after which the `viper.Set` line in `TestResolveSelectionWeights_CLIFlagBeatsPreset` should be deletable and the test should still pass. That deletion is the real assertion, as you said.

avitenzer (approved):

PR #263, opened 2026-08-10
refactor(provideroptimizer): rename WeightedSelector to ProviderSelector

Tomelia1999:
Stacked on #262 — **review that one first**. Base is `feat/provider-selection-mode` so this diff shows only the rename; retarget to `main` once #262 merges.

## Why

#262 added a best-score selection policy, so `WeightedSelector` now names a type that does weighted random selection *and* highest-score selection. Scoring and reporting are shared; only the final pick varies by mode. The name promises one of the two.

I deliberately kept this out of #262 — a repo-wide mechanical rename mixed into a feature diff makes the feature impossible to review.

## Identifier map

| before | after |
|---|---|
| `WeightedSelector` | `ProviderSelector` |
| `WeightedSelectorConfig` | `ProviderSelectorConfig` |
| `NewWeightedSelector` | `NewProviderSelector` |
| `DefaultWeightedSelectorConfig` | `DefaultProviderSelectorConfig` |
| `ConfigureWeightedSelector` | `ConfigureProviderSelector` |
| `GetWeightedSelectorConfig` | `GetProviderSelectorConfig` |
| `UpdateWeightedSelectorStrategy` | `UpdateProviderSelectorStrategy` |
| `po.weightedSelector` | `po.providerSelector` |
| `weighted_selector.go` | `provider_selector.go` (and `_test`) |

## Deliberately NOT renamed

These genuinely describe the weighted draw and stay correct:

- `pickWeightedIndex` — this one really is the weighted selection
- `SelectionModeWeightedRandom` / the `weighted_random` flag value
- `updateDecayingWeightedAverage` — unrelated EWMA helper
- `"weighted selection"` prose in tests that exercise the default mode
- `"weighted composite score"` / `"weighted contributions"` comments — the composite genuinely is a weighted sum of the QoS axes

## The six non-mechanical edits

Everything else is a pure identifier substitution. These six exist only because the old name had leaked into prose:

1. The `ProviderSelector` doc comment still claimed the type "implements continuous weighted random selection" — now only half true, so it's rewritten to say scoring is shared and only the pick varies.
2–4. Three weight-validation warning strings named the old type (`"invalid weighted selector weight"` → `"invalid provider selector weight"`, etc.). These are startup config-validation warnings, not hot-path logs.
5. The `SetDeterministicSeed` comment named the old type.
6. The fallback comment in `NewProviderSelector` now lists `SelectionMode` alongside `Strategy`/`MinSelectionChance` as user choices preserved when invalid weights reset to defaults.

[...]

PR #264, opened 2026-08-10
feat(provideroptimizer): add --qos-selection-priority weight presets

Tomelia1999:
Third in the stack — #262 → #263 → **this**. Base is `refactor/rename-provider-selector`; retarget down the chain as each merges.

Implements [FAILOVER-TASKS](../blob/main/agent_docs/bug-reports/failover/FAILOVER-TASKS.md) section 1's `selection` setting.

## What

```
```

|---|---|---|---|---|

**Dominant, not exclusive.** The chosen axis takes 0.70, but availability keeps real weight in every preset so a fast-but-flaky provider cannot outrank a fast-and-solid one on latency alone. `balanced` is byte-identical to today's defaults — it's the flag's default value, so anything else would silently move every existing deployment.

## Why the axis presets zero stake

`CalcWeightsByStake` hands every static provider an identical weight (`maxWeight * 10`, and `maxWeight` stays 1 in an all-static deployment). So the stake term is **the same constant for every candidate**:

- under `weighted_random` it actively flattens the lottery — every provider sits on the same pedestal, compressing the probability ratio between a great provider and a mediocre one

## Composes with `--qos-selection-mode`, doesn't duplicate it

| flag | decides |
|---|---|

## Precedence

## Drive-by fix the new tests surfaced

`ParseSelectionMode` (from #262) rejected the empty string, which tied startup correctness to the flag being viper-bound: any call site that hadn't bound it reads `""` and aborts with a confusing `invalid selection mode:` instead of using the default. **Empty now means "not specified"** and resolves to the default in both parsers. Genuinely wrong values (`bset`, `quickest`) still hard-fail — that's the property that actually matters.

This supersedes the "invalid values abort at startup" note in #262 on the empty-string case only.

## `--strategy` deliberately untouched

I measured what it does today: its only scoring effect is an exponent nudge worth **at most +0.015 on the composite** (typically +0.006), and it does not move the weights at all. `StrategyBalanced` — the default, and the only reachable value for YAML users since the flag has no `BindPFlag` — is a genuine no-op in the switch. So it cannot interfere with this flag.

## Verification

- `go build ./...`, `go vet ./...` — clean
- `go test ./...` — zero failures repo-wide
- `--help` output verified

Tests include the end-to-end proof that the preset is more than a table of numbers: given the same two providers, `fastest` and `most-reliable` pick **different** ones. Plus the sum-to-1.0 invariant (a drifted preset would be silently renormalised by `NewProviderSelector`), the `balanced == today's defaults` backward-compat pin, and both halves of the precedence rule.

### DIFF
--- a/protocol/common/cobra_common.go
+++ b/protocol/common/cobra_common.go
@@ -96,6 +96,7 @@ const (
 	ProviderOptimizerStakeWeight        = "qos-stake-weight"         // weight for stake (default: 0.1)
 	ProviderOptimizerMinSelectionChance = "qos-min-selection-chance" // minimum selection probability for any provider (default: 0.01)
 	ProviderOptimizerSelectionMode      = "qos-selection-mode"       // how the winner is picked from the scored providers: weighted-random (default) or best
+	ProviderOptimizerSelectionPriority  = "qos-selection-priority"   // preset over the four weights above: balanced (default), most-reliable, fastest, freshest
 
 	// optimizer qos sampling cadence — drives the in-memory /metrics
 	// selection-score cache and the OTel optimizer_qos emit.
--- a/protocol/provideroptimizer/selection_mode.go
+++ b/protocol/provideroptimizer/selection_mode.go
@@ -45,7 +45,16 @@ func SelectionModeNames() []string {
 // Unknown values are rejected rather than silently defaulted: the mode dispatch treats
 // anything that is not SelectionModeBest as weighted-random, so a typo'd flag would
 // otherwise leave an operator who asked for "best" quietly running the old policy.
+//
+// An empty string is the one exception — it means "not specified" and resolves to the
+// default. Rejecting it would tie startup to the flag being viper-bound: any call site
+// that has not bound the flag reads "" and would abort with a confusing "invalid
+// selection mode:" rather than simply using the default.
 func ParseSelectionMode(str string) (SelectionMode, error) {
+	if strings.TrimSpace(str) == "" {
+		return SelectionModeWeightedRandom, nil
+	}
+
 	normalized := strings.ReplaceAll(str, "-", "_")
 	for _, mode := range []SelectionMode{SelectionModeWeightedRandom, SelectionModeBest} {
 		if strings.EqualFold(normalized, mode.String()) {
--- a/protocol/provideroptimizer/selection_mode_test.go
+++ b/protocol/provideroptimizer/selection_mode_test.go
@@ -28,11 +28,19 @@ func TestParseSelectionMode(t *testing.T) {
 	// Unknown values must be rejected, not defaulted: the selection dispatch treats
 	// anything that is not SelectionModeBest as weighted-random, so a silent fallback
 	// would leave an operator who asked for "best" running the old policy unnoticed.
-	for _, in := range []string{"", "bset", "greedy", "highest"} {
+	for _, in := range []string{"bset", "greedy", "highest"} {
 		t.Run("invalid/"+in, func(t *testing.T) {
 			_, err := ParseSelectionMode(in)
 			require.Error(t, err)
 		})
+	}
+
+	// Empty means "not specified" and must resolve to the default rather than error —
+	// otherwise any caller that has not viper-bound the flag reads "" and aborts.
+	for _, in := range []string{"", "   "} {
+		got, err := ParseSelectionMode(in)
+		require.NoError(t, err)
+		require.Equal(t, SelectionModeWeightedRandom, got)
 	}
 }
 
--- a/protocol/provideroptimizer/selection_priority.go
+++ b/protocol/provideroptimizer/selection_priority.go
@@ -1,0 +1,149 @@
+package provideroptimizer
+
+import (
+	"fmt"
+	"strings"
+)
+
+// SelectionPriority is a named preset over the four QoS weights that feed
+// CalculateScore. It answers "what should the router optimise for", while
+// SelectionMode answers "how is the winner picked from those scores" — the two are
+// orthogonal and compose:
+//
+//	--qos-selection-priority fastest --qos-selection-mode best
+//	  → sort providers by latency and always send to the head of the queue
+//
+//	--qos-selection-priority fastest --qos-selection-mode weighted_random
+//	  → a lottery biased heavily toward the fastest providers
+//
+// The preset only sets weights; it adds no scoring machinery. Operators who want
+// something other than the four presets can still set the individual
+// --qos-*-weight flags, which take precedence over whatever the preset chose.
+type SelectionPriority int
+
+const (
+	// SelectionPriorityBalanced is the zero value and keeps the historical default
+	// weights, so an unset flag changes nothing for existing deployments.
+	SelectionPriorityBalanced SelectionPriority = iota
+
+	// SelectionPriorityMostReliable favours the provider that answers successfully
+	// most often.
+	SelectionPriorityMostReliable
+
+	// SelectionPriorityFastest favours the provider that answers quickest.
+	SelectionPriorityFastest
+
+	// SelectionPriorityFreshest favours the provider closest to the head of the chain.
+	SelectionPriorityFreshest
+)
+
+// priorityWeights holds the four QoS weights a preset applies. Each set sums to 1.0,
+// which keeps NewProviderSelector's normaliser from rescaling them.
+type priorityWeights struct {
+	availability float64
+	latency      float64
+	sync         float64
+	stake        float64
+}
+
+// presetWeights maps each priority to its weights.
+//
+// The three axis presets are deliberately dominant rather than exclusive: the chosen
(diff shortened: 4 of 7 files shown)


avitenzer (commented):
Reviewed at `40b7811`. Verified locally: `go build ./...` and `go vet` clean, new priority tests plus the sibling `rpcsmartrouter` suites all pass.

Scope note: the diff is against `refactor/rename-provider-selector` (#263), so this covers the priority feature only, not the rename.

**Verdict: approve with changes.** The design is the right shape — a named preset over the existing weights rather than a new scoring path, with `balanced` byte-identical to today's defaults so it's a true no-op until someone opts in. Three code items and two process items below.

---

## What I verified

- **No viper env binding anywhere in `protocol/` or `cmd/`**, so `Changed() || InConfig()` genuinely covers every input path. The precedence rule has no third-channel gap — worth stating since that's the kind of thing that silently breaks later.
- **`withCleanViper`'s `viper.Reset()` is safe.** The two sibling tests touching viper in this package (`cross_validation_policy_test.go`, `static_provider_endpoints_test.go`) all use isolated `viper.New()` instances, so the global reset can't clobber them.
- `ApplyTo` really does leave `SelectionMode`, `MinSelectionChance`, `Strategy` and the adaptive wiring alone, and there's a test pinning exactly that.

`TestSelectionPriorityChangesTheWinner` is the test that earns its keep here. A preset is just a map literal, so it's easy to write four tests that assert the map contains the numbers you typed into the map — proving nothing. Constructing two providers and showing `fastest` and `most-reliable` pick *different* ones is the only claim an operator cares about. The comment explaining why the latency gap is deliberately extreme (20s vs 0.05s) is also the right instinct: with the fixed-max fallback the normalised latency range is compressed, so a realistic sub-second gap would be swamped by the availability rescale of `[0.8,1.0] → [0,1]`. Encoding that in the test comment means whoever next touches the normaliser will understand why the numbers look artificial.

---

## 1. Should fix: the doc comment for `routerConfigOptimizerWeights` got orphaned

`resolveSelectionWeights` was inserted *between* an existing type's doc comment and the type itself (`rpcsmartrouter.go:689-700`):

```go
// routerConfigOptimizerWeights is the JSON shape for a single provider-optimizer's
// active selection weights, used for each per-chain entry in the PerChainOptimizer map
// returned by GET /debug/runtime-config. Fields carry no json tags, so each key marshals
// as the bare Go identifier — tests grep the same string in test code and router source.
// resolveSelectionWeights turns --qos-selection-priority plus the individual   // <- new
...
func resolveSelectionWeights(...)

type routerConfigOptimizerWeights struct {   // <- now undocumented
```

Godoc merges those into one block attached to `resolveSelectionWeights`, and the type loses its docs. Moving the function below the type (or above the comment) fixes it. Purely mechanical, but the orphaned comment is actively misleading — it describes JSON marshalling that has nothing to do with the function it now sits on.

## 2. Should fix: the headline precedence workflow silently renormalises

The PR sells "the preset is a starting point, never a cage." But `NewProviderSelector` still renormalises any weight set that doesn't sum to 1.0 (`provider_selector.go:186`), and overriding one axis of a preset breaks the sum *by construction*:

```
--qos-selection-priority fastest --qos-latency-weight 0.5
  config:     avail 0.20  lat 0.50   sync 0.10  stake 0.00   (sum 0.80)
  effective:  avail 0.25  lat 0.625  sync 0.125 stake 0.00   <- rescaled
```

The operator asked for `0.5` and gets `0.625`. And `resolveSelectionWeights` logs the **pre-normalisation** number:

```go
utils.LavaFormatInfo("Working with provider selection priority: "+priority.String(),
    utils.LogAttr("latencyWeight", config.LatencyWeight),   // prints 0.5, selector runs 0.625
```

The normaliser is pre-existing and not this PR's doing — but this PR is what moves it from an edge case onto the main path. Before, all four weights came from defaults summing to 1.0, so an operator overriding one was visibly off-script. Now "pick a preset, tweak one axis" is the *documented* workflow, and it produces a non-unit sum every time.

`TestResolveSelectionWeights_CLIFlagBeatsPreset` asserts `config.LatencyWeight == 0.5`, which is true of the config struct and false of the effective weight — so it passes either way. It's pinning the wrong layer.

Cheapest fix: renormalise inside `resolveSelectionWeights` and log the effective values, so the Info line matches what the selector actually runs. Failing that, note in the flag help that weights are relative and get rescaled to sum to 1.0.

## 3. Should fix: the stake-zeroing justification is over-broad

The description argues the axis presets can zero stake because `CalcWeightsByStake` hands every static provider an identical `maxWeight * 10`. That holds only when no explicit stake is configured (`consumer_types.go:1358`):

```go
if cswp.StaticProvider && stakeOmitted {      // stakeOmitted == (stakeAmount == 0)
    staticProvidersToBoost = append(...)
    continue
}
...
weights[cswp.PublicLavaAddress] = stake       // :1365 — differs per provider
```

If an operator sets explicit per-provider stake — a path the code deliberately supports, with the comment "treat the static provider like any other provider" — then stake **is** a discriminating signal, and `fastest` / `freshest` / `most-reliable` silently discard it.

The design choice may well still be right; someone asking for `fastest` plausibly wants stake ignored. It's the stated reasoning that doesn't hold, and it's load-bearing for three of the four presets. Suggest narrowing the claim to "in a static deployment with no explicit stake."

## 4. Process: CI is not running lint, govulncheck, or the Jira gate

`gh pr checks 264` returns **10 checks; #262 returned 20.** Missing: `lint`, `govulncheck`, `Validate Jira ticket`, `jira-ticket/validated`. Those workflows are gated on `base: main`, and this PR targets `refactor/rename-provider-selector`.

So the all-green tick is misleading — this code has not been linted or vulnerability-scanned, and the Jira gate that blocked #262 hasn't evaluated it. All of them fire on retarget to main. Worth re-checking at that point rather than reading the current green as ready.

## 5. Process: the `FAILOVER-TASKS` link 404s

The body links `[FAILOVER-TASKS](../blob/main/agent_docs/bug-reports/failover/FAILOVER-TASKS.md)`. `agent_docs/` is gitignored (`.gitignore:62`), and `git ls-files | grep -i failover` returns nothing — the file isn't in the repo, so no reviewer can open it.

Since that doc is the stated source of the requirement ("implements section 1's `selection` setting"), the rationale it carries should be inlined in the PR body instead of linked.

## Nits

- **Canonical spellings diverge across sibling flags.** Mode is snake_case (`weighted_random`), priority is kebab-case (`most-reliable`) — and the two parsers normalise in *opposite* directions (`-`→`_` vs `_`→`-`). Both work, but `/debug/runtime-config` will end up reporting one of each style. Cheap to align now, awkward once either is public.
- **`/debug/runtime-config` doesn't expose the priority.** #262 added `SelectionMode` there specifically for observability. The four weights are exposed so the preset is mostly inferable — but not once an axis has been overridden, which is exactly when you'd want to know.
- **Empty-string change is well-argued** and the comment explains it properly. One consequence not stated: an operator explicitly passing `--qos-selection-mode ""` now silently gets `weighted_random` rather than an error. Negligible in practice, but it is a real narrowing of #262's fail-fast guarantee.

## Tests

Strong overall. `TestParseSelectionPriority` rejecting `latency` and `sync-freshness` is a nice touch — those are the `--strategy` spellings and the single most plausible wrong guess for this flag. The sum-to-1.0 invariant test is well-motivated (a drifted preset would be silently rescaled). `TestResolveSelectionWeights_ConfigFileBeatsPreset` explicitly asserts `Changed()` is false for every flag before calling, which pins the exact bug the `InConfig` half exists to prevent.

The one gap is the layer issue in finding 2: nothing asserts the *effective* weights after `NewProviderSelector`, so the renormalisation is invisible to the suite.

---

**Summary:** move `resolveSelectionWeights` below the type it displaced; decide between renormalise-and-log or documenting the rescale; narrow the stake justification; inline the `agent_docs` content; and re-check CI after retargeting to main.

avitenzer (approved):

avitenzer:
Superseded by #282, which consolidates this stack (#262 → #263 → #264) into a single PR rebased onto current `main` (zero conflicts).

One change versus this stack: the rename in #263 now targets `EndpointSelector` rather than `ProviderSelector`, addressing @nimrod-teich's request to move away from "provider" terminology. "Endpoint" is the term the codebase is already converging on (`protocol/endpointstate`, `EndpointMetrics`, `GetAllDirectRPCEndpoints`, `EndpointLagThreshold`).

Tracked under [MAG-2780](https://magmadevs.atlassian.net/browse/MAG-2780) (new subtask of MAG-2599 Failover v2), which also clears the Jira gate that was failing here.

Closing in favour of #282.

[MAG-2780]: https://magmadevs.atlassian.net/browse/MAG-2780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #265, opened 2026-08-11
fix(smart-router): boot on any usable provider instead of exiting (MAG-2525)

avitenzer:
# Description

The router exited at startup when every static provider failed verification, even with healthy backups configured. A restart during a failover — exactly when a restart is most likely — turned a handled failover into an outage. An all-down window at boot became a CrashLoopBackOff that never self-healed, because a crash-looping pod can't adopt a provider that comes back.

Reported twice: MAG-2532 (customer-facing, Kraken/BSC — primaries down, backups healthy, deploy restarted the pod, router refused to start) and MAG-2525 (total wipe-out at boot → crash loop). Both trace to the same check. MAG-2532 is a strict subset, so this closes both.

**Provider health is a runtime condition, not a configuration error.** The router already tolerates every provider dropping while it runs: it stays up, returns 5xx, and re-adopts providers as they recover. Boot now behaves the same way. Only a chain with nothing *configured* still exits.

Per chain + api-interface:

| healthy primaries | healthy backups | result |
| --- | --- | --- |
| ≥1 | any | serves primaries |
| 0 | ≥1 | serves backups (degraded) |
| 0 | 0 | boots, reports unhealthy, 500s on relays, retries |
| *none configured* | *none configured* | exits non-zero |

## Replacing the signal that fail-fast carried

Removing the fatal also removes what it told operators, so two things replace it:

- **The relays monitor is seeded from the boot verdict.** It defaults optimistic (`isHealthy: 1`), which would have answered `200` on `/lava/health` for a chain that came up dark, until the first health relay resolved an interval later. `/readyz` was already fail-closed and is unaffected.
- **New `smartrouter_endpoint_serving_tier` gauge** — 2=primaries, 1=degraded on backups only, 0=dark. Alert on `<2` and `==0`, now that pod restarts no longer signal a dark chain. Republished from every path that mutates the live pairing, not just boot: a stale reading would page forever after a recovery, or stay silent after a demotion.

## Recovery cadence

No longer a flat 3m. A **dark** chain retries from ~2s doubling to the same 3m ceiling, then decays back once serving — a boot-time blip costs seconds instead of minutes. A merely **degraded** chain keeps the flat interval; retries there buy redundancy, not availability. Backups joined that loop too; they previously recovered only on the 15m epoch reverification, which was tolerable only while a chain could never boot on backups alone.

## Three adjacent defects on the same path

All get worse once backups are load-bearing at boot:

- **Backup validation had no timeout** — `context.WithCancel` where the static tier used a 30s `WithTimeout`, so a blackholed backup hung bring-up indefinitely. Both tiers now share `BootValidateTimeout`.
- **Boot validation was sequential** at up to 30s per provider, so three dead primaries cost 90s before a healthy backup was even dialled. Both tiers now validate in parallel, bounded by `SpecReVerifyConcurrency`.
- **Boot validation didn't clone the chain parser**, which the epoch reverifier already does to stop gRPC verification mutating the live parser. Sharing `validateProvider` picks that up.

That sharing also collapses three near-identical copies of the verification logic (both boot tiers and the retry loop) into one — most of the 400 deleted lines.

## Review follow-up

Two defects Anna's review surfaced, fixed in `54832ad`:

- **Subscription tiers were frozen at the boot-time healthy lists.** Nothing updated them afterwards, which was survivable only while a chain could not boot without a healthy static provider. Once it can, both scenarios above became permanent subscription outages: a dark boot installed the NoOp WS manager and left `grpcSubscriptionManager` nil for the process lifetime, and a backups-only boot never promoted a recovered primary. HTTP relays recover in both cases, so nothing surfaced it until a client opened a subscription. Which manager is installed is now decided from the *configured* providers, and `SetEndpoints` repopulates the tiers from every path that mutates the live pairing. The tiers stay healthy-only — seeding them from the configured list would hand a subscription a dead endpoint with no fallback and suppress the primary→backup cascade.
- **The epoch promote path left providers in the failed lists**, so the retry loop later merged a second session for the same provider (`mergeRecoveredSessions` keys by index, not by `PublicLavaAddress`), doubling its weight in `validAddresses`. `applyReverification` now surfaces promoted names and `updateEpoch` prunes them per tier.

## Testing

- Full suite: 33/33 packages pass. `-race` clean on `rpcsmartrouter`. `go vet` (the repo's `make lint`) and `golangci-lint` clean on changed files.
- New unit tests in `boot_resilience_test.go` and `serving_tier_metrics_test.go` cover tier partitioning/ordering/concurrency, the dark verdict, adaptive cadence, re-admission merging, the health seed, and the gauge (including that it falls back down after recovery rather than acting as a high-water mark).
- **Verified end-to-end on k3d** against a live cluster. Two integration tests in `smart-router-automation` (`mag-2525-2532-boot-resilience`, PR to follow) pass against this branch and **fail** against a router built without it, so they're real regression guards:
  - MAG-2532 — primaries down, backups healthy → `rollout completed in 7.2s on backups alone`, then `HTTP 200 served by backup tier: ['simbackup1']`
  - MAG-2525 — everything down → boots, 500s cleanly, adopts primaries once healthy

Two pre-existing issues found and deliberately left alone: a data race in `relays_monitor_test.go` (`TestHappyFlow`, confirmed on main) and two `golangci-lint` findings in untouched code (`spec_reverifier.go:153`, `rpcsmartrouter.go:2723`, both confirmed on main).

## Not included

`errCh` handling at `rpcsmartrouter.go:330-333` still exits the process on the first endpoint error, so one chain's *config* error takes down every chain. The availability half of that is gone — only config errors reach `errCh` now — and whether a config error should still fail the whole process reads like an intentional choice, so I left it. Happy to change if not.

## Jira ticket

Jira ticket: MAG-2525

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change — n/a, no API or client break. `smartrouter_endpoint_serving_tier` is additive; the boot behavior change is the fix itself and is documented above.
* [x] provided a link to the relevant issue or specification — MAG-2525 carries the agreed behavior spec in a comment; MAG-2532 tracks the customer report.
* [ ] confirmed all CI checks have passed — pending first run.

## Reviewers Checklist

I have...

### DIFF
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -258,6 +258,34 @@ assert `/debug/reset-all` emptied each store (see MAG-1762). All drop to `0` aft
 | `smartrouter_csm_sticky_sessions` | Gauge | `spec`, `apiInterface` | Live sticky-session affinities. |
 | `smartrouter_csm_reported_providers` | Gauge | `spec`, `apiInterface` | Size of the reported-providers register. |
 
+#### Serving tier (availability)
+
+| Metric | Type | Labels | Description |
+| --- | --- | --- | --- |
+| `smartrouter_endpoint_serving_tier` | Gauge | `spec`, `apiInterface` | Which provider tier the endpoint is serving from: `2` = primaries, `1` = degraded (backups only), `0` = dark (no healthy providers). |
+
+Before MAG-2525 an endpoint with no healthy provider exited the process, so a
+CrashLoopBackOff was the de-facto alert. It now boots and reports unhealthy instead,
+which is what makes a restart during a failover survivable — and it means **this gauge,
+not pod restarts, is the signal that a chain cannot serve**. It is republished from
+every path that mutates the live pairing (boot, background retry, epoch
+promote/demote, config reload), so it tracks the current state rather than the boot
+verdict.
+
+Suggested alerts:
+
+| Condition | Meaning |
+| --- | --- |
+| `smartrouter_endpoint_serving_tier == 0` | Endpoint is dark — all relays 5xx. Page. |
+| `smartrouter_endpoint_serving_tier < 2` | Serving on backups only. Redundancy is gone; the next failure is an outage. |
+
+Sizing the `for:` window: recovery from `0` is not one cadence. A chain that was **dark
+at boot** is retried on an adaptive schedule starting at ~2s and doubling to a 3m
+ceiling, so it typically self-heals in seconds. A chain that booted healthy and was
+later **demoted to dark** has no retry loop — demoted providers are not fed into the
+failed-provider lists — and is only re-checked by the 15m epoch re-verifier. Alert
+windows should assume the 15m case.
+
 ---
 
 ## Shared metrics
--- a/protocol/metrics/relays_monitor.go
+++ b/protocol/metrics/relays_monitor.go
@@ -31,6 +31,27 @@ func NewRelaysMonitor(interval time.Duration, chainID, apiInterface string) *Rel
 	}
 }
 
+// SeedInitialHealth overrides the optimistic default before Start runs.
+//
+// The default assumes healthy because, for a normally-booting chain, nothing is
+// known until the first health relay completes and reporting 503 in that window
+// would fail readiness for every rollout. But when startup validation already
+// found zero usable providers, that optimism is a lie: the endpoint would answer
+// 200 on its health path while every relay 503s, until the first health relay
+// resolves an interval later. Since MAG-2525 a chain in that state boots instead
+// of exiting, so this is the difference between being pulled from rotation and
+// silently accepting traffic it cannot serve.
+//
+// Must be called before Start; Start's immediate probe overwrites this with the
+// real verdict, which is the intent — the seed only covers the boot window.
+func (sem *RelaysMonitor) SeedInitialHealth(healthy bool) {
+	if sem == nil {
+		return
+	}
+
+	sem.storeHealthStatus(healthy)
+}
+
 func (sem *RelaysMonitor) SetRelaySender(relaySender func() (bool, error)) {
 	if sem == nil {
 		return
--- a/protocol/metrics/serving_tier_metrics_test.go
+++ b/protocol/metrics/serving_tier_metrics_test.go
@@ -1,0 +1,101 @@
+package metrics
+
+import (
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+func newSmartRouterForServingTierTest() *SmartRouterMetricsManager {
+	return &SmartRouterMetricsManager{
+		endpointServingTier: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{Name: "t_sr_endpoint_serving_tier"},
+			[]string{"spec", "apiInterface"},
+		),
+		urlToProviderNames: make(map[string][]string),
+	}
+}
+
+func TestServingTier_Mapping(t *testing.T) {
+	for _, tc := range []struct {
+		name                         string
+		healthyStatic, healthyBackup int
+		want                         int
+	}{
+		{"serving from primaries", 3, 2, ServingTierPrimary},
+		{"primaries healthy, no backups configured", 1, 0, ServingTierPrimary},
+		{"all primaries down, backups healthy", 0, 2, ServingTierDegraded},
+		{"backup-only configuration", 0, 1, ServingTierDegraded},
+		{"nothing healthy", 0, 0, ServingTierDark},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			require.Equal(t, tc.want, ServingTier(tc.healthyStatic, tc.healthyBackup))
+		})
+	}
+}
+
+// Operators alert on `< 2` for degraded and `== 0` for dark, so the ordering is
+// part of the contract rather than an implementation detail.
+func TestServingTier_ConstantOrdering(t *testing.T) {
+	require.Equal(t, 0, ServingTierDark)
+	require.Equal(t, 1, ServingTierDegraded)
+	require.Equal(t, 2, ServingTierPrimary)
+	require.Less(t, ServingTierDark, ServingTierDegraded)
+	require.Less(t, ServingTierDegraded, ServingTierPrimary)
+}
+
+func TestSetEndpointServingTier_PublishesGauge(t *testing.T) {
+	m := newSmartRouterForServingTierTest()
+	gauge := m.endpointServingTier.WithLabelValues("BSC", "jsonrpc")
+
+	m.SetEndpointServingTier("BSC", "jsonrpc", 3, 2)
+	require.Equal(t, float64(ServingTierPrimary), testutil.ToFloat64(gauge))
+
+	m.SetEndpointServingTier("BSC", "jsonrpc", 0, 2)
+	require.Equal(t, float64(ServingTierDegraded), testutil.ToFloat64(gauge),
+		"all primaries down but backups serving is degraded, not dark")
+
+	m.SetEndpointServingTier("BSC", "jsonrpc", 0, 0)
+	require.Equal(t, float64(ServingTierDark), testutil.ToFloat64(gauge))
+}
+
+// The gauge is a gauge, not a high-water mark: it has to fall back down when a
+// provider recovers. It is set once at boot and then republished from every path
+// that mutates the session maps, so a stale reading would either page forever
+// after a recovery or stay silent after a demotion.
+func TestSetEndpointServingTier_RecoversAfterDark(t *testing.T) {
+	m := newSmartRouterForServingTierTest()
+	gauge := m.endpointServingTier.WithLabelValues("BSC", "jsonrpc")
+
+	m.SetEndpointServingTier("BSC", "jsonrpc", 0, 0) // booted dark
+	require.Equal(t, float64(ServingTierDark), testutil.ToFloat64(gauge))
+
+	m.SetEndpointServingTier("BSC", "jsonrpc", 0, 1) // a backup came back
+	require.Equal(t, float64(ServingTierDegraded), testutil.ToFloat64(gauge))
+
+	m.SetEndpointServingTier("BSC", "jsonrpc", 2, 1) // primaries followed
+	require.Equal(t, float64(ServingTierPrimary), testutil.ToFloat64(gauge))
+
+	m.SetEndpointServingTier("BSC", "jsonrpc", 0, 0) // and back down on demotion
+	require.Equal(t, float64(ServingTierDark), testutil.ToFloat64(gauge))
+}
+
+func TestSetEndpointServingTier_ChainsAreLabelledSeparately(t *testing.T) {
+	m := newSmartRouterForServingTierTest()
+
+	m.SetEndpointServingTier("BSC", "jsonrpc", 0, 0)
+	m.SetEndpointServingTier("ETH1", "jsonrpc", 3, 0)
+
+	require.Equal(t, float64(ServingTierDark),
+		testutil.ToFloat64(m.endpointServingTier.WithLabelValues("BSC", "jsonrpc")))
+	require.Equal(t, float64(ServingTierPrimary),
+		testutil.ToFloat64(m.endpointServingTier.WithLabelValues("ETH1", "jsonrpc")),
+		"one dark chain must not drag down another chain's reading")
+}
+
+func TestSetEndpointServingTier_NilManagerIsSafe(t *testing.T) {
+	var m *SmartRouterMetricsManager
+	require.NotPanics(t, func() { m.SetEndpointServingTier("BSC", "jsonrpc", 0, 0) })
+}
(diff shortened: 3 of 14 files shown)


AnnaR-prog:
## Review

The design is right and well argued — boot now matches the runtime posture the router already had, and treating provider health as a runtime condition rather than a config error is clearly correct. I verified the readiness story holds up: `/readyz` is fail-closed and reflects serving capacity without restarting the pod (`smartrouter_metrics_manager.go:741-745`), `/lava/health` nil-guards to `false` when `--relays-health-enable=false` (`relays_monitor.go:112-116`), so no configuration answers 200 for a chain that came up dark. `UpdateAllProviders` handles an empty pairing without error, and `relevantStaticProviderList` is already interface-filtered at `rpcsmartrouter.go:1929-1935`, so dropping the redundant inner filter is correct and the healthy-count arithmetic is sound.

`go test ./protocol/rpcsmartrouter/ ./protocol/metrics/ -count=1` passes, the new tests are `-race` clean, `go build ./...` clean.

Three things below. I've left out a handful of small ones (a stale comment, an orphaned doc comment, a missing `docs/METRICS.md` row for the new gauge, a wording fix in the PR description) — happy to add those separately if useful.

---

### 1. A dark boot permanently disables subscriptions

`rpcsmartrouter.go:2358-2359, 2398-2399`

```go
wsEndpoints := collectWSEndpoints(healthyStaticProviders, "primary")
wsBackupEndpoints := collectWSEndpoints(healthyBackupProviders, "backup")
```

Both subscription managers are built once, from the **boot-time healthy** lists, and never updated. `wsEndpoints` is stored in the constructor (`direct_ws_subscription_manager.go:176`) and only ever read by `selectFromTier` (`:332`). There is exactly one construction site (`:2364` / `:2386`) and one assignment into the server (`rpcsmartrouter_server.go:128`). Nothing in `retryFailedProviders`, `readmitRecoveredProviders`, `updateEpoch`, or `rebuildPairingFromConfig` touches them.

So in this PR's own target scenarios:

- **MAG-2525 (everything down at boot):** both lists are empty → `NewNoOpWSSubscriptionManager` is installed and `grpcSubscriptionManager` stays nil (`:2405-2416` never runs). Every `eth_subscribe` returns *"no WebSocket endpoints configured"* **for the lifetime of the process**, and `GetReflectionConnection` takes the nil branch (`rpcsmartrouter_server.go:308`) forever — even after every provider recovers and HTTP relays are serving normally.
- **MAG-2532 (primaries down, backups healthy):** the WS manager is built with an empty primary tier. Recovered primaries never join it, so subscriptions stay pinned to backups permanently.

Before this PR the all-empty case couldn't occur — the process exited instead. The partial case existed, but always left at least one healthy WS endpoint. This change makes the fully-empty variant reachable *and* permanent, which converts a self-healing boot outage into a silent, non-self-healing subscription outage — against the PR's stated goal that the router "re-adopts providers as they recover".

**The fix is an update path, not a wider initial list.** A `SetEndpoints`-style update (guarded, rebuilding `endpointsByURL` alongside the tier slices), called from `readmitRecoveredProviders` and `updateEpoch`'s promote/demote — the same places that already call `publishServingTierLocked`.

Worth flagging the tempting one-liner explicitly, because it looks right and isn't: **do not build the lists from the configured providers instead of the healthy ones.** `selectEndpoint` is called exactly once per subscription, with `ignoredEndpoints == nil` (`:559`), and there is no retry-across-endpoints loop — a failed dial reports to the optimizer and returns an error to the client (`:574-583`). And `selectFromTier` only errors when the tier is **empty or fully ignored** (`:353-358`), which is the sole trigger for the primary→backup cascade (`:332-345`). So listing dead primaries would hand some subscriptions a dead endpoint with no fallback *and* suppress the backup cascade entirely — strictly worse than today in the MAG-2532 shape, where the empty primary tier is exactly what makes the cascade fire.

If the update path doesn't fit this PR's scope, this belongs in **Not included** with a ticket — it shouldn't ship silently, because the failure is invisible until a customer opens a subscription.

---

### 2. The epoch promote path doesn't prune the failed lists → duplicate sessions for one provider

`rpcsmartrouter.go:3335-3336` vs `:3491-3492`, `spec_reverifier.go:178, 237`

Only three sites write the failed lists: boot (`:2309-2314`), `readmitRecoveredProviders` (`:3335-3336`), and `pruneRestoredFromFailed` from `rebuildPairingFromConfig` (`:3491-3492`). `applyReverification`'s **promote** path is not among them — it moves a provider back into `providerSessions` (`spec_reverifier.go:237+`) but leaves it in `failedStaticProviders`.

Sequence:

1. Provider P fails boot verification → `failedStaticProviders[key] = [P]`, retry loop launched.
2. Epoch tick (~15m): `applyReverification` finds P healthy and absent from `fresh` → promotes it into `providerSessions`.
3. Retry loop's next tick still sees P in the failed list → revalidates → succeeds → `mergeRecoveredSessions` appends a **second** session for the same `PublicLavaAddress` at a fresh index (no dedupe by address).

`csm.pairing` is keyed by address so it dedupes, but `pairingAddresses` / `pairingAddressesLength` (`consumer_session_manager.go:319-326`) and `validAddresses` (`:860-865`) do not — P lands in `validAddresses` twice, doubling its selection weight and inflating the pairing length. The superseded `ConsumerSessionsWithProvider` is dropped without closing its `DirectRPCConnection`.

This is **pre-existing for the static tier**, and I'd normally leave it alone — except that this PR:

- extends the same window to the backup tier (new surface), and
- introduces `pruneRestoredFromFailed` specifically to enforce this invariant, with a comment at `:3480-3483` stating it ("so `retryFailedProviders` won't later merge a duplicate session for the same name"), then wires it into only one of the two admission paths.

The fix is the helper that's already in the diff: surface the promoted names out of `applyReverification` and call `pruneRestoredFromFailed` on both maps in `updateEpoch`, right where `publishServingTierLocked(chainKey)` is called (`:3117`). If you'd rather not widen the PR, a comment noting the asymmetry would at least stop the next reader assuming it's covered.

---

### 3. Scope note: the adaptive cadence only ever applies to a chain that was dark *at boot*

Same root as #2 — the failed lists are written from three places, and `applyReverification`'s **demote** path isn't one of them either. So a chain that boots healthy and later loses every provider to demotion has no retry loop running at all (it either never started, or already self-terminated at `:3245-3251`), and waits up to a full 15m epoch for the reverifier.

Nothing here is wrong — the PR's framing ("a boot-time blip costs seconds instead of minutes") is accurate as scoped. But operators are being told to page on `smartrouter_endpoint_serving_tier == 0`, and that alert fires identically for boot-dark (recovers in ~2s) and demoted-dark (up to 15m). Worth a line in the PR description or the gauge's `Help` text so the runbook doesn't assume seconds. Feeding demoted providers into the failed lists would close the gap properly, but that's a larger change and reasonable to defer.

---

### Testing and CI

The unit tests are good where they reach — `validateProviderTier` partitioning / configured-order / concurrency, the dark verdict, `retryIntervalFor`, the health seed, `mergeRecoveredSessions`, `pruneRestoredFromFailed`, both-tier re-admission, chain isolation, and the gauge including its fall-back-down behaviour. All pass, `-race` clean.

The gap: **nothing in this repo exercises `CreateSmartRouterEndpoint`'s non-fatal path**, which is the headline behaviour change. The claimed regression guards (`mag-2525-2532-boot-resilience`) live in an unmerged PR in `smart-router-automation`, so until that lands, reinstating the fatal breaks no test here. Acceptable if the automation PR lands close behind — worth saying so explicitly rather than leaving the checklist reading as covered.

`govulncheck` is red, but it's a stale base rather than this PR: the branch sits on `golang.org/x/net v0.57.0` while `main` has since bumped to `v0.58.0`, and the same workflow is green on `main`. A rebase should clear it.

---

### Verdict

Approve the direction. #1 is the one I'd want resolved before merge — it turns the scenario this PR exists to fix into a permanent subscription outage, and the fix is contained. #2 is pre-existing but the PR both widens it and already ships the helper that closes it. #3 is a documentation/runbook note.

avitenzer:
Thanks — this was a genuinely useful review. All three findings reproduce; #1 and #2 are fixed in 54832ad, #3 is documented. One correction on your suggested fix for #1, and one on the CI note.

## 1. Dark boot permanently disables subscriptions — fixed

Confirmed exactly as described. Worth stating precisely for the record, though: those two `collectWSEndpoints` lines are byte-identical on `main`, so "built once, never updated" is entirely pre-existing. What this PR changed is *reachability* — the fatal at `rpcsmartrouter.go:2263` guaranteed `healthyStaticProviders` was non-empty, so the fully-empty variant couldn't occur. Doesn't change your conclusion: the PR is what makes the worst form reachable, so it's the PR's job to close it.

**Your fix sketch is incomplete on its own**, and it's worth spelling out why since it's the same shape of trap as the one-liner you flagged. A `SetEndpoints` called from `readmitRecoveredProviders`/`updateEpoch` does nothing for MAG-2525: in that shape boot already installed `NewNoOpWSSubscriptionManager` and left `grpcSubscriptionManager` nil, so there is no Direct manager to call the setter on. The fix needs both halves:

1. **Manager type from the configured providers, not the healthy ones.** A dark boot now gets a real Direct manager with empty tiers. This is not the thing you warned against — the *tiers* stay healthy-only. An empty tier already errors at `selectFromTier`, so behaviour while dark is unchanged; it's just recoverable now. The NoOp manager survives only for the case where no `ws://` URL is configured at all, where it's permanent by definition and its error is the honest answer.
2. **`SetEndpoints`**, wired into `readmitRecoveredProviders`, `updateEpoch`, and `rebuildPairingFromConfig` — every path that calls `publishServingTierLocked`.

Your warning about seeding from configured providers is in the code as a comment, with the reasoning (single `selectEndpoint` call, no ignore set, no retry loop, and the empty primary tier being precisely what fires the cascade). It's the kind of thing that gets "simplified" back in otherwise.

Two knock-on details:

- The tiers are no longer immutable after construction, so selection reads both through one snapshot under the manager's existing `RWMutex`. One snapshot per cascade, not per tier — otherwise a swap between tier 1 and tier 2 yields a mixed generation.
- Existing upstream pools are deliberately left alone on a tier change. A pool whose URL just left the tier still serves its live subscriptions until they close, and `performCleanup` reaps it once idle. Tearing them down on demotion would kill working subscriptions on a provider that merely failed spec re-verification.

The gRPC side got the same treatment, which also unsticks reflection — `GetReflectionConnection` selects through the same cascade.

## 2. Promote path doesn't prune the failed lists — fixed

Confirmed down the chain: `mergeRecoveredSessions` appends at `maxIdx` with no dedupe by `PublicLavaAddress`; `csm.pairingAddresses` is keyed by index (`consumer_session_manager.go:360`); `setValidAddressesToDefaultValue` iterates it and appends, so the address lands twice in `validAddresses`.

`applyReverification` now returns promoted names and `updateEpoch` prunes with the existing `pruneRestoredFromFailed`. One deviation from your sketch: pruned **per tier**, not with the union. `rebuildPairingFromConfig` uses the union, which is only reachable if a name is configured in both tiers — but in that case a recovered static twin would drop a still-failing backup out of the retry loop. There's a test for it.

## 3. Scope note — documented, not fixed

Agreed on both the diagnosis and that feeding demoted providers into the failed lists is a bigger change than belongs here. `docs/METRICS.md` now has a serving-tier section with the two suggested alerts and an explicit note that the `for:` window should be sized for the 15m demoted-dark case, not the ~2s boot-dark one. The gauge's `Help` text says the same, so it's visible from `/metrics` without the doc.

## The small ones you left out

`docs/METRICS.md` had no row for the gauge — added, as above. Please do send the stale comment and orphaned doc comment; I'd rather fix them here than leave them.

## Testing

Your test-coverage criticism was fair and I didn't want to repeat it. Each of the three new behaviours was verified to **fail** with only the corresponding fix reverted — the two promote/prune guards and the `updateEpoch` republish guard all go red, so they're real guards rather than assertions that happen to hold.

Still true that nothing here exercises `CreateSmartRouterEndpoint`'s non-fatal path directly; that remains on the `smart-router-automation` PR, which I'll land close behind.

## CI

The `govulncheck` note is stale — the branch is on `golang.org/x/net v0.58.0` and it's green, along with everything else.

Full suite passes, `-race` clean on `rpcsmartrouter`, `go vet` and `golangci-lint` introduce nothing new (the two findings you'd expect, `spec_reverifier.go` `copyloopvar` and the `appendAssign` in `updateEpoch`, are on pre-existing lines my diff only shifted).

avitenzer:
Rather than wait, I went looking for the three you'd held back. Fixed in 942320e — please confirm these are the ones you meant, since I inferred them:

**Stale comment** — `rpcsmartrouter.go:2481`. The `SilenceUsage: true` comment cited *"all static providers failed verification"* as its example startup failure, which is precisely the case this PR stopped being a startup failure. The flag itself is still right and still worth keeping for the config errors that remain fatal; only the example was wrong. Now names what actually reaches it.

**Orphaned doc comment** — `spec_reverifier.go:304`. This PR inserted `BootValidateTimeout` and `validateProviderTier` between `validateProvider`'s doc comment and `validateProvider`, leaving the doc ~90 lines above its function and reading as a preamble to the timeout var. Moved back onto the function.

**PR description** — the dark row of the behaviour table said the router `503`s. It doesn't: a relay against a dark chain fails with `PairingListEmptyError` with no upstream status to propagate, so `jsonRPC.go:521-526` falls through to `StatusInternalServerError`. 500, which is what the branch's own k3d run reported two sections lower — the table and the test evidence contradicted each other. The 503 in `chainlib/common.go:664` is the *health* endpoint, which is a different thing and stays correct. Fixed to "500s on relays".

While in there I also added a short **Review follow-up** section to the description covering 54832ad, since 450 lines of subscription-manager code with no mention in the description would ambush the next reviewer.

If any of the three you had in mind is a different one, send it over — the METRICS.md row is already in from the last round.

AnnaR-prog (approved):

PR #266, opened 2026-08-11
ci: strip retired periodic-probe args in PR gate

avitenzer:
Jira ticket: MAG-2733

Unblocks the `dev-sim-prtests` **rollout step** for #260. See "Remaining work" — this is necessary but not sufficient.

## The failure

The rollout step swaps only the image on the pre-existing `eth-sim-router` Deployment (`kubectl set image`), so the live container args outlive the binary that understood them. #260 removes the legacy synthetic prober and with it two flags the deployed args still pass. Cobra rejects the first and exits 1 before the router starts:

```
Error: unknown flag: --periodic-probe-providers-interval
```

The pod CrashLoopBackOffs, the gate fails and rolls back. Seen on runs [31487560320](https://github.com/Magma-Devs/smart-router/actions/runs/31487560320) and [31488457933](https://github.com/Magma-Devs/smart-router/actions/runs/31488457933).

## The change

The gate already sanitizes previously-retired flags out of the live args before rollout — `--optimizer-qos-listen`, `--show-provider-address-in-metrics`, `--geolocation`, `--concurrent-providers`, `--provider-optimizer-*`. #260's two flags simply aren't on that list. This adds them to both predicates: `retired_flag_name` for the `--flag value` form and `retired_flag_assignment` for `--flag=value`, which is the form the chart uses.

| Flag | Removed by #260 | In deployed args | Listed here |
|---|---|---|---|
| `--periodic-probe-providers-interval` | yes | yes (chart line 153, literal) | yes |
| `--enable-periodic-probe-providers` | yes | yes (chart line 156, literal) | yes |
| `--probe-update-weight` | **no** | yes (chart line 207, values-driven) | **no** |

### Why `--probe-update-weight` is not listed

An earlier revision of this PR listed it. That was wrong on both counts, per review:

- **#260 keeps it.** It only looks deleted because gofmt re-aligned the constant block when its two neighbours went — name and value are identical on both sides. On #260's branch it is still registered (`rpcsmartrouter.go:3087`) and still applied (`:2906`, via `scoreutils.SetProbeUpdateWeight`).
- **It is in the deployed args**, rendered from a values key at `charts/smart-router/templates/routers_deployment.yaml:207`. Stripping it is a no-op only by coincidence — chart default and binary default are both `0.25`. Tune `probeUpdateWeight` and the gate would silently reset it, validating a router whose provider-scoring weight differs from production's.

Only flags the binary under test rejects belong on this list. A pre-emptive entry inverts what the gate is for: instead of failing when chart and binary disagree, it hides the disagreement.

## Verification

Extracted the jq program from the workflow and ran it against a realistic arg list drawn from the chart:

- strips both retired flags, in the `--flag=value` form the chart uses
- leaves every retained flag and its value intact, including `--min-relay-timeout=5s` immediately after a stripped flag, and `--probe-update-weight=0.9` (a deliberately non-default value, to prove a tuned setting survives)
- control run with `main`'s current sanitizer lets both through, confirming this is the delta that fixes it

## Remaining work — this alone does not get #260 green

The sanitizer applies only to the `kubectl set image` path. The next step in the same job, "Deploy and validate latest Helm chart with PR image", installs `smart-router-helm-chart@main` fresh with the PR image, and that chart **hardcodes both flags as literals** (not values-driven, so no values file suppresses them):

```
charts/smart-router/templates/routers_deployment.yaml:153  - --periodic-probe-providers-interval=120s
charts/smart-router/templates/routers_deployment.yaml:156  - --enable-periodic-probe-providers=true
```

The install runs `--wait --atomic`, so a pod that won't start rolls back and fails the step. That step has never executed for #260 (the run dies at the rollout step first), so this is read from chart source plus a successful run of the same step on #259 — not an observed #260 failure.

**A change in `smart-router-helm-chart` guarding or removing lines 153 and 156 — leaving 207 alone — must land on that repo's `main` before the gate is re-dispatched for #260.**

## Ordering

The gate is `workflow_dispatch` and the sanitizer is inline in the workflow YAML, so the version that executes comes from the dispatched ref — not from the PR under test. This has to be on `main` before re-dispatching for #260; putting it on #260's branch alone would not help.

### DIFF
--- a/.github/workflows/pr-gate-build-artifact.yml
+++ b/.github/workflows/pr-gate-build-artifact.yml
@@ -496,12 +496,16 @@ jobs:
               ($arg == "--show-provider-address-in-metrics") or
               ($arg == "--geolocation") or
               ($arg == "--concurrent-providers") or
+              ($arg == "--enable-periodic-probe-providers") or
+              ($arg == "--periodic-probe-providers-interval") or
               (($arg | startswith("--provider-optimizer-")) and (($arg | contains("=")) | not));
             def retired_flag_assignment($arg):
               ($arg | startswith("--optimizer-qos-listen=")) or
               ($arg | startswith("--show-provider-address-in-metrics=")) or
               ($arg | startswith("--geolocation=")) or
               ($arg | startswith("--concurrent-providers=")) or
+              ($arg | startswith("--enable-periodic-probe-providers=")) or
+              ($arg | startswith("--periodic-probe-providers-interval=")) or
               (($arg | startswith("--provider-optimizer-")) and ($arg | contains("=")));
             def sanitize_arg($arg):
               if retired_flag_name($arg) then


AnnaR-prog:
## Review

The four lines for the two flags #260 actually removes are correct and needed. Two things to change: the `--probe-update-weight` lines should come out, and this PR does not finish unblocking #260 — the same job's Helm step fails on the same flags, and the sanitizer doesn't reach it.

Confirming the diagnosis first, because it's right. Run [31488457933](https://github.com/Magma-Devs/smart-router/actions/runs/31488457933), step "Deploy and validate dev-sim-prtests Kubernetes rollout":

```
Error: unknown flag: --periodic-probe-providers-interval
```

The `jq` reducer handles both spellings correctly, including the case called out in the description — a kept flag immediately after a stripped one survives, because `drop_next` only swallows a following token that doesn't start with `--`.

The ordering note is also correct and worth crediting: `on:` is `workflow_dispatch` only, so the workflow file that runs comes from the dispatched ref. This has to be on `main` before re-dispatching for #260.

---

### 1. `--probe-update-weight` doesn't belong on the list

The description says all three flags are ones #260 removes, and that this one isn't in the deployed args so listing it costs nothing. Both halves are wrong.

**#260 does not remove it.** On #260's own branch the flag is still registered and still consumed:

- `protocol/rpcsmartrouter/rpcsmartrouter.go:3087` — registered as a `Float64` flag
- `protocol/rpcsmartrouter/rpcsmartrouter.go:2906` — read back and applied via `scoreutils.SetProbeUpdateWeight(...)`

It looks deleted in the diff because of formatting. `ProbeUpdateWeightFlagName` shows as one `-` line and one `+` line only because its two neighbours were deleted and the formatter re-aligned the remaining constant — name and value are identical on both sides. #260 removes two flags, not three.

**It is in the deployed args.** The chart this gate deploys renders it unconditionally:

```
smart-router-helm-chart charts/smart-router/templates/routers_deployment.yaml:207
  - --probe-update-weight={{ $.Values.miscellaneous.routers.probeUpdateWeight }}
```

with `probeUpdateWeight: 0.25` as the chart default. (I read this from chart source — I don't have cluster access to read the live `eth-sim-router` Deployment directly.)

The effect today is small, because the binary default is also `0.25`, so stripping it lands on the same number. That's coincidence, not design. As soon as anyone tunes `probeUpdateWeight` away from the default, the gate silently resets it and validates a router whose provider-scoring weight differs from the one production runs — which is exactly what the simulator tests measure.

The general point: this list should only contain flags the binary under test rejects. A pre-emptive entry inverts what the gate is for — instead of failing when chart and binary disagree, it hides the disagreement. Suggest dropping the two `--probe-update-weight` lines and keeping the other four.

### 2. This doesn't finish unblocking #260

The sanitizer only applies to the `kubectl set image` path. The next step in the same job — "Deploy and validate latest Helm chart with PR image" — installs the chart fresh from `smart-router-helm-chart@main` with the PR image, and that chart hardcodes both flags #260 removes:

```
charts/smart-router/templates/routers_deployment.yaml:153
  - --periodic-probe-providers-interval=120s
charts/smart-router/templates/routers_deployment.yaml:156
  - --enable-periodic-probe-providers=true
```

These are literal lines, not driven by a values key, so no values file can suppress them. And the routers really do render: in the last successful gate run ([31487918987](https://github.com/Magma-Devs/smart-router/actions/runs/31487918987), for #259) that step reported `STATUS: deployed` and then `Smoke result: eth_chainId=0x1` — a router pod from this chart started and served a request. The install runs `--wait --atomic`, so a pod that won't start doesn't just log an error; Helm rolls back and the step fails.

To be precise about what's observed: this step has never executed for #260, because the run died at the rollout step first. That's a mechanism read from chart source plus a successful run of the same step, not an observed #260 failure.

So after merging this as-is, #260 clears the rollout step and then fails at the Helm step on the same flags. The remaining piece is a change in `smart-router-helm-chart` — guard or delete lines 153 and 156, leave 207 alone — landing on that repo's `main` before the gate is re-dispatched. That blocks the goal, not this merge.

### Smaller notes

- **Nothing tests the sanitizer.** `retired_flag_name` appears nowhere in the repo except inside this workflow file. It's been patched four times now (#167, #168, #169, this one). Extracting the `jq` and running it by hand was the right instinct, but it leaves nothing behind for the next person. A small script plus a fixture arg list, run in CI, would make the next addition self-checking.
- **The root cause is the rollout step's design.** Swapping only the image guarantees this failure every time a flag is retired. Deploying from the chart — which the next step already does — keeps args and binary in step. Out of scope here; worth its own ticket.

avitenzer:
Both points verified independently and acted on. Thanks — the `--probe-update-weight` catch was a real defect, not a style nit.

**1. `--probe-update-weight` removed from the list** (fac7c04).

You're right that #260 doesn't remove it, and my extraction was the bug: I grepped for `^-.*FlagName.*=` and never checked for a matching `+` line, so a gofmt re-alignment read as a deletion. Confirmed the flag survives on #260's branch at `rpcsmartrouter.go:3087` (registered) and `:2906` (applied via `SetProbeUpdateWeight`).

Confirmed the chart side too, from `smart-router-helm-chart@main`:

```
charts/smart-router/templates/routers_deployment.yaml:207
  - --probe-update-weight={{ $.Values.miscellaneous.routers.probeUpdateWeight }}
charts/smart-router/values.yaml:209
  probeUpdateWeight: 0.25
```

against `DefaultProbeUpdateWeight float64 = 0.25` in `utils/score/score_config.go:39` — so the two defaults coincide exactly as you said, and the silent-reset hazard is real the moment anyone tunes it. Your framing is the right one and I've put it in the commit message: a pre-emptive entry inverts what the gate is for.

Re-ran the extracted jq with `--probe-update-weight=0.9` in the input to confirm a tuned value now survives; only the two retired flags are stripped.

**2. Helm step — confirmed, and it does block #260.**

Verified lines 153 and 156 are literals, not values-driven, so no values file suppresses them. I've rewritten the PR description to stop claiming this unblocks #260 outright — it now says rollout-step-only, with the chart change called out as required before re-dispatch. Agreed the fix is guard-or-delete 153/156 and leave 207 alone.

**On the two smaller notes** — both fair, and both genuinely out of scope for this PR:

- *No test for the sanitizer.* Agreed, and the fourth patch to a list with no coverage is exactly when it should stop. I'd rather not bolt it onto this PR since this one is blocking; happy to do it as a follow-up if you want to file it.
- *The rollout step's design is the root cause.* Agreed — deploying from the chart, as the next step already does, keeps args and binary in step and would make this whole allowlist unnecessary. Worth its own ticket.

One thing I did **not** do: commit 5d5886c still has the wrong three-flag rationale in its message, and this repo rebase-merges, so it would land on `main` as-is. I left it rather than force-push a branch under review. Say the word and I'll squash the two into one correct commit before merge.

avitenzer:
Squashed to a single commit (0e28e49) so the incorrect three-flag rationale in the old 5d5886c doesn't land on `main` — this repo rebase-merges, so that message would have persisted verbatim.

Content is unchanged: tree hash `8d8f69d` before and after, and `git diff fac7c04 HEAD` is empty. The final diff is the four lines for the two flags #260 actually removes.

The commit message now carries the `--probe-update-weight` reasoning from your review — why it stays off the list, and the coincident 0.25 defaults that make stripping it look harmless — plus the note that this unblocks the rollout step only and the chart change is still required.

Tomelia1999 (approved):

PR #269, opened 2026-08-12
fix(metrics): bound the method label for unmatched (Default-*) apis

nimrod-teich:
#Closes MAG-2950

Jira ticket: MAG-2950

## Why

The `method` label on every `smartrouter_*` metric family is unbounded for requests that miss the spec: the parser mints a synthetic api named `Default-<raw request name>` (protocol/chainlib/base_chain_parser.go:375) and that raw name becomes the label value. For REST chains the raw name is the URL with **concrete values** where a matched template would carry `{placeholders}` — so every distinct block number or address mints a brand-new series.

Live evidence from production (2026-08-12): the `method` label held **46,067 distinct values, 45,942 of them** of the form `Default-/chains/main/blocks/<n>/header/` — one client polling Tezos block headers with a trailing slash the spec template doesn't match. That put ~79k series on each `smartrouter_requests_*` family and was the root cause behind the war-room dashboard timing out (dashboard side fixed separately in smart-router-standalone — PR #168 "check in the Customer Experience War Room dashboard").

Batches had the identical disease and were already cured by the signature admission registry (`batch_method_label.go`). This PR extends the same treatment to unmatched single-method names, at the same normalization funnel.

## What changed

- `protocol/metrics/default_method_label.go` (new) — `normalizeDefaultMethodLabel`: the first 32 distinct `Default-*` names per spec pass through; everything past the cap collapses to `Default-other`. The pass-through budget is deliberate: a genuinely-missing spec API produces **one stable name**, which is a valuable spec-gap signal — only concrete-ID floods blow the budget.
- `smartrouter_default_method_overflow_total{spec}` (new counter) — reports when the cap binds, so a lossy breakdown is detectable and alertable.
- `protocol/metrics/batch_method_label.go` — `normalizeMethodLabel` routes single-method names through the new bounding; `batchSignatureRegistry` cap becomes a field (`newBatchSignatureRegistry()` keeps 64, `newDefaultMethodRegistry()` uses 32) instead of a hardcoded constant.
- `protocol/metrics/smartrouter_metrics_manager.go` — registry + counter wiring.
- `protocol/metrics/default_method_label.go` — **shape normalization, applied before admission.** A path segment that is a *value* rather than part of the route folds to `{}`, so `/blocks/1/header` and `/blocks/12/header` become the one shape `/blocks/{}/header` and cost a single budget entry. A segment is read as a value when it is all digits, `0x`-hex, or ≥26 chars carrying a digit over an identifier charset (base58 hashes ~43, bech32 ~39, Tezos block hashes 51). The flood that motivated the cap now never reaches it; the cap stays the backstop for a spec that genuinely produces >32 distinct routes.
  - **Only paths are reshaped.** JSON-RPC names are already stable identifiers and several carry digits that belong to the name (`web3_clientVersion`, `eth_getBlockByNumber`), so anything not starting with `/` is returned untouched.
  - **Route elements survive by construction** — `v1`, `v1beta1`, `head`, `latest`, `ledger_info` are neither all-digit nor long enough to read as opaque. There is a test enumerating them.
  - The placeholder is bare `{}`, not a named one: an unmatched path has no spec template to take a parameter name from, and inventing one would make the label look like a template it is not. `{}` is not itself a concrete id, so reshaping is idempotent — which the delegating recorders rely on.
- `protocol/metrics/default_method_label_test.go` (new) — matched methods untouched, stable names admitted and stable across budget exhaustion (the in-flight gauge Add/Sub pairing hazard), per-spec budgets, cap collapse + overflow reporting, idempotence of `Default-other`.

Admission is monotone (never evicts), same as batches — a raw name maps to the same label value for the process lifetime.

The two layers compose: reshaping keeps the budget spent on **shapes** rather than instances, and the cap still bounds a spec that produces too many distinct shapes. The cap tests previously reached the cap via a block-number flood, which reshaping now prevents by design; they use distinct routes instead.

## How to verify

```bash
go test ./protocol/metrics/ -count=1
```

After deploy, on any affected box the method-label cardinality stops growing:

```bash
curl -s http://<prometheus>/api/v1/label/method/values | jq '.data | length'   # stabilizes at ~(real methods + 32×specs + 1)
```

`sum by (spec) (rate(smartrouter_default_method_overflow_total[5m]))` shows which specs have clients hitting unmatched URLs.

Note on the TEZOS case above: the follow-up investigation found it is a **matcher** bug rather than a spec gap — the spec already carries `/chains/main/blocks/{block_id}/header`, and only the client's trailing slash misses the anchored pattern. That is fixed in #312, which removes the flood at its source. This PR stays the backstop for names that genuinely miss the spec.

### DIFF
--- a/protocol/metrics/batch_method_label.go
+++ b/protocol/metrics/batch_method_label.go
@@ -91,15 +91,21 @@ func buildBatchSignature(apiName string) (signature string, size int, tooWide bo
 	return BatchMethodPrefix + strings.Join(distinct, batchSignatureElementSeparator), size, false
 }
 
-// batchSignatureRegistry bounds the set of batch signatures allowed to become label
-// values, tracked per spec so a chatty chain can't consume another chain's budget.
+// batchSignatureRegistry bounds a set of label values, tracked per spec so a chatty
+// chain can't consume another chain's budget. Batch signatures were the first user;
+// unmatched-API names (default_method_label.go) reuse it with their own cap.
 type batchSignatureRegistry struct {
 	lock     sync.RWMutex
-	admitted map[string]map[string]struct{} // spec -> set of admitted signatures
+	cap      int
+	admitted map[string]map[string]struct{} // spec -> set of admitted values
 }
 
 func newBatchSignatureRegistry() *batchSignatureRegistry {
-	return &batchSignatureRegistry{admitted: make(map[string]map[string]struct{})}
+	return &batchSignatureRegistry{cap: maxBatchSignaturesPerSpec, admitted: make(map[string]map[string]struct{})}
+}
+
+func newDefaultMethodRegistry() *batchSignatureRegistry {
+	return &batchSignatureRegistry{cap: maxDefaultMethodsPerSpec, admitted: make(map[string]map[string]struct{})}
 }
 
 // admit reports whether signature may be used as a label value for spec.
@@ -127,13 +133,13 @@ func (r *batchSignatureRegistry) admit(spec, signature string) bool {
 
 	signatures, ok := r.admitted[spec]
 	if !ok {
-		signatures = make(map[string]struct{}, maxBatchSignaturesPerSpec)
+		signatures = make(map[string]struct{}, r.cap)
 		r.admitted[spec] = signatures
 	}
 	if _, alreadyAdmitted := signatures[signature]; alreadyAdmitted {
 		return true
 	}
-	if len(signatures) >= maxBatchSignaturesPerSpec {
+	if len(signatures) >= r.cap {
 		return false
 	}
 	signatures[signature] = struct{}{}
@@ -148,8 +154,13 @@ func (r *batchSignatureRegistry) admit(spec, signature string) bool {
 // RecordRelaySuccess → AddEndpointRelayServiced), and each normalizes at its own
 // boundary so that calling any of them directly is still safe.
 func (m *SmartRouterMetricsManager) normalizeMethodLabel(spec, method string) string {
-	if m == nil || !strings.Contains(method, batchMethodSeparator) {
+	if m == nil {
 		return method
+	}
+	if !strings.Contains(method, batchMethodSeparator) {
+		// Single-method names have their own unbounded case: the synthetic
+		// "Default-<raw request>" apis minted for spec misses (default_method_label.go).
+		return m.normalizeDefaultMethodLabel(spec, method)
 	}
 
 	signature, _, tooWide := buildBatchSignature(method)
--- a/protocol/metrics/default_method_label.go
+++ b/protocol/metrics/default_method_label.go
@@ -23,11 +23,122 @@ const (
 	// spec-gap signal this cap preserves — while a concrete-ID flood blows through
 	// the budget immediately and every later name collapses to DefaultMethodOther.
 	maxDefaultMethodsPerSpec = 32
+
+	// defaultMethodIDPlaceholder stands in for a path segment that carries a concrete
+	// value. It deliberately does not name the parameter: unlike a matched api, an
+	// unmatched path has no spec template to take a name from, and inventing one would
+	// make the label look like a template it is not.
+	defaultMethodIDPlaceholder = "{}"
+
+	// minOpaqueIDLen is the length at or above which an alphanumeric segment carrying a
+	// digit is read as an identifier rather than a path element. It clears the longest
+	// real segments seen in the specs (getBlockTransactionCountByNumber, 32 chars, has
+	// no digit) and sits below the shortest identifiers (base58 tx hashes ~43, bech32
+	// addresses ~39, Tezos block hashes 51).
+	minOpaqueIDLen = 26
 )
 
+// defaultMethodShape collapses concrete identifiers in an unmatched REST path to
+// defaultMethodIDPlaceholder, so /blocks/1/header and /blocks/12/header become the one
+// shape /blocks/{}/header. The per-spec budget is then spent on path SHAPES rather than
+// on one entry per block number, which is what keeps a flood from reaching the cap at
+// all — the cap stays the backstop for genuinely pathological specs.
+//
+// Only paths are reshaped. A JSON-RPC method name is already a stable identifier and
+// several carry digits that are part of the name (web3_clientVersion,
+// eth_getBlockByNumber), so anything not starting with "/" is returned untouched.
+//
+// The result is idempotent: defaultMethodIDPlaceholder is not itself a concrete
+// identifier, so reshaping an already-reshaped path is a no-op. That matters because
+// the public recorders delegate to one another and normalize at each boundary.
+func defaultMethodShape(method string) string {
+	raw := strings.TrimPrefix(method, defaultMethodPrefix)
+	if !strings.HasPrefix(raw, "/") {
+		return method
+	}
+
+	segments := strings.Split(raw, "/")
+	reshaped := false
+	for i, segment := range segments {
+		if isConcreteIDSegment(segment) {
+			segments[i] = defaultMethodIDPlaceholder
+			reshaped = true
+		}
+	}
+	if !reshaped {
+		return method
+	}
+	return defaultMethodPrefix + strings.Join(segments, "/")
+}
+
+// isConcreteIDSegment reports whether one path segment is a value rather than part of
+// the route: a block number, a hex address, or an opaque hash. A segment that mixes
+// letters without carrying an identifier's length is left alone, which is what keeps
+// route elements such as "v1", "head", "latest" and "ledger_info" intact.
+func isConcreteIDSegment(segment string) bool {
+	if segment == "" || segment == defaultMethodIDPlaceholder {
+		return false
+	}
+	if isAllDigits(segment) {
+		return true
+	}
+	if len(segment) > 2 && (strings.HasPrefix(segment, "0x") || strings.HasPrefix(segment, "0X")) && isHex(segment[2:]) {
+		return true
+	}
+	return len(segment) >= minOpaqueIDLen && hasDigit(segment) && isIDCharset(segment)
+}
+
+func isAllDigits(s string) bool {
+	for _, r := range s {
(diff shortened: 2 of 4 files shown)


avitenzer (commented):
Second pass, at `3385dae`. All 20 checks green. The core design still looks right to me — reshape-before-admit is the correct ordering, and the cap is a sound backstop. One new finding, from a check I had not run before: feeding every real spec route through the reshaping predicate.

**Disclosure first:** I authored `3385dae` on this branch and edited the description to add the Jira line. Treat my read of that commit as self-review rather than independent verification — it needs someone else's eyes before this merges.

## The reshaping heuristic eats real route segments

I extracted all **2,007 distinct literal route segments** from the spec repo — excluding `{placeholder}` segments, where collapsing is the intended behaviour — and ran each through `isConcreteIDSegment`. Six are collapsed:

```
[TRX]          getdelegatedresourceaccountindexv2        (34 chars)
[TRX]          gettriggerinputforshieldedtrc20contract   (39)
[TRX]          isshieldedtrc20contractnotespent          (32)
[TRX]          scanshieldedtrc20notesbyivk               (27)
[TRX]          scanshieldedtrc20notesbyovk               (27)
[SQDSUBGRAPH]  1        (from /subgraphs/name/1)
```

Run through the shipped `defaultMethodShape`, five distinct TRON endpoints become one label:

```
/wallet/scanshieldedtrc20notesbyivk              -> Default-/wallet/{}
/wallet/scanshieldedtrc20notesbyovk              -> Default-/wallet/{}
/wallet/isshieldedtrc20contractnotespent         -> Default-/wallet/{}
/wallet/gettriggerinputforshieldedtrc20contract  -> Default-/wallet/{}
/wallet/getdelegatedresourceaccountindexv2       -> Default-/wallet/{}
/wallet/getnowblock                              -> Default-/wallet/getnowblock   <- short one survives
/subgraphs/name/1                                -> Default-/subgraphs/name/{}
```

`...byivk` and `...byovk` differ by one letter and are genuinely different endpoints.

### Where the reasoning went wrong

From the comment on `minOpaqueIDLen`:

> It clears the longest real segments seen in the specs (getBlockTransactionCountByNumber, 32 chars, has no digit)

The threshold was set above the longest segment *without* a digit, but the predicate is `len >= 26 && hasDigit && isIDCharset`. Long segments *with* digits were never swept — and TRON's `trc20` / `v2` names are 27-39 chars and all-alphanumeric.

That also means this claim from the description does not hold:

> **Route elements survive by construction** — `v1`, `v1beta1`, `head`, `latest`, `ledger_info` are neither all-digit nor long enough to read as opaque. There is a test enumerating them.

`TestDefaultMethodShapeKeepsRouteElements` passes because everything it enumerates is short. The corpus has 27-39 char route segments carrying digits.

### Impact

Bounded but real. It only bites paths that miss the spec, so it needs a TRX or SQDSUBGRAPH client hitting one of these six via a trailing slash, case difference, or query string — exactly the MAG-2943 shape. When it does, the operator loses the signal saying *which* endpoint is missing, which is the specific value the 32-name budget exists to preserve. The breakdown also goes inconsistent: short `/wallet/*` names keep their identity while long ones merge.

### The fix looks like one constant

The numbers separate cleanly:

```
longest real route segment eaten:   39  gettriggerinputforshieldedtrc20contract
shortest identifier to still catch: 44  lava@17ym998u666u8w2qgjd5m7w7ydjqmu3mlgl7ua2
                                    45  cosmos1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5lzv7xu
                                    50  BKiHLREqU3JkXfzEDYAkmmfX48gBDtYqbugTxCcv9YrK1H1EAy
```

There is a clean gap between 39 and 44, so raising `minOpaqueIDLen` from 26 to ~40 clears all five TRX segments while still catching every real identifier in the corpus. The comment's "bech32 ~39" was pessimistic — actual bech32 in the specs is 44-45, so there is headroom.

Two caveats:

* `TestDefaultMethodShapeCollapsesConcreteIDs` asserts the 44-char bech32 case collapses; that still holds at 40, but the boundary is worth a deliberate look rather than a re-run.
* `/subgraphs/name/1` is **not** fixed by any length change — a bare `1` is caught by `isAllDigits` regardless. I would accept that one rather than contort the predicate; a numeric subgraph id is arguably a value anyway.

Worth noting the direction of error: **under-collapse is safe** (costs budget entries, the cap backstops it), **over-collapse is lossy and unrecoverable** (two endpoints become one series). Biasing the threshold upward is the right way to be wrong.

### Worth making permanent

The corpus sweep is ~15 lines against a checked-in segment list, and it is what turns "route elements survive by construction" from a claim into a guarantee. I ran it as a scratch test and removed it; happy to contribute it if useful.

## Also outstanding

`3385dae` has no test. The existing tests exercise `normalizeMethodLabel` directly rather than per-recorder, so a one-off test for that recorder would break the pattern — but nothing will catch the next setter that forgets the call. A table test asserting every method-labelled setter normalizes would cover it, and would have caught the straggler gap in the first place.

## Verified fine

* `admit` handles a nil receiver, so a manager built without `defaultMethods` degrades to `Default-other` rather than panicking.
* No zero-value `batchSignatureRegistry{}` anywhere — the constant-to-field change for `cap` looked like a footgun (a zero-value registry would have `cap: 0` and admit nothing), but both constructors are the only construction sites.
* Idempotence holds: `isConcreteIDSegment` excludes `{}`, and matched templates never enter the path for want of the `Default-` prefix.
* Trailing slashes shape correctly — `Split` yields an empty final segment, which is rejected.
* The tests that matter are there: the in-flight gauge Add/Sub pairing hazard, per-spec budgets, and the route-element safety list (which is the one that needs widening).

---

Posting as a comment rather than request-changes, since I have a commit on this branch and would rather not formally gate it. For the record I do think the over-collapse should be resolved before merge — but it is one constant, and the production bleed this PR stops (157k series, ~1,100 new values/hour per MAG-2943) is the bigger problem, so I would not sit on this for long. Happy to convert to request-changes if you would rather it gate.

avitenzer (changes requested):
Converting my previous review to request-changes, as discussed — the detail is in that comment, this is just the gating signal.

**The blocker:** `defaultMethodShape` collapses five real TRX route segments (27-39 chars, carrying digits from `trc20` / `v2`) into `Default-/wallet/{}`, merging five distinct endpoints into one label. Introduced by this PR, and lossy in the unrecoverable direction.

**The fix:** raise `minOpaqueIDLen` from 26 to ~40. The corpus separates cleanly — longest real route segment eaten is 39, shortest identifier that must still be caught is 44.

Not gating, but worth resolving in the same pass:

* `3385dae` is mine and needs a second pair of eyes; it also has no test.
* The corpus sweep is worth landing as a permanent test — happy to contribute it.

To be clear about priority: the production bleed this PR stops (157k series, ~1,100 new values/hour) is a bigger problem than the six segments. I am gating because the fix is one constant, not because I think this should wait. If it turns out to be more than that, say so and I will clear the block and file a follow-up instead.

nimrod-teich:
Addressed in becc1fe: `minOpaqueIDLen` 26 → 43, with the 39-char TRX segment added to `TestDefaultMethodShapeKeepsRouteElements` to pin the boundary from below.

Went to the top of the gap rather than ~40: base58 of a 32-byte hash is 43-44 chars, so 43 still catches every identifier class while maximizing route headroom — and over-collapse is the unrecoverable direction, under-collapse is bounded by the budget cap. `/subgraphs/name/1` accepted as-is per your note (all-digits is a value).

avitenzer (approved):
Approving — `becc1fe` resolves the blocker, and I re-ran the corpus sweep against the shipped predicate rather than taking the constant on faith.

## Verified

Worktree checkout of `becc1fe2` (not my working dir), sweeping every literal REST route segment across all 155 specs — 3,044 `(spec, segment)` pairs, 2,008 distinct segments — through `isConcreteIDSegment` as shipped:

**One segment collapses: `/subgraphs/name/1`.** The one we agreed to accept. All five TRX segments survive.

The margin is clean on both sides, which is the part I actually cared about:

```
longest route segment carrying a digit:  39  gettriggerinputforshieldedtrc20contract
40-42 dead zone (route side):            EMPTY
minOpaqueIDLen:                          43
tightest identifier still caught:        43  base58 of a 32-byte hash
```

Identifier classes still collapse at 43: base58 43 and 44, bech32 44/45, Tezos 51, Solana sig 88. Eth addresses (42) are unaffected because they go through the `0x` branch, not the length branch — so the threshold move never had them in scope.

Going to the top of the gap rather than my suggested ~40 was the better call, and for the reason you gave: over-collapse is the unrecoverable direction, so the threshold should sit as high as the identifier classes allow. `go test ./protocol/metrics/ -count=1` passes at head.

One incidental finding worth recording: the longest *surviving* segment is 56 chars (`estimate_superfluid_delegation_amount_by_validator_denom`), which survives on the `hasDigit` conjunct alone. That conjunct is load-bearing, not just a narrowing convenience — worth knowing before anyone simplifies the predicate later.

## Straggler concern — closed on substance

My last review said "nothing will catch the next setter that forgets the call." I ran that audit by hand across the package: every method-labelled `WithLabelValues`, checked against whether its enclosing setter normalizes.

**10 of 10 normalized, 0 stragglers.** All method labels live in `smartrouter_metrics_manager.go`; none elsewhere in the package. So there is no gap today — what's missing is only the regression guard.

## Two follow-ups, neither gating

1. **`7cb4f98` is still mine and still wants a second pair of eyes.** My approval doesn't clear that, and I'd rather not have it merge on my signature alone. It's a one-line change and I've now audited all 10 setters around it, so whoever looks is checking a one-liner against a sweep that's already been done — not redoing the analysis. @nimrod-teich if you're happy with it, say so on the PR and I'll consider it covered.
2. **The corpus sweep is worth landing as a permanent test.** ~15 lines against a checked-in segment list. It's what turns "route elements survive by construction" from a claim into a guarantee, and it's the check that would have caught the original 26 before review rather than after. Happy to open it as a follow-up rather than hold this.

Branch is behind `main` — needs an update before it can go in.

PR #275, opened 2026-08-16
fix(smart-router): route two providers configured with one name (MAG-2724)

avitenzer:
Jira ticket: MAG-2724

Fixes [MAG-2724](https://magmadevs.atlassian.net/browse/MAG-2724).

A provider's configured name is its routing identity — the key of `csm.pairing` and of the retry skip-list. Two providers sharing one collapse into a single entry: one upstream serves every request while the other sits idle, and once that name is set aside after a failure both go with it, so the customer gets an error while a healthy provider stands by.

`dffc0f2` established that duplicate names must be tracked separately, but only for the startup failed-provider set. Everything downstream was still keyed by name.

## Approach

**The router refuses to start on that configuration.** `ValidateUniqueProviderNames` runs on the boot path, over static and backup together, before any listener binds:

```
Error: invalid direct-rpc endpoints definition ErrMsg: duplicate provider names:
"SimTwin" on chain ETH1 api-interface jsonrpc is shared by 2 providers
(urls: https://one..., https://two...) — a provider name is the router's identity
for a node, so two nodes sharing one on the same chain and api-interface cannot be
told apart; give each a distinct name
```

Every collision is named at once and sorted, so one edit fixes the config and the same config always produces the same message (the grouping is a Go map — unsorted output would vary per run).

Scoped to chain + api-interface on purpose: a session manager is built per chain+interface and owns its own pairing map, so that is exactly where names collapse. Reusing one operator label across chains (`alchemy` on ETH1 and POLYGON1) is a normal way these configs are written and stays legal. Both lists go in one call because static and backup land in separate maps but are looked up across both by address — a per-list check would miss a name shared between the two.

## What this deliberately does not do

An earlier revision of this PR also gave colliding providers derived identities (`SimTwin#0`, `SimTwin#1`) so both stayed routable. That is removed. It kept a broken config running under names the operator never chose — including in the metrics `endpoint_id`, breaking any dashboard keyed on it — and it split "name" from "identity" across the whole codebase, so every map keyed by a provider had to key on the derived value instead. That invariant was violated four times in this PR alone (twice caught in review, twice more caught by a rebase). Refusing the config makes that entire class of bug unrepresentable: if names are unique, keying by name is simply correct.

## Where the check lives

On the boot path, not in `ParseStaticProviderEndpoints`. `smart-router health` is what an operator reaches for to work out why a config will not boot, so it has to be able to load the config the router rejects. It logs the collision as a warning and probes every node anyway:

```
$ smartrouter health dupcfg --use-static-spec specs/
rows: [('SimTwin', 'https://one...'), ('SimTwin', 'https://two...')]
stderr: warn  the router will REFUSE TO START on this config — probing it anyway
```

Both rows keep the configured name — the router's identity for them really is ambiguous, and inventing a label here would misrepresent what it does with this config. They are told apart by `url`, which is what identifies the node.

The chart-side check in smart-router-helm-chart catches the same mistake one step earlier, at template render, where it fails the release rather than the pods — and catches what this check cannot see: `SimTwin` vs `simtwin`, or `My Node` vs `my-node`, which look distinct in `values.yaml` and arrive at the router as one name after the configmap's `lower | replace " " "-"`.

## On boot behaviour

This makes a duplicate name fatal at boot, which cuts against the MAG-2525 direction of booting on any usable provider. Calling it out rather than leaving it implicit: the config is unroutable as written, the operator has to change it either way, and failing loudly at startup is a better outcome than silently serving at half capacity. **No live config trips it** — rendered `values_local.yml`, `values.yml`, `values_internal.yml` and `values_sim.yml` from smart-router-standalone plus the chart's own values, and scanned the repo's `direct-rpc` configs, so nothing deployed or in the dev gate crashloops.

## Testing

`go build ./protocol/...`, `go vet`, and the full `lavasession` + `rpcsmartrouter` suites pass.

Verified against the built binary rather than only in unit tests: boot exits with the message above, and `health` loads the same config, warns, and probes both nodes.

Heads-up: the lavasession suite has a **pre-existing flake** (`Failed locking selection after MaximumNumberOfSelectionLockAttempts`) — it failed 1 of 5 runs with these changes stashed, so it is not from this work, but it may show up in CI.

[MAG-2724]: https://magmadevs.atlassian.net/browse/MAG-2724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/protocol/lavasession/provider_names_mag2724_test.go
+++ b/protocol/lavasession/provider_names_mag2724_test.go
@@ -1,0 +1,127 @@
+package lavasession
+
+import (
+	"testing"
+
+	"github.com/magma-Devs/smart-router/protocol/common"
+	"github.com/stretchr/testify/require"
+)
+
+// MAG-2724: two providers configured under one name collapsed into a single entry — one upstream
+// served every request while the other sat idle, and setting that name aside after a failure left
+// the router with nothing to retry against.
+//
+// A provider's name is its routing identity, so the fix is to refuse the configuration rather than
+// to invent a second identity for it: the router will not start, and the error tells the operator
+// which names to change. These tests cover the check that makes that decision.
+
+func mag2724Endpoint(name, chainID, apiInterface, url string) *RPCStaticProviderEndpoint {
+	return &RPCStaticProviderEndpoint{
+		Name:         name,
+		ChainID:      chainID,
+		ApiInterface: apiInterface,
+		NodeUrls:     []common.NodeUrl{{Url: url}},
+	}
+}
+
+func TestValidateUniqueProviderNames_RejectsCollisionWithinChainAndInterface(t *testing.T) {
+	err := ValidateUniqueProviderNames([]*RPCStaticProviderEndpoint{
+		mag2724Endpoint("SimTwin", "ETH1", "jsonrpc", "http://one:8545"),
+		mag2724Endpoint("SimTwin", "ETH1", "jsonrpc", "http://two:8545"),
+	})
+
+	require.Error(t, err)
+	// The operator has to be able to find the offending pair from the boot log alone — the router
+	// is down at this point, so the message is the only thing they have to work from.
+	require.Contains(t, err.Error(), "SimTwin")
+	require.Contains(t, err.Error(), "ETH1")
+	require.Contains(t, err.Error(), "jsonrpc")
+	require.Contains(t, err.Error(), "http://one:8545")
+	require.Contains(t, err.Error(), "http://two:8545")
+}
+
+func TestValidateUniqueProviderNames_ReportsEveryCollisionDeterministically(t *testing.T) {
+	// Refusing to boot is worth little if it names one collision at a time and the operator has to
+	// redeploy to find the next. It also must not vary run to run — the grouping is a Go map, and
+	// map order is exactly the nondeterminism that made the original bug hard to pin down.
+	endpoints := []*RPCStaticProviderEndpoint{
+		mag2724Endpoint("alpha", "ETH1", "jsonrpc", "http://a1:8545"),
+		mag2724Endpoint("alpha", "ETH1", "jsonrpc", "http://a2:8545"),
+		mag2724Endpoint("beta", "ETH1", "jsonrpc", "http://b1:8545"),
+		mag2724Endpoint("beta", "ETH1", "jsonrpc", "http://b2:8545"),
+		mag2724Endpoint("gamma", "POLYGON1", "rest", "http://g1:8545"),
+		mag2724Endpoint("gamma", "POLYGON1", "rest", "http://g2:8545"),
+	}
+
+	err := ValidateUniqueProviderNames(endpoints)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "alpha")
+	require.Contains(t, err.Error(), "beta")
+	require.Contains(t, err.Error(), "gamma")
+
+	for i := 0; i < 20; i++ {
+		require.Equal(t, err.Error(), ValidateUniqueProviderNames(endpoints).Error(),
+			"same config must produce the same message on every run")
+	}
+}
+
+func TestValidateUniqueProviderNames_AllowsRepeatedNameOutsideOneScope(t *testing.T) {
+	// One label per upstream operator, reused across chains and interfaces, is the normal way these
+	// configs are written. Each combination builds its own session manager with its own pairing map,
+	// so nothing collapses — rejecting these would refuse to start working deployments.
+	tests := []struct {
+		name      string
+		endpoints []*RPCStaticProviderEndpoint
+	}{
+		{
+			name: "same name on different chains",
+			endpoints: []*RPCStaticProviderEndpoint{
+				mag2724Endpoint("alchemy", "ETH1", "jsonrpc", "http://eth:8545"),
+				mag2724Endpoint("alchemy", "POLYGON1", "jsonrpc", "http://polygon:8545"),
+			},
+		},
+		{
+			name: "same name on different api-interfaces of one chain",
+			endpoints: []*RPCStaticProviderEndpoint{
+				mag2724Endpoint("alchemy", "ETH1", "jsonrpc", "http://eth:8545"),
+				mag2724Endpoint("alchemy", "ETH1", "rest", "http://eth:8546"),
+			},
+		},
+		{
+			name: "distinct names in one scope",
+			endpoints: []*RPCStaticProviderEndpoint{
+				mag2724Endpoint("primary", "ETH1", "jsonrpc", "http://one:8545"),
+				mag2724Endpoint("secondary", "ETH1", "jsonrpc", "http://two:8545"),
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			require.NoError(t, ValidateUniqueProviderNames(tt.endpoints))
+		})
+	}
+}
+
+func TestValidateUniqueProviderNames_RejectsCollisionAcrossStaticAndBackup(t *testing.T) {
+	// Static and backup providers live in separate maps but are looked up across both by address,
+	// so a name shared between the two lists is as ambiguous as one shared within either. This is
+	// why the boot path passes both lists in one call instead of checking each on its own.
+	static := []*RPCStaticProviderEndpoint{mag2724Endpoint("shared", "ETH1", "jsonrpc", "http://primary:8545")}
+	backup := []*RPCStaticProviderEndpoint{mag2724Endpoint("shared", "ETH1", "jsonrpc", "http://backup:8545")}
+
+	require.NoError(t, ValidateUniqueProviderNames(static), "each list is unique on its own")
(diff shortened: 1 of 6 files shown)


Tomelia1999:
1.  One part of the system still reads the old badge

Every 15 minutes there is a health re-check. It looks at the couriers currently working, asks each one "are you still alive?", and updates the list.

technical review:
spec_reverifier.go is a missed identity site (Major)

applyReverification does exactly what the PR body identifies as the bug class in mergeAbsentProviders, and it wasn't touched:

- spec_reverifier.go:164 — activeNames := byName(fresh), and byName (:258-264) keys by PublicLavaAddress.
- :172 — looked up with p.Name.
- :176, :203 — healthyNames filled with p.Name.
- :228 — read back as healthyNames[s.PublicLavaAddress].

On the degraded path layer 2 exists to protect, p.Name is SimTwin but s.PublicLavaAddress is SimTwin#0. Every epoch tick (15m):

- both colliding providers read !wasActive → appended to toAdmit → convertProvidersToSessions opens fresh DirectRPCConnections;
- both existing sessions miss healthyNames[s.PublicLavaAddress] → demoted → closeDemotedDirectConnections tears their transports down.

So the pair churns its connections and resets its QoS forever. That is worse than the "one silently dropped" the fallback is supposed to degrade to.

Separately, :173 keys demoteFailStreak by tier.String() + "|" + p.Name. Two colliding providers share one failure streak, so one flapping provider can push its healthy twin past reverifyDemoteThreshold.


2. The health command stops working on the config you need it for -  smart-router health now refuses to diagnose the config it's for

health_cmd.go:218 calls ParseStaticProviderEndpoints, which now hard-errors on duplicates. The command an operator reaches for to diagnose a bad config now refuses to run on it.

avitenzer:
Both findings were real. Fixed in `aa729b3`.

## 1. `spec_reverifier.go` — missed identity site

Confirmed, including the mechanism you traced. Reproduced it as a test before fixing: with the config side read as `p.Name` and the session side as `PublicLavaAddress`, a healthy collided pair comes back from cycle 1 as

```
demoted: [0x14a25c8ba090 0x14a25c8ba120]   // both sessions, every tick
```

both promoted *and* both demoted on the same tick — connections rebuilt and torn down every 15m, QoS reset each time. You're right that this is worse than the "one silently dropped" layer 2 is supposed to degrade to.

All four sites now key on `Address()` (`:172` active lookup, `:173` streak key, `:176`/`:203` healthy set); `:228` keeps reading `PublicLavaAddress`, which agrees again. `byName` is renamed `byAddress` with a note that callers must never look up by `Name` — the old name is what invited the mismatch, and `fakeConvert` in the test helpers was keyed by `Name` too, which is why the existing tests couldn't have caught this.

The `demoteFailStreak` point was also correct and is a separate bug: keyed `"<tier>|<address>"` now. Pinned behaviourally — twins alternate failures across two cycles, and under the shared counter the second increment reaches the threshold and demotes a provider whose own failure was a one-off.

## 2. `smart-router health` refuses the config it's for

Agreed, and the check was in the wrong place rather than needing a special case. A name collision is a property of the whole set of lists a router loads, not of one list — so the per-list check couldn't catch a name shared *across* static and backup anyway, and the boot path's `ValidateUniqueProviderNames(directRPCEndpoints, backupDirectRPCEndpoints)` already covers all three cases. It's now the only duplicate-name check on the boot path; `ParseStaticProviderEndpoints` is a parser again.

`health` resolves collisions instead of rejecting them, so it labels the rows apart — which also answers "which of my two nodes is the broken one":

```
$ smartrouter health dupcfg --use-static-spec specs/
rows: [('SimTwin#0', 'https://one...', False), ('SimTwin#1', 'https://two...', False)]
stderr: warn  duplicate provider names — the router REJECTS this config at startup;
              probing the colliding providers under distinct labels
```

and boot is still fail-fast:

```
$ smartrouter bootdup --use-static-spec specs/
Error: invalid direct-rpc endpoints definition ErrMsg: duplicate provider names: "SimTwin" on chain ETH1 ...
```

## Also

Swept for other `Name`-keyed identity maps after your catch — none left. One display-only site changed: the failed-provider retry log built its list from `Name`, so a collided pair printed the same name twice, which defeats the line's purpose.

Four new tests in `protocol/rpcsmartrouter/provider_names_mag2724_test.go`, each confirmed to fail with its own fix reverted (outputs above are from those runs). Full `rpcsmartrouter` and `lavasession` suites pass; the pre-existing selection-lock flake didn't reproduce this time.

avitenzer:
Rebased onto `main` (`12428dd`) — conflicts resolved, PR is mergeable again.

Both conflicts were in `rpcsmartrouter.go` against the MAG-2525 work, and in both cases main's version is the one that survived: `pruneRestoredFromFailed` now prunes both tiers, and the failed-provider retry log now covers static + backup. My change was only ever "match on `Address()` instead of `Name`", so it moved into the helper where that logic now lives rather than staying inline.

Worth flagging, since it wasn't in what you reviewed: rebasing turned up **two more sites of the same class** in code that landed on main after this branch, both real rather than cosmetic.

**`applyReverification`'s new third return.** `promoted` was built from `p.Name` and feeds `pruneRestoredFromFailed(rpsr.failedStaticProviders, chainKey, nameSet(promotedStatic))`, which matches on `Address()`. For a collided provider the two never meet, so the promotion doesn't supersede the pending failed-init retry — which is exactly the duplicate-session case the comment above that call warns about: `retryFailedProviders` revalidates, succeeds, and `mergeRecoveredSessions` appends a *second* session, landing the provider twice in `validAddresses` with double selection weight and leaking the superseded `DirectRPCConnection`. Now returns addresses; doc comment updated to say so.

**`activeProviders` (`rpcsmartrouter.go:3446`).** Builds its set from `s.PublicLavaAddress` and then looks up `p.Name` — the same mismatch as `applyReverification`. A collided provider reads as not live, so its node urls drop out of the subscription endpoint set. Now reads through `Address()`.

Neither was reachable before this branch (the addresses were identical strings), so they're not regressions in main — but they'd have shipped broken alongside it, and it's the same "an identity map keyed by a display label" pattern you called out. The `promoted` one is covered by an added assertion in the steady-state test; a spurious promotion there now fails the test for the same reason it would misbehave in production.

Also switched the two boot-validation log lines to `Address()` for the same reason as the retry log — a collided pair otherwise prints the same name twice.

`go build ./protocol/...`, `go vet`, and the full `rpcsmartrouter` (93s) and `lavasession` (32s) suites all pass on the rebased tree.

avitenzer:
@Tomelia1999 both findings were real, thank you — but the fix for the first one is that the code it applied to is gone. Rewritten as a single commit (`e5f3cef`), 326 lines instead of 744. Earlier comments on this PR described the previous approach and would only mislead now, so I removed them; this is the current state.

## The `<name>#<index>` renaming is removed

The router now refuses to start on a duplicate name and logs which ones to fix. Your review is most of the reason. `spec_reverifier.go` was the second place in one PR where a map keyed by provider identity had been left keyed on `Name`, and rebasing onto main turned up two more — `applyReverification`'s new `promoted` return, and `activeProviders`. Four instances of one mistake is the invariant telling you it doesn't hold. Dropping the name/identity split makes that entire class unrepresentable instead of merely fixed: if names are unique, keying by name is correct everywhere.

So finding 1 no longer applies — `spec_reverifier.go` is untouched by this PR, back at main. The churn you described can't occur because the two strings can never diverge.

## Finding 2 stands and is kept

`ParseStaticProviderEndpoints` does not reject duplicates. The check lives on the boot path only, so `smart-router health` can still load the config the router refuses:

```
$ smartrouter health dupcfg --use-static-spec specs/
rows: [('SimTwin', 'https://one...'), ('SimTwin', 'https://two...')]
stderr: warn  the router will REFUSE TO START on this config — probing it anyway
```

Both rows keep the configured name. The router's identity for them genuinely is ambiguous, and labelling them apart would claim a distinction the router doesn't make — they're separated by `url`, which is what identifies the node.

## The cost worth your eye

A duplicate name is now **fatal at boot**, which cuts against MAG-2525's "boot on any usable provider instead of exiting". The argument for accepting it: the config is unroutable as written, the operator has to change it either way, and failing loudly beats silently serving at half capacity. No live config trips it — checked standalone's four values files, the chart's own, and the repo's `direct-rpc` configs — so nothing deployed or in the dev gate crashloops.

The chart-side check in smart-router-helm-chart catches the same mistake a step earlier at template render, failing the release rather than the pods, and catches what the Go side structurally cannot see: `SimTwin` vs `simtwin`, or `My Node` vs `my-node`, which look distinct in `values.yaml` and arrive as one name after the configmap's `lower | replace " " "-"`.

## Verification

`go build ./protocol/...`, `go vet`, and the full `lavasession` + `rpcsmartrouter` suites pass. Both behaviours checked against the built binary, not only in unit tests: boot exits naming the collision and both urls; health loads the same config, warns, and probes both nodes.

Pre-existing flake unrelated to this work: `Failed locking selection after MaximumNumberOfSelectionLockAttempts` in the lavasession suite (1 of 5 runs with these changes stashed).

nimrod-teich (approved):

PR #282, opened 2026-08-16
feat(provideroptimizer): configurable endpoint selection policy (mode + priority presets)

avitenzer:
Jira ticket: MAG-2780

Consolidates the three-PR stack #262 → #263 → #264 into one PR, rebased onto current `main`. One behavioural change versus the stack: the rename now targets `EndpointSelector`, not `ProviderSelector` — see [The rename changed target](#the-rename-changed-target).

Implements FAILOVER-TASKS section 1: let the operator decide **what** the router optimises for when choosing who serves a request, and **how** the winner is picked.

Three commits, reviewable independently and in order.

---

## 1. `--qos-selection-mode` — how the winner is picked

```
--qos-selection-mode weighted_random|best   (default: weighted_random)
```

`best` always routes to the highest-scoring endpoint. The default is unchanged, so **production behaviour is identical unless the flag is set**.

### Why the split lands where it does

Only one step in the selection flow is mode-dependent. Scoring, the stats payload, the metrics emit, and the `wantedProviders` sampling loop are all shared:

| layer | mode-dependent? |
|---|---|
| CSM branch + `wantedProviders` loop | no |
| `ChooseProviderWithStats` | no |
| `CalculateProviderScores` | no |
| **turning `[]ProviderScore` into one address** | **yes — the whole delta** |

So `SelectProviderWithStats` is split along that seam rather than forked wholesale:

- `pickWeightedIndex` — today's cumulative-probability draw, verbatim
- `pickBestIndex` — the new argmax
- `buildSelectionResult` — the debug breakdown and `SelectionStats`, previously duplicated across **four** return paths

Hoisting the reporting is what makes the second mode cost ~12 lines. `SelectProviderWithStats` went from 108 lines to ~50 across four focused functions.

The guards stay mode-independent. Notably the all-zero-scores path stays *uniform* for both modes: with no signal to rank on, an argmax would be an arbitrary N-way tie, so uniform is the honest answer either way.

### Tie-breaking is not defensive polish

`pickBestIndex` breaks ties uniformly at random via one-pass reservoir sampling rather than first-wins.

`CalculateScore` collapses every endpoint below `MinAcceptableAvailability` to *exactly* `minSelectionChance` (MAG-2237). On a degraded chain that means exact N-way ties are routine — and a first-wins scan would pin 100% of traffic onto whichever address happens to sort first in the pairing list. That is precisely the starvation the floor exists to prevent.

Known limitation, deliberately accepted: when every endpoint is collapsed to the floor, `best` mode has nothing to rank on and degrades to a uniform random pick. Ranking on the unfloored composite would re-open the MAG-2237 hole (latency/sync freeze at last-healthy values for a failing endpoint, so a dead node can look fast). Recovery is driven by the prober, not by routing.

### Observability

`SelectionStats` gains `Mode`, surfaced in the selection-stats header ahead of RNG:

```
[provider1: 0.900, 0.800, 0.700, 0.100, 0.750] | Mode: best | RNG: 0.000000 | Selected: provider1
```

An `RNGValue` of `0` is otherwise ambiguous — it means *no draw happened* under `best`, but *single candidate* or *all scores zero* under `weighted_random`. `/debug/runtime-config` reports the mode per chain alongside the weights.

### Drive-by: removed `ChooseBestProvider` / `ChooseBestProviderWithStats`

Despite the name, these were semantically identical to `ChooseProvider` / `ChooseProviderWithStats` — same scoring, same weighted draw, same return shape, differing only in two trace strings. "Best" never meant highest-scoring; sticky sessions were getting an ordinary weighted draw.

The sticky-session path now calls `ChooseProviderWithStats` directly with no behaviour change, and the non-stats variant had no callers at all.

---

## 2. `WeightedSelector` → `EndpointSelector`

Once commit 1 lands, `WeightedSelector` names a type that does weighted random selection *and* highest-score selection. Scoring and reporting are shared; only the final pick varies by mode. The name promises one of the two.

Kept as its own commit deliberately — a repo-wide mechanical rename mixed into a feature diff makes the feature impossible to review.

### The rename changed target

The original #263 renamed to `ProviderSelector`. @nimrod-teich requested changes on that PR: *"We want to move away from the 'provider' terminology."*

This PR renames to **`EndpointSelector`** instead. "Endpoint" is the term the codebase is already converging on — `protocol/endpointstate`, `EndpointMetrics`, `GetAllDirectRPCEndpoints`, `EndpointLagThreshold`. So this moves *with* that direction rather than against it.

Scope note: this renames the selector type only. The surrounding package is still `provideroptimizer` and the wider `provider*` vocabulary is untouched — a full terminology migration is a much larger change and belongs in its own ticket.

### Identifier map

| before | after |
|---|---|
| `WeightedSelector` | `EndpointSelector` |
| `WeightedSelectorConfig` | `EndpointSelectorConfig` |
| `NewWeightedSelector` | `NewEndpointSelector` |
| `DefaultWeightedSelectorConfig` | `DefaultEndpointSelectorConfig` |
| `ConfigureWeightedSelector` | `ConfigureEndpointSelector` |
| `GetWeightedSelectorConfig` | `GetEndpointSelectorConfig` |
| `UpdateWeightedSelectorStrategy` | `UpdateEndpointSelectorStrategy` |
| `po.weightedSelector` | `po.endpointSelector` |
| `weighted_selector.go` | `endpoint_selector.go` (and `_test`) |

### Deliberately NOT renamed

These genuinely describe the weighted draw and stay correct:

- `pickWeightedIndex` — this one really is the weighted selection
- `SelectionModeWeightedRandom` / the `weighted_random` flag value
- `updateDecayingWeightedAverage` — unrelated EWMA helper
- `"weighted selection"` prose in tests that exercise the default mode
- `"weighted composite score"` / `"weighted contributions"` comments — the composite genuinely is a weighted sum of the QoS axes

### The six non-mechanical edits

Everything else is a pure identifier substitution. These six exist only because the old name had leaked into prose:

1. The `EndpointSelector` doc comment still claimed the type "implements continuous weighted random selection" — now only half true, so it's rewritten to say scoring is shared and only the pick varies.
2. – 4. Three weight-validation warning strings named the old type (`"invalid weighted selector weight"` → `"invalid endpoint selector weight"`, etc.). These are startup config-validation warnings, not hot-path logs.
5. The `SetDeterministicSeed` comment named the old type.
6. The fallback comment in `NewEndpointSelector` now lists `SelectionMode` alongside `Strategy`/`MinSelectionChance` as user choices preserved when invalid weights reset to defaults.

**Note for reviewers:** three log-message strings changed (the weight-validation warnings). If anything greps those, it needs updating — though they only fire on invalid startup config.

---

## 3. `--qos-selection-priority` — what to optimise for

```
```

|---|---|---|---|---|

**Dominant, not exclusive.** The chosen axis takes 0.70, but availability keeps real weight in every preset so a fast-but-flaky endpoint cannot outrank a fast-and-solid one on latency alone. `balanced` is byte-identical to today's defaults — it's the flag's default value, so anything else would silently move every existing deployment.

### Why the axis presets zero stake

`CalcWeightsByStake` hands every static endpoint an identical weight (`maxWeight * 10`, and `maxWeight` stays 1 in an all-static deployment). So the stake term is **the same constant for every candidate**:

- under `weighted_random` it actively flattens the lottery — everyone sits on the same pedestal, compressing the probability ratio between a great endpoint and a mediocre one

### Composes with `--qos-selection-mode`, doesn't duplicate it

| flag | decides |
|---|---|

### Precedence

### Drive-by fix the new tests surfaced

`ParseSelectionMode` (from commit 1) rejected the empty string, which tied startup correctness to the flag being viper-bound: any call site that hadn't bound it reads `""` and aborts with a confusing `invalid selection mode:` instead of using the default. **Empty now means "not specified"** and resolves to the default in both parsers. Genuinely wrong values (`bset`, `quickest`) still hard-fail — that's the property that actually matters.

### `--strategy` deliberately untouched

Measured: its only scoring effect is an exponent nudge worth **at most +0.015 on the composite** (typically +0.006), and it does not move the weights at all. `StrategyBalanced` — the default, and the only reachable value for YAML users since the flag has no `BindPFlag` — is a genuine no-op in the switch. So it cannot interfere with this flag.

---

## Verification

- `go build ./...` — clean
- `go vet ./...` — clean repo-wide
- `make test` (`go test ./... -count=1 -timeout 300s`) — zero failures across all 33 test packages
- `--help` output verified for both flags

Rebased onto current `main` with **zero conflicts**, including `protocol/rpcsmartrouter/rpcsmartrouter.go`, which moved on `main` since the stack was cut.

Tests include the end-to-end proof that the preset is more than a table of numbers: given the same two endpoints, `fastest` and `most-reliable` pick **different** ones. Plus the sum-to-1.0 invariant (a drifted preset would be silently renormalised by `NewEndpointSelector`), the `balanced == today's defaults` backward-compat pin, and both halves of the precedence rule.

For commit 1: best-scorer-placed-last (fails a first-wins scan), uniform tie-break across three floor-collapsed endpoints, leader tracking, mode stamping across all six mode × short-circuit combinations, header formatting, parse/reject coverage, and a regression pin that the default mode still spreads 80/20. `TestDebugRuntimeConfig_SmartRouter_PerChainOptimizer` gives its two chains *different* modes, so the field is proven to travel per-chain rather than from a global default.

For commit 2, verified mechanical beyond the test suite: the same transform was applied to the pre-rename tree and diffed against this one. Only the six prose edits above remain, and the test-file diff is empty — so no logic moved.

## Notes for review

- The selection-stats header gained a `| Mode: <mode> |` segment. No test parses it, but external tooling reading it positionally would need a look.
- The mode and priority are global across chains (one flag → every chain's optimizer), consistent with the existing `qos-*` weight flags. Going per-chain later is config plumbing only; the selection seam does not move.

Supersedes #262, #263 and #264.

### DIFF
--- a/protocol/common/cobra_common.go
+++ b/protocol/common/cobra_common.go
@@ -96,6 +96,7 @@ const (
 	ProviderOptimizerStakeWeight        = "qos-stake-weight"         // weight for stake (default: 0.1)
 	ProviderOptimizerMinSelectionChance = "qos-min-selection-chance" // minimum selection probability for any provider (default: 0.01)
 	ProviderOptimizerSelectionMode      = "qos-selection-mode"       // how the winner is picked from the scored providers: weighted-random (default) or best
+	ProviderOptimizerSelectionPriority  = "qos-selection-priority"   // preset over the four weights above: balanced (default), most-reliable, fastest, freshest
 
 	// optimizer qos sampling cadence — drives the in-memory /metrics
 	// selection-score cache and the OTel optimizer_qos emit.
--- a/protocol/common/cobra_common.go
+++ b/protocol/common/cobra_common.go
@@ -95,6 +95,7 @@ const (
 	ProviderOptimizerSyncWeight         = "qos-sync-weight"          // weight for sync score (default: 0.2)
 	ProviderOptimizerStakeWeight        = "qos-stake-weight"         // weight for stake (default: 0.1)
 	ProviderOptimizerMinSelectionChance = "qos-min-selection-chance" // minimum selection probability for any provider (default: 0.01)
+	ProviderOptimizerSelectionMode      = "qos-selection-mode"       // how the winner is picked from the scored providers: weighted-random (default) or best
 
 	// optimizer qos sampling cadence — drives the in-memory /metrics
 	// selection-score cache and the OTel optimizer_qos emit.
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -1359,7 +1359,7 @@ func (csm *ConsumerSessionManager) getValidProviderAddresses(ctx context.Context
 		providers = csm.getTopTenProvidersForStatefulCalls(validAddresses, ignoredProvidersList)
 	} else if stickiness != "" {
 		var selectionStats *provideroptimizer.SelectionStats
-		providers, selectionStats = csm.providerOptimizer.ChooseBestProviderWithStats(ctx, validAddresses, ignoredProvidersList, cu, requestedBlock)
+		providers, selectionStats = csm.providerOptimizer.ChooseProviderWithStats(ctx, validAddresses, ignoredProvidersList, cu, requestedBlock)
 		if selectionStats != nil {
 			csm.setSelectionStats(selectionStats)
 		}
--- a/protocol/lavasession/consumer_types.go
+++ b/protocol/lavasession/consumer_types.go
@@ -115,8 +115,6 @@ type ProviderOptimizer interface {
 	AppendRelayDataConsensus(providerAddress string, latency time.Duration, cu, syncBlock uint64, syncRef provideroptimizer.SyncReference)
 	ChooseProvider(ctx context.Context, allAddresses []string, ignoredProviders map[string]struct{}, cu uint64, requestedBlock int64) (addresses []string)
 	ChooseProviderWithStats(ctx context.Context, allAddresses []string, ignoredProviders map[string]struct{}, cu uint64, requestedBlock int64) (addresses []string, stats *provideroptimizer.SelectionStats)
-	ChooseBestProvider(ctx context.Context, allAddresses []string, ignoredProviders map[string]struct{}, cu uint64, requestedBlock int64) (addresses []string)
-	ChooseBestProviderWithStats(ctx context.Context, allAddresses []string, ignoredProviders map[string]struct{}, cu uint64, requestedBlock int64) (addresses []string, stats *provideroptimizer.SelectionStats)
 	GetReputationReportForProvider(string) (*pairingtypes.QualityOfServiceReport, time.Time)
 	Strategy() provideroptimizer.Strategy
 	UpdateWeights(map[string]int64, uint64)
--- a/protocol/provideroptimizer/dead_provider_starved_test.go
+++ b/protocol/provideroptimizer/dead_provider_starved_test.go
@@ -18,7 +18,7 @@ import (
 // endpoint therefore kept a composite of ~2/3 of a healthy peer and was still selected ~7–18% of
 // the time (measured), instead of being starved to the MinSelectionChance floor.
 //
-// Fix (weighted_selector.CalculateScore): once availability falls below the acceptable minimum
+// Fix (endpoint_selector.CalculateScore): once availability falls below the acceptable minimum
 // (availabilityScore == 0, i.e. raw availability < score.MinAcceptableAvailability), collapse the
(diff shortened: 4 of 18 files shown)


avitenzer:
Closing in favour of #262, which is now the single consolidated PR.

I opened this before noticing that #263 and #264 had already been merged up into `feat/provider-selection-mode` during the session, which had already made #262 the consolidated one. The `EndpointSelector` rename, the `Jira ticket: MAG-2780` line, and the rebase onto current `main` have all been folded into #262 instead.

Verified equivalent: all 16 touched files are byte-identical between this branch and #262's head.

PR #283, opened 2026-08-16
fix(rpcsmartrouter): score retryable node errors against availability (MAG-2156)

avitenzer:
# Description

The optimizer never penalized providers that return JSON-RPC node errors over HTTP 200, so a provider failing most of its relays kept a perfect availability score and stayed in the first-pick rotation.

A JSON-RPC node error is HTTP 200 with `{"error":...}` in the body, so the direct-RPC completion gate — `err != nil || statusCode >= 500 || statusCode == 429` — never fired for it. The relay landed on `OnSessionDone`, which fed the optimizer an availability sample of **1** for a response that had failed.

The classification was never the problem: `direct_rpc_relay.go` already tags these `IsNodeError=true`, and the optimizer's penalty machinery (including the MAG-2237 collapse) works correctly. The gate simply never read the flag, so the availability store never saw the failures. `OnSessionDone` / `OnSessionFailure` are the only two production QoS write paths, both in this goroutine, so there was no second path that could rescue the score. The `NodeErrorRate` in `consumer_optimizer_qos_client.go` is telemetry only — nothing reads it for selection.

## The fix

Extract the gate into `shouldFailSessionForResult` and have it score node errors, with two carve-outs on **two different axes**:

- **`IsNonRetryable`** — "retrying elsewhere would not help". Deterministic caller-fault errors (unsupported method, invalid params, execution reverted) come back from *every* endpoint for the same request. Scoring them would drive the whole pairing below `MinAcceptableAvailability` on a burst of malformed client requests and collapse every endpoint to the selection floor — punishing healthy nodes for a bad request instead of identifying a bad node.
- **`IsRateLimited`** — "the endpoint is healthy but busy". `SubCategoryRateLimit` is contractually backoff-without-marking-unhealthy, and this does **not** follow from retryability: `NODE_RATE_LIMITED` (2005) is `Retryable: true` while `NODE_LIMIT_EXCEEDED` (2011) is `Retryable: false`, and both must stay out of the availability signal.

The second carve-out is new in this PR's second revision. The first revision gated on `IsNonRetryable` alone, which conflated "would a retry help" with "is this the node's fault" and silently started scoring 2005 against availability — widening the very `SubCategoryRateLimit` contradiction the PR called out as out of scope.

`RelayResult.IsRateLimited` is the new plumbing. The four transports now stamp the whole flag set through one `RelayResult.ApplyNodeErrorClassification` call instead of assigning fields individually, which is how `IsUnsupportedMethod` came to be classified in the registry but never assigned on the direct path — reachable in theory, dead in production.

### What is deliberately scored

Documented at the gate rather than left implicit:

1. **Unclassified error bodies.** `ClassifyNodeErrorForRetry` returns all-false when nothing matches, and a false flag is an absence of information, not a verdict. Narrowing the gate to positively-classified errors would exempt exactly the novel failures the ticket is about, so these are scored.
2. **Retryable but request-shaped errors** — `CHAIN_BLOCK_NOT_FOUND` (3201), `CHAIN_TX_NOT_FOUND` (3202), `CHAIN_RECEIPT_NOT_FOUND` (3203), `CHAIN_DATA_NOT_AVAILABLE` (3205), `NODE_RESOURCE_NOT_FOUND` (2012), `NODE_RESOURCE_UNAVAILABLE` (2013). `Retryable: true` is defined as "another endpoint has a chance of succeeding", so the endpoint is the differentiator and demoting it is defensible. Where it is not — every endpoint lacks the tx — all demote equally and `weighted_selector` degrades to uniform random rather than starving. If that proves wrong in production, tagging these with a fault-axis `SubCategory` fixes it with no change to the gate.

## Measured effect

The ticket's scenario driven **through the gate** — 120 relays × 5 seeds, P1 returning `-32603` on 90% of its own traffic, P2/P3 healthy. "Before" runs the pre-fix predicate verbatim against the same relay stream:

| | P1 availability | P1 first-pick share |
|---|---|---|
| Before | ~1.0000 | 32.0% (192/600) |
| After | 0.20–0.33, below the 0.80 floor | 2.8% (17/600) |

This clears the ticket's acceptance criterion (share below 20%, and below both healthy peers).

- `protocol/rpcsmartrouter/rpcsmartrouter_server.go` — the gate and its two carve-outs.
- `protocol/rpcsmartrouter/direct_rpc_relay.go` — the gRPC status-error branch, which previously returned `IsNodeError` with `err == nil` and no classification at all, so the gate had no carve-out to consult there.

## Tests

- `node_error_scoring_mag2156_test.go` — 13-case table over the gate. Reverting the production change fails exactly the `node error over HTTP 200` case.
- `node_error_demotion_mag2156_test.go` (`protocol/rpcsmartrouter`) — the ticket's acceptance criterion driven through the real chain: upstream error code → `ClassifyNodeErrorForRetry` → `RelayResult` flags → `shouldFailSessionForResult` → optimizer → `ChooseProvider`. Subtests differ **only** in which error the endpoint returns; whether anyone is demoted is the gate's verdict. Verified: reverting the gate reproduces the bug inside the measurement (P1 holds `1.0000` at a 90% error rate). Deterministically seeded across 5 seeds rather than relying on repeat runs.
- `unsupported_method_scoring_invariant_test.go` (`protocol/common`) — the gate delivers the `SubCategoryUnsupportedMethod` "no provider scoring" contract via `IsNonRetryable` and never reads the subcategory. That works only because all four such codes happen to be `Retryable: false`. This iterates the registry and fails if one is ever registered retryable, which is what actually guards the contract.

Full suites pass for `rpcsmartrouter`, `provideroptimizer`, `lavasession`, `relaycore`, `common`; `go test ./protocol/...`, `go build ./...` and `go vet ./...` clean.

## Reviewer note — rollout

`OnSessionFailure` does more than score, and three things step up together on first deploy:

1. **Session-failure metrics.** `qosManager.AddFailedRelay` and `updateMetricsManager(..., false)` now fire on scored node errors. Dashboards will show a step change independent of any selection effect. Worth pre-announcing so it is not read as a regression.
2. **`ConsecutiveErrors` / blocklisting.** At 15 consecutive failures the *session* is blocklisted. Provider-level blocking additionally requires zero successful relays — so a 90%-error provider is not blocked outright, but an endpoint failing *every* relay is blocked on the 16th, and the gate runs per-endpoint inside one request, so a deterministic burst reaches all N endpoints at once. The pool resets once drained. The two carve-outs are exclusions rather than score adjustments precisely because this escalation is sharp; the residual exposure is the unclassified-error case above.
3. **Selection under total demotion.** If every endpoint falls below `MinAcceptableAvailability`, `weighted_selector.go:260` collapses each to `minSelectionChance` — selection does not starve, but latency and sync differentiation is lost until the prober recovers them.

Suggested first deploy: watch session-failure rate and provider-block counts on one chain before rolling wider.

### Still out of scope

- `statusCode == 429` fails the session, contradicting the same `SubCategoryRateLimit` contract. Pre-existing and untouched — but note this PR no longer *widens* that contradiction, which was the reason to reconsider it.
- `rest_integration_test.go:434` documents a separate open bug where a REST 501 becomes a transport error.
- gRPC `UNAUTHENTICATED` (16) classifies Unknown and is therefore scored. Deliberate: credentials are configured per endpoint, so an endpoint rejecting ours is unusable to us. Noted at the call site.

The ticket's noted stale `0.90` comment in `weighted_selector.go` was already fixed on main — it reads `0.80` today.

The xfail on `test_selection_demotes_error_prone_provider` (per the 2026-06-25 ticket comment) lives outside this repo and is the remaining end-to-end verification.

## Jira ticket

Jira ticket: MAG-2156

---

## Author Checklist

I have...

* [ ] confirmed all CI checks have passed — pending CI run


AnnaR-prog:
## Review

The bug is real and the fix is in the right place. I confirmed the load-bearing part of the diagnosis: `OnSessionDone` and `OnSessionFailure` have exactly two call sites in the entire non-test tree, both inside `sendRelayToDirectEndpoints`, so there was genuinely no second path that could have rescued the score — the classification was already correct upstream and the gate simply never read it.

**What I ran:** `go build ./...` + `go vet` clean; `rpcsmartrouter` / `provideroptimizer` / `lavasession` / `common` suites pass; the demotion test 30× (not flaky); a test-merge against current `main` (14 commits ahead, 2 touching this file) — clean. Reverting the production change fails the gate test on exactly one case (`node error over HTTP 200`), as claimed.

---

### 1. The demotion test does not exercise the fix

`node_error_demotion_mag2156_test.go` is in package `provideroptimizer` and calls `po.AppendRelayFailure` / `po.AppendRelayData` directly — it never reaches `shouldFailSessionForResult`. I reverted the production change and **it still passes**.

Three consequences:

- Its doc comment says "this subtest fails without it" — verified false.
- The PR body calls it "the ticket's acceptance criterion end to end" — it is an optimizer property test, not end-to-end.
- The **Measured effect** table (36.7% → ~2%) is therefore a simulation of the optimizer given assumed inputs, not a measurement of this fix. The numbers aren't wrong about the optimizer; they just don't measure what the body says they measure.

Cheapest fix: build `RelayResult`s, route them through `shouldFailSessionForResult`, and feed the optimizer from the gate's verdict. Otherwise relabel the test and drop the end-to-end claim. The gate test itself is genuinely good and I'd keep it as-is.

### 2. `IsNonRetryable` is the wrong predicate for "not the node's fault"

The gate needs *is this the node's fault*; it is using *would retrying elsewhere help*. Those coincide for most errors and diverge here:

- Unmatched errors default to `Retryable: true` **by documented policy** (`error_codes.go` — "retrying unknowns is the deliberate choice"), so every **unclassified** JSON-RPC error body is now scored against availability.
- Explicitly retryable but request-shaped, i.e. returned identically by every endpoint: `CHAIN_BLOCK_NOT_FOUND` (3201), `CHAIN_TX_NOT_FOUND` (3202), `CHAIN_RECEIPT_NOT_FOUND` (3203), `CHAIN_DATA_NOT_AVAILABLE` (3205), `NODE_RESOURCE_NOT_FOUND` (2012), `NODE_RESOURCE_UNAVAILABLE` (2013).

To be fair to the PR: the three examples it names are all genuinely `Retryable: false` (2001, 4003, 3101), and all four `SubCategoryUnsupportedMethod` errors (2001 / 2008 / 2009 / 2010) are non-retryable — that contract does hold. `CHAIN_STATE_PRUNED` and `NODE_SYNCING` being scored is desirable. The issue is that the description claims the general property ("deterministic caller-fault errors are carved out") while the flag only delivers the specific instances. Either narrow the claim, or gate on `Category` / `SubCategory` alongside `Retryable`.

### 3. A sustained node-error burst drains the pairing pool

I drove `OnSessionFailure` with the exact error this change generates against a 3-provider pairing, every relay failing. A provider is blocked at the **16th** consecutive failure (`MaximumNumberOfFailuresAllowedPerConsumerSession = 15` → session `BlockListed` → `usedComputeUnits <= LatestRelayCu` → `blockProvider = reportProvider = true`, `consumer_session_manager.go:2036-2050`). Observed run:

```
iter 42: valid=[provider5 provider6]  blocked=[provider4]
iter 46: valid=[provider6]            blocked=[provider4 provider5]
iter 48: valid=[]                     blocked=[provider4 provider5 provider6]
iter 49: pool reset
```

The reviewer note only addresses the 90%-error case ("will not be blocked outright") — true, but the zero-successful-relay case is the one that bites, and it is exactly the deterministic-burst scenario finding 2 leaves uncovered. Caveat: my harness feeds failures directly, so this demonstrates the mechanism, not the classification link; that link is finding 2.

### 4. Rate-limit scope inconsistency

The description flags `statusCode == 429` as contradicting the `SubCategoryRateLimit` "healthy but busy, do not mark unhealthy" contract and declines to fix it as out of scope — while newly extending that same contradiction through the body path: `MessageContains("rate limit")` → `NODE_RATE_LIMITED` (2005), which carries that subcategory but is `Retryable: true`, so it is now scored. `NODE_LIMIT_EXCEEDED` (2011) shares the subcategory but is `Retryable: false` and stays carved out — good evidence that the two axes are orthogonal. Either both instances are acceptable or both are problems; it's odd to call one out of scope while widening it.

### Nits

- "`IsNonRetryable` is already computed on this exact result object for all four transports" is inaccurate. `direct_rpc_relay.go:773` (gRPC status-error branch) sets `IsNodeError: grpcErr.Code >= 13`, returns `err == nil`, and never sets `IsNonRetryable` — no carve-out is possible there, so gRPC 16 `UNAUTHENTICATED` now scores as an availability failure.
- The `unsupported method` test case carries no signal: the gate never reads `IsUnsupportedMethod`, so it passes purely on `IsNonRetryable` and is byte-equivalent to the case above it — its stated rationale misattributes why it passes. Compounding this, `relayResult.IsUnsupportedMethod` is never copied at `rpcsmartrouter_server.go:3867-3873`, so the flag cannot be set on the direct path at all and the combination is unreachable in production.
- The `relayResult == nil → false` guard is unreachable from the only call site: line 1860 dereferences `localRelayResult.StatusCode` before line 1861 calls the helper.
- Cross-validation amplifies findings 2 and 3 in the same direction — the gate runs per-endpoint goroutine, so one deterministic error hits all N endpoints within a single request.
- `OnSessionFailure` also drives `qosManager.AddFailedRelay` and `updateMetricsManager(..., false)`, so session-failure metrics and dashboards will step up on first deploy independently of the optimizer effect. Worth pre-announcing.
- When *all* providers fall below `MinAcceptableAvailability`, `weighted_selector.go:260` collapses each to `minSelectionChance` — selection doesn't starve, but it degrades to uniform random with all latency/sync differentiation lost until the prober recovers them.

---

### Verdict

I'd hold merge on **finding 1** only — a test that advertises itself as the regression guard and isn't shouldn't ship, and it's cheap to fix. Findings 2–4 want an author response: at minimum narrow the description's safety claim to what `IsNonRetryable` actually means, and either address the unclassified-error exposure or accept it explicitly with a rollout plan. The rest are nits.

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

avitenzer:
Thanks — this was a good catch on finding 1 in particular, and finding 2's framing ("the gate needs *is this the node's fault*; it is using *would retrying elsewhere help*") is the right diagnosis. All four findings verified and addressed in e35c0f4. PR description rewritten, since several of the claims you flagged were in it.

### 1. The demotion test did not exercise the fix — fixed

Confirmed exactly as described: it was in package `provideroptimizer`, called `AppendRelayFailure` directly, and its subtests branched on a test-controlled `scoreable` bool. Reverting the production change left it green.

Rewritten in package `rpcsmartrouter` and driven through the real chain:

```
upstream error code/message
  -> common.ClassifyNodeErrorForRetry   (real registry)
  -> common.RelayResult flags           (as direct_rpc_relay.go sets them)
  -> shouldFailSessionForResult         (the fix)
  -> AppendRelayFailure / AppendRelayData
  -> ProviderOptimizer.ChooseProvider
```

The subtests now differ *only* in which error the endpoint returns — `-32603` vs `-32601` vs a rate limit. Whether anyone gets demoted is the gate's verdict, not the test's. Verified both ways: reverting the gate fails the test, and with the precondition assertion also removed it fails inside the measurement itself with `p1=1.0000` / `"1" is not less than "0.8"` — i.e. it reproduces the ticket's bug rather than a bookkeeping error.

Also switched from `rand.InitRandomSeed()` to `po.SetDeterministicSeed` across five seeds. A test asserting a share threshold should be reproducible rather than needing 30 runs to establish it isn't flaky.

The **Measured effect** table is now measured through the gate, with "before" running the pre-fix predicate verbatim against the same relay stream: **32.0% → 2.8%** first-pick share, availability `~1.0000` → `0.20–0.33`. Close to the old numbers, but now they measure what the caption says.

### 2. `IsNonRetryable` is the wrong predicate — agreed, split the axes

Both bullets confirmed. `ClassifyNodeErrorForRetry` returns the zero value on `LavaErrorUnknown`, so every unclassified body had `IsNonRetryable=false` and was scored.

I've done both things you offered as alternatives, because they turned out to apply to different halves:

- **Gated on the SubCategory** for the rate-limit half — that's finding 4, below.
- **Narrowed the claim** for the rest. The description no longer asserts the general property. The gate documents scoring of unclassified bodies and of the six request-shaped-but-retryable codes as deliberate, with the reasoning: for unclassified, restricting to positively-classified errors would exempt precisely the novel failures the ticket is about; for 3201/3202/3203/3205/2012/2013, `Retryable: true` is *defined* as "another endpoint has a chance of succeeding", so the endpoint is the differentiator. Where it isn't — every endpoint lacks the tx — all demote equally and selection degrades to uniform rather than starving. If that's wrong in production, tagging those codes with a fault-axis `SubCategory` fixes it with no change to the gate.

On the unsupported-method contract holding "by coincidence" — you're right that nothing enforced it. Added `TestUnsupportedMethodCodesAreAllNonRetryable_MAG2156` in `protocol/common`, which iterates the registry and fails if a `SubCategoryUnsupportedMethod` code is ever registered `Retryable: true`. That's the assertion that breaks instead of the contract silently.

### 4. Rate-limit scope inconsistency — fixed (taking this before 3, since it changes 3's answer)

Fair hit, and the sharpest one: declaring 429 out of scope while widening the same contradiction through the body path isn't a defensible position. `NODE_RATE_LIMITED` (2005) / `NODE_LIMIT_EXCEEDED` (2011) sharing a subcategory across opposite retryability is, as you say, exactly the evidence that the axes are orthogonal.

Added `RelayResult.IsRateLimited`, populated from `SubCategoryRateLimit`, and the gate now carves it out independently. Pinned by `TestRateLimitCarveOutIsIndependentOfRetryability_MAG2156` (same subcategory, opposite retryability, same verdict) and `TestRateLimitSubCategorySpansBothRetryabilities_MAG2156`.

`statusCode == 429` stays out of scope as pre-existing — but this PR no longer widens it, which was the actual objection.

### 3. Sustained node-error burst drains the pairing pool — accepted, documented, partially reduced

Reproduction accepted; the 16th-consecutive-failure path and the zero-successful-relay condition are as you traced them. Two responses:

- The exposure is narrower than it was: the rate-limit carve-out removes a whole class of burst (a rate-limited upstream is now a non-event for scoring rather than a pool-drain vector). What remains is the unclassified-error case from finding 2, which is one item, not two.
- The rest is accepted rather than fixed. It's documented at the gate — including the point you raised that the gate runs per-endpoint inside one request, so a deterministic burst reaches all N at once — and the PR now carries a rollout section instead of a one-line reviewer note. Also pre-announced there: `AddFailedRelay` and `updateMetricsManager(..., false)` stepping up on first deploy, and the `weighted_selector.go:260` uniform-collapse behaviour under total demotion.

### Nits

- **gRPC branch never set `IsNonRetryable`** — correct, and the "all four transports" claim was wrong; removed. That branch returns `err == nil` with `IsNodeError` set, the same shape as the body-error paths, but never classified — so there was no carve-out to consult there *at all*. It now classifies like the others. gRPC 16 still scores (it classifies Unknown), which I've kept deliberately and documented at the call site: credentials are per-endpoint, so an endpoint rejecting ours is genuinely unusable to us; if all reject, all demote together and selection goes uniform rather than starving.
- **`unsupported method` test case carried no signal** — right on both counts. `relayResult.IsUnsupportedMethod` was never copied at `rpcsmartrouter_server.go:3867-3873`, and it was never *assigned* in `direct_rpc_relay.go` either, so the flag was dead end to end on the direct path. Both fixed. The four transports now stamp the flag set through one `RelayResult.ApplyNodeErrorClassification` call rather than assigning fields individually — that per-field assignment is how the flag went missing, and it would have taken `IsRateLimited` the same way. The test case's comment now says plainly that it passes on `IsNonRetryable` and points at the registry invariant as the thing that actually guards it.
- **`relayResult == nil` guard unreachable** — correct. Kept, with a comment saying so: the helper is independently tested, and I'd rather it be total on its own signature than depend on a caller invariant. Happy to drop it if you'd prefer.
- **Cross-validation amplification** — folded into the gate's doc comment.

### Not changed

`rest_integration_test.go:434` (REST 501 → transport error) stays out of scope as a separate pre-existing bug.

Full suites pass for `rpcsmartrouter`, `provideroptimizer`, `lavasession`, `relaycore`, `common`; `go test ./protocol/...`, `go build ./...`, `go vet ./...` clean.

🤖 Response assisted by [Claude Code](https://claude.com/claude-code)

AnnaR-prog (approved):

PR #284, opened 2026-08-17
feat(smart-router/debug): add cross-validation-events (MAG-2772)

avitenzer:
Jira ticket: MAG-2772

## What

`GET /debug/cross-validation-events` — a read-only, per-request record of the cross-validation dissent the router observed. Built to Victoria's design from the ticket comments, after the 2026-08-16 decision to build the endpoint rather than repoint the parked tests at the Go tests.

## Why the existing surfaces don't work

The router already recorded this dissent twice, and neither can be asserted on from an automated run:

- **The info logs** (`cross-validation outlier detected` / `cross-validation straggler resolved`) — diagnostic evidence only by team rule, never a gating assertion.
- **`smartrouter_cross_validation_mismatch_total`** — on the unpublished metrics port. Reading it needs a hand-held port-forward, and when the tunnel is down the read comes back empty and the test passes having measured nothing. That is the silent pass `/debug/provider-scores` was added to end (MAG-2707, #259); this applies the same pattern.

The counter also can't answer what the parked tests ask. Its labels are `{spec, apiInterface, method, group, finality}` — no provider, no request id — so counts are derivable from events, but events are not derivable from the counter. Hence event-shaped.

## The endpoint

Flat JSON array of self-describing rows (each carries its own `ChainID` + `ApiInterface`), oldest first, from both recording paths:

| `Source` | Meaning |
| --- | --- |
| `reply-time` | a content outlier seen before the reply — it is in `lava-cross-validation-disagreeing-providers` |
| `straggler` | a provider that lost the race to quorum (it was in `pending-providers`) whose late answer the async watcher resolved |

Per row: `RequestID` (the **lava-guid**, i.e. the `Lava-Guid` response header value), `ProviderAddress`, `ProviderGroup`, `Method`, `Finality`, `ConsensusHash` + `OutlierHash` (full 32-byte hex; the log's short form is a prefix), `Outcome`, `DelayMs`, `Seq`/`RecordedAt`, and `MismatchCounted`.

`MismatchCounted` says whether **this** row is the one that moved `smartrouter_cross_validation_mismatch_total`. That makes the counter's once-per-distinct-group-per-request dedup visible instead of implicit — the exact rule two of the parked tests exist to pin.

**Every straggler resolution is recorded, not only dissent.** An `agreed` row is the positive control you asked for; `node-error` / `protocol-error` / `not-received` say a late answer arrived broken or never arrived.

Filters (optional, ANDed): `request_id`, `chain_id`, `outcome`, `limit`. Plus `POST /debug/cross-validation-events/clear` — the optional reset, included because it's the same one-liner `/debug/logs/clear` already offers.

```bash
curl -s "http://127.0.0.1:$DEBUG_PORT/debug/cross-validation-events?request_id=$LAVA_GUID" | jq
```

### Status codes

Following MAG-2707 rather than the flat-array endpoints' unconditional `200 []`:

- **200** — rows follow; `[]` means the recorder was **live** and saw no dissent.
- **503** — the recorder is not installed, so nothing was being recorded and an empty answer would mean nothing.

The ring is bounded (4096) and `X-Cross-Validation-Events-Dropped` reports evictions, so truncation cannot pass for absence.

## Cost when not in use

Recording is debug-mode only: the ring is installed next to `utils.EnableDebugLogBuffer` under the same `--debug-address` condition. A production router pays one atomic nil load per dissent and stores nothing.

## Two things worth a reviewer's eye

- **`CrossValidationStragglerResult` now carries `ResponseHash`.** Without it, straggler rows would have an empty outlier hash — one of the fields the ticket names. `classifyStragglerResponse` returns the hash it compared, zeroed for outcomes with no content, so "answered differently" and "had no answer" stay distinguishable.
- **The reply-time outlier loop is now one pass** (log + event + metric) so the event can name which outlier carried the group's single increment. The metric contract is unchanged — once per distinct outlier group — and emitting in outlier order rather than map-iteration order makes it deterministic. Its guard now requires only `listenEndpoint` and no longer requires a metrics manager: a router that cannot count the dissent can still record it.

## Known gap, deliberately left

A cross-validated request where **everyone agreed at reply time** and there were no stragglers records **no rows**. Reading `[]` for such a request therefore does not by itself prove the request ran — but the test already holds proof in its own response headers (`lava-cross-validation-status`, `-agreeing-providers`). Inventing a per-request anchor row was not asked for, so it is not here. Say the word if you'd rather have one.

Also out of scope, as the ticket says: reading the effective post-clamp cross-validation parameters, and provider-pin-header visibility. Neither fell out of this design.

## Testing

`go build ./...`, `go test ./protocol/rpcsmartrouter/ ./protocol/relaycore/` and `go test -race ./protocol/rpcsmartrouter/ -run CrossValidation` all pass locally.

New tests cover three layers:

- **the ring** — bounded and keeps the newest, filters, drop accounting, off by default, clear;
- **the HTTP contract** — 405, 503-when-off, 200-`[]`-when-live, the full row shape, filters, clear, the dropped header;
- **the production glue on both recording paths, against the real relay code** — reply-time outliers (including that two same-group outliers each get a row while the counter moves once, and that an all-agreeing request records nothing), and the straggler watcher (late dissent, late agreement as positive control, same-group dedup).

## For the automation side

The four parked tests in `tests/simulator/simulator_production_tests/cross_validation_enhancements/` can be repointed here and switched on. The correlation key is the `Lava-Guid` response header value, passed as `request_id`. Note that this is **not** `/debug/logs`' `request_id`, which is the caller's `X-Request-Id` — a different identifier.

### DIFF
--- a/docs/CROSS-VALIDATION.md
+++ b/docs/CROSS-VALIDATION.md
@@ -91,6 +91,11 @@ failure — the structured signal lets the client decide whether to retry (quoru
 fall back (structural reasons). Header names, the closed failure-reason enum, and the gRPC-trailer
 caveat are tabulated in the [in-package reference](../protocol/rpcsmartrouter/README.md#response-headers).
 
+A router started with `--debug-address` additionally serves `GET /debug/cross-validation-events`, a
+read-only per-request record of the dissent it observed (which provider, which group, which
+recording path) for test suites that cannot scrape the metrics port — see
+[Reading recorded dissent](../protocol/rpcsmartrouter/README.md#reading-recorded-dissent).
+
 ## Caveats
 
 - **Writes.** An *operator policy* on a stateful (write) method is rejected at startup.
--- a/protocol/relaycore/relay_processor.go
+++ b/protocol/relaycore/relay_processor.go
@@ -659,6 +659,11 @@ type CrossValidationStragglerResult struct {
 	ProviderGroup   string        // "" folds to common.DefaultProviderGroup at the metric layer
 	Outcome         string        // one of common.CrossValidationStragglerOutcome*
 	Delay           time.Duration // how long after the reply the straggler resolved (watch duration for not-received)
+	// ResponseHash is the late response's own content hash, the value compared against the consensus
+	// to produce Outcome. Zero when there was no content to hash (node-error, protocol-error,
+	// not-received), which the debug event surface renders as "no hash" rather than as a differing
+	// one (MAG-2772).
+	ResponseHash [32]byte
 }
 
 // WatchCrossValidationStragglers consumes responses that arrive on the processor's channel AFTER the
@@ -698,11 +703,13 @@ func (rp *RelayProcessor) WatchCrossValidationStragglers(ctx context.Context, pe
 			return
 		}
 		delete(pending, addr)
+		outcome, responseHash := classifyStragglerResponse(response, consensusHash, protocolMessage)
 		record(CrossValidationStragglerResult{
 			ProviderAddress: addr,
 			ProviderGroup:   response.RelayResult.ProviderInfo.ProviderGroup,
-			Outcome:         classifyStragglerResponse(response, consensusHash, protocolMessage),
+			Outcome:         outcome,
 			Delay:           time.Since(start),
+			ResponseHash:    responseHash,
 		})
 	}
 	// resolveFromRecorded resolves pending providers whose response never reaches this watcher's
@@ -718,7 +725,8 @@ func (rp *RelayProcessor) WatchCrossValidationStragglers(ctx context.Context, pe
 		}
 		// resolveOne records a watched provider from its stored result and drops it from pending;
 		// a no-op for providers not in the pending set (already resolved, or never watched).
-		resolveOne := func(addr, group, outcome string) {
+		// responseHash is the stored content hash, zero for the outcomes with nothing to hash.
+		resolveOne := func(addr, group, outcome string, responseHash [32]byte) {
 			if _, watched := pending[addr]; !watched {
 				return
 			}
@@ -728,6 +736,7 @@ func (rp *RelayProcessor) WatchCrossValidationStragglers(ctx context.Context, pe
 				ProviderGroup:   group,
 				Outcome:         outcome,
 				Delay:           time.Since(start),
+				ResponseHash:    responseHash,
 			})
 		}
 		successResults, nodeErrorResults, protocolErrorResults := rp.GetResultsData()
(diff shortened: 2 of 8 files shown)


avitenzer:
## Verified against a live router

Item left open at review time — the Go tests drive the real recording paths, but nothing had run them through a running router. Added `scripts/pre_setups/init_smartrouter_eth_cv_events.sh` (9e72899), which boots a router with `--debug-address`, proves the surface end to end, and leaves it up with a command cheat sheet.

**The reason this mattered:** whether a chart's `group-label` actually reaches `ProviderInfo.ProviderGroup` on a live relay is the one thing the Go tests cannot prove — they inject the group directly. That was the residual gap I named on MAG-2772 back when I was arguing to close it. It is now closed by observation:

```json
{
  "Seq": 2,
  "Source": "reply-time",
  "ChainID": "ETH1",
  "ApiInterface": "jsonrpc",
  "RequestID": "381654785364225774",
  "Method": "eth_getBalance",
  "ProviderAddress": "sim-3",
  "ProviderGroup": "external",
  "Outcome": "disagreed",
  "Finality": "finalized",
  "ConsensusHash": "fe37ce1a1d5567bbb6db1860d08693527a691221b6abead9243a91929e145c39",
  "OutlierHash":   "438ee3eb2e8bcc7ab5cba8067d09c335a3a8b34746d7a370359ac4c2a688f8ab",
  "MismatchCounted": true,
  "DelayMs": 0,
  "RecordedAt": "2026-08-17T08:29:30Z"
}
```

`"ProviderGroup": "external"` is the `group-label` from the generated config, carried through a real relay onto the row. `RequestID` is the `Lava-Guid` that request returned.

### What the harness covers

Upstreams are the provider_simulator rather than public RPC — the feature only produces rows when providers **disagree**, and a shared-truth fleet can never manufacture a dissent. Which recording path sees a dissent is otherwise a race (all three upstreams are local and answer in microseconds), so the harness uses the simulator's latency knob to choose it deterministically:

| Scenario | Setup | Expected |
| --- | --- | --- |
| 1. positive control | nobody dissents, one provider late | `agreed` straggler row, `MismatchCounted: false`, `OutlierHash == ConsensusHash` |
| 2. reply-time | dissenter answers first, honest pair held back | `reply-time` row, `MismatchCounted: true` |
| 3. straggler | dissenter delayed 1500ms past the early-exit | `straggler` row, `DelayMs: 1503`, `MismatchCounted: true` |

**21 checks, all passing.** Scenario 3's row, showing the async path carries the group label and a measured delay:

```json
{"Seq":3,"Source":"straggler","Outcome":"disagreed","ProviderAddress":"sim-3",
 "ProviderGroup":"external","MismatchCounted":true,"DelayMs":1502}
```

### The rows reconcile with the counter they replace

Two dissents recorded as events; two increments on the counter, correctly labelled:

```
smartrouter_cross_validation_mismatch_total{spec="ETH1",apiInterface="jsonrpc",method="eth_getBalance",group="external",finality="finalized"} 2
smartrouter_cross_validation_straggler_total{...,outcome="agreed"} 1
smartrouter_cross_validation_straggler_total{...,outcome="disagreed"} 1
```

That is the "counts are derivable from events" claim, checked rather than asserted.

### Also confirmed live

- `X-Cross-Validation-Events-Dropped: 0` on a ring that has not evicted.
- The debug listener answering **200** is what the harness waits on for readiness — that single check proves both that `--debug-address` took effect and that the recorder is installed, since an uninstalled recorder answers 503.
- `POST /debug/cross-validation-events/clear` isolates scenarios; each of the three starts from an empty ring.

### Two harness bugs found and fixed on the way (both mine, neither in the router)

- `awk BEGIN{IGNORECASE=1}` is a gawk extension; macOS ships BWK awk, where it silently sets an unused variable and the `Lava-Guid` match never fires.
- `jq '.MismatchCounted // ""'` reads `false` as empty — `//` treats `false` and `null` alike — so the positive control's `false` looked like a missing field.

### Usage

```bash
scripts/pre_setups/init_smartrouter_eth_cv_events.sh              # boot + smoke, leaves router up
SKIP_SMOKE=1 scripts/pre_setups/init_smartrouter_eth_cv_events.sh # just a debug-enabled router
```

Ports are distinct from `init_smartrouter_eth.sh` (3360/7779) and `test_uc4_smartrouter_sim.sh` (3392/7798), so it can run alongside them.

Tomelia1999 (approved):

PR #285, opened 2026-08-17
fix(lavasession): match lava-select-provider case-insensitively

nimrod-teich:
#Closes MAG-2816

Jira ticket: MAG-2816

## Summary

Provider names are free-form config strings, and the deployment pipeline does not preserve their case. The helm chart folds each node's name into the router's own config:

```yaml
# smart-router-helm-chart · charts/smart-router/templates/routers_configmap.yaml:52
name: {{ $node.name | lower | replace " " "-" }}
```

…while the values file (and the dashboard reading it) spells the name for people. So `lava-select-provider: Lava` against a registered `lava` names the right provider — and the exact match rejected it outright, since header selection deliberately does not fall back to another provider.

Observed on `gk8-prod`:

```
$ curl -X POST https://iota-mainnet-jsonrpc.gk8.magmadevs.com \
    -H 'lava-select-provider: Lava' -d '{…"method":"iota_getChainIdentifier"…}'
{"error":{"code":-32000,"message":"… {selectedProvider:Lava,validProviders:lava,…}"}}

# same request, lowercase header → {"jsonrpc":"2.0","id":1,"result":"6364aad5"}
```

The upstream was healthy throughout. Case-sensitivity bought nothing here and cost a rejected relay.

## The change

Both lookups in `getValidProviderAddresses` now fold case:

- **The valid-address match.** Resolution returns the **canonical** address rather than the header's spelling, so everything downstream — the pairing map, the session store, the metrics labels — keeps keying on the router's own name instead of whatever the caller typed. An exact hit still wins outright, so a config that registers two names differing only in case resolves the way the caller spelled it.
- **The in-request ignored set.** It is keyed by canonical address, so a differently-cased header would otherwise miss the very provider the check exists to bound — reinstating the unbounded re-selection spin that guard was added for.

Folding case does **not** soften the contract: a name matching no provider in any case is still an error, not a silent fallback to someone else.

## Test plan

- `go build ./...` clean
- `go test ./protocol/lavasession/...` — pass
- `go test ./protocol/rpcsmartrouter/...` — pass
- New: case-insensitive resolution returns the canonical address (upper / lower / mixed), the ignored-set check folds case, an unknown name is still rejected, exact-spelling-wins on a case-colliding config, plus unit tables for both helpers

**Pre-existing flake, unrelated:** `TestEndpointHealthRecovery` fails intermittently — reproduced on a clean `main` worktree at 1 failure in 8 runs with none of this branch's code applied. Not touched here.

## Relationship to #286

**Independent — either can merge first.** #286 splits the shared sentinel so "not available" and "already failed" stop reporting under one string; this one fixes the case-sensitive match that produced the example above. They touch adjacent lines in `getValidProviderAddresses`, so whichever lands second needs a trivial conflict resolution.

The dashboard currently works around this by pre-folding node names ([smart-router-dashboard#123](https://github.com/Magma-Devs/smart-router-dashboard/pull/123)); that shim can go once this ships.

### DIFF
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -1278,37 +1278,40 @@ func convertSelectionStatsToMetrics(stats *provideroptimizer.SelectionStats) (al
 	return allScores, rngValue
 }
 
-// Provider names are free-form config strings and the pipeline does not preserve
-// their case — the helm chart folds node names into the router's config, the
-// values file spells them for humans. Both lookups below fold case.
-
-// resolveSelectedProviderAddress maps a header-supplied name onto the address the
-// router registered, so downstream lookups key on the router's spelling.
-func resolveSelectedProviderAddress(selectedProvider string, addresses []string) (string, bool) {
-	// Exact hit wins: two config names differing only in case resolve as spelled.
+// resolveSelectedProviderAddress maps a header-supplied provider name onto the address
+// the router registered. Provider names are free-form config strings and the pipeline
+// does not preserve their case — the helm chart folds node names into the router's
+// config, the values file spells them for humans — so the match folds case.
+//
+// It returns the canonical address rather than the header's spelling because everything
+// past this point keys on the router's own name: csm.pairing, the session store, the
+// in-request ignored set, the metrics labels. csm.pairing in particular is a
+// LavaFormatFatal on a miss, so handing the caller's spelling downstream would turn a
+// mis-cased header from a rejected relay into a dead router.
+//
+// An exact hit wins outright, so a config registering two names that differ only in case
+// still resolves the way the caller spelled it. Absent one, a fold matching more than one
+// address resolves to nothing and returns those candidates instead: validAddresses is
+// built by ranging a map and is reordered as providers are blocked and recovered, so
+// picking one would serve the same pinned name from a different upstream run to run —
+// the opposite of what pinning is for. The caller reports the collision so the operator
+// can rename a provider.
+func resolveSelectedProviderAddress(selectedProvider string, addresses []string) (address string, ambiguous []string) {
 	if slices.Contains(addresses, selectedProvider) {
-		return selectedProvider, true
+		return selectedProvider, nil
 	}
-	for _, address := range addresses {
-		if strings.EqualFold(address, selectedProvider) {
-			return address, true
+	var folded []string
+	for _, candidate := range addresses {
+		if strings.EqualFold(candidate, selectedProvider) {
+			folded = append(folded, candidate)
 		}
 	}
-	return "", false
-}
-
-// providerInSetFold reports whether a provider name is in a set keyed by
-// canonical addresses. Exact hit is the fast path; the scan handles a case diff.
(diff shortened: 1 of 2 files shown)


avitenzer (approved):
Approving. The core fix is right and lands in the one place it needs to — I traced `lava-select-provider` end to end and `resolvePinDirectives` → `GetSessions` → `getValidProviderAddresses` is its only consumer, so there's no second site still matching case-sensitively.

Worth recording that returning the **canonical** address rather than the header's spelling is load-bearing rather than just tidy: `consumer_session_manager.go:1736` does `csm.pairing[providerAddress]` and hits `LavaFormatFatal` on a miss, so handing the caller's spelling downstream would have turned a mis-cased header from a rejected relay into a dead router. That's a stronger justification than the metrics-labels argument in the description.

I pushed one follow-up commit (83f66ba) for two things the original missed, both of which only bite on a config registering two names that differ only in case — legal config, since `ValidateUniqueProviderNames` (MAG-2724) groups by exact name, and reachable from a hand-written `routers.yaml` though not via the chart, which lowercases:

1. **`providerInSetFold` could reject a healthy provider.** With `Lava` and `lava` both registered, header `Lava`, ignored set `{lava}`: exact miss, fold hit, relay rejected as "already failed" — but `Lava` is a different upstream that never failed. Resolution honoured the exact spelling and the ignored-set check didn't, so the two disagreed on identity. Fixed by resolving to the canonical address first and then looking the ignored set up by that address exactly; the set is only ever written with addresses that came out of the pairing, so folding bought nothing there. `providerInSetFold` is deleted — net −12 lines.

2. **`TestResolveSelectedProviderAddressPrefersTheExactSpelling` asserted an ordering guarantee that doesn't exist.** The no-exact-hit case pinned `"LAVA"` → `"Lava"` as "the first registered spelling". There is no first: `validAddresses` is built by ranging `csm.pairingAddresses`, a `map[uint64]string` (`:863`, from `UpdateAllProviders` at `:407`), and is further reordered as providers are blocked (`:1873`) and recovered (`:2134`). The test passed only because it was handed a literal slice. An ambiguous fold now resolves to nothing and reports the colliding names, sorted — same call MAG-2724 makes, and silently serving a pin from an arbitrary one of two upstreams is exactly what the no-fallback contract exists to prevent.

One behaviour change to be aware of: resolution now precedes the ignored-set check, so a pinned provider that is both ignored and no longer in `validAddresses` reports "not available" instead of "already failed". Same `SelectedProviderUnavailableError` sentinel so `errors.Is` callers are unaffected, and both paths still return an error so the spin bound holds — but worth a glance when #286 splits that sentinel. Both existing message strings are preserved verbatim; `chainlib/common_test.go:150,179` asserts on one of them.

Still open, not blocking: in `TestSelectedProviderMatchesCaseInsensitively` the `strings.ToLower(pinned)` case is a no-op against `provider0`, so that table reads as three fold cases when it's two. The description also needs a small edit — the ignored-set bullet describes folding behaviour that's now gone.

`go build ./...` and `go vet` clean; `protocol/lavasession`, `protocol/chainlib` and `protocol/rpcsmartrouter` all pass locally, and CI is green on 83f66ba.

PR #287, opened 2026-08-17
fix(smart-router): clone chain parser in background provider recovery

avitenzer:
# Description

Background provider recovery breaks live gRPC relays until the pod restarts.

The 3-minute retry loop passed the **live** `chainParser` into `GetChainRouter` and then cancelled the scoped verification context. For gRPC, `NewGrpcChainProxy` → `setupForProvider` rebinds the parser's `registry`/`codec` to the verification connection, so cancelling it left the live parser pointing at a dead connection — every subsequent gRPC relay on that provider failed with a nil connector until the pod was restarted. Because a timer drives it, the outage recurred every 3 minutes.

**The fix already existed.** `CloneChainParserForValidation` (`protocol/chainlib/chainlib.go:103`) is written, documented and unit-tested, and was already applied in the spec re-verification path (`spec_reverifier.go:333`). `retryFailedStaticProviders` held a hand-copied duplicate of the same verification sequence and never got it.

So rather than clone in both copies, this deletes the duplicate: the retry loop now calls `validateProvider`, the helper the re-verifier already uses. One code path, so this can only ever be wrong in one place. The two blocks were already functionally identical — the apparent difference between `MaximumStreamsOverASingleConnection` and `DefaultMaximumStreamsOverASingleConnection` is a `uint64` cast of the same value (`consumer_types.go:83-90`).

- `protocol/rpcsmartrouter/rpcsmartrouter.go` (+12 / −48) — the dedupe
- `protocol/chainlib/grpc_parser_validation_isolation_test.go` (new) — the regression pin

## Testing

The new test stands up a gRPC server with reflection on a loopback port and pins both halves of the invariant:

1. handing over the **live** parser rebinds its registry — the hazard is real
2. handing over the **clone** leaves the live parser untouched

**Confirmed to fail without the fix.** Reverting the clone produces:

```
Expected nil, but got: &dyncodec.Registry{remote:(*dyncodec.GRPCReflectionProtoFileRegistry)...}
MAG-2538: verification must not rebind the live parser's registry
```

Worth recording for whoever extends this: a first attempt at the test pointed at an unreachable address and passed on both fixed and broken code. `NewGRPCConnector` blocks on the first connection and errors out before `newGrpcChainProxy` runs, so `setupForProvider` is never reached and there is nothing to mutate. A live reflection server is required for the test to mean anything.

| Check | Result |
| --- | --- |
| `go build ./...` | pass |
| `go vet ./protocol/rpcsmartrouter/` | pass |
| `go test ./protocol/chainlib/` | pass |
| `go test ./protocol/rpcsmartrouter/` | pass |
| `go test -race` (retry goroutine tests) | pass |

## Behaviour change to note in review

The retry loop previously logged two distinct warnings — `retry: static provider chain router creation still failing` and `retry: static provider verification still failing`. It now logs only the second, since `validateProvider` returns a single error. The two cases stay distinguishable from the `error` field, but anyone grepping on the first string will need to adjust.

## Jira ticket

Jira ticket: MAG-2538

---

## Author Checklist

I have...

* [x] provided a link to the relevant issue or specification — MAG-2538 above
* [x] updated the relevant documentation or specification, including comments for documenting Go code — the new call site carries a comment explaining why the clone is load-bearing
* [ ] confirmed all CI checks have passed — pending first CI run

### DIFF
--- a/protocol/chainlib/grpc_parser_validation_isolation_test.go
+++ b/protocol/chainlib/grpc_parser_validation_isolation_test.go
@@ -1,0 +1,64 @@
+package chainlib
+
+import (
+	"context"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/magma-Devs/smart-router/protocol/common"
+	"github.com/magma-Devs/smart-router/protocol/lavasession"
+	spectypes "github.com/magma-Devs/smart-router/types/spec"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+// startReflectionGRPCServer brings up a bare gRPC server that serves server
+// reflection on a random loopback port. Reflection is the only thing this test
+// needs: setupForProvider builds its registry from the reflection connection.
+func startReflectionGRPCServer(t *testing.T) string {
+	t.Helper()
+	srv := grpc.NewServer()
+	reflection.Register(srv)
+	lis, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+	go func() { _ = srv.Serve(lis) }()
+	t.Cleanup(srv.Stop)
+	return lis.Addr().String()
+}
+
+func TestGetChainRouter_CloneProtectsLiveGrpcParser(t *testing.T) {
+	addr := startReflectionGRPCServer(t)
+	endpoint := &lavasession.RPCProviderEndpoint{
+		ChainID:      "TEST",
+		ApiInterface: "grpc",
+		NodeUrls:     []common.NodeUrl{{Url: addr}},
+	}
+
+	// Part 1 — the hazard is real. Passing the LIVE parser mutates its registry.
+	direct, err := NewGrpcChainParser()
+	require.NoError(t, err)
+	direct.SetSpec(spectypes.Spec{})
+	require.Nil(t, direct.registry, "fresh parser starts with no registry")
+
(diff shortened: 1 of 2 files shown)


avitenzer:
Closing — **superseded by MAG-2525**, which landed on main while this was open.

`retryFailedStaticProviders` no longer exists. Commits `b16fd25` / `6c6a0f5` renamed it to `retryFailedProviders` and routed its verification through:

```go
var retryValidateFn = func(ctx context.Context, provider *lavasession.RPCStaticProviderEndpoint, chainParser chainlib.ChainParser) error {
	return validateProvider(ctx, provider, chainParser, BootValidateTimeout)
}
```

`validateProvider` clones via `CloneChainParserForValidation` — which is exactly what this PR did: collapse the hand-copied verification block into the shared helper rather than adding a second clone call. There is no longer any `GetChainRouter` call in `rpcsmartrouter.go` at all.

The merge conflict reported on this PR *is* that convergence — both changes rewrote the same block the same way, independently.

MAG-2538 is closed as fixed on main.

**Deliberately not carried forward:** `protocol/chainlib/grpc_parser_validation_isolation_test.go`. It stood up a gRPC server with reflection and triggered the *real* parser mutation through `GetChainRouter`, rather than simulating it by assigning `registry`/`codec` directly. Main keeps only `TestCloneChainParserForValidation_GrpcIsolation`, which would still pass if the helper stopped isolating — so that invariant, which MAG-2525's refactor now depends on, is pinned only by inspection. The branch `fix/mag-2538-clone-parser-on-provider-recovery` is left in place if anyone wants to revive it.

PR #288, opened 2026-08-17
fix(smart-router): classify gRPC status errors and cancellations correctly

avitenzer:
# Description

**Two tickets, one PR: MAG-2549 and MAG-2687 are the same three lines of `sendGRPCRelay`.** The Jira gate accepts one key per PR, so MAG-2549 holds the gate line below; MAG-2687 is closed by this change too.

### MAG-2549 — a gRPC status error with a response was served as a success

The error branch left `StatusCode` at its zero default, and `GrpcMessage.CheckResponseError` reports "no error" whenever `StatusCode == 0`, so `results_manager.setValidResponse` routed it into the **success** results. Any upstream error — `INVALID_ARGUMENT` on a malformed transaction, `NOT_FOUND` on an object — reached the client as a success carrying the error payload, and was cacheable.

The same branch derived `IsNodeError` from `grpcErr.Code >= 13`, so every code below `INTERNAL` — `INVALID_ARGUMENT`, `DEADLINE_EXCEEDED`, `NOT_FOUND`, `RESOURCE_EXHAUSTED` — was marked not-a-node-error, which is what gates the cache write and the `lava-identified-node-error` header.

### MAG-2687 — cancelled relays recorded as successes with fabricated latency

A relay the router itself cancelled was recorded as a completed, successful relay carrying a latency sample that was never taken — feeding the optimizer and calling `ResetHealth()` on endpoints that were never measured. Two causes, both fixed:

1. grpc-go returns `status.Error(codes.Canceled, ...)`, which does **not** wrap `context.Canceled`, so the relay-race carve-out could never see it.
2. `handleGRPCError` rebuilt the error from a bare code and message, discarding the cause, so nothing downstream could recover the sentinel either.

> A third contributing cause was originally listed here — a substring collision in the classifier guard. On a closer look it turned out to be already covered by cause 1, and the classifier change made for it has been reverted. See "Review follow-up (round 2)" below.

## The four changes

| # | File | Change |
| --- | --- | --- |
| 1 | `lavasession/direct_rpc_connection.go` | `handleGRPCError` returns `(nil, …%w context.Canceled)` on a router-cancelled relay. The **nil response** is load-bearing: it makes the phantom-success arm unreachable, and it resolves local cancellation *structurally*, in `DetectConnectionError`'s layer-1 `errors.Is` check. |
| 2 | `lavasession/direct_rpc_connection.go` | `GRPCStatusError` gains `cause` + `Unwrap()`, so it stops being a dead end for `errors.Is` / `errors.As`. |
| 3 | `common/error_classifier.go` | `codes.InvalidArgument` (3) gets a registry row → `USER_INVALID_PARAMS`. The remaining gRPC codes are documented as deliberately unregistered, each with its own reason. **The canceled/deadline guard stays at the broad `"rpc error"` — see round 2.** |
| 4 | `rpcsmartrouter/direct_rpc_relay.go` | The error branch now mirrors its own success sibling: `StatusCode`, `CheckResponseError`, and `ApplyNodeErrorClassification` — the whole flag set, not `IsNonRetryable` alone. |

**Change 4 came out smaller than MAG-2549 described.** The success path 40 lines below already classified correctly, so the error branch simply adopts it. That also retires the `IsNodeError: Code >= 13` threshold — the open scope question on MAG-2687 — without inventing a new rule. Nothing is lost: `handleGRPCError` builds `response.StatusCode` and `GRPCStatusError.Code` from one `errorCode`, so they are equal by construction.

## Two findings reviewers should see

**MAG-2687's step 0 is a non-issue, and this PR pins why.** That step warns that adding `Unwrap` makes `status.Code(err)` resolve via `errors.As`, and that `IsSessionSyncLoss` reads it. It does not: `IsSessionSyncLoss` lives in `lavasession/common.go`, which imports **`github.com/gogo/status`**, not `google.golang.org/grpc/status` — and gogo's `Code()` does not unwrap. Measured directly: `status.Code(wrapped)` is 677 under grpc-go while `IsSessionSyncLoss(wrapped)` is `false`. The test now pins this as a tripwire for anyone swapping that import.

**There is a real side effect in the other direction.** `ExtractNodeErrorDetails` (`node_error_handler.go:95`) *does* use grpc-go's `status.FromError`, and the direct-RPC path reaches it via `error_mapper.go:37`. For a `*GRPCStatusError` it previously returned code `0`; it now returns the real code (`3`). That is the improvement — Tier-1 `CodeEquals` matchers can finally fire — and it is a behaviour change reviewers should see.

> **Corrected during review.** An earlier version of this description also claimed the *message* got cleaner (`"malformed transaction"` instead of `"gRPC error 3: malformed transaction"`). It does not. grpc-go's `status.FromError` overwrites the status message with the **outer** error's `Error()` when it resolves via `errors.As`, and `ExtractNodeErrorDetails` already defaulted `errorMessage` to `nodeError.Error()` — so the message is identical before and after. The code is the whole of the change. `TestGRPCStatusError_ResolvesThroughStatusFromError` pins both halves, including the message *not* changing, and is what caught the wrong claim.

## Testing

New tests start from **real gRPC error shapes**, not `*url.Error`. Every test added for MAG-2648 built a `*url.Error` — including one written to prove the fix was transport-general — which is exactly why the identical bug on the gRPC transport stayed invisible.

Every fix is verified **red-first** by reverting it in isolation and re-running:

| Reverted | Test that goes red | Observed |
| --- | --- | --- |
| ctx-cancel early return | `TestHandleGRPCError_CancelledContextReturnsNilResponse` | FAIL → PASS |
| `cause` preservation | `TestHandleGRPCError_PreservesCause` | FAIL → PASS |
| broad `"rpc error"` guard | `TestHandleGRPCError_RemoteCancellationIsNotLocal` | FAIL (classified 1008 / 1007) → PASS |
| broad `"rpc error"` guard | `TestGRPCStatusError_IsNeverReadAsLocalCancellation`, `TestRemoteStatusGuard_AppliesAcrossTransports` | FAIL (both transports) → PASS |
| gRPC code-3 row | `TestSendGRPCRelay_StatusErrorClassification/InvalidArgument` | FAIL → PASS |
| `ApplyNodeErrorClassification` → `IsNonRetryable` alone | `TestSendGRPCRelay_StatusErrorClassification/{ResourceExhausted,Unimplemented}` | FAIL → PASS |
| node-message fallback | `TestSendGRPCRelay_ClassificationMessageIsTheNodesOwn` | FAIL → PASS |
| `StatusCode` on `RelayResult` | `TestResultsManager_GRPCStatusResponseIsANodeError` | pinned by its own `REGRESSION` subtest |

`TestGRPCRemoteStatusError_StillNotLocalCancellation` and the `Unavailable` row of `TestSendGRPCRelay_StatusErrorClassification` are the collateral guards, and the most important tests here: carving out caller-fault codes must not go so far that a genuinely **remote** deadline, cancel, or `UNAVAILABLE` stops being blamed on the endpoint.

Full suites green: `common`, `lavasession`, `chainlib`, `rpcsmartrouter`, `relaycore`, `metrics`. Plus `go build ./...`, `gofmt -d`, `git diff --check`, and `go vet` on all four touched packages.

## Review follow-up (commit `40c6912`)

1. **`ctx.Err() != nil` → `errors.Is(ctx.Err(), context.Canceled)`** in `handleGRPCError`. The old guard also matched an *expired deadline*, so a timeout could have been relabelled a client cancellation — letting a slow endpoint escape blame.
2. **Bare type assertion → `errors.As`** in `sendGRPCRelay`. This PR adds `Unwrap` precisely to make the type inspectable through wrapping; its main consumer should use it.
3. **The classifier edit is transport-global and is pinned as such.** `stringConnectionFallbacks` is one package-level table and `detectConnectionErrorFromString` takes no transport argument, so the width of that guard changes classification for JSON-RPC and REST too. `TestRemoteStatusGuard_AppliesAcrossTransports` covers it. *(Partly superseded by round 2: the guard is no longer narrowed, but the reach argument stands and the test now pins the opposite direction.)*
4. **The `ExtractNodeErrorDetails` claim above was wrong and is corrected inline.**

## Review follow-up (round 2 — commit `2669397`)

Rebased onto current `main`; the branch is no longer conflicting.

**1. The classifier guard is restored to its original width.**

This PR had widened the canceled/deadline rows from `mustNotContain: "rpc error"` to `"rpc error: code ="` so that `GRPCStatusError`'s rendering — `"gRPC error 1: context canceled"`, which lowercased contains `"rpc error"` inside the word `"gRPC"` — would match them. Revisiting that: the wrapper is only built *after* `handleGRPCError`'s local-cancellation branch has returned, so in practice a value of that shape carries a status the **endpoint** reported. Since `PROTOCOL_CONTEXT_CANCELED` is `Retryable=false`, matching it on those rows would read an upstream fault as local orchestration and skip a retry that could have reached a healthy endpoint. Re-running with the narrower guard shows this directly: a remote `codes.Canceled` classifies as `1008 PROTOCOL_CONTEXT_CANCELED`, a remote `DeadlineExceeded` as `1007`.

The narrowing also turns out not to be needed. Change 1 above already handles local cancellation structurally: `DetectConnectionError` catches the real `context.Canceled` sentinel in its layer-1 `errors.Is` check, well before the string table is consulted. That table only sees errors whose sentinel chain was already lost, and for those the broader "contains an rpc-error marker" is the conservative reading. The guard is restored with the reasoning recorded inline, so the trade-off is visible to whoever looks at it next.

The two tests written against the narrower guard are replaced by their inverses — a cause-less `GRPCStatusError` should **not** read as a local cancellation, and a bare `"rpc error"` substring should **still** exclude on every transport.

**2. The rebase conflict reconciled two independent fixes to the same arm.** Each side had addressed a different half, so neither "ours" nor "theirs" alone was complete:

| | this branch | `main` |
| --- | --- | --- |
| what counts as a node error | `CheckResponseError`, every non-OK status | `grpcErr.Code >= 13` |
| what happens once it is one | `IsNonRetryable` alone | `ApplyNodeErrorClassification`, the full set |

The resolution takes `CheckResponseError` from this branch and `ApplyNodeErrorClassification` from `main`, so `IsRateLimited` and `IsUnsupportedMethod` survive.

**3. One detail the merged version depends on.** `GrpcMessage.CheckResponseError` decides `hasError` from the status code alone and always returns an **empty message**. Classifying on `""` would reduce the gRPC registry to its code matchers and leave every message-based row unreachable, so a `RESOURCE_EXHAUSTED` whose message says `"rate limit"` would be scored as an availability failure rather than exempted as healthy-but-busy. The classification message now falls back to the node's own status message, then to the wrapped rendering.

**4. `codes.InvalidArgument` gets a registry row.** gRPC defines it as arguments "problematic regardless of the state of the system", so every endpoint rejects the same malformed request identically. Left unregistered it classified `UNKNOWN_ERROR`, and a burst of malformed client requests demoted the whole healthy pairing toward the selection floor — blocklisting on the 16th consecutive one. `USER_INVALID_PARAMS` is `Retryable=false`, which both stops the pointless retry and keeps the availability gate off the endpoint.

The other codes are deliberately left unregistered rather than added as a range — registering by range is what the `Code >= 13` threshold did, and the aim here is to avoid repeating it. Each code's reasoning is recorded at the table: `4 DeadlineExceeded` (this endpoint was slow, demote it), `5 NotFound` / `11 OutOfRange` (pruned vs archive nodes genuinely disagree, so a retry elsewhere can succeed), `7 PermissionDenied` / `16 Unauthenticated` (credentials are per-endpoint, so demote), `8 ResourceExhausted` (ambiguous in the spec — quota vs out-of-disk; the `"rate limit"` message row catches the quota case), `1 Canceled` (a local cancel never reaches this table).

**5. New call-path tests drive the real `sendGRPCRelay`, not the registry.** That distinction matters here: most of what these changes fix is a call site not consuming a classification, so a test that calls `ClassifyNodeErrorForRetry` directly can pass while the relay path is still wrong.

| Case | Asserted |
| --- | --- |
| `InvalidArgument` | `StatusCode=3`, `IsNodeError`, `IsNonRetryable`, **not** scored |
| `ResourceExhausted` + "rate limit" | `IsRateLimited`, **not** scored |
| `ResourceExhausted`, no rate-limit wording | unclassified, **scored** — proves the flag came from the message, not the code |
| `Unimplemented` | `IsUnsupportedMethod`, non-retryable, **not** scored |
| `Unavailable` | retryable and **scored** — the collateral guard |

Plus `TestResultsManager_GRPCStatusResponseIsANodeError`, which drives the real `ResultsManager.SetResponse` and carries a `REGRESSION` subtest reproducing the pre-fix `StatusCode: 0` shape landing in `successResults`.

## Review follow-up (round 3 — commit `afaa30f`)

Rebased onto current `main` (`ec7c0a1`); the three commits it picked up touch only `direct_grpc_subscription_manager.go` and are disjoint from this change. Addresses all four findings in the review above; the point-by-point reply is in the comment thread.

**1. gRPC `NOT_FOUND` (5) and `OUT_OF_RANGE` (11) no longer demote the endpoint — and this supersedes two claims made above.**

Replacing `Code >= 13` with `CheckResponseError` does not only fix codes ≥ 13; it makes every non-OK code a node error, which newly exposed the most common codes on Cosmos and Sui gRPC to the availability gate. `NOT_FOUND` on a missing object scored one availability failure per endpoint asked, per retry, for a query whose answer is simply "no" — and `relayInnerDirect` called `ResetHealth()` on the same relay, because no gRPC code satisfies its `statusCode >= 500 || == 429` gate. The endpoint was demoted and certified healthy at once.

Resolved by adding a third fault axis rather than by registering 5 and 11 `Retryable=false`: non-retryable would keep them out of the availability signal, but only by ending the request at the first endpoint that answers — which is the pruned-node-to-archive-node retry.

* `SubCategoryDataScope` — "the endpoint answered truthfully that it does not hold this data".
* `NODE_DATA_NOT_HELD` (2017), `Retryable: true`, carrying that subcategory; gRPC 5 and 11 registered to it.
* `RelayResult.IsDataScope` plumbed through `ApplyNodeErrorClassification` and `relayInnerDirect`; `shouldFailSessionForResult` reads it as a third carve-out.
* `relayInnerDirect`'s `ResetHealth` is now gated on `shouldResetEndpointHealth`, defined as the exact negation of the scoring gate so the two cannot diverge.

Scoped to the gRPC rows on purpose: `NODE_RESOURCE_NOT_FOUND` (2012) and the `CHAIN_*_NOT_FOUND` codes stay on the scoring side, since only the gRPC status codes were shown to be a routine query outcome rather than an error the node raised.

> **Corrected.** Change 3's line above — "the remaining gRPC codes are documented as deliberately unregistered" — and round 2's item 4 entry for `5 NotFound` / `11 OutOfRange` are both superseded here. The pruned-vs-archive reasoning recorded there was right about retryability and wrong to conclude the codes therefore needed no row: it left them scoring against the endpoint. Codes 4, 7, 16 and message-less 8 are unchanged and still score.

**2. The cancellation guard drops its `status.Code(err) == codes.Canceled` conjunct.** grpc-go does not promise `Canceled` when the router cancels — a connection torn down in the same instant surfaces `Unavailable` — and on any other code the relay took the node-error arm, returned `err == nil`, and the `err != nil` guard at the relay-race carve-out never fired. Same bug class as MAG-2687, narrower window. The local context is authoritative on its own; the `context.Canceled` distinction from an expired deadline stays, with a collateral test now that the branch is wider. The originating status is wrapped alongside the sentinel (`%w` on both).

**3. `results_manager.setValidResponse` gets the empty-message fallback `sendGRPCRelay` already had.** `GrpcMessage.CheckResponseError` always returns `""`, so classifying on it left every message-based registry row unreachable one layer up — a `RESOURCE_EXHAUSTED` saying "rate limit" was filed as `UNKNOWN_ERROR` — and logged the node error with an empty error string. It now falls back to the node's own reply body, then to a synthetic description.

**4. Tests now cover the codes the change actually moved.** 4, 5 and 11 were not node errors at all under `Code >= 13`, so they were the untested half.

| Case | Asserted |
| --- | --- |
| `NotFound` / `OutOfRange` | `IsDataScope`, **not** scored, **still retryable** |
| `DeadlineExceeded` | scored — the counterweight, so the carve-out does not drag the sub-13 range out with it |
| health reset | agrees with the scoring gate in both directions, and a plain success still resets |

Plus `TestDataScopeCodesAreAllRetryable_MAG2549`, which fails the build if a data-scope code is ever registered non-retryable — at that point the axis has stopped buying anything and the archive fallback is gone.

**Smaller notes from the review**, all applied: the cancel branch keeps its cause; the "classify the way the success path does" comment now states outright that the success path's `hasError` branch is unreachable on this transport; and the code-3 row's cross-path reach into `GrpcErrorHandler.HandleNodeError` is recorded at the row, along with the same reach for the two new rows.

Every fix verified red-first by reverting it in isolation. `docs/ERROR-REGISTRY-DESIGN.md` updated for 2017 and the new subcategory.

## Jira ticket

Jira ticket: MAG-2549

---

## Author Checklist

I have...

* [x] provided a link to the relevant issue or specification — MAG-2549 and MAG-2687
* [x] confirmed all CI checks have passed

### DIFF
--- a/docs/ERROR-REGISTRY-DESIGN.md
+++ b/docs/ERROR-REGISTRY-DESIGN.md
@@ -70,6 +70,8 @@ Errors returned by the blockchain node itself (not execution/state errors).
 | 2013 | `NODE_RESOURCE_UNAVAILABLE` | Resource exists but unavailable | Yes | JSON-RPC -32002 |
 | 2014 | `NODE_GATEWAY_TIMEOUT` | Gateway timeout (HTTP 504 from provider) | Yes | HTTP 504 |
 | 2015 | `NODE_BAD_GATEWAY` | Bad gateway (HTTP 502 from provider) | Yes | HTTP 502 |
+| 2016 | `NODE_UNAUTHORIZED` | Upstream rejected router credentials (HTTP 401) | No | HTTP 401 |
+| 2017 | `NODE_DATA_NOT_HELD` | Endpoint does not hold the requested data — pruned or never existed (SubCategoryDataScope) | Yes | gRPC 5, gRPC 11 |
 | **Bitcoin/UTXO Node Errors (2100-2149)** |||||
 | 2101 | `NODE_BITCOIN_WARMUP` | Node still warming up (Bitcoin -28 / RPC_IN_WARMUP) | Yes | — |
 | 2102 | `NODE_BITCOIN_INITIAL_DOWNLOAD` | Node in initial block download (Bitcoin -10 / RPC_CLIENT_IN_INITIAL_DOWNLOAD) | Yes | — |
@@ -411,10 +413,12 @@ const (
     SubCategoryNone              ErrorSubCategory = iota
     SubCategoryUnsupportedMethod                         // zero retries, zero CU, cached response, no provider scoring
     SubCategoryRateLimit                                 // endpoint is healthy but busy; apply backoff, do not mark unhealthy
+    SubCategoryDataScope                                 // endpoint does not hold the data; keep retrying elsewhere, but do not score it
 )
 
 func (sc ErrorSubCategory) IsUnsupportedMethod() bool { return sc == SubCategoryUnsupportedMethod }
 func (sc ErrorSubCategory) IsRateLimit() bool         { return sc == SubCategoryRateLimit }
+func (sc ErrorSubCategory) IsDataScope() bool         { return sc == SubCategoryDataScope }
 
 
 // LavaError is the central error definition
@@ -594,7 +598,7 @@ _Blocked on protocol upgrade: sdkerrors carry ABCI codes used in the gRPC wire f
 
 9. **Transparent hop: original errors pass through unchanged.** The router/consumer is a transparent hop — the user always receives the original error from the node, unmodified. `LavaError` classification is metadata for internal use only (logging, metrics, endpoint health). Unknown/unmatched errors default to `CategoryExternal` because they are node pass-throughs. _(Clarification added in Phase 7: `handleAndClassify` wraps classified errors in `LavaWrappedError` on the **Go error return path** — this is internal plumbing for retry/health decisions and never reaches the user. The actual node response body travels separately and is always returned unmodified to the user. The "transparent hop" principle applies to the response body, not the internal Go error return.)_
 
-10. **Retryable is the primary retry signal, not SubCategory.** The consumer and smart-router retry state machines short-circuit on `LavaError.Retryable=false` via `RelayResult.IsNonRetryable`, populated at classification time on both paths (consumer: `rpcconsumer_server.go`; smart-router: `direct_rpc_relay.go` / `rpcsmartrouter_server.go`). This covers every terminal classification — `CHAIN_EXECUTION_REVERTED`, `CHAIN_OUT_OF_GAS`, `CHAIN_DOUBLE_SPEND`, `CHAIN_INVALID_SIGNATURE`, all of 3000-range — not only unsupported methods. SubCategory continues to govern adjacent policy: `SubCategoryUnsupportedMethod` triggers the zero-CU carve-out and caching, and `SubCategoryRateLimit` drives backoff without marking the endpoint unhealthy. Layer D user-input errors are non-retryable but charge normal CU — the provider does real work because responses are not cached. `relay_processor.HasNonRetryableUserFacingErrors` keys off the `IsNonRetryable` flag rather than re-classifying, so adding a new non-retryable error type requires no state-machine changes.
+10. **Retryable is the primary retry signal, not SubCategory.** The consumer and smart-router retry state machines short-circuit on `LavaError.Retryable=false` via `RelayResult.IsNonRetryable`, populated at classification time on both paths (consumer: `rpcconsumer_server.go`; smart-router: `direct_rpc_relay.go` / `rpcsmartrouter_server.go`). This covers every terminal classification — `CHAIN_EXECUTION_REVERTED`, `CHAIN_OUT_OF_GAS`, `CHAIN_DOUBLE_SPEND`, `CHAIN_INVALID_SIGNATURE`, all of 3000-range — not only unsupported methods. SubCategory continues to govern adjacent policy: `SubCategoryUnsupportedMethod` triggers the zero-CU carve-out and caching, and `SubCategoryRateLimit` drives backoff without marking the endpoint unhealthy. `SubCategoryDataScope` is the third such axis and the one Retryable cannot express: the errors are retryable (a pruned endpoint's miss is an archive endpoint's hit) yet must stay out of the availability signal, because the endpoint answered truthfully about its own data scope. `rpcsmartrouter.shouldFailSessionForResult` reads all three. Layer D user-input errors are non-retryable but charge normal CU — the provider does real work because responses are not cached. `relay_processor.HasNonRetryableUserFacingErrors` keys off the `IsNonRetryable` flag rather than re-classifying, so adding a new non-retryable error type requires no state-machine changes.
 
 ## 8. Chains Analyzed
 
--- a/protocol/common/endpoints.go
+++ b/protocol/common/endpoints.go
@@ -448,6 +448,12 @@ type RelayResult struct {
 	// Orthogonal to IsNonRetryable — NODE_RATE_LIMITED (2005) is retryable,
 	// NODE_LIMIT_EXCEEDED (2011) is not, and both set this flag.
 	IsRateLimited bool
+	// IsDataScope is true when the node error carries SubCategoryDataScope: the
+	// endpoint answered truthfully that it does not hold the requested data
+	// (pruned height, object that never existed). Retryable — an archive node may
+	// hold it — but the endpoint is healthy, so the direct-RPC availability gate
+	// excludes it. Orthogonal to both flags above.
+	IsDataScope bool
 }
 
 // ApplyNodeErrorClassification stamps every registry-derived policy flag onto a node-error
@@ -465,6 +471,7 @@ func (rr *RelayResult) ApplyNodeErrorClassification(family ChainFamily, transpor
 	rr.IsNonRetryable = classification.IsNonRetryable
 	rr.IsUnsupportedMethod = classification.IsUnsupportedMethod
 	rr.IsRateLimited = classification.IsRateLimited
+	rr.IsDataScope = classification.IsDataScope
 }
 
 func (rr *RelayResult) GetReplyServer() pairingtypes.Relayer_RelaySubscribeClient {
--- a/protocol/common/error_classifier.go
+++ b/protocol/common/error_classifier.go
@@ -271,8 +271,39 @@ var genericErrorMappings = map[TransportType][]errorMapping{
 
 	TransportGRPC: {
 		// gRPC status code matchers (codes from google.golang.org/grpc/codes)
+		//
+		// A code earns a row here only when its verdict is the SAME at every
+		// endpoint. Rows are what let the direct-RPC availability gate
+		// (rpcsmartrouter.shouldFailSessionForResult) exempt a status from
+		// demoting the endpoint, and an unregistered code scores — which is the
+		// right default for "we have not catalogued this failure".
+		//
+		// codes.InvalidArgument is the one caller-fault code in the set: gRPC
+		// defines it as arguments "problematic regardless of the state of the
+		// system", so every endpoint rejects the same malformed request
+		// identically. Left unregistered it classified UNKNOWN_ERROR, and a burst
+		// of malformed client requests demoted the whole healthy pairing toward
+		// the selection floor — blocklisting on the 16th consecutive one
+		// (MAG-2549). USER_INVALID_PARAMS is Retryable=false, which both stops the
+		// pointless retry and keeps the gate off the endpoint.
+		{GRPCCodeEquals(3), LavaErrorUserInvalidParams},       // codes.InvalidArgument
 		{GRPCCodeEquals(12), LavaErrorNodeUnimplemented},      // codes.Unimplemented
 		{GRPCCodeEquals(14), LavaErrorNodeServiceUnavailable}, // codes.Unavailable
+		// Deliberately NOT registered, each for its own reason — do not add them
+		// as a block, which is how `Code >= 13` went wrong in the first place:
+		//   4  DeadlineExceeded  - this endpoint was too slow; another may not be.
+		//                          Demoting it is the correct signal.
+		//   5  NotFound          - pruned vs archive nodes genuinely disagree, so a
+		//   11 OutOfRange          retry elsewhere can legitimately succeed.
+		//   7  PermissionDenied  - credentials are configured per endpoint, so one
+		//   16 Unauthenticated     rejecting ours is unusable to us and should demote.
+		//   8  ResourceExhausted - ambiguous in the gRPC spec (per-user quota vs.
+		//                          out of disk). The "rate limit" message row below
+		//                          catches the quota case without asserting the code
+		//                          alone means "healthy but busy".
+		//   1  Canceled          - a LOCAL cancellation never reaches this table;
+		//                          handleGRPCError resolves it structurally. One that
+		//                          does reach here is remote and unproven.
 		// Message-based matchers for gRPC errors conveyed without status codes
 		{MessageContains("rate limit"), LavaErrorNodeRateLimited},
 		{MessageContains("enhance_your_calm"), LavaErrorNodeRateLimited}, // HTTP/2 GOAWAY ENHANCE_YOUR_CALM
@@ -636,22 +667,35 @@ func (r stringConnectionFallback) matches(msg string) bool {
 //
 // INVARIANT: written at package-init time (declaration), read-only thereafter.
 var stringConnectionFallbacks = []stringConnectionFallback{
-	// Remote gRPC status errors render as "rpc error: code = ..."; those carry a
-	// *remote* deadline/cancel and must not be classified as a local context error.
+	// Remote gRPC status errors start with "rpc error"; those carry a *remote*
+	// deadline/cancel and must not be classified as a local context error.
+	// The guard is encoded as a mustNotContain on the rpc-error prefix marker.
 	//
-	// The guard matches that full prefix and NOT the bare substring "rpc error".
-	// Our own local wrapper renders as "gRPC error 1: context canceled", which
-	// lowercased contains "rpc error" inside the word "gRPC" — so the bare guard
-	// disqualified local cancellations from the very rows they needed to match,
-	// classifying them as UNKNOWN_ERROR instead (MAG-2687).
+	// The guard is deliberately the BARE substring "rpc error", not the fuller
+	// "rpc error: code =" prefix. Widening it to the full prefix was tried and
+	// reverted (MAG-2687): lavasession.GRPCStatusError renders as
+	// "gRPC error 1: context canceled", which lowercased contains "rpc error"
+	// inside the word "gRPC" but NOT "rpc error: code =". That wrapper is built
+	// for REMOTE statuses — a cancel the *endpoint* reported — so admitting it
+	// here labels an upstream fault as local orchestration and, because
+	// PROTOCOL_CONTEXT_CANCELED is Retryable=false, silently suppresses the
+	// retry that would have reached a healthy endpoint.
+	//
+	// Nothing is lost by keeping the guard broad. A cancellation the router
+	// itself caused never depends on this table: GRPCDirectRPCConnection
+	// .handleGRPCError returns a nil response and an error wrapping the real
+	// context.Canceled sentinel, which DetectConnectionError resolves in its
+	// layer-1 errors.Is check well before the string fallback runs. This table
+	// only ever sees errors whose sentinel chain was already lost, and for those
+	// "contains an rpc-error marker" is the conservative read.
 	{
 		mustContain:    []string{"context deadline exceeded"},
-		mustNotContain: []string{"rpc error: code ="},
+		mustNotContain: []string{"rpc error"},
 		result:         LavaErrorContextDeadline,
 	},
 	{
 		mustContain:    []string{"context canceled"},
-		mustNotContain: []string{"rpc error: code ="},
+		mustNotContain: []string{"rpc error"},
 		result:         LavaErrorContextCanceled,
 	},
 	// HTTP/2 GOAWAY closes the whole connection. Exclude ENHANCE_YOUR_CALM —
--- a/protocol/common/error_classifier.go
+++ b/protocol/common/error_classifier.go
@@ -286,15 +286,33 @@ var genericErrorMappings = map[TransportType][]errorMapping{
 		// the selection floor — blocklisting on the 16th consecutive one
 		// (MAG-2549). USER_INVALID_PARAMS is Retryable=false, which both stops the
 		// pointless retry and keeps the gate off the endpoint.
-		{GRPCCodeEquals(3), LavaErrorUserInvalidParams},       // codes.InvalidArgument
+		//
+		// Reach: this table is keyed by TRANSPORT, not by relay path, so the row also
+		// applies on the provider-based gRPC path via chainlib.GrpcErrorHandler
+		// .HandleNodeError — an INVALID_ARGUMENT becomes non-retryable there too. That
+		// is intended and follows from gRPC's own definition of the code, but it is
+		// wider than "direct RPC only" and should be read as such.
+		{GRPCCodeEquals(3), LavaErrorUserInvalidParams}, // codes.InvalidArgument
+		// codes.NotFound and codes.OutOfRange are the ordinary "I do not have this"
+		// outcomes of a Cosmos or Sui gRPC query — a missing object, an account that
+		// was never funded, a height below the node's pruning window. They are the
+		// most COMMON non-OK codes on those chains, not a failure mode.
+		//
+		// They need a row for a reason the other two do not: sendGRPCRelay marks every
+		// non-OK status IsNodeError, so without one they fall through to UNKNOWN_ERROR
+		// and the availability gate scores them — one demotion per retry, on every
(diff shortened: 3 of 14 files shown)


AnnaR-prog:
## Review

Reviewed at `2669397`. The two root causes are correctly diagnosed and correctly fixed, and the round-2 self-reversal on the classifier guard is the right call — `DetectConnectionError`'s layer-1 `errors.Is(err, context.Canceled)` runs before the string table, so change 1 makes the narrowing unnecessary.

### What I verified

- `go build ./...`, `go vet` on the four touched packages, and the `common` / `lavasession` / `rpcsmartrouter` / `relaycore` / `chainlib` suites all pass.
- `GRPCStatusError` is constructed in exactly one place and consumed in exactly one place, so `Unwrap`'s blast radius is contained.
- The `IsSessionSyncLoss` claim holds: `lavasession/common.go:16` imports `github.com/gogo/status`, while `chainlib/node_error_handler.go:18` imports grpc-go's. The two really do behave differently here.
- The new code-3 row cannot leak into other transports: `ClassifyLegacyError` passes `errorCode = 0`, and `IsUnsupportedMethodError`'s all-transport loop only tests `IsUnsupportedMethod()`. No Tier-2 `chainErrorMappings` entry uses a gRPC code, so the generic table is the whole story.

### 1. Codes 4, 5, 11, 7, 16 and message-less 8 flip from "not a node error" to "scored against the endpoint"

`direct_rpc_relay.go:812` — the main behavioural change, and it is not in the description.

Replacing `Code >= 13` with `CheckResponseError` does not just fix codes >= 13; it newly makes every non-OK code a node error. Traced end to end for `NOT_FOUND` (5):

- `IsNodeError = true`, `StatusCode = 5`.
- Code 5 is unregistered → `ClassifyError` → `LavaErrorUnknown` → `ClassifyNodeErrorForRetry` returns the zero `NodeErrorClassification{}` → `IsNonRetryable = false`, `IsRateLimited = false`.
- `shouldFailSessionForResult` (`rpcsmartrouter_server.go:1495`): the `>= 500 || == 429` gate cannot fire on a gRPC code, so it falls to `IsNodeError && !IsNonRetryable && !IsRateLimited` → **true**. Before this PR it was false. That is `OnSessionFailure`, an availability sample against the endpoint, and consecutive-failure blocklisting.
- `relaycore.setValidResponse` routes it to `nodeResponseErrors`, so `RequiredResults` (successes only) stays unsatisfied and `GetResultsSummary.hasNonRetryableNodeError` is false → the relay is **retried to other endpoints**, each retry repeating the penalty.

`NOT_FOUND` is an ordinary query outcome on Cosmos and Sui gRPC — a missing object or account, not a sick node. The PR's own argument for registering `InvalidArgument` ("a burst of malformed client requests demoted the whole healthy pairing toward the selection floor") applies to code 5 with more force, since 5 is more common. `OUT_OF_RANGE` (11) has the same shape.

The good news: this does not break clients. `processNonCrossValidationResult` (`relay_processor.go:1214`) returns `nodeErrors[0]` with `err == nil` when there are no successes, so the node's error body still reaches the caller. The cost is latency, CU and endpoint scoring — not a functional failure.

**This also makes a pre-existing inconsistency actively contradictory.** `relayInnerDirect`'s gate at `rpcsmartrouter_server.go:3868` is `(statusCode >= 500 && != 501) || == 429`, which no gRPC code satisfies, so it falls through to "Success — reset endpoint health" and calls `targetEndpoint.ResetHealth()`. Before, code 5 gave `shouldFailSession = false` *and* `ResetHealth` — coherent. Now it gives `shouldFailSession = true` *and* `ResetHealth` — the QoS path penalises the endpoint while the health path marks it healthy, for exactly the codes this PR newly scores.

Suggestion: register 5 and 11 as deterministic (they are "this data does not exist here", and the pruned-vs-archive argument covers a minority of cases), or gate the new `IsNodeError` breadth so unregistered sub-13 codes do not reach the availability signal. Either way the `>= 500 || == 429` gates need to learn about gRPC codes now that `RelayResult.StatusCode` carries them.

### 2. The cancellation fix still has the hole it was written to close

`direct_rpc_connection.go:924`

The early return requires **both** `errors.Is(ctx.Err(), context.Canceled)` **and** `status.Code(err) == codes.Canceled`. When the router cancels and grpc-go surfaces something else — `Unavailable` from a connection torn down in the same instant — the conjunct fails, and then:

`handleGRPCError` builds a `GRPCStatusError` with a non-nil response → `sendGRPCRelay` takes the node-error arm → returns `(result, nil)` → `isClientCancel := err != nil && common.IsClientCancellation(err, goroutineCtx)` at `rpcsmartrouter_server.go:1835` sees `err == nil` → **false**. A relay we cancelled is scored against the endpoint. Same bug class as MAG-2687, narrower window.

The risk is asymmetric: missing a local cancel penalises a healthy endpoint, while over-detecting one merely skips scoring for a relay we abandoned anyway. A cancelled local ctx is already the authoritative signal, so the `status.Code` conjunct costs more than it buys — dropping it, or widening it to accept `Unavailable`/`Unknown` when the local ctx is cancelled, closes this.

### 3. The empty-message trap is fixed at one layer and left at the other

`relaycore/results_manager.go:134`

The PR correctly identifies that `GrpcMessage.CheckResponseError` always returns `""`, and fixes it in `sendGRPCRelay` by falling back to `grpcErr.Message`. But `setValidResponse` hits the identical trap: `err := fmt.Errorf("%s", errorMessage)` on an empty string, then classifies with `err.Error()`. Code matchers still fire (it passes `StatusCode` as `errorCode`), but every message-based row is dead — so a `RESOURCE_EXHAUSTED` whose message says "rate limit" classifies as `UNKNOWN_ERROR` there, exactly the failure the PR describes. It also logs `"received node error reply from provider"` with an empty error string.

Impact is mostly observability and `RelayError.LavaError`, since the retry-relevant flags come from `direct_rpc_relay`'s classification. But it is the same bug one layer up, and this PR is the natural place to fix it.

### 4. The tests pin the codes that were already node errors, not the ones that just became node errors

`grpc_status_error_classification_test.go:112`

`TestSendGRPCRelay_StatusErrorClassification` covers 3, 8, 12 and 14. `NotFound` appears only in the not-cacheable loop at line 189, which asserts nothing about scoring. Nothing pins the behaviour of 4, 5 or 11 — where the change is largest.

### Smaller notes

The cancel branch drops the originating error entirely (`fmt.Errorf("gRPC relay cancelled: %w", context.Canceled)`) — slightly ironic in a PR whose other half is about preserving causes; `%w` on both would keep the grpc status for logs.

The comment "classify exactly the way the success path below does" describes a path that is dead for gRPC: the success path sets `StatusCode: http.StatusOK`, so its `CheckResponseError` can never report an error.

The code-3 registry row is transport-global, not direct-RPC-only — it also reaches `GrpcErrorHandler.HandleNodeError` (`node_error_handler.go:460`) on the provider-based gRPC path, where `INVALID_ARGUMENT` becomes non-retryable. Defensible under gRPC's own definition of the code, but a wider change than the description implies.

### Verdict

**Approve with follow-up, contingent on finding 1.** The diagnoses are right, the fixes are right, and clients still get their response bodies — nothing here breaks a request. But this is live direct-RPC code, and finding 1 changes endpoint scoring for ordinary gRPC query outcomes on every direct gRPC chain. I would want 5 and 11 resolved (registered, or excluded from the availability signal) plus tests for 4/5/11 before merge; findings 2 and 3 are small enough to fold into the same pass.

avitenzer:
## Round 3 — all four findings addressed, pushed as `b17fcc5`

Thanks — finding 1 was right, and the ResetHealth contradiction inside it was the part I had not seen at all. Branch is rebased onto current `main` (`c382aef`).

### 1. Codes 5 and 11 — took the second option, plus the coherence fix

I went with **"excluded from the availability signal"** rather than **"registered as deterministic"**, because deterministic buys the exemption at a price I do not think we want to pay: `IsNonRetryable` is read by `relay_processor.hasNonRetryableNodeError`, so marking `NOT_FOUND` non-retryable ends the request at the first endpoint that answers. A height a pruned node discarded is a height an archive node still has — that retry is the archive fallback, and on a router that routes archive traffic deliberately, killing it is a functional regression where the current bug is only a scoring one.

Both properties have to hold at once — retryable **and** unscored — and neither existing axis can express that. So this uses the escape hatch `shouldFailSessionForResult`'s own doc comment prescribes ("tag these with a fault-axis SubCategory if that proves wrong in production"):

- `SubCategoryDataScope` — "the endpoint answered truthfully that it does not hold this data".
- `NODE_DATA_NOT_HELD` (2017), `Retryable: true`, carrying that subcategory.
- gRPC `5` and `11` registered to it; `RelayResult.IsDataScope` plumbed through `ApplyNodeErrorClassification` and `relayInnerDirect`; the gate reads it as a third carve-out.

Deliberately scoped to the gRPC rows: `NODE_RESOURCE_NOT_FOUND` (2012) and the `CHAIN_*_NOT_FOUND` codes stay on the scoring side, because only the gRPC status codes were shown to be a routine query outcome rather than an error the node raised. Reopening MAG-2156's documented decision for JSON-RPC and REST is a separate change with its own evidence.

`TestDataScopeCodesAreAllRetryable_MAG2549` pins the invariant: a data-scope code registered `Retryable=false` fails the build, because at that point the axis has silently stopped buying anything and the archive fallback is gone.

**The ResetHealth contradiction is fixed too.** `relayInnerDirect` now gates on `shouldResetEndpointHealth(result)`, defined as the exact negation of `shouldFailSessionForResult`, so the health path and the QoS path cannot form different opinions about one relay. Keeping it a negation rather than a second rule means a carve-out added to either gate is automatically honoured by the other — a rate-limited, non-retryable or data-scope answer still restores health, an `UNAVAILABLE` no longer does. `TestHealthResetAgreesWithScoring` covers both directions plus the plain-success path.

**On "the `>= 500 || == 429` gates need to learn about gRPC codes":** I deliberately did not do this, and recorded why inline. Every gRPC status already reaches `shouldFailSessionForResult` through the `IsNodeError` branch, where the carve-outs apply. Teaching the status clause that 13/14/15 are "gRPC 5xx" and 8 is "gRPC 429" would re-score them *ahead* of the carve-outs and undo the rate-limit exemption `direct_rpc_relay.go` had just classified. I checked the other readers of `RelayResult.StatusCode` for the same hazard: `fiberCtx.Status(...)` is only reached from the JSON-RPC, REST and tendermint listeners, and `chainlib/grpc.go` notes the field is "used only for rest at the moment", so a gRPC code cannot leak out as an HTTP status.

Codes **4, 7, 16** and message-less **8** still score, per the reasoning already at the table — 4 is about this endpoint's latency, 7/16 about per-endpoint credentials. Flagging that explicitly since only 5 and 11 moved.

### 2. Cancellation hole — conjunct dropped

`status.Code(err) == codes.Canceled` is gone; the cancelled local context is the whole test. Your asymmetry argument is the one I applied. The deadline distinction stays, and `TestHandleGRPCError_ExpiredDeadlineIsNotACancellation` is the collateral guard now that the branch is wider — an expired deadline must still keep its response and reach the node-error arm.

`TestHandleGRPCError_CancelledContextWinsOverTheReportedStatus` runs `Unavailable`, `Unknown`, `Internal` and `Canceled` against a cancelled context; the first three fail under the old conjunct.

### 3. Empty-message trap at the other layer — fixed

`setValidResponse` now falls back to the node's own reply body, then to a synthetic `"node error with no message (status N)"` so `err` is never blank. `TestResultsManager_ClassifiesOnTheNodesBodyWhenTheMessageIsEmpty` uses `RESOURCE_EXHAUSTED` + "rate limit" as the witness, since code 8 has no registry row of its own — `NODE_RATE_LIMITED` there can only come from the body surviving. It was `UNKNOWN_ERROR` before.

### 4. Tests for 4, 5 and 11 — added

Added to `TestSendGRPCRelay_StatusErrorClassification`, with `DeadlineExceeded` as the counterweight (carving out data-scope codes must not drag the rest of the sub-13 range out with them), plus `TestSendGRPCRelay_DataScopeCodesStayRetryable` for the retry half and `OutOfRange` in the not-cacheable loop.

### Smaller notes

- **Cancel branch drops its cause** — fixed, `%w` on both: `fmt.Errorf("gRPC relay cancelled: %w: %w", context.Canceled, err)`. `TestHandleGRPCError_CancelledRelayKeepsItsCause` pins it.
- **"the way the success path below does"** — reworded. The comment now says outright that the success path's `hasError` branch is unreachable on this transport (it is only reached with `StatusCode = http.StatusOK`), and that what is being copied is its shape, not a live path.
- **code-3 row is transport-global** — recorded at the row, naming `GrpcErrorHandler.HandleNodeError` as the other consumer. The two new rows have the same reach: on the provider path a gRPC `NOT_FOUND` now classifies as `NODE_DATA_NOT_HELD` instead of `UNKNOWN_ERROR`, which is a telemetry improvement there and no retry change, since it stays `Retryable=true`.

### Verification

Every fix verified red-first by reverting it in isolation:

| Reverted | Test that goes red |
| --- | --- |
| `IsDataScope` carve-out | `TestSendGRPCRelay_StatusErrorClassification/{NotFound,OutOfRange}`, `TestSendGRPCRelay_DataScopeCodesStayRetryable` |
| `shouldResetEndpointHealth` gating | `TestHealthResetAgreesWithScoring` |
| dropping the `status.Code` conjunct | `TestHandleGRPCError_CancelledContextWinsOverTheReportedStatus/{Unavailable,Unknown,Internal}`, `TestHandleGRPCError_CancelledRelayKeepsItsCause` |
| `setValidResponse` body fallback | `TestResultsManager_ClassifiesOnTheNodesBodyWhenTheMessageIsEmpty` |

Green: `common`, `lavasession`, `relaycore`, `rpcsmartrouter`, `chainlib`, `metrics`. Plus `go build ./...`, `go vet` on the four touched packages, `gofmt -l`, `git diff --check`. `docs/ERROR-REGISTRY-DESIGN.md` updated for 2017 and the new subcategory.

avitenzer:
Rebased onto current `main` (`ec7c0a1`) — the round-3 changes described above are now `afaa30f`, not `b17fcc5`. The three commits picked up touch only `direct_grpc_subscription_manager.go` and are disjoint from this one. All 20 CI checks green on the rebased head; full suites re-run locally after the rebase.

@AnnaR-prog ready for another look whenever you have a moment — the PR is `BLOCKED` only on an approving review (your last one came through as a comment rather than a formal review, so it does not count toward the gate).

AnnaR-prog (approved):

PR #289, opened 2026-08-17
fix(smart-router): guard gRPC connector against use after Close

avitenzer:
# Description

A gRPC relay racing (or following) `Close()` dereferenced a nil connector and took the **whole process** down — every chain served by that pod, not just the gRPC one. Split out of MAG-2332 as its Bug 2; it is a DFNS Phase 2 blocker (Sui is the only gRPC chain in DFNS scope).

**Reported as one unguarded read. The connection had no coherent lifecycle at all.**

| # | Site | Problem |
| --- | --- | --- |
| 1 | `SendRequest` — `g.connector.GetRpc` | Unguarded read. `Close()` nils the field under `connMu` and does **not** clear `initialized`, so `ensureInitialized` short-circuits and the request walks into the nil. **This is the reported panic.** |
| 2 | `SendRequest` — `defer g.connector.ReturnRpc(conn)` | A *second* unguarded read of the field, at the point the `defer` statement executes. Same nil, one line later. |
| 3 | `initialize()` — `g.connector = connector` | Published under `initMu`, while `Close()` guards the same field with `connMu`. **No lock in common** — the two raced outright. |
| 4 | `ensureInitialized` / `Close` ordering | `closed` was read *before* `initMu` was taken. An initializer already past that check could publish its connector **after** `Close` had seen a nil connector and returned — a live pool behind a closed connection. |
| 5 | `GRPCConnector` (chainproxy) | No closed state at all, unlike the HTTP `Connector` beside it. `GetRpc` looped on an empty pool ignoring context cancellation, and every refill path could repopulate a pool `Close` had just drained. |
| 6 | `initialize()` — the connector's lifetime | The pool is built lazily inside the **first relay**, so `NewGRPCConnector` was handed that relay's context — and it treats its context as the connector's *lifetime*, not a dial deadline: `addClientsAsynchronouslyGrpc` ends with `go connectorLoop(ctx)`, which is `<-ctx.Done(); connector.Close()`. `sendGRPCRelay`'s own `defer cancel()` therefore closed the connector on **every** request. |
| 7 | `ensureInitialized` — cached init failure | `initialized` was stored `true` whatever `initialize()` returned, with the error cached in an `initErr` field and replayed forever. A node that was merely down when the first relay arrived poisoned the connection permanently. |

Problems 4 and 5 were found in review of the first pass; they are why a `closed` flag plus a pointer snapshot was not enough. Problem 6 was found by @AnnaR-prog in review of the second pass, and it is why the `closed` flag alone was not enough either. Problem 7 was found while fixing 6 — it is the same defect shape reached by a different route, and it is fixed here because it lands on the same function.

### Why problem 6 mattered

It was latent on `main` and survivable there: `Close()` was effectively a pool drain, and the next `GetRpc` spawned `increaseNumberOfClients` and silently refilled it, so the bug showed up only as a re-dial per relay. The `closed` flag makes the same teardown **terminal** — `GetRpc` returns `ErrGRPCConnectorClosed` forever and the refill paths discard every dial that would have healed it. `GRPCDirectRPCConnection` does not notice (its own `closed` is false, `initialized` is true), and `endpoint.DirectConnections[0]` is cached for the whole pairing, so the endpoint stays dead until the pairing is rebuilt. That is the direct gRPC path — i.e. Sui and the DFNS Phase 2 scope this PR names as its motivation.

The `chainlib` chain-proxy path was never affected: `NewGrpcChainProxy` passes a long-lived startup context, so its `connectorLoop` does not fire in normal operation.

## The fix

**Lock order is `initMu` -> `connMu`**, documented on the type. After `Close` returns: the connector is nil, no later write can install one, every borrowed client has been handed back, and further requests get `ErrGRPCConnectionClosed`.

- **The connector's lifetime and its dial deadline are now separate contexts.** `NewGRPCConnector(lifetimeCtx, dialCtx, ...)`. The root defect was that one `ctx` parameter meant three things inside it — the blocking dial's deadline, the async fill's dial deadline, and the connector's lifetime — and problem 6 was conflating the first with the third. `lifetimeCtx` owns the pool: `connectorLoop` parks on it, and the async fill dials under it, because that fill is background work which should finish rather than work belonging to whoever called. `dialCtx` bounds only the initial blocking dial, which runs in the caller's critical path.
- **The connection owns its lifetime context.** `newGRPCDirectRPCConnection` creates a `connectorCtx` rooted at `context.Background()`, and `Close()` cancels it. Rooting it at `Background` is deliberate — re-parenting on any caller-supplied context is the bug. `initialize()` passes `(g.connectorCtx, relayCtx)`: it is the one caller that genuinely has two. `chainlib`'s `NewGrpcChainProxy` passes the same startup context twice, which the call site now documents rather than hides.
- **Checkout is a lease, not a snapshot.** `SendRequest` holds `connMu` in read mode for the whole of `GetRpc`, not just the field read. Copying the pointer keeps the *pointer* alive, not the *pool* behind it — `Close` could still drain the connector in the gap before `GetRpc` reached it. With the lock held, `Close` waits behind the checkout, and by the time it proceeds the connector has counted the borrowed client and its own `Close` drains it.
- **`Close` serializes with in-flight initialization.** It publishes `closed`, cancels `connectorCtx`, then takes `initMu` to wait out any initializer already running, then detaches under `connMu`. The publish in `initialize()` re-checks `closed` under `connMu` and closes the fresh connector rather than installing it. Cancelling *before* taking `initMu` is what keeps teardown prompt: an initializer already inside a `grpc.WithBlock()` dial aborts on the cancellation instead of making `Close` wait out a full dial.
- **`Close` drains outside `connMu`,** so waiting for borrowed clients does not stall new requests; they fail fast instead.
- **Initialization latches on success only.** A failed `initialize()` is retried by the next relay instead of being replayed from a cache, and `initErr` is gone. Retrying is safe: every error path in `initialize()` leaves no connector installed, including the one that dials successfully and then finds the connection closed, which tears its own connector down. Taking a genuinely dead endpoint out of rotation is `consumer_session_manager`'s job — endpoint health, connection refusals, probe re-enable — not a latch here.
- **A relay whose context expired while queued does not start its own dial.** Retrying alone would trade a permanently dead endpoint for a serialized dial storm, since `initialize()` dials with `grpc.WithBlock()` under `initMu` and every waiter would then run a full dial budget on a context that had already expired. `ensureInitialized` checks `ctx.Err()` after acquiring `initMu`, so the queue drains rather than compounding.
- **`GRPCConnector` becomes closed- and context-aware.** `GetRpc` returns `ErrGRPCConnectorClosed`, honours `ctx`, and the fill paths (`addClient`, `increaseNumberOfClients`) drop clients dialed after `Close` instead of resurrecting the pool. `ReturnRpc` drops rather than pools a client handed back to a closing connector. The startup fill's zero-client check also no longer treats an early teardown as a reason to `LavaFormatFatal` the process.

### Cleanups from the second review round

- `registry` and `codec` were written in `initialize` and nil'd in `Close`, but read nowhere in the package. Removed, along with the locking commentary that claimed to protect them. The descriptor-set load stays as a config check, which is all it ever was — resolution goes through reflection in `getMethodDescriptor` on every path.
- `initialize` no longer re-assigns `descriptorsCache`, which `newGRPCDirectRPCConnection` already sets at construction: it was a redundant, unsynchronized pointer write to a field other methods read.
- `GRPCConnector.ReturnRpc` nil-checked `rpc` in the new closed branch but not in the pre-existing path below it. Hoisted to a single check, placed after the `usedClients` decrement that `Close`'s drain depends on.

### Correction to the earlier description of problem 2

An earlier revision of this PR claimed the deferred call re-reads `g.connector` at function *exit*. It does not: Go evaluates and saves a deferred call's receiver when the `defer` statement executes. The hazard was simply a second unguarded read at that point. Keeping the local snapshot is still correct — a pooled conn must go back to the connector that issued it — but the reasoning in the code comments has been fixed to match Go's actual semantics.

## Testing

Gated with channels and barriers rather than sleeps or scheduler luck. **Every regression test below was verified to fail against the previous behaviour** before being kept:

| Test | Reverting the fix produces |
| --- | --- |
| `TestClose_WaitsForPoolCheckout` | `Close returned while a pool checkout was still in flight` |
| `TestClose_DuringCheckout_BorrowedClientIsReturnedOnce` | same, plus the borrowed client is orphaned |
| `TestClose_DuringFirstInitialization_DoesNotPublishConnector` | a live connector installed after `Close` returned |
| `TestInitialization_ConcurrentWithClose_NeverLeaksConnector` | `Expected nil, but got: &fakeGRPCConnector{closed:false}` |
| `TestGRPCConnectorGetRpcHonoursContextCancellation` | **hangs** — test binary killed at the 60s timeout |
| `TestGRPCConnectorRefillAfterCloseIsDropped` | pool refilled after `Close` |
| `TestConnector_OutlivesTheRelayThatInitializedIt` | `connectorLoop ctx.Done` in the log, then `grpc connector is closed` — *"the pool must not die with the relay that built it"* |
| `TestInitialization_FailureIsRetriedNotLatched` | *"a failed initialize must not latch"* |
| `TestInitialization_SpentContextDoesNotQueueAnotherDial` | the queued relay dials anyway — `context.Canceled` not in the error chain |
| `TestInitialization_RelayDeadlineBoundsTheDial` | **12.02s** against a 300ms relay deadline — *"a relay must not wait out the connector's dial budget past its own deadline"* |

### Why the tests did not catch problem 6, and what closes the gap

Every `lavasession` test injects a `fakeGRPCConnector` through the `grpcConnectorFactory` seam, so none of them ever construct a real `GRPCConnector`; the `chainproxy` tests build connectors and call `Close()` directly, never through `connectorLoop`. Neither combination reached `NewGRPCConnector`'s own context handling — the seam that made this code testable is exactly what routed every test around the defect.

- `TestConnector_OutlivesTheRelayThatInitializedIt` runs the **production** factory against a real loopback gRPC server, cancels the relay context the way `sendGRPCRelay` does, and asserts the pool is still usable — then asserts `g.Close()` *does* end it, so the fix cannot degenerate into "never cancel". This is the regression pin, verified failing above.
- `TestGRPCConnectorLifetimeIsItsLifetimeContext` pins the contract itself: `lifetimeCtx` is the connector's life. New coverage rather than a regression pin — nothing exercised this before, which is the point.
- `TestGRPCConnectorDialContextDoesNotEndTheConnector` pins the other half of the split. `dialCtx` is the caller's own context and ends almost immediately; if it reached `connectorLoop` or the async fill, the split would just reintroduce problem 6.
- `TestGRPCConnectorDialContextBoundsTheBlockingDial` pins the deadline at the connector level, as `TestInitialization_RelayDeadlineBoundsTheDial` does end-to-end.
- `TestGRPCConnectorReturnRpcNilStillDecrements` covers the hoisted nil check, pinning that it does not skip the `usedClients` decrement `Close` spins on.

The two windows the first revision documented as untestable are also covered. Both needed `GetRpc` to *succeed*, which a zero-value `*grpc.ClientConn` cannot survive; a `bufconn` loopback server supplies a real one, and the connector-factory seam (completing the abstraction `grpcConnectorInterface` already opened) makes `initialize()` reachable without dialing a node. `TestSendRequest_ReturnsBorrowedClientExactlyOnce` pins that the client goes back to the connector that issued it, exactly once.

Retained from the first pass: close-before-init, close-after-init, idempotent `Close`, and 50 concurrent relays against a `Close` over 20 rounds.

Green under `-race`: `./protocol/lavasession` and `./protocol/chainlib/...` in full, plus `-run 'Test.*Close|TestConnector_|TestInitialization|TestGRPCConnector' -count=20`. Also `go build ./...`, `go vet ./...`, `git diff --check`, and `golangci-lint` output identical to the pre-change baseline (53 issues over `lavasession` + `chainlib` before and after; the three in touched files are pre-existing and on untouched lines).

## Related loose threads (not addressed here)

- The `ERR "got to setValidResponse with nil Reply"` that preceded the original panic is still on main at `protocol/relaycore/results_manager.go:112`. A relay result with nil Reply and no error suggests another path returns `(nil, nil)` on this failure mode.
- `GetRpc` spawns `go increaseNumberOfClients(ctx, ...)` with the *request's* context, so background refills die with the request that triggered them. Same class as problem 6 — background work bounded by a caller — but benign, since the next `GetRpc` retries.

Both left alone deliberately — separate defects, separate changes.

## Jira ticket

Jira ticket: MAG-2808

---

## Author Checklist

I have...

* [x] provided a link to the relevant issue or specification — MAG-2808
* [x] included the necessary unit and integration tests — the previously documented coverage gaps are now closed
* [ ] confirmed all CI checks have passed — pending CI run on the rebased head

### DIFF
--- a/protocol/chainlib/chainproxy/connector.go
+++ b/protocol/chainlib/chainproxy/connector.go
@@ -177,7 +177,19 @@ type GRPCConnector struct {
 	usedClients int64
 	credentials credentials.TransportCredentials
 	nodeUrl     common.NodeUrl
+
+	// closed is set by Close and never cleared. Unlike the HTTP Connector's atomic
+	// flag it is guarded by lock, because every reader already holds lock: that
+	// makes "is the pool alive" and "take a client" a single atomic step, which a
+	// separate atomic could not guarantee.
+	closed bool
 }
+
+// ErrGRPCConnectorClosed is returned by GetRpc once Close has run. The HTTP
+// Connector has always reported this; GRPCConnector had no closed state at all,
+// so callers racing a Close either blocked forever on a pool that would never
+// refill or had their retry goroutines repopulate it (MAG-2808).
+var ErrGRPCConnectorClosed = errors.New("grpc connector is closed")
 
 func NewGRPCConnector(ctx context.Context, nConns uint, nodeUrl common.NodeUrl) (*GRPCConnector, error) {
 	NumberOfParallelConnections = nConns // set number of parallel connections requested by user (or default.)
@@ -268,6 +280,9 @@ func (connector *GRPCConnector) getTransportCredentials() grpc.DialOption {
 func (connector *GRPCConnector) increaseNumberOfClients(ctx context.Context, numberOfFreeClients int) {
 	utils.LavaFormatDebug("increasing number of clients", utils.Attribute{Key: "numberOfFreeClients", Value: numberOfFreeClients},
 		utils.Attribute{Key: "url", Value: connector.nodeUrl.Url})
+	if connector.isClosed() {
+		return // the pool is being torn down; dialing now would only resurrect it
+	}
 	var grpcClient *grpc.ClientConn
 	var err error
 	for connectionAttempt := 0; connectionAttempt < MaximumNumberOfParallelConnectionsAttempts; connectionAttempt++ {
@@ -282,6 +297,13 @@ func (connector *GRPCConnector) increaseNumberOfClients(ctx context.Context, num
 
 		connector.lock.Lock() // add connection to free list.
 		defer connector.lock.Unlock()
+		if connector.closed {
+			// Close drained the pool while this dial was in flight. Appending now
+			// would hand a live client to a closed connector that nobody will ever
+			// borrow from again, so close it here instead.
+			grpcClient.Close()
+			return
+		}
 		connector.freeClients = append(connector.freeClients, grpcClient)
 		return
 	}
@@ -292,6 +314,10 @@ func (connector *GRPCConnector) GetRpc(ctx context.Context, block bool) (*grpc.C
 	connector.lock.Lock()
 	defer connector.lock.Unlock()
 
+	if connector.closed {
+		return nil, ErrGRPCConnectorClosed
+	}
+
 	numberOfFreeClients := len(connector.freeClients)
 	if numberOfFreeClients <= int(connector.usedClients) { // if we reached half of the free clients start creating new connections
 		go connector.increaseNumberOfClients(ctx, numberOfFreeClients) // increase asynchronously the free list.
@@ -300,18 +326,35 @@ func (connector *GRPCConnector) GetRpc(ctx context.Context, block bool) (*grpc.C
 	if numberOfFreeClients == 0 {
 		if !block {
 			return nil, errors.New("out of clients")
-		} else {
-			for {
-				connector.lock.Unlock()
-				// if we reached 0 connections we need to create more connections
-				// before sleeping, increase asynchronously the free list.
-				go connector.increaseNumberOfClients(ctx, numberOfFreeClients)
-				time.Sleep(50 * time.Millisecond)
-				connector.lock.Lock()
-				numberOfFreeClients = len(connector.freeClients)
-				if numberOfFreeClients != 0 {
-					break
-				}
+		}
+		// Wait for the async fill to produce a client. Both exits below are new:
+		// this loop used to spin unconditionally, so a caller waiting on a pool
+		// that would never refill was stuck forever even after its context was
+		// cancelled, and the goroutines it kept spawning could repopulate a pool
+		// Close had just drained (MAG-2808).
+		for {
+			connector.lock.Unlock()
+			// if we reached 0 connections we need to create more connections
+			// before sleeping, increase asynchronously the free list.
+			go connector.increaseNumberOfClients(ctx, numberOfFreeClients)
+
+			var ctxDone bool
+			select {
+			case <-ctx.Done():
+				ctxDone = true
+			case <-time.After(50 * time.Millisecond):
+			}
+
+			connector.lock.Lock()
+			if ctxDone {
+				return nil, fmt.Errorf("waiting for a free grpc client: %w", ctx.Err())
+			}
+			if connector.closed {
+				return nil, ErrGRPCConnectorClosed
+			}
+			numberOfFreeClients = len(connector.freeClients)
+			if numberOfFreeClients != 0 {
+				break
 			}
 		}
 	}
@@ -328,6 +371,15 @@ func (connector *GRPCConnector) ReturnRpc(rpc *grpc.ClientConn) {
 	defer connector.lock.Unlock()
 
 	connector.usedClients--
+	if connector.closed {
+		// Close is blocked waiting for usedClients to reach zero, so the decrement
+		// above must happen either way. Drop the conn rather than appending it to a
+		// pool that is being torn down, which would leave a live client behind.
+		if rpc != nil {
+			rpc.Close()
+		}
+		return
+	}
 	if len(connector.freeClients) > (int(connector.usedClients) + int(NumberOfParallelConnections) /* the number we started with */) {
 		rpc.Close() // close connection
 		return      // return without appending back to decrease idle connections
@@ -344,6 +396,11 @@ func (connector *GRPCConnector) connectorLoop(ctx context.Context) {
 func (connector *GRPCConnector) Close() {
 	for i := 0; ; i++ {
 		connector.lock.Lock()
+		// Mark closed under the same lock GetRpc, ReturnRpc and
+		// increaseNumberOfClients take. From here on no client can be handed out
+		// and no in-flight dial can refill the pool we are about to drain, so the
+		// usedClients wait below is guaranteed to make progress.
+		connector.closed = true
 		for i := 0; i < len(connector.freeClients); i++ {
 			connector.freeClients[i].Close()
 		}
@@ -364,6 +421,9 @@ func (connector *GRPCConnector) Close() {
 
 func addClientsAsynchronouslyGrpc(ctx context.Context, connector *GRPCConnector, nConns uint, nodeUrl common.NodeUrl) {
 	for i := uint(0); i < nConns; i++ {
+		if connector.isClosed() {
+			break // Close is draining; further dials would only be thrown away
+		}
 		rpcClient, err := connector.createConnection(ctx, nodeUrl, connector.numberOfFreeClients())
 		if err != nil {
 			break
@@ -371,7 +431,14 @@ func addClientsAsynchronouslyGrpc(ctx context.Context, connector *GRPCConnector,
 		connector.addClient(rpcClient)
 	}
 	if (connector.numberOfFreeClients() + connector.numberOfUsedClients()) == 0 {
-		if ctx.Err() != nil {
+		if connector.isClosed() {
+			// A connector closed this early legitimately has no clients — Close
+			// drained them. Reaching the fatal below would turn an ordinary
+			// teardown during startup into a process exit.
+			utils.LavaFormatWarning("gRPC connector closed before the async fill finished", nil,
+				utils.LogAttr("address", nodeUrl.UrlStr()),
+			)
+		} else if ctx.Err() != nil {
 			// Probe-scoped ctx (validateProvider, etc.) was cancelled before
 			// the async fill produced any client. Caller will treat the
 			// returned error as a probe failure; the process must not exit.
@@ -393,7 +460,20 @@ func addClientsAsynchronouslyGrpc(ctx context.Context, connector *GRPCConnector,
 func (connector *GRPCConnector) addClient(client *grpc.ClientConn) {
 	connector.lock.Lock()
 	defer connector.lock.Unlock()
+	if connector.closed {
+		// The startup fill is still running while Close drains. Appending here
+		// would refill the pool behind Close's back, exactly as the retry path in
+		// increaseNumberOfClients could (MAG-2808).
+		client.Close()
+		return
+	}
 	connector.freeClients = append(connector.freeClients, client)
+}
+
+func (connector *GRPCConnector) isClosed() bool {
+	connector.lock.RLock()
+	defer connector.lock.RUnlock()
+	return connector.closed
 }
 
 func (connector *GRPCConnector) numberOfFreeClients() int {
--- a/protocol/chainlib/chainproxy/connector.go
+++ b/protocol/chainlib/chainproxy/connector.go
@@ -191,7 +191,26 @@ type GRPCConnector struct {
 // refill or had their retry goroutines repopulate it (MAG-2808).
 var ErrGRPCConnectorClosed = errors.New("grpc connector is closed")
 
-func NewGRPCConnector(ctx context.Context, nConns uint, nodeUrl common.NodeUrl) (*GRPCConnector, error) {
+// NewGRPCConnector builds a pooled gRPC connector.
+//
+// The two contexts are deliberately separate, because a single one was three
+// different things at once and callers could only get two of them right
+// (MAG-2808):
+//
+//	lifetimeCtx  owns the pool. connectorLoop parks on it and closes the connector
+//	             when it ends, and the async fill dials under it — that fill is
+//	             background work which should finish, not work that belongs to
+//	             whoever happened to call this.
+//	dialCtx      bounds ONLY the initial blocking dial below, which runs in the
+//	             caller's critical path. It is the caller's own deadline, so a
+//	             relay does not wait materially past the point it would have given
+//	             up anyway.
+//
+// Callers with nothing to distinguish (a startup path, where the process's
+// context is both) pass the same context twice. Callers building a pool inside a
+// request — the direct-RPC path — must NOT: passing the request's context as
+// lifetimeCtx tears the pool down the moment that request returns.
+func NewGRPCConnector(lifetimeCtx, dialCtx context.Context, nConns uint, nodeUrl common.NodeUrl) (*GRPCConnector, error) {
 	NumberOfParallelConnections = nConns // set number of parallel connections requested by user (or default.)
 	connector := &GRPCConnector{
 		freeClients: make([]*grpc.ClientConn, 0, nConns),
@@ -202,12 +221,12 @@ func NewGRPCConnector(ctx context.Context, nConns uint, nodeUrl common.NodeUrl) 
 	// unconfigured endpoint to TLS on retry, so this reflects the configured intent.
 	nodeUrl.TokenOverInsecureWarning(nodeUrl.AuthConfig.GetUseTls() || strings.HasPrefix(nodeUrl.Url, "grpcs://"))
 
-	rpcClient, err := connector.createConnection(ctx, nodeUrl, connector.numberOfFreeClients())
+	rpcClient, err := connector.createConnection(dialCtx, nodeUrl, connector.numberOfFreeClients())
 	if err != nil {
 		return nil, utils.LavaFormatError("grpc failed to create the first connection", err, utils.Attribute{Key: "address", Value: nodeUrl.UrlStr()})
 	}
 	connector.addClient(rpcClient)
-	go addClientsAsynchronouslyGrpc(ctx, connector, nConns-1, nodeUrl)
+	go addClientsAsynchronouslyGrpc(lifetimeCtx, connector, nConns-1, nodeUrl)
 	return connector, nil
 }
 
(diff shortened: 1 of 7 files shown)


AnnaR-prog:
## Review

Reviewed at `6422b79`.

### What's right

The `initMu -> connMu` order is consistent across all three sites that take both. The `closed.Load()` check sitting *before* any `connMu` acquisition does deliver the fail-fast the description claims — a request arriving after `Close()` returns `ErrGRPCConnectionClosed` without queueing behind the pending writer. Draining outside `connMu` is a real improvement over the old code, which held the lock across `GRPCConnector.Close()`'s `usedClients` wait. And `GetRpc` wrapping `ctx.Err()` with `%w` composes correctly with the cancellation work in #288 — `IsClientCancellation` can see through it.

Full `lavasession` and `chainlib/...` suites pass here, and `go build ./...` is clean. The description's correction about `defer` semantics is also right: Go evaluates the receiver at the `defer` statement, so the original hazard really was a second unguarded field read at registration, not a re-read at exit.

### Blocking: the `closed` flag turns a masked bug into terminal failure for every direct gRPC connection

`chainproxy/connector.go:396` + `lavasession/direct_rpc_connection.go:814`

`GRPCConnector`'s own lifetime is bound to the context passed to `NewGRPCConnector`. `addClientsAsynchronouslyGrpc` ends unconditionally with `go connector.connectorLoop(ctx)`, and `connectorLoop` is `<-ctx.Done(); connector.Close()`.

On the direct-RPC path that context is **the first relay's request context**:

```
sendGRPCRelay: requestCtx, cancel := nodeUrl.LowerContextTimeoutWithDuration(ctx, relayTimeout); defer cancel()   // direct_rpc_relay.go:695
  -> GRPCDirectRPCConnection.SendRequest(requestCtx, ...)
  -> ensureInitialized(ctx) -> initialize(ctx) -> newConnector(ctx, 10, ...)
  -> chainproxy.NewGRPCConnector(ctx, ...) -> go addClientsAsynchronouslyGrpc(ctx, ...) -> go connectorLoop(ctx)
```

So the connector is closed the moment the first relay returns. On `main` that was survivable: `Close()` was effectively a pool drain, and the next `GetRpc` spawned `increaseNumberOfClients` and silently refilled it — the latent bug showed up only as a re-dial per relay. With a permanent `closed` flag the same teardown becomes terminal: `GetRpc` returns `ErrGRPCConnectorClosed` forever, and `addClient` / `increaseNumberOfClients` throw away every dial that would have healed it.

A/B with the same test on both branches — a real local gRPC server, `NewGRPCConnector(cancellableCtx, 2, ...)`, cancel, then `GetRpc(context.Background(), true)`:

```
main    -> connectorLoop ctx.Done ... increasing number of clients ... GetRpc -> conn=true  err=<nil>
PR head -> connectorLoop ctx.Done                                     GetRpc -> conn=false err=grpc connector is closed
```

`GRPCDirectRPCConnection` does not notice: its own `g.closed` is false and `initialized` is true, so `ensureInitialized` passes and every relay dies at the checkout. And `endpoint.DirectConnections[0]` (`consumer_types.go:1022`, `consumer_session_manager.go:508`) is a cached object reused for every relay to that endpoint — so the connection stays dead until the pairing changes and `DirectConnections` is rebuilt. That is the direct gRPC path, i.e. Sui and the DFNS Phase 2 scope this PR names as its motivation.

The `chainlib` chain-proxy path is unaffected: `NewGrpcChainProxy` passes a long-lived startup context, so its `connectorLoop` never fires in normal operation.

**The fix is not to drop the flag** — the flag is correct in isolation. It is to give the connector a lifetime that is not the first request's: a connection-owned context created in `newGRPCDirectRPCConnection` and cancelled by `g.Close()`, passed to `newConnector` instead of the relay ctx. That makes `connectorLoop` do the thing it was written to do — closing the connection closes the connector — and it is the same "teardown is a lifetime, not a flag" idea the commit message argues for.

### Why the tests did not catch it

The `lavasession` tests all inject `fakeGRPCConnector` through the new `grpcConnectorFactory` seam, so no test there ever constructs a real `GRPCConnector`. The `chainproxy` tests build connectors and call `Close()` directly, never through `connectorLoop`. Neither combination exercises `NewGRPCConnector`'s own context handling — the seam that made this code testable is exactly what routes every test around the defect. A test that goes through `NewGRPCConnector` with a cancellable ctx and asserts the connector is still usable after that ctx ends would have caught it; happy to hand over the throwaway one used above, which is nearly that test.

### Smaller notes

`g.registry` and `g.codec` are written in `initialize` (lines 898/899, 910/911) and nil'd in `Close` — and read nowhere in the package. The new `g.codec = nil` and the locking commentary about protecting those fields therefore guard nothing observable; the fields look vestigial and could go.

`initialize:843` still reassigns `descriptorsCache` even though `newGRPCDirectRPCConnection` now sets it at construction — redundant, and it is an unsynchronized pointer write to a field other methods read.

`ReturnRpc` nil-checks `rpc` in the new closed branch but the pre-existing path below it still calls `rpc.Close()` unguarded — worth making consistent while you are in there.

`Close()` now holds `initMu` across an in-flight `initialize()`, which dials with `grpc.WithBlock()`, so teardown can block for the length of a dial. Both call sites (`closePurgedUnusedPairingsConnections`, `closeDemotedDirectConnections`) are fire-and-forget goroutines, so this is fine today — just worth knowing it is no longer prompt.

### Verdict

**Request changes.** The locking work is sound and the diagnosis of the original panic is correct — but the new `closed` flag, landing on a connector whose lifetime is already tied to the first relay's context, breaks every direct gRPC relay after the first one on the very path this PR exists to protect. Fixing the connector's lifetime (and adding a test that goes through the real `NewGRPCConnector`) resolves it; the rest of the change can stand as is.

avitenzer:
Thanks — the blocking finding is correct, and the diagnosis of *why the tests missed it* was the sharpest part of it. Fixed in `df6d119` + `9595448`; PR description updated to match.

## The lifetime fix

I re-traced every link before touching anything, and it holds all the way down: `sendGRPCRelay` builds `requestCtx` via `LowerContextTimeoutWithDuration` → `CapContextTimeout` → a real `context.WithTimeout` with `defer cancel()`, and that context reached `go connectorLoop(ctx)`. `endpoint.DirectConnections[0]` is genuinely cached per pairing (`consumer_session_manager.go:508`), so the endpoint really does stay dead rather than recovering on the next relay.

Implemented as you prescribed: `newGRPCDirectRPCConnection` creates a `connectorCtx`, `Close()` cancels it, and `initialize()` passes it to the factory instead of the relay's context.

Two choices worth flagging, since neither was forced by the diagnosis:

**Rooted at `context.Background()`, not at the context `NewDirectRPCConnection` receives.** That parameter *is* the long-lived startup context today, so deriving from it would work and would propagate shutdown. I didn't, because re-parenting the pool's life on a caller-supplied context is the exact bug class here, and it would go wrong silently the first time someone passes something shorter. The connection owns the pool, so it owns the context.

**Cancel before `initMu`, not after.** This turns out to resolve your fourth smaller note rather than just documenting it: an initializer already inside a `grpc.WithBlock()` dial now aborts on the cancellation instead of making `Close` wait out a full dial, so teardown is prompt again rather than merely "fine because both call sites are fire-and-forget".

The cost is that the initial dial no longer inherits the relay's deadline. It stays bounded — `createConnection` caps each attempt at `AverageWorldLatency*2` and gives up after `MaximumNumberOfParallelConnectionsAttempts` — so worst case is ~6s against a dead node instead of the relay timeout. I took that as the better trade: cutting pool construction short at the relay deadline is what produced a failed `initialize()` in the first place, and that used to be permanent (see below). Splitting lifetime from dial-deadline properly would mean changing `NewGRPCConnector`'s signature, which reaches the chainlib path you correctly identified as unaffected — happy to do it if you'd rather have the deadline back.

## Tests

Your point that the factory seam routed every test around the defect was exactly right, and it's the reason the new pin deliberately does *not* use it:

- **`TestConnector_OutlivesTheRelayThatInitializedIt`** (lavasession) — production factory, real loopback gRPC server, relay context cancelled the way `sendGRPCRelay` does. Reverting the one-line context change reproduces your A/B verbatim:
  ```
  INF gRPC direct connection initialized url=grpc://127.0.0.1:59520
  2026/08/18 12:42:12 connectorLoop ctx.Done
  --- FAIL: TestConnector_OutlivesTheRelayThatInitializedIt
      Error:    grpc connector is closed
      Messages: the pool must not die with the relay that built it
  ```
  It also asserts the second half — that `g.Close()` *does* end the connector's life — so the fix can't degenerate into "never cancel".
- **`TestGRPCConnectorLifetimeIsItsConstructorContext`** (chainproxy) — pins the contract itself, so the next caller that hands `NewGRPCConnector` a request-scoped context trips a test rather than a production endpoint. New coverage, not a regression pin.

## Smaller notes

All three applied. `registry`/`codec` confirmed write-only across the package and removed, along with the locking commentary that claimed to protect them; the descriptor-set load stays as the config check it always was, since resolution goes through reflection on every path. The redundant `descriptorsCache` reassignment in `initialize` is gone. `ReturnRpc`'s nil check is hoisted to one place, after the `usedClients` decrement `Close`'s drain depends on, with `TestGRPCConnectorReturnRpcNilStillDecrements` pinning that the decrement still happens.

## One more, found while fixing yours

`ensureInitialized` stored `initialized = true` whatever `initialize()` returned, cached the error in `initErr`, and replayed it forever — so a node that was merely *down* when the first relay arrived poisoned the connection for the life of the pairing. Same "dead until `DirectConnections` is rebuilt" shape as your finding, reached by a different route, and on the same function I was already changing. Pre-existing on `main`, but leaving it would have meant fixing one path to a dead endpoint and shipping the other, so it's in `9595448`.

`initialized` now latches on success only. Retrying is safe because every error path in `initialize()` leaves no connector installed — including the one that dials successfully and then finds the connection closed, which tears its own connector down.

Retrying alone would have traded a permanently dead endpoint for a serialized dial storm, since `initialize()` dials with `grpc.WithBlock()` under `initMu` and each queued relay would then run a full dial budget on a context that had already expired while it waited. So `ensureInitialized` checks `ctx.Err()` after acquiring `initMu`: a relay that spent its budget queued fails instead of adding to the queue behind it, and the queue drains rather than compounding. Both halves verified failing against the old behaviour (`"a failed initialize must not latch"`, and `context.Canceled` absent from the error chain).

Keeping it here rather than splitting it out. It is the same defect class as your finding, on the same function, reached through the same cached `DirectConnections` object — landing the connector-lifetime fix while leaving the other route to a permanently dead endpoint open would be a half-fix, and reviewing the two together is what makes the shared shape visible.

## Verification

`./protocol/lavasession` and `./protocol/chainlib/...` green in full under `-race`, plus `-run 'Test.*Close|TestConnector_|TestInitialization|TestGRPCConnector' -count=20`. `go build ./...` and `go vet ./...` clean. `golangci-lint` identical to the pre-change baseline — 40 issues before and after, none in the four touched files.

avitenzer:
Follow-up on the one open question from my last comment: I said the initial dial no longer inherits the relay's deadline, estimated the cost at ~6s, and offered to split lifetime from dial deadline if you'd rather have it back.

**I under-sold it. Measured, it's 12.02s, and it isn't only the dialing relay.** Split done in `d57f9c5`.

## What I got wrong

Two corrections to that estimate:

**~12s, not ~6s.** `createConnection` runs up to `MaximumNumberOfParallelConnectionsAttempts` (10) attempts at `AverageWorldLatency*2` each — and when no credentials are set and the pool is empty, each iteration *also* tries a TLS-upgrade dial with its own 600ms cap. Measured end-to-end against a dead address with a 300ms relay deadline:

```
--- FAIL: TestInitialization_RelayDeadlineBoundsTheDial (12.02s)
    Error:    "12.020846667s" is not less than "2s"
    Messages: a relay must not wait out the connector's dial budget past its own deadline
```

**Every concurrent relay pays it, not just the one dialing.** `initMu` is a plain `sync.Mutex` with no context awareness, and `ensureInitialized` holds it across `initialize()`. A relay arriving during that dial blocks on the lock for its full duration regardless of its own deadline. The `ctx.Err()` check I added with the retry fix stops it starting a *second* dial — but only after it has already waited out the first.

That last part is specifically a product of combining the two fixes, and it's worth naming because it cuts against the retry fix you'd otherwise be right to wave through. On `main` the latch was wrong but it *bounded* the damage: relay 1 failed at its deadline and relays 2..N got the cached error instantly. Removing the latch (correct) and the deadline (a side effect) together means N relays each hang for the full budget. Twelve seconds of delayed failover for every concurrent relay to that endpoint — on the path this PR exists to protect.

## The fix, and why not a revert

Reverting would recreate your bug. The root defect was always that one `ctx` parameter means three things inside `NewGRPCConnector` — the blocking dial's deadline, the async fill's dial deadline, and the connector's lifetime — and problem 6 was conflating the first with the third. So I named the two that actually differ:

```go
func NewGRPCConnector(lifetimeCtx, dialCtx context.Context, nConns uint, nodeUrl common.NodeUrl) (*GRPCConnector, error)
```

`lifetimeCtx` owns the pool: `connectorLoop` parks on it, and the async fill dials under it — that fill is background work which should finish, not work belonging to whoever happened to call. `dialCtx` bounds only the blocking dial in the caller's critical path.

This does touch the chainlib path you confirmed was unaffected, which I want to flag rather than bury. It passes the same context twice — `NewGrpcChainProxy` is a startup path where `ctx` genuinely is both — so behaviour there is unchanged, and the call site now documents that rather than hiding it. The direct-RPC path passes `(g.connectorCtx, relayCtx)`: the one caller that really has two.

The other thing I'd flag: two adjacent `context.Context` params are swappable by mistake. I judged the doc comment plus two reviewed call sites sufficient, but if you'd rather have an options struct or a separate `NewGRPCConnectorWithDialContext`, say so — it's one more function either way.

## Tests

- **`TestInitialization_RelayDeadlineBoundsTheDial`** — end-to-end through the production factory. Dialing under the lifetime context instead produces the 12.02s failure above.
- **`TestGRPCConnectorDialContextDoesNotEndTheConnector`** — the other half, and the one that matters most: cancelling `dialCtx` must *not* close the pool. Without it, a split is just problem 6 with extra steps.
- **`TestGRPCConnectorDialContextBoundsTheBlockingDial`** — the deadline at the connector level.
- `TestGRPCConnectorLifetimeIsItsConstructorContext` renamed to `...IsItsLifetimeContext` and narrowed to the lifetime context alone.

## One more loose thread, not fixed here

`GetRpc` spawns `go increaseNumberOfClients(ctx, ...)` with the *request's* context, so background refills die with the request that triggered them. Same class as your finding — background work bounded by a caller — but benign, since the next `GetRpc` retries. Noted in the description rather than fixed; this change has grown twice already and that one earns its own review.

## Verification

`./protocol/lavasession` and `./protocol/chainlib/...` green in full under `-race`, plus `-run 'Test.*Close|TestConnector_|TestInitialization|TestGRPCConnector' -count=20`. `go build ./...` and `go vet ./...` clean. `golangci-lint` unchanged from baseline over `lavasession` + `chainlib` — 53 issues before and after, and the three that land in touched files are pre-existing and on lines this change does not touch.

AnnaR-prog (approved):

PR #290, opened 2026-08-17
fix(smart-router): stop gRPC reconnect from tearing down its own subscription

avitenzer:
# Description

A transient error on a gRPC upstream stream started a reconnect **and then ran the listener's deferred cleanup anyway.** Cleanup deletes the subscription from the manager's map, closes every client channel and nils `connectedClients` — so on a *successful* restoration:

| Step | Listener goroutine | Reconnect goroutine |
| --- | --- | --- |
| 1 | sees error, spawns reconnect, returns | — |
| 2 | defer → `cleanupSubscription`: deletes from map, `Add(-1)`, **closes all client channels** | reconnecting… |
| 3 | — | succeeds, mutates `activeSub` in place, **restarts the listener** |
| 4 | restarted listener eventually exits → `cleanupSubscription` **again** → `Add(-1)` a second time | — |

Result: the restored stream routes to nobody, the upstream stream leaks, and `totalSubscriptions` double-decrements — below zero, which disables the max-subscriptions limit.

## The fix adopts the WebSocket manager's contract in full, not just its flag

`DirectWSSubscriptionManager` already solved this, and documents it as an **ownership transfer** rather than a flag:

> once `reconnectInFlight` is set, `handleUpstreamDisconnect` is responsible for calling `cleanupSubscription` on every failure path … so a failed restoration does not leak the subscription

The gRPC manager had neither half. So this changes both:

1. `listenForUpstreamMessages` sets `reconnectInFlight` before spawning the reconnect and skips its deferred cleanup when set.
2. `handleUpstreamDisconnect` now owns cleanup on **all three** failure paths — `ReconnectWithBackoff`, `GetConnectionForStream`, `createUpstreamStream` — which previously called only `activeSub.cancel()`.

**Adding the flag alone would have traded a teardown bug for a leak.** The listener stops cleaning up, and nothing else was doing it on those failure paths. The `restoring`-CAS early return deliberately does *not* clean up — another handler is in flight and holds ownership — and now says so in a comment.

Net **+28 / −3** in production code.

## Testing

| Reverted | Test | Result |
| --- | --- | --- |
| cleanup-ownership | `TestGRPCHandleUpstreamDisconnect_ReconnectFailureCleansUp` | **FAIL** ✓ |
| `reconnectInFlight` guard | `TestGRPCListenForUpstreamMessages_NormalExitStillCleansUp` | passes — expected; see below |

That second row is stated rather than hidden: the normal-exit test is a **collateral guard**, there to prove the new flag cannot leak on non-hand-off paths. It should stay green either way. A third test pins the concurrent-restoration early return.

Full `rpcsmartrouter` suite green under `-race` (98s). Tests renamed to a `TestGRPC…` prefix because they otherwise collide with the WebSocket manager's identically-named tests in `-run` patterns.

## The gap, and why it is narrower than it looks

The WS manager *does* have the analogous `TestListenForUpstreamMessages_ReconnectSkipsCleanup`, so the asymmetry needs an explanation rather than silence. Two reasons it does not transfer, both recorded in the test file:

1. WS made its error source injectable **on purpose** — `upstreamErrSource` exists, per its own comment, so tests can drive the error branch with a fake. gRPC's error source is `upstreamStream.RecvMsg`, and `grpc.ClientStream` is also an interface, so that half would fake fine.
2. The blocker is the statement *before* it: `msgFactory.NewMessage(activeSub.methodDescriptor.GetOutputType())` nil-derefs without a real `*desc.MethodDescriptor`, and **this repo has no generated `grpc.ServiceDesc` anywhere** — protos resolve dynamically through reflection, so there is nothing to load one from.

Giving the gRPC listener the same seam WS has is a production change for testability and belongs with MAG-2643 — the work that makes this code reachable in production at all.

## Sequencing note for reviewers

This code is **currently unreachable in production**: `DirectGRPCSubscriptionManager.StartSubscription` has zero callers, and the relay path refuses gRPC streaming at `rpcsmartrouter_server.go:3675`. That makes this a zero-risk merge, and it is deliberately landing *before* MAG-2643 rather than after — wiring streaming first would ship it with a known teardown-and-leak bug on the first transient error.

The `Blocks` link (MAG-2643 blocks this) is about **verification** — this cannot be reproduced end-to-end until streaming is wired — not about implementation order.

## Review follow-up (commit `dd6e6b9`)

Three findings from a review pass. The original change removed the one *known* double-call path; these make the damage class structurally impossible.

1. **`cleanupSubscription` is now idempotent**, matching the guard `DirectWSSubscriptionManager.cleanupSubscription` has had all along. The map `delete` was already a no-op on a missing key, but `totalSubscriptions.Add(-1)` was unconditional — so any second call drove the counter below zero, which is precisely what disables the max-subscriptions limit.

   **This is not merely defensive.** There is a live second path even after the ownership fix: `cleanupStaleSubscriptions` removes any subscription whose ctx is done, and `handleUpstreamDisconnect`'s failure paths call `activeSub.cancel()` **before** cleaning up. A sweep landing in that window decrements, and cleanup decrements again. The guard — not the call ordering — is what closes it.

   It also protects the reconnect-**success** path this PR cannot reach with a test: even if the listener's cleanup did run alongside a restoration, the counter would stay correct.

2. **Cleanup ownership is expressed once**, as a deferred `restored` check, instead of repeated at three early returns. A fourth failure path added later is covered by construction — the same shape of mistake as MAG-2538, where a sequence was copied and one copy missed the fix.

3. **The other `totalSubscriptions.Add(-1)`** (in `cleanupStaleSubscriptions`) is safe as written — it only decrements keys found while iterating the map — and its interaction with `cancel()`-then-cleanup is now documented on `cleanupSubscription` rather than left to be rediscovered.

Both new guards verified **red-first**: reverting the idempotence check makes the counter read `-1`; reverting the deferred ownership leaves the subscription registered after a failed restoration.

## Review follow-up 2 (commit `7020d59`)

Five findings from @AnnaR-prog's review. The through-line: the previous commit made
`cleanupSubscription` idempotent with an early return on a missing map key — but that
early return exits the *whole* function, so it guarded the counter by skipping the
release. Each fix moves a guard onto the thing it is actually a property of.

| # | Finding | Fix |
| --- | --- | --- |
| 1 | Idempotence guard skipped resource release | `cleanupStaleSubscriptions` delegates to `cleanupSubscription` instead of partial-reaping; guard moves onto the subscription object (`cleanedUp`) |
| 2 | Guard checked key presence, not identity | Map delete is gated on `current == activeSub`; counter is decremented per object |
| 3 | CAS early-return could orphan a subscription | `restoring.Store(false)` moved ahead of the restarted listener's spawn |
| 4 | `IsIdempotent` only asserted the counter | New `requireGRPCSubReleased` asserts channel closed, `connectedClients` nil, client untracked, ctx cancelled |
| 5 | Cleanup never untracked `clientSubscriptions` | Untrack loop added |

Two things worth calling out beyond the table.

**The sweep was the real second decrement path.** Finding 1's interleaving is not narrow:
`removeClientFromSubscription` cancels when the last client leaves and defers cleanup to a
listener that can be parked in `RecvMsg` indefinitely — `ctx.Done()` is only checked
*between* receives. On a quiet stream the sweep wins routinely. With the sweep delegating,
there is now exactly one release site and one decrement site, matching
`createNewSubscription`'s single increment.

**`cancel()` moved into `cleanupSubscription`** and off the three `handleUpstreamDisconnect`
failure paths, so the sweep can never observe a cancelled-but-unreleased subscription. Same
placement as `DirectWSSubscriptionManager.cleanupSubscription`.

The diff also takes `clientRouterIDs` and `upstreamConnection` under `activeSub.lock` —
both were read unsynchronised while `handleUpstreamDisconnect` writes them under it.

### Verification

All five guards verified red-first:

| Reverted | Test that goes red |
| --- | --- |
| sweep delegates to cleanup | `CleanupStaleSubscriptions_ReleasesResources` |
| identity check on map delete | `CleanupSubscription_DoesNotEvictSuccessor` |
| identity check as a whole-function early return | `DoesNotEvictSuccessor` — counter reads 2, not 1 |
| `cleanedUp` latch → map-existence guard | `DoesNotEvictSuccessor` |
| per-client untrack | `CleanupSubscription_IsIdempotent` |
| `cancel()` in cleanup | `CleanupSubscription_IsIdempotent` |

Full `rpcsmartrouter` suite green under `-race` (104s) on the rebased base.

**Not covered:** finding 3. The latch-release ordering only matters on the
reconnect-*success* path, which is unreachable in a unit test for the reason already
documented in the test file — `methodDescriptor` nil-derefs and this repo has no generated
`grpc.ServiceDesc` to build one from. Stated in the test file's NOT COVERED block rather
than left implied.

**Still open, deliberately:** the AB-BA lock inversion between `Unsubscribe` (dgm→sub) and
`removeClientFromSubscription` (sub→dgm). Out of scope here and left for MAG-2643; the new
untrack loop is placed *outside* `activeSub.lock` specifically so it does not add a third
site to that inversion.

**On CI:** correct that no workflow in `.github/workflows/` runs `go test` — verified. The
checklist item below is about the configured checks passing and should not be read as
evidence these tests run in CI. They do not; the `-race` runs above are the only evidence.

## Jira ticket

Jira ticket: MAG-2540

---

## Author Checklist

I have...

* [x] provided a link to the relevant issue or specification — MAG-2540
* [x] included the necessary unit and integration tests — unit tests added; the one integration-only gap is documented above and in the test file
* [ ] confirmed all CI checks have passed — pending first CI run

### DIFF
--- a/protocol/rpcsmartrouter/direct_grpc_subscription_manager.go
+++ b/protocol/rpcsmartrouter/direct_grpc_subscription_manager.go
@@ -52,6 +52,11 @@ type grpcActiveSubscription struct {
 	// Restoration state
 	restoring atomic.Bool
 
+	// Set once cleanupSubscription has released this subscription's resources. Keyed on
+	// the subscription object rather than on its presence in activeSubscriptions, so
+	// release runs exactly once no matter who reaches cleanup first.
+	cleanedUp atomic.Bool
+
 	// Message sequence counter
 	messageSeq atomic.Uint64
 
@@ -273,29 +278,39 @@ func (dgm *DirectGRPCSubscriptionManager) cleanupLoop(ctx context.Context) {
 	}
 }
 
-// cleanupStaleSubscriptions removes subscriptions with cancelled contexts
+// cleanupStaleSubscriptions removes subscriptions with cancelled contexts.
+//
+// It delegates to cleanupSubscription rather than deleting the entry itself. A partial
+// reap — map delete plus counter, and nothing else — leaves the client channels open,
+// the connection's stream count inflated so the pool cannot scale down, and the idMapper
+// entries live, with no one left to release them: the listener's own cleanup arrives to
+// find the key already gone (MAG-2540).
+//
+// The window is not narrow. removeClientFromSubscription cancels the subscription when
+// its last client leaves and leaves cleanup to the listener, which may be parked in
+// RecvMsg indefinitely — ctx.Done() is only checked between receives — so on a quiet
+// stream this sweep routinely gets there first.
 func (dgm *DirectGRPCSubscriptionManager) cleanupStaleSubscriptions() {
-	dgm.lock.Lock()
-	defer dgm.lock.Unlock()
-
-	var toRemove []string
-	for hashedParams, sub := range dgm.activeSubscriptions {
+	dgm.lock.RLock()
+	var stale []*grpcActiveSubscription
+	for _, sub := range dgm.activeSubscriptions {
 		select {
 		case <-sub.ctx.Done():
-			toRemove = append(toRemove, hashedParams)
+			stale = append(stale, sub)
 		default:
 			// Still active
 		}
 	}
+	dgm.lock.RUnlock()
 
-	for _, hashedParams := range toRemove {
-		delete(dgm.activeSubscriptions, hashedParams)
-		dgm.totalSubscriptions.Add(-1)
+	// Called unlocked: cleanupSubscription takes dgm.lock itself.
+	for _, sub := range stale {
+		dgm.cleanupSubscription(sub.hashedParams, sub)
 	}
 
-	if len(toRemove) > 0 {
+	if len(stale) > 0 {
 		utils.LavaFormatDebug("DirectGRPC: cleaned up stale subscriptions",
-			utils.LogAttr("count", len(toRemove)),
+			utils.LogAttr("count", len(stale)),
 		)
 	}
 }
@@ -623,18 +638,14 @@ func (dgm *DirectGRPCSubscriptionManager) createUpstreamStream(
 	return stream, nil
 }
 
-// listenForUpstreamMessages listens for messages from upstream gRPC stream
-// On an upstream error we hand off to handleUpstreamDisconnect, which either restores the
-// subscription (mutating activeSub in place and spawning a fresh listener) or tears it down.
-// Cleanup must NOT run from this goroutine in that case: it deletes the entry from
-// dgm.activeSubscriptions, closes every client channel and nils connectedClients — so a
-// successful restoration would route into nobody, leak the new upstream stream, and
-// decrement totalSubscriptions a second time when the restarted listener later exits,
-// driving the counter below zero and disabling the max-subscriptions limit (MAG-2540).
+// listenForUpstreamMessages listens for messages from upstream gRPC stream.
 //
-// Cleanup ownership: once reconnectInFlight is set, handleUpstreamDisconnect owns
-// cleanupSubscription on every failure path, so a failed restoration still releases the
-// subscription. This mirrors the contract DirectWSSubscriptionManager already documents.
+// On an upstream error, ownership of the subscription transfers to
+// handleUpstreamDisconnect, which restores it in place or releases it. Cleanup must not
+// also run here: it closes every client channel and nils connectedClients, so a
+// successful restoration would route into nobody and leak the new stream (MAG-2540).
+// Once reconnectInFlight is set, handleUpstreamDisconnect owns cleanup on every failure
+// path — the same contract DirectWSSubscriptionManager documents.
 func (dgm *DirectGRPCSubscriptionManager) listenForUpstreamMessages(
 	ctx context.Context,
 	hashedParams string,
@@ -733,9 +744,13 @@ func (dgm *DirectGRPCSubscriptionManager) handleUpstreamDisconnect(
 ) {
 	// Prevent concurrent restoration.
 	//
-	// Returning here without cleanup is correct: another handleUpstreamDisconnect is
-	// already in flight for this subscription and holds cleanup ownership, so cleaning
-	// up here would tear down a restoration that is still in progress.
+	// Returning here without cleanup is correct: another handleUpstreamDisconnect holds
+	// the latch and owns cleanup for this subscription, so tearing down here would kill a
+	// restoration still in progress. That is only true because the latch is released
+	// before the restarted listener is spawned (see the end of this function) — if it
+	// outlived the hand-off, a handler spawned by that listener would lose the CAS and
+	// return, leaving the subscription registered with no listener, nobody reconnecting,
+	// and an uncancelled ctx that keeps the stale sweep from collecting it.
 	if !activeSub.restoring.CompareAndSwap(false, true) {
 		return
 	}
@@ -745,6 +760,7 @@ func (dgm *DirectGRPCSubscriptionManager) handleUpstreamDisconnect(
 	// that handed off skipped its own deferred cleanup, so every way out of this
 	// function short of a completed restoration must release the subscription — and a
 	// failure path added later gets that for free instead of having to remember it.
+	// cleanupSubscription cancels, so the failure paths below do not.
 	restored := false
 	defer func() {
 		if !restored {
@@ -759,7 +775,6 @@ func (dgm *DirectGRPCSubscriptionManager) handleUpstreamDisconnect(
 	// Reconnect pool
 	if err := activeSub.upstreamPool.ReconnectWithBackoff(ctx); err != nil {
 		utils.LavaFormatWarning("DirectGRPC: failed to reconnect", err)
-		activeSub.cancel()
 		return
 	}
 
@@ -767,7 +782,6 @@ func (dgm *DirectGRPCSubscriptionManager) handleUpstreamDisconnect(
 	newConn, err := activeSub.upstreamPool.GetConnectionForStream(ctx)
 	if err != nil {
 		utils.LavaFormatWarning("DirectGRPC: failed to get new connection", err)
-		activeSub.cancel()
 		return
 	}
 
@@ -781,7 +795,6 @@ func (dgm *DirectGRPCSubscriptionManager) handleUpstreamDisconnect(
 	)
 	if err != nil {
 		utils.LavaFormatWarning("DirectGRPC: failed to create new stream", err)
-		activeSub.cancel()
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


AnnaR-prog:
## Review

Reviewed at `dd6e6b9`. The design intent — adopt the WS manager's ownership contract rather than just its flag — is right, and the two halves of the ownership fix are correct.

### What I verified

- All four new tests pass under `-race`.
- Both red-first claims hold. Removing the existence guard → `IsIdempotent` fails with `actual: -1`. Removing the deferred ownership → `ReconnectFailureCleansUp` fails with the subscription still registered.
- "Unreachable in production" is accurate: `DirectGRPCSubscriptionManager.StartSubscription` has zero non-test callers, and `rpcsmartrouter_server.go:3675` refuses gRPC streaming.

### 1. The idempotence guard skips resource release, not just the counter

`direct_grpc_subscription_manager.go:831`

The early return exits the *whole* function. It's reachable through the exact interleaving the new comment cites as justification, because `cleanupStaleSubscriptions:288-294` does a **partial reap** — `delete` + `Add(-1)` and nothing else:

1. Something cancels the sub ctx.
2. The sweep ticks, sees `ctx.Done()`, removes the entry and decrements — never closing client channels, never calling `NotifyStreamRemoved`, never calling `RemoveMapping`.
3. The listener finally returns → `cleanupSubscription` → key not found → **early return**.

Net: counter correct, but the client's `replyChan` is never closed (`SafeChannelSender` only closes on `Close()`; ctx cancellation just makes sends drop), the connection's stream count stays inflated so the pool can't scale down, and `idMapper` entries leak. Before this PR that same interleaving gave you `-1` on the counter but *did* release everything. As written it trades a counter bug for a silent resource leak.

The widest window isn't the reconnect path (cancel → cleanup back-to-back). It's `removeClientFromSubscription:960` — the last-client branch calls `sub.cancel()` + `close(closeSubChan)` and never calls cleanup itself, leaving that to the listener. The listener can be parked in `RecvMsg` indefinitely: `ctx.Done()` is only checked *between* receives (the `select` has a `default:`), and the initial stream is bound to the caller's `ctx` at line 504, not the `subCtx` created at line 513 — so `cancel()` doesn't abort it. On a quiet stream the sweep at `CleanupInterval` wins routinely.

Suggested fix, and it lands further inside the WS contract this PR is adopting: have `cleanupStaleSubscriptions` delegate to `cleanupSubscription` instead of partial-reaping, and/or move `activeSub.cancel()` into `cleanupSubscription` the way WS does (`direct_ws_subscription_manager.go:1570`) and drop the three explicit `cancel()` calls. Either closes the race rather than guarding it.

### 2. The guard should check pointer identity, not just key presence

`direct_grpc_subscription_manager.go:832`

`hashedParams` is `hash(methodPath + requestData)` — deterministic, so a client re-subscribing to the same method gets the same key. This PR moves cleanup to *after* `ReconnectWithBackoff`, so a stale handler can now sit for seconds and then tear down a *newer* subscription registered under that key, closing its clients' channels. Previously the listener's cleanup ran promptly, before the backoff, so the window was much narrower. One line, and the signature already carries `activeSub`:

```go
if cur, found := dgm.activeSubscriptions[hashedParams]; !found || cur != activeSub {
```

This is a place gRPC can be strictly better than WS, whose cleanup only takes `hashedParams` and can't do the check.

### 3. The CAS early-return can orphan a subscription

`direct_grpc_subscription_manager.go:741` / `810`

`defer activeSub.restoring.Store(false)` is registered *before* the cleanup defer, so it runs last — after `go dgm.listenForUpstreamMessages(...)` has already spawned. In that window: the new listener errors immediately, sets `reconnectInFlight`, spawns handler B; B's CAS fails and it returns without cleanup; handler A has `restored = true` so it won't clean up either. The subscription stays registered with no listener and nobody reconnecting, and since its ctx was never cancelled the stale sweep won't collect it. Previously the listener's unconditional cleanup caught this.

The comment at line 736 ("another handleUpstreamDisconnect is already in flight … and holds cleanup ownership") is what's false here — the in-flight handler has already finished restoring. Fix: `activeSub.restoring.Store(false)` explicitly before the `go` spawn (the trailing defer then becomes a harmless no-op). Narrow window, and WS has the identical hole — a contract-level gap, not a gRPC regression.

### 4. `IsIdempotent` only asserts the counter

`direct_grpc_subscription_reconnect_test.go:117-134`

It would pass unchanged with finding #1's leak in place. Worth asserting that the first call actually released resources (`connectedClients` nil, sender closed), and adding a sweep-then-cleanup case.

### 5. `cleanupSubscription` never untracks `clientSubscriptions`

After a listener-side teardown, `checkClientSubscriptionLimit:1105` keeps counting a dead subscription and ratchets toward the per-client cap. Pre-existing, but on-theme: WS's cleanup does clean its per-client map, and the new comment claims the guard "covers more than the counter."

### Not this PR, but flagging since this file is about to get wired up

There's an AB-BA lock inversion between `Unsubscribe:876` (holds `dgm.RLock`, takes `sub.RLock`) and `removeClientFromSubscription:942` (holds `sub.Lock`, takes `dgm.Lock` via `untrackClientSubscription`). A pending writer blocking new readers on `dgm.lock` is enough to wedge both.

Two smaller notes: no workflow in `.github/workflows/` runs `go test`, so these tests won't be enforced by CI — the checklist's "all CI checks passed" isn't evidence they run. And the comment-to-code ratio is steep (~30 lines of comment on ~10 lines of change); the ticket ref plus the mechanism would carry it.

### Verdict

**Approve.** The core fix is correct and the ownership contract is the right model. Nothing here blocks the merge — the code has no production callers, so the findings can't bite today. Findings 1 and 2 are one-liners and worth folding in now; all of them should be closed before MAG-2643 wires streaming up, since that's the moment they become live.

avitenzer:
Thanks — all five are fixed in `7020d59`. The review found the real shape of it: the
previous commit guarded the counter by placing an early return in front of the release.
Each fix below moves a guard onto the thing it is actually a property of.

### 1. Idempotence guard skipped resource release

Fixed, taking the first of your two suggestions and the second along with it.
`cleanupStaleSubscriptions` now delegates to `cleanupSubscription` instead of partial-reaping,
and `activeSub.cancel()` moved *into* `cleanupSubscription`, off all three
`handleUpstreamDisconnect` failure paths. So there is now one release site, one decrement
site, and no window in which the sweep can observe a cancelled-but-unreleased subscription.

The `removeClientFromSubscription` → parked-in-`RecvMsg` path you identified as the wide
window is what made me do both rather than pick one: with cleanup deferred to a listener
that may not check `ctx.Done()` for a long time, the sweep isn't a rare interleaving, it's
the common one. That's recorded on the sweep now.

The guard itself moved onto the subscription object (`cleanedUp atomic.Bool`) rather than
its presence in the map — "released exactly once" is a property of the subscription, not of
a map entry, and keying it there is what makes the release unskippable regardless of who
gets to cleanup first.

### 2. Pointer identity — fixed, but not as the early return

Adopted the check, applied to the map delete only rather than as a whole-function guard.
Your literal one-liner reproduces finding #1 inside finding #2's fix: the superseded
subscription returns before closing its own client channels. I wrote the test first and it
catches exactly that —

```
--- FAIL: TestGRPCCleanupSubscription_DoesNotEvictSuccessor
    Error: Not equal: expected: 1, actual: 2
```

— the successor stays registered correctly, but the superseded subscription strands its
counter slot and never releases. So it splits: identity gates the *delete*, `cleanedUp`
gates everything else including the decrement. The counter is then per subscription object,
one decrement matching `createNewSubscription`'s one increment, which also fixes the
counter staying permanently inflated after `Stop()`.

### 3. CAS early-return orphan

Fixed as suggested — `restoring.Store(false)` before the `go` spawn, trailing defer left as
a no-op. You're right that the comment was the false part; it now states the ordering
dependency explicitly, since the "another handler holds ownership" claim is only true
*because* the latch is released before the hand-off.

Honest gap: this one has no test. It only manifests on the reconnect-**success** path, which
is unreachable in a unit test for the reason already documented in the test file
(`methodDescriptor` nil-derefs, no generated `grpc.ServiceDesc` in the repo to build one
from). Written into the NOT COVERED block rather than left implied.

Left the identical hole in `DirectWSSubscriptionManager` (line 1386) alone — agreed it's a
contract-level gap, and widening this PR into that file seemed worse than filing it.

### 4. `IsIdempotent` only asserted the counter

Fixed. The fixture now carries a real client with a live channel, and
`requireGRPCSubReleased` asserts the channel is *closed* (not merely drained),
`connectedClients` nilled, the client untracked, and the ctx cancelled. Every cleanup test
runs it. Added the sweep-then-cleanup case you asked for
(`TestGRPCCleanupStaleSubscriptions_ReleasesResources`) plus the successor case for #2.

Five guards verified red-first, including the two reverts that only the new assertions
catch:

| Reverted | Goes red |
| --- | --- |
| sweep delegates to cleanup | `CleanupStaleSubscriptions_ReleasesResources` |
| identity check on map delete | `DoesNotEvictSuccessor` |
| identity check as whole-function early return | `DoesNotEvictSuccessor` (counter 2, not 1) |
| `cleanedUp` → map-existence guard | `DoesNotEvictSuccessor` |
| per-client untrack | `IsIdempotent` |
| `cancel()` in cleanup | `IsIdempotent` |

### 5. `clientSubscriptions` never untracked

Fixed. Deliberately placed *outside* `activeSub.lock` — `untrackClientSubscription` takes
`dgm.lock`, and putting it inside would have added a third site to the inversion you flagged
below. Client keys are snapshotted under the sub lock, then untracked unlocked.

Also took `clientRouterIDs` and `upstreamConnection` under `activeSub.lock` while I was in
there; both were read unsynchronised against `handleUpstreamDisconnect`'s writes.

### On the rest

**AB-BA inversion** — left alone, agreed it's not this PR. Noted in the PR body as owed
before MAG-2643, and the new untrack loop is positioned so it doesn't make it worse.

**CI doesn't run `go test`** — verified, you're right: none of the 9 workflows in
`.github/workflows/` invoke it. Said so explicitly in the PR body so the checklist isn't
read as evidence these tests are enforced. The `-race` runs are the only evidence there is.

**Comment-to-code ratio** — fair. Trimmed the `listenForUpstreamMessages` block from 12
lines to 7 and rewrote the `cleanupSubscription` doc as two named guards instead of a
narrative. It's still comment-heavy, which I'd defend for this particular function — but
the ratio is better and the mechanism is stated once rather than three times.

Full `rpcsmartrouter` suite green under `-race` (104s), re-run after rebasing onto the
updated base.

AnnaR-prog (approved):

PR #292, opened 2026-08-18
feat(smart-router): serve gRPC server-streaming methods (MAG-2643)

avitenzer:
Jira ticket: MAG-2643

## Why

`DirectGRPCSubscriptionManager.StartSubscription` had **zero production callers**. It could open an upstream stream and fan it out to per-client channels, but nothing on the serving side held a client connection open to read one — so gRPC server-streaming could not be served in any mode.

Worse than "unsupported": when the streaming guard didn't fire, a server-streaming method fell through to a unary `Invoke`, which reads the first message and errors on the second. The caller waited out the full (usually generous, `hanging_api`) timeout and got a **truncated stream reported as success**.

## What

**The listener half.** `grpcproxy`'s handler already receives a `grpc.ServerStream`, so this needed a pump, not a new transport: forward until the producer closes or the client goes away, then run `StreamResponse.Close` to release the client upstream. `GrpcChainListener` wires it to the manager through a `GRPCSubscriptionProvider` optional interface on `RelaySender` — the same shape `GRPCReflectionProvider` already uses, so no signature churn.

The acknowledgement payload deliberately does **not** go on the wire: every frame of a gRPC stream must decode as the method's output type, so the router's subscription id rides in the response headers instead.

**Streaming is decided from the spec**, via the `FUNCTION_TAG_SUBSCRIBE` parse directive — the same signal the WebSocket path already uses. Not from a live reflection lookup: reflection is throttled or disabled on many public gRPC gateways, and that was the only classifier, so a failed lookup silently produced the truncated-stream behaviour above. Requires [lava-specs#115](https://github.com/Magma-Devs/lava-specs/pull/115) (merged).

**The relay-path gate is now a backstop, not the mechanism.** Spec-first, so it depends on neither the manager nor reflection; reflection remains a second line of defence for methods the spec doesn't declare, with an error naming the spec fix. Its old message pointed users at provider-based relay, which cannot serve streaming and is unreachable in this repo.

## ⚠️ `main` is currently broken for Sui gRPC

Now that lava-specs#115 is merged, `sui.json` carries SUBSCRIBE directives. On `main` that makes the chain router demand a WebSocket URL for a **gRPC** collection, so every Sui gRPC provider is rejected at startup:

```
provider validation failed — excluding from provider list   provider=sui-grpc-1
provider validation failed — excluding from provider list   provider=sui-grpc-2
ATTENTION: no healthy providers for endpoint — serving unavailable
```

Not even reflection works. gRPC server-streaming rides the same HTTP/2 connection as every other gRPC call — there is no WebSocket in that picture. This PR scopes that requirement to the interfaces that genuinely subscribe over a socket, and is what re-syncs the two repos.

## Defects found while making this path reachable

| | |
|---|---|
| **Shared stream died with its creator** | `conn.NewStream` binds an upstream stream to the context it was created with — the subscribing client's request context. Sharing is on by default, so the first client's disconnect cancelled the stream underneath every client that had joined. Now rooted in the manager. |
| **Shutdown panic** | `Stop` and `removeClientFromSubscription` both closed `closeSubChan` with no latch. The listener only notices between receives, so a subscription stays registered long after its last client left — a `Stop` in that window closed an already-closed channel and killed the process. |
| **Empty request never sent** | A server-streaming call owes the server exactly one request message before half-closing. An all-default request marshals to zero bytes, and the send was skipped on that basis — so `SubscribeCheckpoints` with no options half-closed without sending anything and the upstream failed with `Internal: Missing request message`. Found by running against a live node. |
| **Joiner got someone else's id** | A joining client was handed the creator's cached acknowledgement, telling it to unsubscribe with an id registered to another client. |
| **JSON request bodies rejected** | The parser used the known-type factory, which returns a linked Go type when one is registered, and the JSON branch rejected it outright — anything cosmos-shaped took that path. |

## Also: provider-relay subscription plumbing removed

`SendNodeMsg` carried a `ch chan interface{}` for a provider relay's subscription frames plus `subscriptionID` / `*rpcclient.ClientSubscription` returns. Every call site passed `nil` and discarded both. jsonrpc and tendermint carried a full second code path for it; rest and grpc only a guard rejecting it. All removed — `SendNodeMsg` is now one request in, one reply out. No behaviour change; the branches were unreachable.

## Verification

**End to end against live Sui testnet** (`fullnode.testnet.sui.io:443`), on the rebased branch:

- reflection and unary `GetServiceInfo` through the router
- `SubscribeCheckpoints` streaming continuously — 65 messages in 25s
- subscription sharing: second client logs `client joined existing subscription`, both receive
- teardown: `subscription cleaned up` on disconnect, no leak

**Tests.** Six guards verified red-first by mutating the implementation. Full `./protocol/...` + `./types/...` green (25 packages); `chainlib` and `rpcsmartrouter` also under `-race`.

The end-to-end manager tests drive a real upstream stream — `ServerReflectionInfo` is borrowed purely as an already-linked server-streaming descriptor, seeded into the connection cache, so no reflection happens in test.

## Out of scope / follow-ups

- **gRPC cross-validation does not pass.** Each provider fills its own descriptor cache by reflection on first use; against Sui testnet that outlives the 2s relay timeout, so a cold provider fails at exactly 2.000s, never caches, and fails identically next time — CV stalls at `successCount:1`. Pre-existing on the unary path, unrelated to streaming, documented in the Sui script with the log line to match on. Deserves its own ticket.
- Per-client subscription limits (subscribe rate, max-per-client) don't bind for gRPC: one stream is one subscriber, so each stream is a new client key. The global cap is what bounds this interface. Rationale is in the code.
- The router rejects an absolute config path (`viper.SetConfigName` + search paths). Every pre-setup script works around it with a relative path.

## Test script

`scripts/pre_setups/init_smartrouter_sui.sh` — the Sui/gRPC counterpart of the ETH one. Fetches `sui.json` (not bundled here) into gitignored `debugging/`, defaults to Sui's public testnet fullnode because it streams anonymously, and supports keyed gateways via per-endpoint `SUI_GRPC_API_KEY_N`. Keys are per endpoint on purpose — one shared key would be rendered into every node-url.

[MAG-2643]: https://magmadevs.atlassian.net/browse/MAG-2643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


protocol/chainlib/grpc.go
```diff
@@ -498,6 +515,141 @@ func (apil *GrpcChainListener) Serve(ctx context.Context, cmdFlags common.Consum
...
+		metricsData := metrics.NewRelayAnalytics(dappID, apil.endpoint.ChainID, apil.endpoint.ApiInterface)
+		metricsData.SetProcessingTimestampBeforeRelay(startTime)
+
+		utils.LavaFormatDebug("in <<< GRPC stream subscribe",
+			utils.LogAttr("GUID", ctx),
+			utils.LogAttr("_method", method),
+			utils.LogAttr("dappID", dappID),
+		)
+
+		firstReply, repliesChan, err := subscriptionManager.StartSubscription(ctx, protocolMessage, dappID, consumerIp, connectionUniqueId, metricsData)
```
nimrod-teich, line 568: **Blocking — `GRPCClientRateLimiter` is not goroutine-safe; concurrent subscribes crash the router.**

This call lands in `AllowSubscribe` (`grpc_streaming_config.go:101`), which reads and writes plain maps with no mutex — and it runs on every incoming gRPC stream's own handler goroutine. `UnsubscribeAll` calls `CleanupClient` (a bare `delete`) concurrently from stream teardown.

I reproduced it: a 50-goroutine probe calling `AllowSubscribe` with distinct keys under `-race` produces DATA RACE warnings and then `fatal error: concurrent map read and map write` — an unrecoverable runtime fault, not a caught panic. The PR's sharing tests pass `-race` only because they subscribe sequentially.

Pre-existing code, but this PR is what gives it production callers. Minimal fix is a mutex in the limiter. Alternatively, since the comment above documents that per-stream-unique client keys mean these limits never bind for gRPC anyway, the subscribe-side limiter call could just be dropped for this path.

protocol/rpcsmartrouter/direct_grpc_subscription_manager.go
```diff
@@ -456,14 +468,11 @@ func (dgm *DirectGRPCSubscriptionManager) joinExistingSubscription(
...
-		// Create acknowledgement
-		firstReply = dgm.createStreamAcknowledgement(routerID, sub.methodPath)
-	}
-
-	return firstReply, replyChan, nil
+	// The acknowledgement is minted per client, never shared. It carries the router
+	// id this client must quote to unsubscribe, and clientRouterIDs holds a distinct
+	// one per client — handing a joiner the creator's copy would tell it to
+	// unsubscribe using an id that is not its own.
+	return dgm.createStreamAcknowledgement(routerID, sub.methodPath), replyChan, nil
```
nimrod-teich: **Blocking — join-vs-teardown race: nil-map write panic, or a silent instant-EOF stream.**

`joinExistingSubscription` writes `sub.connectedClients[clientKey] = sender` (line 460) with no liveness check, while `cleanupSubscription` sets `connectedClients = nil` (line 892) after removing the map entry. `StartSubscription` grabs the sub reference under `RLock`, releases the lock, then joins — so when the last client leaves just as a new client subscribes to the same params (the sharing hot case):

- if cleanup completes before the joiner takes `sub.lock`, the join panics on assignment to a nil map and kills the process;
- if the joiner lands inside the cancel→cleanup window, cleanup closes its freshly-added sender and the client receives an instant clean EOF on a subscription — exactly the silent-truncation class this PR exists to eliminate.

Fix: under `sub.lock`, bail out to the create-new path (or retry) when `sub.cleanedUp` is set / `connectedClients` is nil / `sub.ctx.Err() != nil`.

protocol/chainlib/grpc.go
```diff
@@ -498,6 +515,141 @@ func (apil *GrpcChainListener) Serve(ctx context.Context, cmdFlags common.Consum
...
+
+		// Snapshot before the emit goroutine starts. AddMetricForGrpc writes Success and
+		// Origin on the struct it is handed, and the per-delivery emits work from these
+		// same fields for the life of the subscription — so each side needs its own
+		// copy. ConsumerWebsocketManager copies for the same reason.
+		subscriptionFields := *metricsData
+		go apil.logger.AddMetricForGrpc(metricsData, err, &metadataValues)
+		if err != nil {
+			apil.logger.LogRequestAndResponse("grpc stream in/out", true, method, string(reqBody), "", err.Error(), connectionUniqueId, time.Since(startTime), err)
+			return nil, utils.LavaFormatError("failed to start gRPC subscription", err,
```
nimrod-teich, line 578: **Unbounded map growth on failed `StartSubscription`.**

Client keys are unique per stream, and the only cleanup path is `UnsubscribeAll` via `StreamResponse.Close` — which runs only when a stream was actually established. Every *failed* subscribe therefore leaks two entries keyed by a GUID that will never be seen again:

- the rate-limiter entry `AllowSubscribe` creates before any other check (`grpc_streaming_config.go:104`);
- the sticky-session entry written in `createNewSubscription` (line 508) before stream creation can still fail.

With endpoints down and clients in a retry loop, both maps grow without bound. This error branch should release them (e.g. call `UnsubscribeAll`/`CleanupClient` + sticky delete for the key on failure).

protocol/chainlib/grpc.go
```diff
@@ -498,6 +515,141 @@ func (apil *GrpcChainListener) Serve(ctx context.Context, cmdFlags common.Consum
...
+				utils.LogAttr("_method", method),
+				utils.LogAttr("dappID", dappID),
+			)
+		}
+
+		return &grpcproxy.StreamResponse{
+			Replies: apil.forwardSubscriptionReplies(ctx, repliesChan, subscriptionFields, snapshotMetricsHeaders(metadataValues)),
+			// firstReply's payload is a JSON acknowledgement, which would not decode as
+			// the method's output type — only its subscription id is carried, as headers.
+			Metadata: convertRelayMetaDataToMDMetaData(firstReply.GetMetadata()),
```
nimrod-teich: **Minor — `content-type: application/json` leaks into the gRPC response headers.**

`createStreamAcknowledgement` puts `content-type` in the ack metadata, and this forwards the whole thing to `stream.SetHeader` — so a duplicate content-type header can go on the wire alongside the transport's own `application/grpc`. It evidently worked against live Sui, but strict clients or intermediary proxies may object. Suggest filtering that key (or only forwarding the subscription-id key, which is the one that matters here).

nimrod-teich (commented):
Reviewed the full diff, read `direct_grpc_subscription_manager.go` end-to-end on the branch, built it, and ran the new test files under `-race` (all green). The design is sound and the PR body checks out — I verified all five claimed defect fixes in code (manager-rooted stream context, `closeOnce` latch, unconditional first `SendMsg`, per-client ack minting, `dynamic.NewMessage` for both encodings), and the `SendNodeMsg` plumbing removal is genuinely mechanical with every call site and mock updated. The spec-first classification plus the two-tier backstop in `relayInnerDirect` is the right shape, and the `chain_router.go` ws-requirement scoping precisely fixes the Sui provider rejection.

However: making `StartSubscription` reachable for the first time also makes two dormant crash bugs in the surrounding manager code hot. I'd fix the two inline crashers (rate-limiter map race — reproduced as a fatal runtime error, and the join-vs-teardown race) before merge; the rest can ride or follow up.

One finding with no diff anchor:

**Latent lock-order inversion in `Unsubscribe`** (`direct_grpc_subscription_manager.go` ~line 923). `Unsubscribe` holds `dgm.lock.RLock` while acquiring `sub.lock`, whereas `joinExistingSubscription` / `removeClientFromSubscription` hold `sub.lock` while acquiring `dgm.lock` inside track/untrack. Today `Unsubscribe` has no production caller — gRPC has no unsubscribe frame, the listener only uses `UnsubscribeAll` — so it can't fire. But the first caller someone adds can deadlock the manager (an RWMutex with a pending writer blocks new readers). Suggest deleting the dead method or fixing the ordering while it's cheap.

Also worth knowing, not blocking: every request on a subscription-bearing gRPC chain is now parsed twice (stream callback, then unary path). Gated to spec-with-SUBSCRIBE chains and fine at current volumes.

protocol/chainlib/grpc.go
avitenzer: Fixed in 2090e6f — mutex in the limiter.

Took the mutex rather than dropping the subscribe-side call, because `AllowUnsubscribe` and `CleanupClient` race each other regardless of what the subscribe path does; removing one caller would have left the type unsafe. `Allow()` runs outside the lock — `rate.Limiter` has its own, and holding both would serialise every stream's admission behind one mutex.

Reproduced your finding before fixing: stripping the mutex again gives `WARNING: DATA RACE` on `mapaccess1_faststr` inside `AllowSubscribe`, from two goroutines in the same insert.

Also rewrote `TestGRPCClientRateLimiter_ConcurrentAccess`. It looped sequentially and carried a comment documenting the missing synchronisation as a known limitation — which is a fair part of why this survived to having callers. It now runs 50 goroutines with distinct keys, the shape the listener produces.

protocol/rpcsmartrouter/direct_grpc_subscription_manager.go
avitenzer: Fixed in bd3de2c — liveness check under `sub.lock`, bailing out to the create-new path.

Took the bail-out rather than a retry: retrying would have to loop against a subscription that is only getting deader, and the create path is where the client needs to end up anyway. `joinExistingSubscription` returns `errSubscriptionUnavailable` and `StartSubscription` falls through, so the race resolves into a working stream rather than an error the caller never asked for.

Three conditions, because teardown has three entry points that set different things first:

- `cleanedUp` — `cleanupSubscription` latches this before it touches anything else;
- `connectedClients == nil` — only true once cleanup's own critical section has run;
- `ctx.Err() != nil` — `removeClientFromSubscription` and `Stop` cancel without latching.

Two guards, verified red-first by reverting the check:

- `TestGRPCJoinExistingSubscription_RefusesSubscriptionBeingTornDown` — subtests for both states you named. Fails with `panic: assignment to entry in nil map` on the first, and a silently-successful join on the second.
- `TestGRPCStartSubscription_ReplacesSubscriptionThatDiedBeforeJoin` — the other half: being refused must cost the client nothing, so it opens its own stream and receives on it.

protocol/chainlib/grpc.go
avitenzer: Fixed in bd3de2c — released in the manager rather than at the listener, since the manager owns both maps and every caller then gets it.

`StartSubscription` now runs `releaseClientStateIfIdle` on any failure, via a named-return `defer` so a failure path added later gets it without having to remember. It drops the sticky entry and the limiter entry, but only while the client holds no subscription at all — a client with a live stream keeps its affinity. Deliberately narrower than a full `UnsubscribeAll`, which would also tear down that client's other subscriptions.

One placement note: the defer sits *after* the rate-limit check, not before. A rate rejection is the limiter remembering that the client just spent a token, so releasing there would let a retry loop reset its own limit. That strands nothing on this path — with a per-stream key the limiter is always fresh and always admits its first call, so the branch cannot fire for gRPC at all.

`TestGRPCStartSubscription_ReleasesClientStateOnFailure` stages the failure at the `IsServerStreaming` check, which is the last thing `createNewSubscription` does before the stream exists and is past the point where the sticky entry has been written. Reverting the defer gives 5 stranded limiter entries from 5 failed subscribes.

protocol/chainlib/grpc.go
avitenzer: Fixed in a238bee — filtered, via a new `streamResponseHeaders`.

Went with a filter rather than forwarding only the subscription-id key, so a header added to the acknowledgement later still reaches the client instead of being silently swallowed. Filtered at the wire boundary rather than at `createStreamAcknowledgement`, because that is where the constraint actually lives: `content-type: application/json` is true of the ack payload and only wrong once it is put on a response whose payload is never sent.

The deny-list is the headers the HTTP/2 gRPC transport owns — `content-type`, `content-length`, `grpc-status`, `grpc-message`, `grpc-encoding`, `grpc-accept-encoding`, `te`. It also lowercases keys, which `metadata.MD` expects and `convertRelayMetaDataToMDMetaData` was not doing.

Two guards: `TestStreamRelayCallback_DropsTransportOwnedHeaders` at the callback level with the exact metadata the manager mints, and `TestStreamResponseHeaders_FiltersReservedKeys` on the filter itself so the set stays honest. Reverting the call gives `Should be empty, but was [application/json]`.

avitenzer:
Thanks — all five addressed, pushed as `2090e6f`, `bd3de2c`, `a238bee`. The four inline threads have replies and are resolved; this covers the one with no diff anchor.

## Lock-order inversion in `Unsubscribe`

Fixed the ordering rather than deleting the method, in `bd3de2c`.

Fixing the ordering turned out to be the smaller change, because the convention already existed and only two functions were breaking it: `cleanupSubscription` was carefully finishing its `sub.lock` section before calling `untrackClientSubscription`, with a comment saying exactly why. So `dgm.lock` outer is the established order, and `Unsubscribe` was the one already following it — `joinExistingSubscription` and `removeClientFromSubscription` were the violators. Both now finish their `sub.lock` section before calling `track`/`untrackClientSubscription`, and the invariant is written down on the `lock` field rather than left implicit in one function's comment.

That keeps `Unsubscribe` as the single-subscription counterpart to `UnsubscribeAll` and makes it safe for whoever adds the first caller, which was your concern. Its doc comment now says it has no production caller and why.

`TestGRPCManager_JoinAndUnsubscribeRespectLockOrder` drives both directions concurrently. A deadlock has no assertion of its own, so the workload runs in its own goroutine and a 30s timeout is the failure — the fixed ordering finishes in milliseconds. Reintroducing the inversion in the join path deadlocks it reliably and it fails at the guard instead of hanging the run.

## On the double parse

Noted, not changed. Agreed it is a perf observation rather than a defect, and it is gated to spec-with-`SUBSCRIBE` chains.

## Verification

Every guard was verified red-first by mutating the implementation back:

| Guard | Under mutation |
|---|---|
| `TestGRPCClientRateLimiter_ConcurrentAccess` | `WARNING: DATA RACE` in `AllowSubscribe` — your repro |
| `TestGRPCJoinExistingSubscription_RefusesSubscriptionBeingTornDown` | `panic: assignment to entry in nil map`; silent join on the second subtest |
| `TestGRPCStartSubscription_ReplacesSubscriptionThatDiedBeforeJoin` | same panic, through `StartSubscription` |
| `TestGRPCStartSubscription_ReleasesClientStateOnFailure` | 5 stranded limiter entries from 5 failed subscribes |
| `TestGRPCManager_JoinAndUnsubscribeRespectLockOrder` | deadlocks; caught at the 30s guard |
| `TestStreamRelayCallback_DropsTransportOwnedHeaders` | `Should be empty, but was [application/json]` |

`./protocol/rpcsmartrouter/...` and `./protocol/chainlib/...` are green under `-race`, run from a detached worktree at `a238bee` rather than the working directory.

nimrod-teich (approved):

PR #293, opened 2026-08-18
fix(smart-router): consult grpc-config's descriptor-source on every request path

avitenzer:
Jira ticket: MAG-2350

## Problem

`grpc-config`'s descriptor escape hatch — `descriptor-source: reflection|file|hybrid` plus `descriptor-set-path` — has been in the config schema with constants, a validator and docs, but **no request path ever read it**. Every gRPC descriptor lookup resolved through live server reflection.

The result is worse than a missing feature: `descriptor-source: file` passes validation at connect time and is then silently ignored on every request. A config that looks supported and does nothing is harder to diagnose than one that isn't there.

This blocks any chain whose nodes don't serve reflection for the services its spec uses. **Concordium** is the case that forced it — its public gateways expose reflection for `Health` only, so the router exits at boot with `Symbol not found: concordium.v2.Queries`.

## This is not Concordium-specific

There is no chain branching anywhere in the affected code. 32 specs declare a gRPC interface (~26 concrete chains once shared base specs are excluded), and every one of them depends on upstream reflection with no fallback.

`config/smartrouter_examples/smartrouter_cosmos.yml` already documents the milder form of the same failure in its own header: public Cosmos gateways *throttle* reflection, so the gRPC startup check degrades to best-effort. Same root cause, different trigger. Sui is unaffected today only because its mainnet node happens to serve full reflection — an upstream operational choice, not a router guarantee.

## The fix

Two selectors, one per descriptor abstraction, both memoized by path:

| Helper | Interface | Drives |
|---|---|---|
| `rpcInterfaceMessages.DescriptorSourceForNode` | `grpcurl.DescriptorSource` | request/response marshalling |
| `dyncodec.ProtoFileRegistryForNode` | `ProtoFileRegistry` | `dynamicResolve` → `GetParams` → block parsing |

Five reflection-only sites now select through them:

| # | Site | Path |
|---|---|---|
| 1 | `chainlib/grpc.go` `SendNodeMsg` | boot validation, verifications, chain fetcher |
| 2 | `direct_rpc_connection.go` `getMethodDescriptor` | unary relay |
| 3 | `direct_rpc_connection.go` `parseInputMessage` | JSON request parsing |
| 4 | `upstream_grpc_pool.go` `GetMethodDescriptor` | streaming |
| 5 | `chainlib/grpc.go` `setupForProvider` | parser registry |

**Site 5 is not in the ticket.** Without it, the first four let a reflection-less node boot clean and then fail at *parse* time on binary-proto requests — a worse failure than the loud one it replaces.

`newGrpcChainProxy` also never populated `NodeUrl` on its `BaseChainProxy` literal, so `cp.NodeUrl.GrpcConfig` was always the zero value and the config was invisible on path 1 no matter what the operator wrote.

## Usage

```yaml
node-urls:
  - url: "grpcs://grpc.mainnet.concordium.software:20000"
    auth-config: { use-tls: true }
    grpc-config:
      descriptor-source: "file"                        # or "hybrid"
      descriptor-set-path: "config/concordium.protoset"
```

All of a chain's gRPC node-urls must agree on `descriptor-source` and `descriptor-set-path` — see decision 4.

## Behaviour when `grpc-config` is omitted

Descriptor resolution is unchanged: reflection mode returns the server source untouched, and the reflection-mode tests assert identity (`require.Same`), not equivalence.

One exception, which an earlier revision of this description got wrong by claiming "byte-for-byte". Populating `NodeUrl` on the proxy also makes `BaseChainProxy.CapTimeoutForSend` → `NodeUrl.LowerContextTimeout` see a per-node-url `timeout:` on the gRPC provider path, where it previously always read the zero value. This is desirable — it matches jsonrpc/rest — but it is a behaviour change for any gRPC node-url that already sets `timeout:`, independent of `grpc-config`.

## Decisions worth reviewer attention

**1. `hybrid` is file-first, not reflection-first.** `GrpcConfig`'s doc comment claimed reflection-first; the ticket's acceptance criterion says file-then-reflection. I implemented file-first and corrected the comment. The deployments that need hybrid are precisely those whose reflection is absent, partial or rate-limited — asking the node first pays that cost on *every* lookup before arriving at an answer that was on local disk the whole time.

**2. Caching required a `Close()` guard.** `FileDescriptorSetRegistry.Close` latches the registry shut, and every later lookup returns `registry is closed`. Once one parsed registry is shared by every proxy naming that path, an honoured `Close` from any single consumer would poison the rest. `sharedFileRegistry` makes it a no-op — the data is immutable and process-lifetime, so there is nothing to release. `TestSharedFileRegistry_CloseDoesNotPoisonOtherConsumers` is the regression guard.

**3. A descriptor source that cannot resolve is fatal, not advisory.** `direct_rpc_connection.go` used to warn `"will use reflection"` and continue. That was accurate only *before* this PR: in `file` mode sites 2 and 3 now resolve through the same loader and return the same error, and `LoadProtoset` caches failures — so there is no fallback and nothing recovers. Continuing bought one WARN promising a fallback that does not exist, then total relay failure on that endpoint. The check also moved ahead of the dial, since it needs no connection; a config that can never work now fails before spending a dial budget. It cannot fire for the reflection default — `DescriptorSourceForGrpcConfig` only errors for `file` and for an unrecognised `descriptor-source`, both operator config — and the test pins that, because it is what makes fatality safe.

**4. gRPC node-urls on one chain that disagree are rejected at boot.** `registry`/`codec` live on the parser, which is per *chain*; `grpc-config` is per *node-url*. `newChainRouter` builds one proxy per batch entry and passes the same parser to each, so every `setupForProvider` call overwrites the last — and the batch is a `map`, so "last" is whatever Go's randomized iteration order picks that boot. That was benign while each iteration built an equivalent reflection registry; with site 5 wired, a chain with one `file` node-url and one reflection node-url parses blocks against a different descriptor set on each restart. Rejecting beats warning: a warning still leaves the router serving a randomly chosen descriptor set. Nothing that works today is affected, since `grpc-config` was never read before this PR. Normalization happens before the comparison, so an omitted `descriptor-source` and an explicit `"reflection"` do not conflict, and neither do `allow-insecure` / `reflection-timeout`.

**5. Hybrid's descriptor-set half is consulted strictly.** `FileDescriptorSetRegistry.ProtoFileContainingSymbol` walks up the parent chain and, on a miss, returns the *ancestor's* file with a nil error. Since `compositeProtoFileRegistry` falls back on error alone, reflection was never consulted and `Registry.FindDescriptorByName` returned NotFound — at exactly the staleness case hybrid exists to cover. The composite now confirms the descriptor-set half declares the symbol itself, via the registry's existing `HasSymbol` index. The live trigger is a **new rpc**, not a rare nested type: `dynamicResolve` passes a method full name (`grpcMessage.go:93`), so `pkg.Query.Balance` resolves through `pkg.Query` whenever a node gains an rpc after the protoset was cut. Strictness costs the protoset no hits — every symbol kind it contains is directly indexed, so the walk only ever fires on a genuine miss.

Also note `initializeDescriptorSource` is a net **deletion**: it had reimplemented mode dispatch locally, validated with a different parser than the request path used, and discarded the result — so a passing check proved nothing about what would actually be consumed. It now runs through the shared loader, which makes the boot check predictive *and* warms the cache.

## Testing

- **52 tests**, race-clean, no `protoc`/`buf` dependency — fixtures hand-build `FileDescriptorSet` protos in Go
- Covers mode selection, error paths (missing/unreadable/corrupt), cache identity under concurrent first-touch, hybrid fallback and reflection outage, the shared-`Close` regression, and **call-counter assertions that `file` mode never queries reflection at all**
- The three tests added for decisions 3–5 were each run against a deliberately broken version of their own fix, to confirm they catch the bug rather than merely pass
- `go build ./...`, `go vet`, and the `chainlib` / `common` / `lavasession` / `rpcsmartrouter` suites all pass
- `gofmt`/`gofumpt` clean

## Not in this PR

- **No Concordium protoset.** That belongs in `lava-specs` next to `config/concordium.yml`; this PR is the router-side unblock.
- **No live-gateway verification.** Booting `config/concordium.yml` against `grpc.mainnet.concordium.software:20000` remains the acceptance test and needs a reviewer with network access to that endpoint.
- **The protoset cache still latches failures.** Raised in review and deliberately deferred pending a scope call. Decision 3 makes the failure loud instead of silent, but a transient first-touch failure — a protoset not yet mounted, an `EMFILE`, an NFS blip — still poisons that path for the process. Caching successes but not failures keeps the perf win without the latch. Note `083605d`, two commits below this branch point, is the same anti-pattern one layer up.
- **`GrpcConfig.Validate()` has zero callers repo-wide**, so an invalid `descriptor-source` string is only caught at request time (decision 3 makes that fatal). Wiring the validator in is a separate change.
- **Two advisory `SA1019` lint warnings** for `jhump/protoreflect/desc` in the new files — unavoidable, since `grpcurl.DescriptorSource` is built on it. There are already 7 identical ones on `main`, and `.golangci.yml` excludes the sibling files by path. A targeted exclusion mirroring the existing `golang/protobuf/proto` rule would clear all 9; I left shared lint config alone. Lint is advisory on this repo per `lint.yml`.

### DIFF
--- a/config/smartrouter_examples/smartrouter_cosmos.yml
+++ b/config/smartrouter_examples/smartrouter_cosmos.yml
@@ -17,6 +17,23 @@
 # check is best-effort here. A real Cosmos Hub node (or a gateway that allows
 # full reflection) verifies gRPC cleanly; swap one in for production.
 #
+# If you must stay on a gateway that throttles or omits reflection, point the
+# gRPC endpoint at a pre-built descriptor set instead — descriptors then resolve
+# from disk and reflection is never queried:
+#
+#   node-urls:
+#     - url: "grpcs://cosmos-grpc.publicnode.com:443"
+#       grpc-config:
+#         descriptor-source: "file"     # or "hybrid" to fall back to reflection
+#         descriptor-set-path: "config/cosmoshub.protoset"
+#
+# Build the protoset from the chain's proto tree (either tool works):
+#   buf build -o cosmoshub.protoset
+#   protoc -I . --include_imports --descriptor_set_out=cosmoshub.protoset <root>.proto
+#
+# It is a static snapshot: rebuild it when a node upgrade adds rpcs or fields the
+# spec starts using, or those methods fail with "symbol not found" in file mode.
+#
 # Run it:
 #   smartrouter config/smartrouter_examples/smartrouter_cosmos.yml --use-static-spec specs/
 
--- a/config/smartrouter_examples/smartrouter_cosmos.yml
+++ b/config/smartrouter_examples/smartrouter_cosmos.yml
@@ -34,6 +34,11 @@
 # It is a static snapshot: rebuild it when a node upgrade adds rpcs or fields the
 # spec starts using, or those methods fail with "symbol not found" in file mode.
 #
+# All of a chain's gRPC node-urls must agree on descriptor-source and
+# descriptor-set-path. Block parsing resolves through one registry per chain, not
+# one per node-url, so a chain whose node-urls disagree is rejected at boot rather
+# than served by whichever one happened to be applied last.
+#
 # Run it:
 #   smartrouter config/smartrouter_examples/smartrouter_cosmos.yml --use-static-spec specs/
 
--- a/protocol/chainlib/chainproxy/rpcInterfaceMessages/fileDescriptorSource.go
+++ b/protocol/chainlib/chainproxy/rpcInterfaceMessages/fileDescriptorSource.go
@@ -1,0 +1,203 @@
+package rpcInterfaceMessages
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/fullstorydev/grpcurl"
+	"github.com/jhump/protoreflect/desc"
+	"github.com/magma-Devs/smart-router/protocol/common"
+	"github.com/magma-Devs/smart-router/utils"
+)
+
+// protosetCache memoizes parsed descriptor sets by path.
+//
+// A protoset is a static artifact of the config — the Concordium one is ~335KB —
+// and descriptor lookups happen per connection, per uncached method. Without this
+// the same file is read and unmarshalled on every miss for the life of the process.
+//
+// Failures are cached alongside successes deliberately: a path that is missing or
+// corrupt at boot stays that way, and re-reading it on every relay turns one config
+// mistake into per-request filesystem traffic. Editing a protoset requires a router
+// restart, which is already true of every other field in the config.
+var protosetCache sync.Map // map[string]*protosetEntry
+
+type protosetEntry struct {
+	once   sync.Once
+	source grpcurl.DescriptorSource
+	err    error
+}
+
+// LoadProtoset returns the descriptor source backed by the protoset at path,
+// parsing it at most once per path per process.
+func LoadProtoset(path string) (grpcurl.DescriptorSource, error) {
+	cached, _ := protosetCache.LoadOrStore(path, &protosetEntry{})
+	entry, ok := cached.(*protosetEntry)
+	if !ok {
+		return nil, utils.LavaFormatError("corrupt protoset cache entry", nil,
+			utils.LogAttr("descriptor-set-path", path))
+	}
+
+	entry.once.Do(func() {
+		source, err := grpcurl.DescriptorSourceFromProtoSets(path)
+		if err != nil {
+			entry.err = utils.LavaFormatError("failed loading gRPC descriptor set", err,
+				utils.LogAttr("descriptor-set-path", path))
+			return
+		}
+		entry.source = source
+		utils.LavaFormatDebug("loaded gRPC descriptor set",
+			utils.LogAttr("descriptor-set-path", path))
+	})
+
+	return entry.source, entry.err
+}
+
+// ResetProtosetCacheForTest drops every memoized protoset. Tests that write a
+// descriptor set to a temp path and expect it to be re-read must call this;
+// production never does.
+func ResetProtosetCacheForTest() {
+	protosetCache.Range(func(key, _ any) bool {
+		protosetCache.Delete(key)
+		return true
+	})
+}
+
+// compositeDescriptorSource resolves symbols against a protoset first and falls
+// back to server reflection. This is "hybrid" mode.
+//
+// File-first, not reflection-first, is deliberate. The chains that need hybrid at
+// all are the ones whose upstream reflection is absent, partial, or rate-limited —
+// asking the node first would pay that cost on every lookup only to arrive at an
+// answer that was on local disk the whole time. File-first makes the common path
+// free and leaves reflection as what it actually is here: a supplement for symbols
+// the protoset predates.
+type compositeDescriptorSource struct {
+	file   grpcurl.DescriptorSource
+	server grpcurl.DescriptorSource
+}
+
+var _ grpcurl.DescriptorSource = compositeDescriptorSource{}
+
+func (c compositeDescriptorSource) ListServices() ([]string, error) {
+	services, err := c.file.ListServices()
+	if err != nil {
+		return c.server.ListServices()
+	}
+
+	// A reflection failure is not fatal in hybrid mode — the protoset already
+	// answered. Report what we have rather than losing it to the node being down.
+	serverServices, serverErr := c.server.ListServices()
+	if serverErr != nil {
+		utils.LavaFormatDebug("hybrid descriptor source: reflection ListServices failed, using protoset only",
+			utils.LogAttr("error", serverErr))
+		return services, nil
+	}
+
+	seen := make(map[string]struct{}, len(services)+len(serverServices))
+	merged := make([]string, 0, len(services)+len(serverServices))
+	for _, service := range append(services, serverServices...) {
+		if _, duplicate := seen[service]; duplicate {
+			continue
+		}
+		seen[service] = struct{}{}
+		merged = append(merged, service)
+	}
+	return merged, nil
+}
+
+func (c compositeDescriptorSource) FindSymbol(fullyQualifiedName string) (desc.Descriptor, error) {
+	descriptor, fileErr := c.file.FindSymbol(fullyQualifiedName)
+	if fileErr == nil {
+		return descriptor, nil
+	}
+
+	descriptor, serverErr := c.server.FindSymbol(fullyQualifiedName)
+	if serverErr == nil {
+		return descriptor, nil
+	}
+
+	// Both halves missed. Surface both causes: in practice one of them is the
+	// actionable one (stale protoset vs. node not serving reflection) and which
+	// it is depends on the deployment.
+	return nil, utils.LavaFormatError("symbol not found in protoset or via reflection", nil,
+		utils.LogAttr("symbol", fullyQualifiedName),
+		utils.LogAttr("protoset_error", fileErr),
+		utils.LogAttr("reflection_error", serverErr))
+}
+
+func (c compositeDescriptorSource) AllExtensionsForType(typeName string) ([]*desc.FieldDescriptor, error) {
+	fileExtensions, fileErr := c.file.AllExtensionsForType(typeName)
+	serverExtensions, serverErr := c.server.AllExtensionsForType(typeName)
+
+	if fileErr != nil && serverErr != nil {
+		return nil, fileErr
+	}
+
+	seen := make(map[string]struct{}, len(fileExtensions)+len(serverExtensions))
+	merged := make([]*desc.FieldDescriptor, 0, len(fileExtensions)+len(serverExtensions))
+	for _, extension := range append(fileExtensions, serverExtensions...) {
+		name := extension.GetFullyQualifiedName()
+		if _, duplicate := seen[name]; duplicate {
+			continue
+		}
+		seen[name] = struct{}{}
+		merged = append(merged, extension)
+	}
+	return merged, nil
+}
+
+// DescriptorSourceForNode picks the descriptor source implied by a node's
+// grpc-config. serverSource is the reflection-backed source the caller already
+// holds; in "file" mode it is never consulted, so a node that does not serve
+// reflection is never asked.
+//
+// This is chain-agnostic on purpose: nothing here knows which chain it is serving.
+// Concordium is the case that forced it (its gateways serve reflection for Health
+// only), but public Cosmos gateways that throttle reflection hit the same wall
+// through a different door.
+func DescriptorSourceForNode(mode, descriptorSetPath string, serverSource grpcurl.DescriptorSource) (grpcurl.DescriptorSource, error) {
+	switch mode {
+	case common.GrpcDescriptorSourceFile:
+		if descriptorSetPath == "" {
+			return nil, utils.LavaFormatError("descriptor-set-path is required for descriptor-source: file", nil)
+		}
+		return LoadProtoset(descriptorSetPath)
+
+	case common.GrpcDescriptorSourceHybrid:
+		if descriptorSetPath == "" {
+			// Config validation rejects this, but a hybrid node with no protoset is
+			// still meaningfully reflection-only rather than a hard failure.
+			return serverSource, nil
+		}
+		fileSource, err := LoadProtoset(descriptorSetPath)
+		if err != nil {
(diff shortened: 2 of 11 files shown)


AnnaR-prog:
## Review

Fetched the branch (single commit `91038a9` on `origin/main@4127703`) and verified the claims: `go build ./...` and `go vet` on the four touched trees are clean, the new tests pass under `-race`, and `./protocol/chainlib/... ./protocol/common/... ./protocol/lavasession/...` pass.

The core diagnosis is right and the two-selector shape is the correct one — `grpcurl.DescriptorSource` and `dyncodec.ProtoFileRegistry` really are unrelated abstractions, and wiring only the first would have produced the boot-clean/parse-fail split the commit message describes. Reflection mode is identity-preserving (`require.Same`, not equivalence), which is the right way to test "default behaviour unchanged". Three things I'd want addressed before merge.

---

### 1. `initializeDescriptorSource` is still non-fatal, and its warning is now false

`protocol/lavasession/direct_rpc_connection.go:908`

```go
if err := g.initializeDescriptorSource(); err != nil {
    // Log warning but don't fail - we'll try reflection on first request
    utils.LavaFormatWarning("failed to initialize descriptor source, will use reflection", err, ...)
}
```

That comment and message were true *before* this PR — resolution went through reflection on every path regardless. They are false after it. In `file` mode there is no reflection fallback: `getMethodDescriptor` and `parseInputMessage` both route through `DescriptorSourceForGrpcConfig`, which returns the same error. And because `LoadProtoset` caches failures by design, the failure is latched for the process lifetime.

Net effect of a typo'd `descriptor-set-path` (or a typo'd `descriptor-source`, see item 4): one WARN at boot claiming a fallback that does not exist, then 100% relay failure on that endpoint with no further signal. That is the "config that looks supported and does nothing" failure the PR description opens by condemning — moved, not removed. Either fail `initialize` for `file`/invalid modes, or at minimum stop promising reflection.

### 2. `setupForProvider` is last-wins on a per-chain parser, but `descriptor-source` is per-node-url

`protocol/chainlib/grpc.go:136` / `protocol/chainlib/chain_router.go:88`

`newChainRouter` loops the node-url batch calling `proxyConstructor(ctx, nConns, entry, chainParser)` with **the same parser**, and each `newGrpcChainProxy` calls `setupForProvider`, which overwrites `apip.registry`/`apip.codec`. Before this PR every iteration produced an equivalent reflection registry, so last-wins was benign. Site 5 makes the registry depend on that node-url's `grpc-config`.

So a chain with two gRPC node-urls — a base URL plus an archive-extension URL, say — where only one carries `descriptor-source: file` ends up with whichever landed last in the batch driving block parsing for the whole chain. This is the site added beyond the ticket, and it is the one that lands on a per-chain field. Worth either a warning when a batch has heterogeneous `grpc-config`s, or an explicit doc line that descriptor-source must match across a chain's gRPC node-urls.

### 3. Hybrid's file-first fallback is dead for nested symbols — dyncodec half only (confirmed empirically)

`FileDescriptorSetRegistry.ProtoFileContainingSymbol` (`protocol/chainlib/grpcproxy/dyncodec/remote_file.go`, the parent-walk loop) strips parent components off the end until it hits *any* indexed symbol. For `file.v1.Outer.Added`, where the set has `file.v1.Outer` but not `Added`, it returns `chain.proto` with **nil error**. `compositeProtoFileRegistry.ProtoFileContainingSymbol` only falls back on error, so reflection is never asked. Confirmed with a throwaway test in the package:

```
err=<nil> file="chain.proto" reflectionCalls=0
```

Downstream, `Registry.FindDescriptorByName` registers that file, re-looks-up, and returns NotFound — so the operator sees "not found" exactly where hybrid was supposed to fall back to reflection. That is the staleness case the new `DescriptorSetPath` doc comment describes ("a node upgrade adds rpcs or fields the spec starts using").

The grpcurl half is fine — `fileSource.FindSymbol` is strict, so `compositeDescriptorSource` falls back correctly. The asymmetry between the two halves is the finding.

Blast radius is bounded: in this codebase only message/extension full names reach `ProtoFileContainingSymbol` (via `FindMessageByName` / `FindMessageByURL` / `FindExtensionByName`), not method names — so the trigger is a new nested type inside a message the protoset already knows, not a new rpc. The existing test misses it because `TestProtoFileRegistryForNode_HybridPrefersFileThenFallsBack` probes `server.v1.ServerOnly`, whose entire prefix chain is absent, so the walk fails cleanly. Cheapest fix: have the composite confirm the returned file actually contains the symbol before accepting it.

---

### Non-blocking

**`GrpcConfig.Validate()` has zero callers repo-wide.** So the comment at `fileDescriptorSource.go:169` — "Config validation rejects this" — is false, and an invalid `descriptor-source` string is only caught at request time. It also softens the "constants, a validator and docs" framing in the description: the validator is dead code, which is arguably part of why the feature was never wired.

**Populating `NodeUrl` has a second, unmentioned effect.** `BaseChainProxy.CapTimeoutForSend` calls `NodeUrl.LowerContextTimeout`, which now honours a per-node-url `timeout:` on the gRPC provider path, where it was previously always the zero value. Probably desirable — it matches jsonrpc/rest — and the other `GetChainProxyInformation` consumer is log attributes only (`chain_fetcher.go`). But it means "omit `grpc-config` and behaviour is byte-for-byte what it is today" is not quite true; this is the exception. Separately, `NewGrpcChainProxy` trims a trailing `/` before dialing but `cp.NodeUrl.Url` keeps it, so logs can now disagree with the dialed URL.

**The latched failure cache.** Deliberate and documented, but "editing a protoset requires a router restart" conflates an edit with a transient read failure. A protoset not yet mounted, an EMFILE, or an NFS blip on first touch poisons that path permanently — and `initialize()` is a reconnect path, so re-init will not recover it. Caching successes but not failures keeps the perf win without the latch.

**Hybrid's `ListServices`/`AllExtensionsForType` always hit reflection** (they union both halves), which cuts against the "don't pay reflection cost on every lookup" rationale used to justify file-first, and on a truly reflection-less node makes a doomed round trip each call. Union semantics are defensible for `ListServices` — worth a comment saying why the short-circuit does not apply there. Minor: `AllExtensionsForType` drops `serverErr` when both fail, while `FindSymbol` reports both.

**Nits.** `ResetProtosetCacheForTest`/`ResetFileRegistryCacheForTest` are exported from production files — `export_test.go` would keep them off the public surface. Both cache tests call `require.NoError` inside goroutines; testify's `require` calls `t.FailNow()`, which is documented as unsafe off the test goroutine.

### Checked and clean

- Closing the unused reflection remote in `file` mode is safe: `GRPCReflectionProtoFileRegistry.Close()` is a no-op, so it cannot tear down the pooled `reflectionConnection` that `newGrpcChainProxy`'s teardown goroutine still owns.
- Wrapping in `compositeDescriptorSource` loses grpcurl's `sourceWithFiles` fast path, but `grpcurl.GetAllFiles` is never called in this repo, so it is inert.
- `DescriptorSourceForGrpcConfig(grpcConfig, nil)` passing a nil server source does not panic in any mode — there is a test for it.
- The `initializeDescriptorSource` net-deletion argument holds: the old version validated with dyncodec's parser while the request path used grpcurl's, so a pass genuinely proved nothing.

avitenzer:
Thanks — all three are fixed in 6697083. Verification notes below, including one place where I think the review under-scoped its own finding.

### 1. `initializeDescriptorSource` — now fatal, and moved ahead of the dial

Confirmed exactly as described. The one addition: `setupForProvider` already hard-fails on the same bad path (`newGrpcChainProxy` returns the error and boot dies), so the two paths disagreed about whether a bad `descriptor-set-path` is fatal. Aligning `initialize` with what site 5 already does made this a consistency fix rather than a judgment call.

I also moved the check ahead of the connector construction. It needs no connection — it passes a nil server source — so resolving first means a config that can never work fails before spending a dial budget, and `initialize()`'s "every error path leaves no connector installed" invariant is sidestepped rather than unwound.

The test pins the vacuity argument, since that is what makes fatality safe: `{}`, explicit `reflection`, `reflection` + timeout, and both hybrid shapes all return nil. The new failure cannot fire for any config in production today.

### 2. `setupForProvider` — rejected at boot, not warned

Confirmed, and it is worse than last-wins. `rpcProviderEndpointBatch` is a `map[string]lavasession.RPCProviderEndpoint` and `chain_router.go:251` ranges over it directly, so the winner is not last-in-config-order — Go randomizes map iteration, making it **a different winner per process start**. That pushed me to the stronger of your two options: a warning still leaves the router parsing blocks against a randomly chosen descriptor set, and a config that behaves differently every restart is not one anyone can operate. Rejection is the only outcome that is the same every boot.

Nothing that works today is affected, since `grpc-config` was never read before this PR — anything that would now trip the guard was already silently ignored. Normalization happens before the comparison, so an omitted `descriptor-source` and an explicit `reflection` do not conflict, and neither do `allow-insecure` / `reflection-timeout`. `cloneForValidation` starts its record at nil on purpose: the re-verification loop picks its own node-url subset, so an inherited record could reject a set the clone was never given.

I took the doc line as well as the guard — `smartrouter_cosmos.yml` now states the constraint next to the protoset instructions.

Checked while I was in there: `setRouterKeyInBatch` builds every entry with exactly one node-url (`chain_router.go:236`) and everything downstream reads `NodeUrls[0]`, so there is no second `grpc-config` hiding inside an entry that the guard would miss.

### 3. Hybrid ancestor match — fixed, and the blast radius is bigger than stated

The diagnosis and the recommended fix are both right, and the composite now confirms the descriptor-set half declares the symbol itself via the registry's existing `HasSymbol` index.

One correction. The review bounds it with "only message/extension full names reach `ProtoFileContainingSymbol` … not method names — so the trigger is a new nested type inside a message the protoset already knows, not a new rpc." That does not hold: `grpcMessage.go:93` is `dynamicResolve`, which calls `FindDescriptorByName(strings.ReplaceAll(gm.Path, "/", "."))` and casts to `protoreflect.MethodDescriptor` — a **method** full name, and it is the caller site 5 exists to serve. Reproduced against a protoset declaring `hybrid.v1.Query` with a single rpc:

```
METHOD   err=<nil> file="chain.proto" reflectionCalls=0   # hybrid.v1.Query.Balance
NESTED   err=<nil> file="chain.proto" reflectionCalls=0   # hybrid.v1.PingRequest.Added
CONTROL  err=<nil> file="reflected.proto" reflectionCalls=1
```

So the trigger is a new **rpc** on a service the protoset already knows — the ordinary staleness case `DescriptorSetPath`'s doc comment warns about, not the rarer nested-type one. Same fix, more reason to want it.

Worth knowing for anyone reading the fix: `indexFileSymbols` already indexes nested messages, nested enums, extensions, services *and* methods, so every symbol kind that reaches this path is directly indexed. The parent walk only ever fires on a genuine miss, which is why strictness costs the protoset no hits — the test pins that services, methods, messages and the leading-dot form all still resolve locally with reflection untouched. `ProtoFileByPath` is an exact map lookup with no walk, so it never had this shape.

### Testing

Three regression tests, each run against a deliberately broken version of its own fix to confirm it catches the bug rather than merely passing: the dyncodec one on the ancestor-match assertions, the lavasession one on `dials == 0`, the chainlib one on the hybrid→reflection direction the loader would not have caught by itself. `go build ./...`, `go vet`, and `chainlib/... lavasession/... common/... rpcsmartrouter/...` pass under `-race`; `gofmt`/`gofumpt` clean.

### Non-blocking, untouched

All verified accurate, including `GrpcConfig.Validate()` having zero callers and `CapTimeoutForSend` being live on this path at `grpc.go:636`. Two notes:

- **The latched failure cache is the one I would promote out of non-blocking.** `083605d`, two commits below this branch point, is literally *"retry a failed gRPC initialization instead of latching it"* — this PR reintroduces the same anti-pattern one layer down. Combined with item 1 the two compound: a protoset not yet mounted at boot gave one WARN and permanent failure. Item 1 makes that loud, but caching successes and not failures is still the right shape. Happy to do it here or as a follow-up — your call on scope.
- **Trailing slash** is real but narrower than it reads: `HashedNodeUrl` was already computed from the untrimmed URL before this PR, so only the `GetChainProxyInformation` log attributes are newly affected.

The PR description needs a pass too — the "byte-for-byte" claim, and the new boot-time rejection is not described anywhere. Doing that next.

AnnaR-prog (approved):

PR #294, opened 2026-08-18
fix(smart-router): scope gRPC connector pool capacity per instance, and pin MAG-2538 parser isolation

avitenzer:
Jira ticket: MAG-2538

Rebased onto current `main` (`e9f7c66`). The `grpc_test.go` conflict was both sides appending to the end of the file; both are kept — main's `TestSetupForProvider_RejectsConflictingDescriptorSource` and this PR's `TestCloneChainParserForValidation_AbsorbsRealSetupForProvider`. No descriptor-source / `grpc-config` behaviour was dropped.

> **Scope note.** An earlier revision of this description said "tests only, no production code changes". That is no longer true, and the change that made it untrue is the more important half of the PR — hence `fix(...)`, not `test(...)`. See §2.

## 1. Why the test half exists

The router fix for MAG-2538 already landed, with MAG-2525. `retryFailedProviders` routes verification through `retryValidateFn` → `validateProvider`, and `validateProvider` clones via `CloneChainParserForValidation` before handing the parser to `GetChainRouter`. Verified rather than assumed: there is **no** `GetChainRouter` call left in `rpcsmartrouter.go`, and the only non-test caller of `CloneChainParserForValidation` is `spec_reverifier.go:429`.

The ticket was reopened because the regression test was never written.

`TestCloneChainParserForValidation_GrpcIsolation` stands in for the mutation by assigning `registry`/`codec` directly, so it proves the helper returns a distinct struct but not that a clone absorbs what `newGrpcChainProxy` does to it — and nothing at all about whether callers still clone. A caller dropping the clone is exactly what the outage was.

**`TestRevalidateTier_DoesNotRebindLiveGrpcParser`** (`protocol/rpcsmartrouter`) runs the production recovery pass — `revalidateTier` → `retryValidateFn` → `validateProvider`, with `validateProvider`'s own `attemptCtx` cancelled on the way out — and asserts the live parser is still bound to its serving registry two ways:

1. **pointer identity** — `require.Same` on the registry `ParseMsg` copies onto every message it builds;
2. **behaviour** — `GetParams` → `dynamicResolve`, the *only* consumer of that registry and what block parsing runs through. Against a registry left bound to recovery's torn-down reflection connection it logs `failed to dynamicResolve` and returns nil.

The second is not redundant with the first: it also catches a rebind to a different-but-equally-dead registry, which pointer identity would flag only by accident. The registry is observed through exported API only — no unexported access and no new test-only exports.

**`TestCloneChainParserForValidation_AbsorbsRealSetupForProvider`** (`protocol/chainlib`) covers the helper against the real mutation, so a field added to `cloneForValidation` that aliases mutable state is caught in the package that owns it.

Both end with a control asserting the same call **without** the clone *does* rebind the live parser. That is load-bearing: the ticket records that the first attempt at this test (in the superseded #287) pointed at an unreachable address, `NewGRPCConnector` errored before `newGrpcChainProxy` ran, and it passed against fixed **and** broken code.

## 2. The production change: pool capacity moves onto the connector

Running the recovery test under `-race` surfaced a real defect, not a test artifact:

```
WARNING: DATA RACE
Write at 0x000107484b20 by goroutine 28:
  chainproxy.NewGRPCConnector()            connector.go:214
Previous read at 0x000107484b20 by goroutine 57:
  chainproxy.(*GRPCConnector).ReturnRpc()  connector.go:408
```

`NumberOfParallelConnections` was both configuration and state. `NewConnector` and `NewGRPCConnector` each wrote it with their own `nConns`, and `GRPCConnector.ReturnRpc` read it back as *"the number we started with"*. Two connectors alive at once broke that in both directions:

- **Silent misbehaviour** — the second constructor redefined the first's idle-trim threshold, process-wide. Across differing `nConns` the wrong pool gets trimmed. An HTTP connector could reset a gRPC connector's threshold despite never reading it.
- **Data race** — the write and the read are unsynchronized.

MAG-2538's recovery path is one way in (`validateProvider` tears its verification connector down while the next is constructed) but nothing about the hazard is specific to recovery; any two overlapping connectors reach it.

`GRPCConnector` now owns a `capacity` field, written once in `NewGRPCConnector` before publication and read by `ReturnRpc`. Neither constructor mutates the global, which is now **`const`** — so "configuration, not state" is a compiler guarantee rather than a convention, and no future caller can reintroduce either constructor's write. Its one remaining use is the default for `--parallel-connections` at `testing.go:207`.

Three chainproxy tests pin it, all of which **fail against the old model** — two on assertions, the third with three data races:

| Test | Pins |
|---|---|
| `TestGRPCConnectorCapacityIsPerInstance` | capacity survives a later, larger connector |
| `TestGRPCConnectorReturnRpcTrimsAgainstOwnCapacity` | same pool state, two capacities, opposite trim decisions |
| `TestConnectorConstructionDoesNotRaceReturnRpc` | both constructors concurrent with `ReturnRpc`, race-free |

### Live behaviour change

`validateProvider` passes `100` (`DefaultMaximumStreamsOverASingleConnection`, `spec_reverifier.go:431`), while direct-RPC builds its connectors with a hardcoded `nConns=10` on the **request** path (`direct_rpc_connection.go:887`). Under the old model a serving connector's trim threshold was whichever constructor ran most recently, so it oscillated between 10 and 100 depending on relay/retry interleaving. After this PR every connector trims against its own `nConns`.

That is the intended behaviour and the practical delta is small — a 10-connection pool rarely exceeds `used + 10` free — but it is a live change, so it is called out here rather than left implicit.

## Verification

All clean, run from a `git worktree` checkout of the committed sha:

```
gofmt -l <touched files>
go build ./...
go vet ./protocol/chainlib/... ./protocol/rpcsmartrouter/... ./protocol/lavasession/...
git diff --check
go test -race ./protocol/chainlib/chainproxy ./protocol/chainlib ./protocol/rpcsmartrouter -count=1
go test -race ./protocol/lavasession -count=1
go test -race ./protocol/chainlib/chainproxy -count=3 -shuffle=on
go test -race ./protocol/rpcsmartrouter -run '^TestRevalidateTier_DoesNotRebindLiveGrpcParser$' -count=25
```

Deliberate broken-implementation checks still fail as intended:

| Revert | Fails |
|---|---|
| `validationParser := chainParser` in `validateProvider` | `TestRevalidateTier_DoesNotRebindLiveGrpcParser` — at `require.Same`, and independently at the `GetParams` probe with `require.Same` disabled |
| `CloneChainParserForValidation` returns its argument | `TestCloneChainParserForValidation_AbsorbsRealSetupForProvider` |
| `NumberOfParallelConnections = nConns` in either constructor | does not compile: `cannot assign to NumberOfParallelConnections` |

The one-line import reordering in `grpc_auth_connector_test.go` is `gofmt` output on a file that was already unformatted on main.

## Not in this PR

The end-to-end repro the ticket describes — let the 3-minute retry timer fire, then serve a live gRPC relay through **external ingress** — is still blocked on MAG-2477 (gRPC ingress) or MAG-2093 (sim-aware gRPC test client). The assertions in §1 are not that test: they drive the in-process recovery path directly and do not exercise the timer or ingress. Not worked around; the gate is real.

### DIFF
--- a/protocol/chainlib/chainproxy/connector.go
+++ b/protocol/chainlib/chainproxy/connector.go
@@ -36,7 +36,19 @@ const (
 	MaxCallRecvMsgSize                         = 1024 * 1024 * 512 // setting receive size to 512mb for large debug responses
 )
 
-var NumberOfParallelConnections uint = 10
+// NumberOfParallelConnections is the DEFAULT pool size — the default value of the
+// --parallel-connections flag. It is configuration, not state, and const is what
+// makes that a guarantee rather than a convention: no future caller can reintroduce
+// the write below, because the assignment no longer compiles.
+//
+// It used to be both. NewConnector and NewGRPCConnector each wrote it with their own
+// nConns, and GRPCConnector.ReturnRpc read it back as "the number we started with",
+// so a connector's idle-trim threshold was whichever connector happened to be built
+// last, process-wide. With one connector that is invisible; with two it is a data
+// race (MAG-2538 tears a verification connector down while the next is constructed),
+// and across differing nConns it trims the wrong pool. Capacity belongs to a
+// connector, so it now lives on the connector.
+const NumberOfParallelConnections uint = 10
 
 // Connector manages HTTP/JSON-RPC connections to blockchain nodes.
 // For HTTP connections, a single shared rpcclient.Client is used since
@@ -66,7 +78,11 @@ func HashURL(url string) string {
 // pooling internally). The parameter is kept for API compatibility with
 // gRPC connector which still uses connection pooling.
 func NewConnector(ctx context.Context, nConns uint, nodeUrl common.NodeUrl) (*Connector, error) {
-	NumberOfParallelConnections = nConns // used by gRPC connector, ignored for HTTP
+	// nConns is deliberately not recorded anywhere: HTTP pools inside http.Transport.
+	// This used to assign NumberOfParallelConnections "for the gRPC connector", which
+	// made building an HTTP connector silently reset an unrelated gRPC connector's
+	// idle-trim threshold — and race its ReturnRpc. That constant is now const, so
+	// the assignment cannot come back.
 	connector := &Connector{
 		nodeUrl:       nodeUrl,
 		hashedNodeUrl: HashURL(nodeUrl.Url),
@@ -178,6 +194,15 @@ type GRPCConnector struct {
 	credentials credentials.TransportCredentials
 	nodeUrl     common.NodeUrl
 
+	// capacity is the pool size this connector was built with — the nConns its
+	// caller asked for, not whatever the last-constructed connector asked for.
+	// ReturnRpc trims idle clients against it.
+	//
+	// Written once in NewGRPCConnector before the connector is published and never
+	// again, so readers need no synchronization beyond the happens-before that
+	// publishing already gives them.
+	capacity uint
+
 	// closed is set by Close and never cleared. Unlike the HTTP Connector's atomic
 	// flag it is guarded by lock, because every reader already holds lock: that
 	// makes "is the pool alive" and "take a client" a single atomic step, which a
@@ -211,10 +236,10 @@ var ErrGRPCConnectorClosed = errors.New("grpc connector is closed")
 // request — the direct-RPC path — must NOT: passing the request's context as
 // lifetimeCtx tears the pool down the moment that request returns.
 func NewGRPCConnector(lifetimeCtx, dialCtx context.Context, nConns uint, nodeUrl common.NodeUrl) (*GRPCConnector, error) {
-	NumberOfParallelConnections = nConns // set number of parallel connections requested by user (or default.)
 	connector := &GRPCConnector{
 		freeClients: make([]*grpc.ClientConn, 0, nConns),
 		nodeUrl:     nodeUrl,
+		capacity:    nConns,
 	}
 
 	// MAG-2218: warn once per connector, not per dial. createConnection may still upgrade an
@@ -405,7 +430,7 @@ func (connector *GRPCConnector) ReturnRpc(rpc *grpc.ClientConn) {
 		rpc.Close()
 		return
 	}
-	if len(connector.freeClients) > (int(connector.usedClients) + int(NumberOfParallelConnections) /* the number we started with */) {
+	if len(connector.freeClients) > (int(connector.usedClients) + int(connector.capacity) /* the number THIS connector started with */) {
 		rpc.Close() // close connection
 		return      // return without appending back to decrease idle connections
 	}
--- a/protocol/chainlib/chainproxy/connector_capacity_test.go
+++ b/protocol/chainlib/chainproxy/connector_capacity_test.go
@@ -1,0 +1,155 @@
+package chainproxy
+
+import (
+	"context"
+	"sync"
+	"testing"
+
+	"github.com/magma-Devs/smart-router/protocol/common"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// Pool capacity used to live in the package-global NumberOfParallelConnections,
+// which every constructor wrote and GRPCConnector.ReturnRpc read back. Two
+// connectors existing at once was enough to break that in both directions: the
+// second constructor silently redefined the first's idle-trim threshold, and the
+// write raced the read outright. MAG-2538's recovery test is what surfaced it —
+// validateProvider tears its verification connector down while the next connector
+// is being built — but nothing about the hazard is specific to that path.
+
+// TestGRPCConnectorCapacityIsPerInstance pins the ownership half: a connector's
+// trim threshold is the nConns IT was built with, and building another connector
+// does not move it.
+func TestGRPCConnectorCapacityIsPerInstance(t *testing.T) {
+	server := createGRPCServer(t)
+	defer server.Stop()
+
+	ctx := context.Background()
+	url := common.NodeUrl{Url: listenerAddressGrpc}
+
+	small, err := NewGRPCConnector(ctx, ctx, 1, url)
+	require.NoError(t, err)
+	defer small.Close()
+
+	// Built second and larger: under the global this reassigned the threshold that
+	// `small` had already been constructed with.
+	big, err := NewGRPCConnector(ctx, ctx, numberOfClients, url)
+	require.NoError(t, err)
+	defer big.Close()
+
+	require.Equal(t, uint(1), small.capacity, "the first connector must keep its own capacity")
+	require.Equal(t, uint(numberOfClients), big.capacity)
+	// Nothing here asserts NumberOfParallelConnections survived construction: it is a
+	// const, so "no constructor rewrites it" is enforced by the compiler for every
+	// caller rather than by one test for the two constructors it happens to exercise.
+}
+
+// TestGRPCConnectorReturnRpcTrimsAgainstOwnCapacity is the same ownership property
+// asserted through behaviour rather than the field: identical pool state, two
+// capacities, opposite decisions about whether a returned client is worth keeping.
+func TestGRPCConnectorReturnRpcTrimsAgainstOwnCapacity(t *testing.T) {
+	ctx := context.Background()
+
+	for _, tt := range []struct {
+		name     string
+		capacity uint
+		wantFree int
+	}{
+		// freeClients already holds 2 and the borrow count drops to 0 on return, so
+		// the surplus test is `2 > 0 + capacity`.
+		{name: "capacity below the surplus trims", capacity: 1, wantFree: 2},
+		{name: "capacity above the surplus keeps", capacity: 5, wantFree: 3},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			// A real client to hand back, since the trim path closes it. Dialled
+			// standalone rather than borrowed from a connector: a borrow this test
+			// never returns would leave that connector's Close waiting on it forever.
+			returned, err := grpc.DialContext(ctx, listenerAddressGrpc,
+				grpc.WithTransportCredentials(insecure.NewCredentials()))
+			require.NoError(t, err)
+			t.Cleanup(func() { _ = returned.Close() })
+
+			// Hand-built so pool state is identical across the two cases and capacity
+			// is the only variable. Only len(freeClients) feeds the decision, so the
+			// placeholders are never dereferenced.
+			connector := &GRPCConnector{
+				capacity:    tt.capacity,
+				usedClients: 1,
+				freeClients: []*grpc.ClientConn{nil, nil},
+			}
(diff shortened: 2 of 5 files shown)


AnnaR-prog:
## Review: approve with comments

Verified locally at `8e0755a`. The production change is correct, and all three of the "fails against old code" claims in the description reproduce. One substantive finding.

### Verification run

- `go build ./...`, `go vet` clean; `gofmt -l` clean on all five touched files
- `go test -race` green on `./protocol/chainlib/...`, `./protocol/rpcsmartrouter/...`, and `./protocol/lavasession/...` (the other `NewGRPCConnector` caller)
- `go test -race ./protocol/chainlib/chainproxy -count=2 -shuffle=on` — ok. Worth noting since `connector_capacity_test.go` sorts *before* `connector_test.go`, so the new tests now run first and are what rebinds the package-global `listenerAddressGrpc` for the rest of the package. No fallout.
- Both heavy tests stable at `-count=10 -race`

Mutation testing — all three claims hold:

| Mutation | Result |
|---|---|
| capacity → global (both ctor writes + `ReturnRpc`) | `TestGRPCConnectorCapacityIsPerInstance` FAIL (:44); `.../capacity_below_the_surplus_trims` FAIL (:86); `TestConnectorConstructionDoesNotRaceReturnRpc` → **3× `WARNING: DATA RACE`** + FAIL at `-count=5` |
| `validationParser := chainParser` | `TestRevalidateTier_DoesNotRebindLiveGrpcParser` FAIL (:121) |
| `CloneChainParserForValidation` returns its arg | `TestCloneChainParserForValidation_AbsorbsRealSetupForProvider` FAIL (:519) |

---

### 1. The serving-path assertion is inert — it cannot see the defect it is described as catching

`provider_recovery_parser_isolation_test.go:129` is presented as the symptom half ("relays on the recovered provider hit a nil connector... asserting the pointer alone would pin the cause without ever checking the symptom").

It cannot check that. `GrpcChainProxy.SendNodeMsg` (`protocol/chainlib/grpc.go:565`) never reads `GrpcMessage.Registry`. It uses `nodeMessage.Path`, `.GetHeaders()`, `.Msg`, and builds its descriptor source from its **own** connector — `cp.conn.GetRpc(ctx, true)` plus `cp.NodeUrl.GrpcConfig` (`grpc.go:597-598`). The parser's registry is on a different path entirely.

Proof: neutralise the `require.Same` at :121, then remove the clone from `validateProvider`. `SendNodeMsg` still succeeds and the test **passes against the broken code**.

The genuine symptom path is `Registry` → `dynamicResolve()` → `GetParams()` (`grpcMessage.go:77,92`), which block parsing depends on. That one does discriminate:

| | probe after recovery |
|---|---|
| fixed | `GetParams=map[]` |
| clone removed | `ERR failed to dynamicResolve error="proto file containing symbol: EOF"` → `GetParams=<nil>` |

Suggested replacement (or addition):

```go
require.NotNil(t, craftGetBlockMessage(t, chainParser).GetRPCMessage().GetParams(),
    "MAG-2538: the live parser's registry must still resolve after recovery")
```

A `ChainFetcher.FetchLatestBlockNum` against `servingRouter` would cover both halves if you prefer something closer to production. Keeping `SendNodeMsg` as a smoke check is fine — it's the claim about what it covers that needs to go.

This does not weaken the regression pin: `require.Same` at :121 catches the mutation on its own.

**Related** — the same evidence contradicts the stated failure mode. Both the PR body and the test comment say production hit a *nil connector*; what the probe actually shows is descriptor resolution failing through a registry whose reflection remote is dead. That framing is presumably what made `SendNodeMsg` look like the right assertion, and root-cause narratives get copied into the next ticket, so worth reconciling.

### 2. The HTTP half of the hazard isn't pinned

The description argues the HTTP constructor writing the global was half the bug ("An HTTP connector could reset a gRPC connector's threshold despite never reading it"). But re-adding `NumberOfParallelConnections = nConns` to `NewConnector` alone leaves the whole `chainproxy` package `ok` — `TestGRPCConnectorCapacityIsPerInstance` builds only gRPC connectors, and `TestConnectorConstructionDoesNotRaceReturnRpc` builds HTTP ones but asserts only `serving.capacity`.

Harmless now that the global is inert, so a suggestion rather than a defect: one extra line asserting the global at the end of the race test closes it for both constructors.

### 3. `NumberOfParallelConnections` can be `const` now

Nothing assigns it anymore, and `Flags().Uint` at `testing.go:207` accepts a typed const. That turns the new doc comment's "configuration, not state" into a compiler guarantee instead of a convention — which seems like the whole point of the change.

### 4. A live behaviour change worth naming in the description

`validateProvider` passes `100` (`DefaultMaximumStreamsOverASingleConnection`, `spec_reverifier.go:431`), while direct-RPC builds its connectors with a hardcoded `nConns=10` on the **request** path (`direct_rpc_connection.go:887`). Under the old model a serving connector's trim threshold was whichever constructor ran most recently, so it oscillated between 10 and 100 depending on relay/retry interleaving. After this PR every connector trims against its own `nConns`.

That's the intended behaviour and the practical delta is small (a 10-conn pool rarely exceeds `used + 10` free), but it is a live change currently riding under a `test(...)` commit.

### 5. Minor

- **Commit type.** `test(smart-router):` on a commit that contains a production data-race fix. Given the generated changelogs in this repo (`chore(release): add changelog for v1.3.2`), the connector fix won't appear in release notes. Suggest `fix(...)`, or splitting the connector change into its own commit — the description owns the scope drift in prose, but the commit type doesn't.
- `connector_capacity_test.go:44` — `require.Equal(t, NumberOfParallelConnections, uint(10), ...)` has expected/actual reversed, so failure output reads backwards. Capturing the value at test start and asserting it unchanged would also survive someone retuning the default.
- `TestGRPCConnectorReturnRpcTrimsAgainstOwnCapacity` starts a real gRPC server but dials without `grpc.WithBlock()` and never sends on the conn, so the server isn't needed. Dropping it removes a dependency and one more rebinding of `listenerAddressGrpc`.

- No `GetChainRouter` left in `rpcsmartrouter.go`, as claimed. The other two callers don't need the clone: `health_cmd.go:431` builds its own parser at :367, and `testing.go:71` is the dev command.
- The new chainlib test is genuinely stronger than the adjacent `"validation clone does not inherit the record"` subtest, which passes a nil reflection connection and only asserts `NotNil(live.registry)`. Real `setupForProvider` via `GetChainRouter`, `Same` on registry *and* codec, plus a control — no meaningful overlap.
- `grpc_auth_connector_test.go` was indeed unformatted on `main`; the reorder is real `gofmt` output.

avitenzer:
## All five points resolved — `8e0755a` → `2d5c3c7`

Thanks for the mutation testing; every finding held up when I reproduced it. Detail below, since two of the resolutions differ from what was suggested.

> ⚠️ **History note.** The branch was amended and force-pushed rather than extended, so your reviewed sha `8e0755a` is no longer the tip and the line references in your review (`:44`, `:121`, `:129`) have shifted. `git diff 8e0755a..2d5c3c7` is 31 insertions / 20 deletions across three files if you want to re-review just the delta.

### 1. The serving-path assertion was inert — confirmed, fixed

Reproduced exactly as described: with `require.Same` neutralised and the clone removed from `validateProvider`, the test passed (`ok ... 1.283s`). `SendNodeMsg` builds its descriptor source from the ChainProxy's own connector and never reads `GrpcMessage.Registry`, so it could not have failed for the stated reason. Worth adding that `craftGetBlockMessage` builds with `Msg: []byte{}`, so `SendNodeMsg` skips message formatting altogether — it reads only `.Path` and `.GetHeaders()`.

Your suggested probe is now the regression assertion, and it discriminates on its own:

```
ERR failed to dynamicResolve error="proto file containing symbol: rpc error:
    code = Canceled desc = grpc: the client connection is closing"
--- FAIL: TestRevalidateTier_DoesNotRebindLiveGrpcParser
    Messages: MAG-2538: the live parser's registry must still resolve descriptors after recovery
```

(You saw `EOF`; I get `Canceled` — same failure, teardown timing.)

`SendNodeMsg` stays, demoted to a smoke check with a comment recording *why* it cannot observe this defect, so it doesn't get re-promoted later.

One caveat I'd flag on the framing, since it affects how the description now reads: the `GetParams` probe observes the *same* binding as `require.Same`, just behaviourally rather than by pointer. It's strictly stronger (it also catches a rebind to a different-but-equally-dead registry), but it isn't an independent "symptom" layer — so I've dropped the cause/symptom claim from the description rather than restate it. `FetchLatestBlockNum` against `servingRouter` remains the honest way to get a genuinely independent layer; not done here.

**Related, on the failure mode:** corrected in both the test doc comment and the PR body. It was descriptor resolution through a dead reflection connection, not a nil connector.

### 2 + 3. Folded into one change — `const`

Confirmed #2: re-adding `NumberOfParallelConnections = nConns` to `NewConnector` alone left the whole package `ok` under `-race`.

Rather than add the suggested assertion, I took #3 — which subsumes it. `NumberOfParallelConnections` is now `const`, so that mutation no longer compiles:

```
connector.go:80:2: cannot assign to NumberOfParallelConnections
    (neither addressable nor a map index expression)
```

That covers every call site rather than the two constructors one test happens to exercise, and it's exactly the "compiler guarantee instead of a convention" framing you proposed. `go build ./...` and `go vet` are clean repo-wide; `Flags().Uint` at `testing.go:207` takes the typed const without complaint.

### 4. Live behaviour change — named

`spec_reverifier.go:431` passing `100` vs `direct_rpc_connection.go:887` hardcoding `10` now has its own subsection in the description, including that the old threshold oscillated between the two depending on relay/retry interleaving.

### 5. Minor

- **Commit type** — retyped `fix(...)`, and the PR title now matches. This turned out to matter more than cosmetically: the branch is a single commit ahead of `main`, and GitHub's squash-merge takes its subject from the commit message rather than the PR title in that case, so `test(...)` really would have reached the changelog.
- **`:44` argument order** — deleted rather than fixed. With the global `const`, asserting it still equals 10 is a tautology the compiler already enforces; there's a comment there now saying so.
- **Unused gRPC server** — dropped from `TestGRPCConnectorReturnRpcTrimsAgainstOwnCapacity`. Bonus: that test no longer depends on `listenerAddressGrpc` being bound at all, which closes the ordering coupling you noted in your verification run.

### Verification

From a `git worktree` checkout of `2d5c3c7` rather than the working directory:

- `go build ./...`, `go vet ./protocol/chainlib/... ./protocol/rpcsmartrouter/... ./protocol/lavasession/...` — clean
- `gofmt -l` clean on all three touched files
- `go test -race` green on `chainproxy`, `chainlib`, `rpcsmartrouter`, `lavasession`
- `-count=25` on the recovery test, `-count=3 -shuffle=on` on `chainproxy` — stable

All three original mutation claims still fail as intended, plus the new one:

| Revert | Fails |
|---|---|
| `validationParser := chainParser` | at `require.Same`, and independently at the `GetParams` probe with `require.Same` disabled |
| `CloneChainParserForValidation` returns its arg | `TestCloneChainParserForValidation_AbsorbsRealSetupForProvider` |
| `NumberOfParallelConnections = nConns` in either ctor | does not compile |

AnnaR-prog (approved):

PR #295, opened 2026-08-19
fix(smart-router): scope health-probe state per endpoint and normalize gRPC dial addresses (MAG-2333)

avitenzer:
Jira ticket: MAG-2333

Rebased onto `main` at `d03d1a4`. The connector pool-size commit that used to lead this branch has been **dropped**: #294 landed the same fix independently, naming the field `capacity` and making the global `const`, with three dedicated connector tests. Keeping both would have produced a struct carrying `capacity` *and* `nConns` — the two declarations sat in non-conflicting hunks, so a naive resolution wouldn't have shown it. `TestSkipWebsocketVerificationIsPerParser` still passes clean under `-race` on main's version.

## Problem

`smartrouter health` reported endpoints as unhealthy that the live router verifies and serves fine. The ticket recorded two symptoms; they turned out to be two independent defects.

**Bug A — `connector is closed` on every tendermintrpc URL.** `chainlib.SkipWebsocketVerification` was a process global that `health` flipped per endpoint under a mutex held around `ValidateCollect`. **That lock is the defect.** Holding it serialized every endpoint's verification phase while each endpoint's own `context.WithTimeout` kept running from goroutine launch. An endpoint late in the queue entered `ValidateCollect` with most of its budget already spent, its context expired mid-probe, and `connectorLoop` closed the connector out from under the remaining relays. The same premature expiry on the gRPC path surfaces as the reported `descriptorSource.FindSymbol DeadlineExceeded`, and it explains the ticket's flakiness signature — the archive-tagged leg of an identical URL passing while the plain leg failed in the same run.

> **Correction.** An earlier revision of this description also blamed a cross-endpoint race on the global via its second reader, `newChainRouter`. That is **not reachable under `health`**: the command sets `IgnoreWsEnforcementForTestCommands = true` (`health_cmd.go:115`) before any probing, and that guard is the first operand of the same `&&` as the `SkipWebsocketVerification()` call (`chain_router.go:328`), so Go short-circuits before the flag is ever evaluated there. Within `health` the only remaining reader is `getExtensionsForVerification`, reached from `ValidateCollect`, which is fully synchronous. Thanks to @AnnaR-prog for catching it.
>
> Moving the flag onto the parser is still the right change — it removes the shared cell instead of synchronizing it, and `newChainRouter` does read it on the serving path where the guard is `false`. But the lock's cost, not a torn read, is what users hit.

**Bug B — `descriptorSource.FindSymbol DeadlineExceeded` on gRPC.** `grpc.DialContext` takes `host:port`. `createConnection` stripped the `grpc://` / `grpcs://` prefix into a local `addr`, but `increaseNumberOfClients` — the on-demand pool refill — dialed `connector.nodeUrl.Url` raw. A scheme-prefixed endpoint opened its first connection and then never grew its pool again: `GetRpc` respawns a refill every 50ms while a caller waits, each dial failed on the unresolvable target, and the caller blocked until its context expired.

**Bug B is not health-specific.** `increaseNumberOfClients` is called from `GetRpc`, so pool refill was broken for every `grpc(s)://` endpoint on the serving path. `nConns=1` — what `health` builds — only turns "pool never grows" into "every relay hangs", because the single client is retained for reflection.

## The fix

| # | Commit | Change |
|---|---|---|
| 1 | `660f031` | `SkipWebsocketVerification` moves onto the parser; `health` already builds one parser per endpoint, and both readers already hold one. No shared cell, so no mutex and no serialization |
| 2 | `53ef5c2` | `grpcDialAddress` becomes the single normalizer for every gRPC dial site |
| 3 | `572b97b` | `backfillLatestBlock` reports a real height on legs that passed |
| 4 | `7555c80` | Review follow-ups: stop backfilling failed legs; correct the Bug A narrative in code |

Commit 3 addresses the ticket's "known related quirk": `ValidateCollect` fetches the latest block only when a verification needs it for a `LatestDistance`, so specs without one reported `latestBlock: 0` on passing legs.

Commit 4 closes a gap review found in commit 3: the fetch routes through the router built over **all** of an endpoint's probed URLs, so a row for URL A could be filled with a height obtained via URL B — and it filled failed rows too, printing a plausible height beside a leg's own failures. Only passing rows are backfilled now, and the value's endpoint-level provenance is documented on both the function and the JSON struct field rather than left implied.

## Live acceptance

Three-point A/B against public Akash endpoints. `config/akash.yml` from the ticket does not exist — lava-specs PR #42 only landed `akash.json` — so the config below was written to reproduce its shape, with the tendermintrpc entry deliberately mixing `https` + `wss` in one entry.

| Leg | `main` | + ws fix | **this branch** |
|---|---|---|---|
| tmrpc `https://akash-rpc.publicnode.com` | ❌ `connector is closed` | ✅ | ✅ |
| tmrpc `wss://akash-rpc.publicnode.com/websocket` | ❌ `connector is closed` | ✅ | ✅ |
| tmrpc `https://rpc.lavenderfive.com:443/akash` | ✅ | ✅ | ✅ |
| rest `https://rest.lavenderfive.com:443/akash` | ✅ | ✅ | ✅ |
| grpc `grpcs://akash.lavenderfive.com:443` | ❌ `waiting for a free client` | ❌ same | ✅ |
| **top-level** | **`ok:false`** | **`ok:false`** | **`ok:true`** |

Every leg green including the WSS and `grpcs://` legs, and every `latestBlock` positive:

```
TOP-LEVEL ok: True
  akash-tmrpc-publicnode     tendermintrpc  https://akash-rpc.publicnode.com          ok=True  block=28237143
  akash-tmrpc-publicnode     tendermintrpc  wss://akash-rpc.publicnode.com/websocket  ok=True  block=28237144
  akash-tmrpc-lavenderfive   tendermintrpc  https://rpc.lavenderfive.com:443/akash    ok=True  block=28237143
  akash-rest-lavenderfive    rest           https://rest.lavenderfive.com:443/akash   ok=True  block=28237143
  akash-grpc-lavenderfive    grpc           grpcs://akash.lavenderfive.com:443        ok=True  block=28237144
```

The isolated inline-triplet repro from the ticket reproduces and clears the same way. On `grpcs://akash.lavenderfive.com:443`, before: `ok:false`, both checks failing `descriptorSource.FindSymbol ErrMsg: rpc error: code = DeadlineExceeded` — the ticket's Bug B string verbatim. After: `ok:true`, `latestBlock: 28226553`.

<details>
<summary>Config used (self-contained, not committed)</summary>

```yaml
direct-rpc:
  - name: "akash-tmrpc-publicnode"
    chain-id: "AKASH"
    api-interface: "tendermintrpc"
    node-urls:
      - url: "https://akash-rpc.publicnode.com"
        timeout: 15s
      - url: "wss://akash-rpc.publicnode.com/websocket"
        timeout: 15s
  - name: "akash-tmrpc-lavenderfive"
    chain-id: "AKASH"
    api-interface: "tendermintrpc"
    node-urls:
      - url: "https://rpc.lavenderfive.com:443/akash"
        timeout: 15s
  - name: "akash-rest-lavenderfive"
    chain-id: "AKASH"
    api-interface: "rest"
    node-urls:
      - url: "https://rest.lavenderfive.com:443/akash"
        timeout: 15s
  - name: "akash-grpc-lavenderfive"
    chain-id: "AKASH"
    api-interface: "grpc"
    node-urls:
      - url: "grpcs://akash.lavenderfive.com:443"
        timeout: 15s
```

Run with `cd <dir-with-config> && smartrouter health akash_health.yml --use-static-spec <lava-specs> --timeout 60s` — the config argument resolves through viper's search paths, so it must be a bare filename.
</details>

## Testing

- `go build ./...`, `go vet ./protocol/...` clean
- `go test -race ./protocol/chainlib/... ./protocol/rpcsmartrouter/... ./protocol/lavasession/...` — all pass
- `gofmt` clean on every file this PR touches. `chain_router.go` and `chain_router_test.go` are listed by `gofmt -l`, but they already are on `main` (import ordering); left alone rather than fold an unrelated reorder into this diff.

New regression tests, each verified to fail on the unfixed code:

- `TestSkipWebsocketVerificationIsPerParser` — builds two routers concurrently from parsers with opposite settings, against an http-only endpoint and a subscribe-capable spec. Note this pins the per-parser shape on the **serving** path; under `health` that gate is short-circuited, so it is not a reproduction of the reported symptom.
- `TestGRPCConnectorRefillsPoolForSchemePrefixedURL` / `...ForTLSSchemePrefixedURL` — build a one-connection pool, borrow its only client, and require the async refill to produce another. Covers plaintext `grpc://` and a TLS reflection server behind `grpcs://`; both fail in 10s unfixed.
- `TestGrpcDialAddress` — both schemes, bare `host:port`, and a hostname merely containing `grpc`.
- `TestBackfillLatestBlock_*` — fills only missing values on **passing** rows, fetches at most once, skips failed rows, costs no fetch when only failed rows lack a height, and leaves rows untouched on fetch failure.

### Known test gap

The mechanism that actually fixes Bug A — removing the mutex that serialized `ValidateCollect` while per-endpoint timeouts ran down — has **no regression test**. Every health test is a pure-function unit test, and a faithful timing test needs the full probe machinery (real specs, routers, and two endpoints whose verification phases can be observed overlapping). A cheap seam at the `runHealthProbes` fan-out level would pin concurrency that was never broken and would pass against the old code, so none was added. Worth its own ticket. The live A/B above is the current evidence.

## Not addressed

- **`GrpcConfig.ReflectionTimeout` is still declared-but-unread.** Defined, defaulted to 5s and validated, with zero production readers. Wiring it would make reflection stricter than today's inherited relay timeout and could turn currently-passing slow gateways red, so it wants its own ticket.
- **`IgnoreWsEnforcementForTestCommands` is the same anti-pattern**, left in place — a package global in the very same `if`, written by `health_cmd.go` and `testing.go`. No race today, since both writes land before any goroutine starts, but the argument in this PR applies to it too. Follow-up.
- **`increaseNumberOfClients` discards `impliesTLS`** and keeps a second dial loop without the TLS-upgrade retry. Verified safe today only because `NewGRPCConnector` always runs `createConnection` first. The durable version is for it to call `createConnection`. Follow-up.
- **`grpcs://akash-grpc.publicnode.com:443` still reports `ok:false`, for an unrelated upstream reason.** It now reaches the node (`latestBlock: 28226584`) and fails on the response: allnodes blocks `/cosmos.base.tendermint.v1beta1.Service/GetNodeInfo` on their free tier. AKASH's `chain-id` verification calls that method, so the leg reads red regardless of router behaviour. It was green when the ticket was filed on 2026-07-05, so this is a tightening on their side.
- Worth a follow-up: the router surfaces that rejection as `proto: bad wiretype`, which reads like a parsing bug rather than an access-policy response.

### DIFF
--- a/protocol/chainlib/base_chain_parser.go
+++ b/protocol/chainlib/base_chain_parser.go
@@ -22,8 +22,20 @@ import (
 )
 
 var (
-	SkipWebsocketVerification = false
-	DefaultApiName            = "Default-"
+	// SkipWebsocketVerificationDefault seeds every parser built by NewChainParser and
+	// is bound to --skip-websocket-verification. It is written once during flag parsing,
+	// before any parser or goroutine exists, and is read-only from then on.
+	//
+	// It is deliberately NOT consulted at the point of use. The `health` command probes
+	// each direct-rpc entry concurrently and needs a different answer per entry (ws
+	// augmentation only routes for an entry that actually has a ws:// URL), so it used to
+	// flip a package global under a mutex around ValidateCollect. That missed the second
+	// reader — newChainRouter — which runs outside that mutex, so entries raced each other
+	// into the wrong ws enforcement and healthy legs reported red (MAG-2333). Per-parser
+	// state removes the shared cell entirely.
+	SkipWebsocketVerificationDefault = false
+
+	DefaultApiName = "Default-"
 )
 
 type PolicyInf interface {
@@ -52,6 +64,30 @@ type BaseChainParser struct {
 	allowedExtensions map[string]struct{}
 	extensionParser   extensionslib.ExtensionParser
 	active            bool
+
+	// skipWebsocketVerification is this parser's own answer to "should verifications be
+	// augmented with the websocket extension, and should the router enforce ws support".
+	// Seeded from SkipWebsocketVerificationDefault; overridden per-endpoint by callers
+	// that probe several endpoints concurrently. Guarded by rwLock like every other
+	// mutable field here.
+	skipWebsocketVerification bool
+}
+
+// SkipWebsocketVerification reports whether this parser's endpoint opts out of
+// websocket augmentation and enforcement.
+func (bcp *BaseChainParser) SkipWebsocketVerification() bool {
+	bcp.rwLock.RLock()
+	defer bcp.rwLock.RUnlock()
+	return bcp.skipWebsocketVerification
+}
+
+// SetSkipWebsocketVerification overrides the process default for this parser only.
+// Callers probing multiple endpoints concurrently must build one parser per endpoint
+// and set it here rather than reaching for the package-level default.
+func (bcp *BaseChainParser) SetSkipWebsocketVerification(skip bool) {
+	bcp.rwLock.Lock()
+	defer bcp.rwLock.Unlock()
+	bcp.skipWebsocketVerification = skip
 }
 
 func (bcp *BaseChainParser) Activate() {
--- a/protocol/chainlib/chain_fetcher.go
+++ b/protocol/chainlib/chain_fetcher.go
@@ -281,7 +281,7 @@ func getExtensionsForVerification(verification VerificationContainer, chainParse
 		ConnectionType: verification.ConnectionType,
 	}
 
-	if chainParser.IsTagInCollection(spectypes.FUNCTION_TAG_SUBSCRIBE, collectionKey) && !SkipWebsocketVerification {
+	if chainParser.IsTagInCollection(spectypes.FUNCTION_TAG_SUBSCRIBE, collectionKey) && !chainParser.SkipWebsocketVerification() {
 		if verification.Extension == "" {
 			extensions = []string{WebSocketExtension}
 		} else {
--- a/protocol/chainlib/chain_router.go
+++ b/protocol/chainlib/chain_router.go
@@ -325,7 +325,7 @@ func newChainRouter(ctx context.Context, nConns uint, rpcProviderEndpoint lavase
 			break
 		}
 	}
-	if !IgnoreWsEnforcementForTestCommands && hasSubscriptionInSpec && apiCollection.Enabled && !webSocketSupported && !SkipWebsocketVerification {
+	if !IgnoreWsEnforcementForTestCommands && hasSubscriptionInSpec && apiCollection.Enabled && !webSocketSupported && !chainParser.SkipWebsocketVerification() {
 		return nil, utils.LavaFormatError("subscriptions are applicable for this chain, but websocket is not provided in 'supported' map. By not setting ws/wss your provider wont be able to accept ws subscriptions, therefore might receive less rewards and lower QOS score.", nil,
 			utils.LogAttr("apiInterface", apiCollection.CollectionData.ApiInterface),
 			utils.LogAttr("supportedMap", supportedMap),
--- a/protocol/chainlib/chain_router_test.go
+++ b/protocol/chainlib/chain_router_test.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"os"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -387,6 +388,70 @@ func TestChainRouterWithDisabledWebSocketInSpec(t *testing.T) {
 			}
 		})
 	}
+}
+
+// TestSkipWebsocketVerificationIsPerParser pins the fix for MAG-2333.
+//
+// newChainRouter's websocket-enforcement branch used to read a package global that
+// `smartrouter health` flipped per endpoint while probing endpoints concurrently. Two
+// endpoints wanting opposite answers raced, and whichever wrote last decided for both —
+// so healthy legs reported red. The flag now lives on the parser and health builds one
+// parser per endpoint, so concurrent routers must each get the answer they asked for.
+//
+// The endpoint is deliberately http-only against a subscribe-capable spec: that is the
+// exact shape that trips enforcement, and the shape the existing ws tests never reach
+// because they always pair an http URL with a ws URL.
+func TestSkipWebsocketVerificationIsPerParser(t *testing.T) {
+	newSubscribeParser := func(skip bool) ChainParser {
+		t.Helper()
+		parser, err := NewChainParser(spectypes.APIInterfaceJsonRPC)
+		require.NoError(t, err)
+
+		spec := CreateMockSpec()
+		spec.ApiCollections = []*spectypes.ApiCollection{{
+			Enabled:        true,
+			CollectionData: spectypes.CollectionData{ApiInterface: spectypes.APIInterfaceJsonRPC},
+			ParseDirectives: []*spectypes.ParseDirective{{
+				FunctionTag: spectypes.FUNCTION_TAG_SUBSCRIBE,
+			}},
+		}}
+		parser.SetSpec(spec)
+		parser.SetSkipWebsocketVerification(skip)
+		return parser
+	}
+
+	httpOnlyEndpoint := func() *lavasession.RPCProviderEndpoint {
+		return &lavasession.RPCProviderEndpoint{
+			ChainID:      CreateMockSpec().Index,
+			ApiInterface: spectypes.APIInterfaceJsonRPC,
+			NodeUrls:     []common.NodeUrl{{Url: listenerAddressHttp}},
+		}
+	}
+
+	skipping := newSubscribeParser(true)
+	enforcing := newSubscribeParser(false)
+
+	ctx := context.Background()
+	var wg sync.WaitGroup
+	var skipErr, enforceErr error
+	wg.Add(2)
+	go func() {
+		defer wg.Done()
+		_, skipErr = GetChainRouter(ctx, 1, httpOnlyEndpoint(), skipping)
+	}()
+	go func() {
+		defer wg.Done()
+		_, enforceErr = GetChainRouter(ctx, 1, httpOnlyEndpoint(), enforcing)
+	}()
+	wg.Wait()
+
+	require.NoError(t, skipErr, "a parser that opted out of ws verification must accept an http-only endpoint")
+	require.Error(t, enforceErr, "a parser that did not opt out must still refuse an http-only endpoint for a subscribe-capable spec")
+
+	// Neither router run may have leaked into the other parser or into the process default.
+	require.True(t, skipping.SkipWebsocketVerification())
+	require.False(t, enforcing.SkipWebsocketVerification())
+	require.False(t, SkipWebsocketVerificationDefault, "GetChainRouter must not write the package default")
 }
 
 func TestChainRouterWithEnabledWebSocketInSpec(t *testing.T) {
--- a/protocol/chainlib/chainlib.go
+++ b/protocol/chainlib/chainlib.go
@@ -85,6 +85,8 @@ type ChainParser interface {
 	ExtractDataFromRequest(*http.Request) (url string, data string, connectionType string, metadata []pairingtypes.Metadata, err error)
 	SetResponseFromRelayResult(*common.RelayResult) (*http.Response, error)
 	ParseDirectiveEnabled() bool
+	SkipWebsocketVerification() bool
+	SetSkipWebsocketVerification(skip bool)
 }
 
 // CloneChainParserForValidation returns a parser instance safe to pass into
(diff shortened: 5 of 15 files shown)


AnnaR-prog:
## Review: approve with comments — but this collides with #294 and needs a merge decision first

Verified locally at `962b8f0`, with mutation testing on each new test.

### Verification run

- `go build ./...`, `go vet ./protocol/...` clean
- `go test -race` green on `./protocol/chainlib/...`, `./protocol/rpcsmartrouter/...`, `./protocol/lavasession/...`
- `go test -race ./protocol/chainlib -count=2 -shuffle=on` — ok. Worth running because the new `chain_router_test.go` case ends with `require.False(t, SkipWebsocketVerificationDefault, ...)`; nothing in the package writes it, so the assertion is order-independent.
- `gofmt -l` flags only `chain_router.go` and `chain_router_test.go`, and both are already unformatted on `main` (import ordering). Confirmed — the description's claim holds.

Mutation testing, one per new test:

| Mutation | Result |
|---|---|
| revert `0565490` alone | `TestSkipWebsocketVerificationIsPerParser` → **`WARNING: DATA RACE`** + FAIL |
| `SkipWebsocketVerification()` returns the package default | same test FAILs on `"a parser that opted out of ws verification must accept an http-only endpoint"` |
| `increaseNumberOfClients` dials the raw URL again | `TestGRPCConnectorRefillsPoolForSchemePrefixedURL` and `...ForTLSSchemePrefixedURL` both FAIL (10s each, as described) |

---

### 1. Commit `0565490` is the same change as #294, and a naive merge keeps both fields

#294 makes `NumberOfParallelConnections` per-connector too — same fix, field named `capacity` there and `nConns` here. Both are open against the same base.

`git merge-tree` between the two branches shows that only `ReturnRpc` and the two comment blocks conflict. The **field declarations land in separate, non-conflicting hunks** — #294 adds `capacity` after `nodeUrl`, this PR adds `nConns` after `closed` — so a careless resolution produces a struct carrying both, with `capacity: nConns` *and* `nConns: nConns` in the constructor and one of them permanently dead:

```
272:+	nConns uint
281: 		capacity:    nConns,
283:+		nConns:      nConns,
```

Suggestion: drop commit 1 from this PR and rebase on #294. The coverage there is stronger — three dedicated tests, including `TestGRPCConnectorReturnRpcTrimsAgainstOwnCapacity`, which pins the trim threshold behaviourally rather than only pinning the race.

To be fair to this PR: the commit message's claim of coverage checks out. Reverting `0565490` alone *does* trip `-race` in `TestSkipWebsocketVerificationIsPerParser`, exactly as it says. But that coverage is incidental — it comes from that test happening to build two HTTP connectors concurrently, it pins the race and not the threshold, and it would silently disappear if the test's endpoint or interface ever changed.

### 2. Bug A's websocket-enforcement mechanism is unreachable in `health`

`health_cmd.go:115` sets `chainlib.IgnoreWsEnforcementForTestCommands = true` at the top of `RunE`, before any probing — and has done since the command's first commit (`5b5679b`, 2026-06-23), which predates the ticket. `chain_router.go:328` is:

```go
if !IgnoreWsEnforcementForTestCommands && hasSubscriptionInSpec && apiCollection.Enabled && !webSocketSupported && !chainParser.SkipWebsocketVerification() {
```

`&&` short-circuits on the first operand, so under `health` that branch is unreachable and the ws flag is never even *evaluated* in `newChainRouter`.

That leaves exactly two accesses in the health command, and both were already inside `wsVerificationMu`: the write in `probeProvider`, and the read in `getExtensionsForVerification` — whose only caller path is `ValidateCollect`, a fully synchronous loop with no goroutines. So there was no cross-endpoint race on that bool in `health`.

The *other* mechanism in the description — the mutex serializing every endpoint's `ValidateCollect` while each endpoint's own `context.WithTimeout` kept running, so late endpoints entered with their budget spent and `connectorLoop` closed the connector mid-probe — fully explains the `connector is closed` symptom, and removing the mutex is what fixes it.

The change itself is still right: it's correct hardening, and the unsynchronized global genuinely matters on the serving path. It's the causal story that needs a correction, in the description and in the new `probeProvider` comment, which repeats it: *"set here — before GetChainRouter, which reads it too"*.

### 3. `backfillLatestBlock` breaks per-URL attribution

`healthEndpointResult` is per node URL, and `applyValidation` sets `LatestBlock` from `validations[i]` — that URL's own result. But `chainFetcher.FetchLatestBlockNum` routes through the router built over **all** of `probedUrls`, so a row for URL A can be filled with a height fetched via URL B.

It also fills rows with `Ok: false`, so a leg that failed every check now reports a plausible height next to its failures.

The three new tests pin fetch-once, don't-overwrite, and failure-is-invisible, but none asks whether a failed row should be backfilled at all. Your call — skipping `!row.Ok` rows, or labelling the value as endpoint-level rather than per-URL in the JSON, would both resolve it.

### 4. Smaller points

- **`IgnoreWsEnforcementForTestCommands` is the same anti-pattern, left in place** — a package global in the very same `if` (`chain_router.go:18,328`), written by `health_cmd.go:115` and `testing.go:38`. No race today, since both writes land before any goroutine starts, but the argument in this PR ("per-parser state removes the shared cell entirely") applies to it too. Follow-up ticket at most.

- **`cloneForValidation` now takes `RLock` for exactly one field.** `grpc.go:129` calls `apip.SkipWebsocketVerification()` inside a struct literal where every other field is read bare — including `spec` and `taggedApis`, which `SetSpec` mutates under the write lock. Locking one field doesn't make the clone a consistent snapshot; it only adds deadlock surface if a future caller ever holds the write lock across a clone. Reading `apip.skipWebsocketVerification` bare, like its neighbours, is more consistent with the trade-off already documented in that comment block.

- **`increaseNumberOfClients` discards `impliesTLS` — verified safe, but fragile.** `dialAddr, _ := grpcDialAddress(...)` followed by `connector.getTransportCredentials()` works only because `NewGRPCConnector` always runs `createConnection` first, and that calls `connector.setCredentials(...)` both when the scheme implies TLS and on the plaintext→TLS upgrade path. Correct today. But the new doc comment says "EVERY dial site must go through this" with nothing enforcing it, and two dial sites drifting apart *is* the bug being fixed. The durable version is for `increaseNumberOfClients` to call `createConnection` rather than keep a second dial loop that also lacks the TLS-upgrade retry. Follow-up.

- Commit `0565490` carries `Took 19 minutes` / `Took 21 minutes` trailers (twice). Tooling noise that would live in `main`.

- Bug B is correctly framed as not health-specific — `increaseNumberOfClients` is on the `GetRpc` path, so pool refill was broken for every `grpc(s)://` endpoint, and `nConns=1` only turns "pool never grows" into "every relay hangs".
- `BaseChainParser.Construct` assigns fields individually, so `SetSpec` does not clobber `skipWebsocketVerification` — setting it before `RegisterForSpecUpdatesOrSetStaticSpecsWithToken` in `probeProvider` is safe.
- `SkipWebsocketVerificationDefault` is read only in the four `New*ChainParser` constructors. The only production parser built by struct literal is `cloneForValidation`, which carries the value across explicitly.

avitenzer:
## Resolved — `962b8f0` → `7555c80` (rebased, force-pushed; your line refs have shifted)

All seven verified and reproduced. Thanks — #1 and #2 both changed the shape of this PR.

**1. Collision with #294** — #294 landed as `d03d1a4`, so `0565490` is **dropped** and the branch rebased onto it. `connector.go` now carries main's `const` + `capacity` plus this PR's `grpcDialAddress`. Confirmed the hazard first: dropping it *before* #294 landed put `TestSkipWebsocketVerificationIsPerParser` back under `WARNING: DATA RACE`, so it couldn't have gone earlier. Passes at `-count=5 -race` on main's version now.

**2. ws gate unreachable in `health`** — confirmed; `health_cmd.go:115` precedes probing at `:150`, and `&&` short-circuits before `SkipWebsocketVerification()` is evaluated. Corrected in all three places: the `probeProvider` comment, the amended commit message, and the description. Fix kept as hardening. Also noted that `TestSkipWebsocketVerificationIsPerParser` therefore pins the *serving*-path shape, not the reported symptom.

**3. `backfillLatestBlock`** — took both: skips `!row.Ok`, and the endpoint-level provenance is now documented on the function and the JSON field rather than implied. Two tests, both verified to fail against the old guard (`SkipsFailedRows`, `NoFetchWhenOnlyFailedRowsAreMissing`).

**4a–c** — agreed, all three recorded under "Not addressed" as follow-ups.
**4d** — `6fa05e6` had one too, not just `0565490`. They come from a local `prepare-commit-msg` hook; the three inherited commits are clean, the newest keeps its trailer by choice.

### Two things I'm not doing

- **No test for the mutex fix.** Since Bug A's real mechanism is the serialization, that's the untested half — but a seam at the `runHealthProbes` fan-out would pin concurrency that was never broken and passes against the old code. Not adding an inert test. Filed as a "Known test gap" in the description with what a faithful version needs.

Verification post-rebase: `build` / `vet` / `gofmt` clean; `go test -race` green on `chainlib/...`, `rpcsmartrouter/...`, `lavasession/...`.

AnnaR-prog (approved):

PR #297, opened 2026-08-19
fix(smart-router): dial the upstream that serves the api's internal path, root included

nimrod-teich:
Jira ticket: MAG-2881

#Closes MAG-2881

## Summary

A chain serving two API versions over two node-urls gets one direct-RPC endpoint per version, and session selection never looked at which. Whichever endpoint it picked answered every api — so on TON a `/v3` call was dialed at the `/v2` upstream and came back as that vendor's 404, indistinguishable from the chain not supporting it.

Two code paths turn node-urls into dialable things, and they disagreed:

- **`chainRouterImpl.BatchNodeUrlsByServices`** builds one chain proxy per internal path, appending the path to an unpinned url (`autoGenerateMissingInternalPaths`: `nodeUrl.Url = baseUrl + internalPath`). The tracker and the verifications run on these, and they are correct.
- **`convertProvidersToSessions`** builds the direct-RPC endpoints relays run on. It recorded url, addons and extensions — and dropped the internal path.

`GetSessions` takes addon + extensions, so there was nothing left to match on.

## Reproduction

Local `smartrouter` on the TON spec, one shared-root node-url (`https://toncenter.com/api`), so the chain router auto-generates `/api/v2` and `/api/v3` correctly:

```console
$ curl -s 'localhost:3460/addressBook?address=EQAAFhjXz…'
{"ok":false,"error":"Not Found","code":404,"@extra":"_"}
```

`/addressBook` is a v3-only api. That is toncenter **v2**'s error shape — `@extra`, and `X-Api-Version: v2.1.13` on the response. The debug log shows the chain router holding the right proxies (`||internal-path:/v2| → …/api/v2`, `||internal-path:/v3| → …/api/v3`) while the relay dialed a different connection entirely.

## Change

- **`expandInternalPaths`** mirrors `autoGenerateMissingInternalPaths`: a pinned url is taken as it stands, an unpinned one additionally yields one url per internal path with the path appended.
- **A path is generated only on the transport that serves it.** The chain router generates a path url only on a transport matching the collection's subscription-ness (`chain_router.go:145-152`); `subscriptionCollectionProbe` reads the same `SUBSCRIBE` tag out of the parser (addons separated, `POST` and `""` connection types) and the expansion applies the same rule. Without it a STRK config carrying both schemes also yields the crossed pair — `https://host/ws/rpc/v0_8`, `wss://host/rpc/v0_9` — urls the chain router refuses to build and no upstream answers.
- **`lavasession.Endpoint`** carries `InternalPath`.
- **`GetSessionsOptions.InternalPath`** carries the resolved collection's path from `rpcsmartrouter_server.go` to the endpoint filter. Variadic, so the 38 existing test call sites are unchanged.
- **The filter matches exactly, in both directions**, and restricts **only when the provider has a matching endpoint**. A provider with none — a provider-relay pool, where the path lives on the provider's side rather than in our url — keeps every endpoint it had. A routing refinement must not become an outage.
- `internalPath` is a `*string`: **`nil`** means "no collection to match on" (probes), **`""`** is a real path meaning the spec's root collection.
- **A url that already ends in the path is that path's endpoint**, not a parent to append to. An operator can bake the version into the url instead of declaring `internal-path` (`https://vendor/rpc/v0_8`, no field) — a config that serves fine today, because before the endpoint list carried paths at all every relay went to that one url. Expanding it blindly would generate `…/rpc/v0_8/rpc/v0_8` and, with the filter now preferring an exact match, aim every `/rpc/v0_8` relay at that 404. It is registered as that path's endpoint as it stands, and stays the root endpoint too.

That last distinction is the one that matters for Starknet — see below.

## Every internal path in the spec repo

Swept all 247 enabled specs, resolving `imports` with child-overrides-parent (which is how a spec disables an inherited collection — AVAX declares `""` `enabled: false` to kill ETH1's root). **12 spec × interface pairs across 6 chain families** declare any internal path; every other chain has none and is untouched by every line of this PR.

| Spec | Interface | Shape | Paths |
|---|---|---|---|
| AVAX / AVAXT | jsonrpc | paths only | `/C/avax` `/C/rpc` `/P` `/X` |
| AVALANCHEC / CT | jsonrpc | paths only | `/C/avax` `/C/rpc` |
| AVALANCHEP / PT | jsonrpc | paths only | `/P` |
| MONERO / MONEROT | jsonrpc | paths only | `/json_rpc` |
| TON / TONT | rest | paths only | `/v2` `/v3` |
| STRK / STRKS | jsonrpc | **root + paths** | `/rpc/v0_8` `/rpc/v0_9` `/rpc/v0_10`, the three `/ws/...` twins, `/rpc/pathfinder/v0.1` |

Two shapes, and they behave differently:

**paths only** (10 of 12) — no enabled root collection, so every relay carries a path, the filter is always active and always has a matching endpoint, and the unpinned root url the expansion keeps is simply never chosen. Strictly a fix: AVAX's `platform.*` used to be dialed at the C-chain url, and MONERO's whole api used to be dialed at the base rather than `/json_rpc`.

**root + paths** (STRK only) — this is where the first cut of this PR **would have regressed**. STRK's `""` collection carries the whole 26-method set and each `/rpc/v0_x` carries it again, so most STRK relays resolve to `""`:

```
""                    ← HTTP-ONLY + WS-ONLY   (enabled, the whole method set)
/rpc/v0_8             ← HTTP-ONLY
/ws/rpc/v0_8          ← /rpc/v0_8 + WS-ONLY
/rpc/v0_9  /v0_10     …same
/rpc/pathfinder/v0.1  3 apis of its own
```

Treating `""` as "don't filter" would have left root traffic eligible for all **eight** endpoints the expansion generates from one unpinned url, and latency sorting could promote `https://host/rpc/v0_9` — or `https://host/ws/rpc/v0_8`, a ws door generated over an http url — ahead of the real root. Hence the `nil` / `""` distinction. Root traffic now stays on the root url, and a caller that asks for `/rpc/v0_9` gets that version at dial time too, where before its choice was honoured at parse time and then discarded.

| Relay's collection | Eligible endpoints |
|---|---|
| `""` — STRK root, and every chain with no internal paths | the root url only; on a chain with no paths that IS every endpoint, so no change |
| `/rpc/v0_9` | that version's url |
| `/P` — AVAX | `/P` |
| `nil` — probes | all, unchanged |

## Future-proofing

The sweep also produced two guards for specs that do not exist yet:

- **A collection label is not a url path.** Specs use the internal-path field as a plain label for inheritance ingredients — STRK's `HTTP-ONLY` / `WS-ONLY`. They are disabled today so the parser never reports them, but the idiom is in the spec repo, and an enabled one would generate `https://host` + `HTTP-ONLY`. Only a path starting with `/` is appended; anything else yields no endpoint, which the filter reads as "not populated for this provider" and falls back to today's behaviour. (Confirmed: no enabled path in the repo today fails that check.)
- **No duplicate emission.** An operator can declare the base url AND a pinned url equal to base+path. Both are legitimate; the generated twin would be probed and registered in the metrics twice. Emission is deduped on (url, internal path).

A brand-new chain that adds internal paths needs no code change: the path set is read from `chainParser.GetAllInternalPaths()` at construction, which is also what the chain router reads.

## Verified live

One router serving TON **and** ETH1 off the same build — ETH1 is the control, a chain with no internal paths going through the same selection code:

```
ETH1  eth_blockNumber          -> {"jsonrpc":"2.0","result":"0x18980c9","id":1}
TON   /getMasterchainInfo      -> {"ok":true,"result":{"@type":"blocks.masterchainInfo"…   (toncenter v2)
TON   /addressBook?address=…   -> {"EQAAFhjXz…":{"user_friendly":"UQAAFhjXz…","domain":null…  (tonindex v3)
```

Trackers green on both: `chainID=TON failed=0 success=3`, `chainID=ETH1 failed=0 success=1`. TON was also verified on both config shapes DFNS uses — one shared-root url, and one url pinned per version — with the same results; before this PR `/addressBook` returned the v2 404 under both. The shared-root boot log shows the expansion:

```
created direct RPC connection url=https://toncenter.com/api
created direct RPC connection url=https://toncenter.com/api/v2
created direct RPC connection url=https://toncenter.com/api/v3
```

**Every other internal-path family, live, all on shared-root urls so the expansion has to do the work:**

```
AVAX   eth_blockNumber    (/C/rpc)   -> {"result":"0x58dca22"}                     C-chain
AVAX   platform.getHeight (/P)       -> {"result":{"height":"25385361"}}           P-chain
AVAX   avm.getHeight      (/X)       -> {"result":{"height":"526970"}}             X-chain
MONERO get_block_count    (/json_rpc)-> {"result":{"count":3743520,"status":"OK"}}
```

Three Avalanche sub-chains, three upstream urls, each answering its own height. Trackers `failed=0 success=5` (AVAX: base + 4 paths) and `failed=0 success=2` (MONERO).

**STRK** — verified against both a real endpoint (`rpc.starknet.lava.build`) and the Lava gateway.

STRK's spec has subscriptions, so a valid config carries both schemes, and each scheme carries the collections it serves — `/ws/rpc/v0_x` inherits `WS-ONLY` (which holds `SUBSCRIBE`), `/rpc/v0_x` inherits `HTTP-ONLY` (which does not):

```
https://<host>                      ← root collection
  …/rpc/v0_8   …/rpc/v0_9   …/rpc/v0_10   …/rpc/pathfinder/v0.1

wss://<host>                        ← root collection
  …/ws/rpc/v0_8   …/ws/rpc/v0_9   …/ws/rpc/v0_10
```

Nine urls — the same nine the chain router builds proxies for, asserted exactly in `TestExpandInternalPaths`.

> ⚠ The live STRK run below predates the transport guard, which was added in `1f87a35` in response to review. Its endpoint enumeration and `failed=0 success=8` tracker line describe the pre-guard expansion off a single url. The per-url dial table that follows is unaffected — every path in it is an http collection — but the boot-time endpoint list wants a re-run before merge.

**Which url each client call lands on** is the thing this PR changes, and nothing in the router's logs or metrics records it — the per-endpoint metrics label by provider, not url. So the STRK run used a local upstream that answers well enough to boot and appends every request path to a file. `starknet_chainId` is never polled by the tracker, so a post-boot occurrence is always a client relay:

| client sends to | **unpatched** dials | **patched** dials |
|---|---|---|
| `<root>` | `/base` | `/base` |
| `/rpc/v0_9` | `/base` ❌ | `/base/rpc/v0_9` ✅ |
| `/rpc/v0_10` | `/base` ❌ | `/base/rpc/v0_10` ✅ |
| `/rpc/pathfinder/v0.1` | `/base` ❌ | `/base/rpc/pathfinder/v0.1` ✅ |

On the unpatched build all four calls returned the **same** answer, so a caller asking for a specific version had no way to tell the choice was discarded. That is the shape of the bug on a chain with a root collection: not an error, a silently wrong door. Root traffic stays on root in both — the property the `nil` / `""` distinction protects.

## Test plan

- `go build ./...`, `go vet ./protocol/...` — clean
- `go test ./protocol/lavasession/... ./protocol/rpcsmartrouter/... ./protocol/chainlib/...` — pass
- `TestExpandInternalPaths` — 10 cases: no-paths passthrough, shared-root expansion, pinned untouched, mixed provider, addons carried, label not appended, no duplicate emission, stable ordering, **transport guard (STRK's real path set over both schemes, asserting the exact 9-url output)**, **unparseable scheme (bare gRPC `host:port`) still expands its http paths**
- `TestFetchEndpointConnection_InternalPath` — 6 cases: versioned api reaches only its own upstream, **root-collection api stays on the root url (STRK)**, **AVAX shape where no root collection exists**, a chain with no internal paths is unaffected, probes unfiltered, provider with no matching endpoint not filtered to nothing

**Pre-existing flake, not from this PR:** `TestCheckAndUnblock_BackupUnblockedWhenHealthy` fails intermittently on the precondition at `consumer_session_manager_test.go:2521`. Reproduced on clean `main` with the same 20-run stress (`-count=20`), so it is not introduced here.

## Side effects worth knowing

- **Metrics cardinality.** Each generated url registers with the metrics manager, so a chain with N internal paths registers up to N+1 urls per unpinned node-url instead of 1. Bounded by the spec's path count, and only for the 6 families above.
- **Probe traffic.** `probeDirectRPCEndpoints` iterates endpoints, so those chains get one probe per path — matching what the chain router already does for trackers.
- Connections are cheap: HTTP shares one transport singleton, WS is dialed lazily on first send. The expansion opens no idle sockets.

## The tracker-retry double-prefix, also fixed here

A pinned url was being re-expanded into `…/api/v2/v2` — first seen when toncenter rate-limited a tracker attempt. I originally filed this as a separate defect on the theory that the retry path re-runs expansion. **That theory was wrong**, and the A/B says this PR fixes it.

Reproduction: TON with both urls pinned per version, `/v3` pointed at a refused port so its tracker fails and retries. Three runs of each binary, same config:

| build | run | chain routers | `…/vN/vN` urls | tracker retries |
|---|---|---|---|---|
| patched | 1 | 1 | **0** | 6 |
| patched | 2 | 1 | **0** | 7 |
| patched | 3 | **2** | **0** | 6 |
| unpatched | 1 | 2 | **4** | 9 |
| unpatched | 2 | 2 | **4** | 10 |
| unpatched | 3 | 1 | 0 | 0 |

What the table says, read carefully:

- The **second chain router is triggered by tracker-startup retries** — it appears in every run that retried and in neither run that did not (unpatched run 3 retried zero times and built one router).
- The second construction is not itself the bug. Patched run 3 built a second router and still produced **zero** double-prefixed urls.
- The bug is that on the unpatched build that second construction reads an endpoint list where the internal path has been flattened into the url, so it treats a pinned `…/api/v2` as a root and re-expands it. With `Endpoint.InternalPath` carried through, it no longer can.

Zero doubled urls across all three patched runs, including the one that exercised the second construction.

## Related

MAG-2880 / smart-router-dashboard#146 — the dashboard half of the same TON internal-path work.

### DIFF
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -620,7 +620,9 @@ func (csm *ConsumerSessionManager) probeProvider(ctx context.Context, consumerSe
 		return csm.probeDirectRPCEndpoints(ctx, consumerSessionsWithProvider, consumerSessionsWithProvider.PublicLavaAddress)
 	}
 
-	connected, endpoints, providerAddress, err := consumerSessionsWithProvider.fetchEndpointConnectionFromConsumerSessionWithProvider(ctx, tryReconnectToDisabledEndpoints, true, "", nil)
+	// A probe measures reachability of the provider, not of one api collection —
+	// no internal path to match on, so every endpoint stays eligible.
+	connected, endpoints, providerAddress, err := consumerSessionsWithProvider.fetchEndpointConnectionFromConsumerSessionWithProvider(ctx, tryReconnectToDisabledEndpoints, true, "", nil, "")
 	if err != nil || !connected {
 		if errors.Is(err, AllProviderEndpointsDisabledError) {
 			csm.blockProvider(ctx, providerAddress, true, epoch, MaxConsecutiveConnectionAttempts, 0, false, csm.GenerateReconnectCallback(consumerSessionsWithProvider)) // reporting and blocking provider this epoch
@@ -976,6 +978,15 @@ type GetSessionsOptions struct {
 	// agreement threshold so each group can independently reach its internal quorum; default 0/1 keeps the
 	// one-provider-per-group behavior (group-blind fill).
 	PerGroupTarget int
+	// InternalPath is the internal path of the api collection this relay
+	// resolved to ("/v2", "/P", "" for the root). Endpoint selection matches on
+	// it, because in direct mode the path lives in the upstream URL: a chain
+	// serving two API versions over two node-urls (TON's toncenter v2 +
+	// tonindex v3) has one endpoint per version, and dialing the wrong one
+	// returns that vendor's 404 as if it were an answer. Empty for every caller
+	// that doesn't resolve an api collection (probes, tests) — selection is
+	// then unfiltered, exactly as before.
+	InternalPath string
 }
 
 func (csm *ConsumerSessionManager) GetSessions(ctx context.Context, wantedProviderNumber int, cuNeededForSession uint64, usedProviders UsedProvidersInf, requestedBlock int64, addon string, extensions []*spectypes.Extension, stateful uint32, virtualEpoch uint64, stickiness string, selectedProvider string, opts ...GetSessionsOptions) (
@@ -985,6 +996,7 @@ func (csm *ConsumerSessionManager) GetSessions(ctx context.Context, wantedProvid
 	// groups. Variadic so the many existing non-cross-validation callers are unchanged.
 	minGroups := groupBlindMinGroups
 	perGroupTarget := groupBlindPerGroupTarget
+	internalPath := ""
 	if len(opts) > 0 {
 		if opts[0].MinGroups > 1 {
 			minGroups = opts[0].MinGroups
@@ -992,6 +1004,7 @@ func (csm *ConsumerSessionManager) GetSessions(ctx context.Context, wantedProvid
 		if opts[0].PerGroupTarget > 1 {
 			perGroupTarget = opts[0].PerGroupTarget
 		}
+		internalPath = opts[0].InternalPath
 	}
 	// set usedProviders if they were chosen for this relay
 	timeoutCtx, cancel := context.WithTimeout(ctx, time.Second)
@@ -1040,7 +1053,7 @@ func (csm *ConsumerSessionManager) GetSessions(ctx context.Context, wantedProvid
 			sessionEpoch := sessionWithProvider.CurrentEpoch
 
 			// Get a valid Endpoint from the provider chosen
-			connected, endpoints, _, err := consumerSessionsWithProvider.fetchEndpointConnectionFromConsumerSessionWithProvider(ctx, false, false, addon, extensionNames)
+			connected, endpoints, _, err := consumerSessionsWithProvider.fetchEndpointConnectionFromConsumerSessionWithProvider(ctx, false, false, addon, extensionNames, internalPath)
 			if err != nil {
 				// verify err is AllProviderEndpointsDisabled and report.
 				if errors.Is(err, AllProviderEndpointsDisabledError) {
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -622,7 +622,7 @@ func (csm *ConsumerSessionManager) probeProvider(ctx context.Context, consumerSe
 
 	// A probe measures reachability of the provider, not of one api collection —
 	// no internal path to match on, so every endpoint stays eligible.
-	connected, endpoints, providerAddress, err := consumerSessionsWithProvider.fetchEndpointConnectionFromConsumerSessionWithProvider(ctx, tryReconnectToDisabledEndpoints, true, "", nil, "")
+	connected, endpoints, providerAddress, err := consumerSessionsWithProvider.fetchEndpointConnectionFromConsumerSessionWithProvider(ctx, tryReconnectToDisabledEndpoints, true, "", nil, nil)
 	if err != nil || !connected {
 		if errors.Is(err, AllProviderEndpointsDisabledError) {
 			csm.blockProvider(ctx, providerAddress, true, epoch, MaxConsecutiveConnectionAttempts, 0, false, csm.GenerateReconnectCallback(consumerSessionsWithProvider)) // reporting and blocking provider this epoch
@@ -979,14 +979,17 @@ type GetSessionsOptions struct {
 	// one-provider-per-group behavior (group-blind fill).
 	PerGroupTarget int
 	// InternalPath is the internal path of the api collection this relay
-	// resolved to ("/v2", "/P", "" for the root). Endpoint selection matches on
-	// it, because in direct mode the path lives in the upstream URL: a chain
-	// serving two API versions over two node-urls (TON's toncenter v2 +
-	// tonindex v3) has one endpoint per version, and dialing the wrong one
-	// returns that vendor's 404 as if it were an answer. Empty for every caller
-	// that doesn't resolve an api collection (probes, tests) — selection is
-	// then unfiltered, exactly as before.
-	InternalPath string
+	// resolved to ("/v2", "/P", or "" for a spec's root collection). Endpoint
+	// selection matches on it EXACTLY, because in direct mode the path lives in
+	// the upstream URL: a chain serving two API versions over two node-urls
+	// (TON's toncenter v2 + tonindex v3) has one endpoint per version, and
+	// dialing the wrong one returns that vendor's 404 as if it were an answer.
+	//
+	// nil — not "" — is the "no collection to match on" case (probes, tests),
+	// which leaves selection unfiltered. "" is a real path: a chain that
+	// enables a root collection alongside versioned ones (STRK) must keep its
+	// root traffic on the root url.
+	InternalPath *string
 }
 
 func (csm *ConsumerSessionManager) GetSessions(ctx context.Context, wantedProviderNumber int, cuNeededForSession uint64, usedProviders UsedProvidersInf, requestedBlock int64, addon string, extensions []*spectypes.Extension, stateful uint32, virtualEpoch uint64, stickiness string, selectedProvider string, opts ...GetSessionsOptions) (
@@ -996,7 +999,7 @@ func (csm *ConsumerSessionManager) GetSessions(ctx context.Context, wantedProvid
 	// groups. Variadic so the many existing non-cross-validation callers are unchanged.
 	minGroups := groupBlindMinGroups
 	perGroupTarget := groupBlindPerGroupTarget
-	internalPath := ""
+	var internalPath *string
 	if len(opts) > 0 {
 		if opts[0].MinGroups > 1 {
 			minGroups = opts[0].MinGroups
--- a/protocol/lavasession/consumer_session_manager_test.go
+++ b/protocol/lavasession/consumer_session_manager_test.go
@@ -1803,6 +1803,7 @@ func TestDeadConnectionCleanup(t *testing.T) {
 		false, // getAllEndpoints
 		"",    // addon
 		nil,   // extensionNames
+		"",    // internalPath
 	)
 
 	// Should successfully connect (we have one valid connection)
@@ -2264,6 +2265,7 @@ func TestCanceledContextDoesNotPenalizeEndpoint(t *testing.T) {
 		false, // getAllEndpoints
 		"",    // addon
 		nil,   // extensionNames
+		"",    // internalPath
 	)
 
 	// Connection should fail (context canceled), but the endpoint must not be penalized.
--- a/protocol/lavasession/consumer_session_manager_test.go
+++ b/protocol/lavasession/consumer_session_manager_test.go
@@ -1803,7 +1803,7 @@ func TestDeadConnectionCleanup(t *testing.T) {
 		false, // getAllEndpoints
 		"",    // addon
 		nil,   // extensionNames
-		"",    // internalPath
+		nil,   // internalPath
 	)
 
 	// Should successfully connect (we have one valid connection)
@@ -2265,7 +2265,7 @@ func TestCanceledContextDoesNotPenalizeEndpoint(t *testing.T) {
 		false, // getAllEndpoints
 		"",    // addon
 		nil,   // extensionNames
-		"",    // internalPath
+		nil,   // internalPath
 	)
 
 	// Connection should fail (context canceled), but the endpoint must not be penalized.
--- a/protocol/lavasession/consumer_types.go
+++ b/protocol/lavasession/consumer_types.go
@@ -1105,30 +1105,43 @@ func (cswp *ConsumerSessionsWithProvider) sortEndpointsByLatency(endpointInfos [
 
 // fetching an endpoint from a ConsumerSessionWithProvider and establishing a connection,
 // can fail without an error if trying to connect once to each endpoint but none of them are active.
-func (cswp *ConsumerSessionsWithProvider) fetchEndpointConnectionFromConsumerSessionWithProvider(ctx context.Context, retryDisabledEndpoints bool, getAllEndpoints bool, addon string, extensionNames []string, internalPath string) (connected bool, endpointsList []*EndpointAndChosenConnection, providerAddress string, err error) {
+//
+// internalPath is the collection the relay resolved to, or nil when the caller
+// has no collection to match on (probes). "" is a REAL path — the spec's root
+// collection — and matches only the endpoint serving the root, so a chain that
+// enables both a root and versioned collections (STRK) keeps its root traffic
+// on the root url.
+func (cswp *ConsumerSessionsWithProvider) fetchEndpointConnectionFromConsumerSessionWithProvider(ctx context.Context, retryDisabledEndpoints bool, getAllEndpoints bool, addon string, extensionNames []string, internalPath *string) (connected bool, endpointsList []*EndpointAndChosenConnection, providerAddress string, err error) {
 	getConnectionFromConsumerSessionsWithProvider := func(ctx context.Context) (connected bool, endpointPtr []*EndpointAndChosenConnection, allDisabled bool) {
 		endpoints := make([]*EndpointAndChosenConnection, 0)
 		cswp.Lock.Lock()
 		defer cswp.Lock.Unlock()
 		// Restrict to the endpoints serving the api's internal path — but only
-		// when this provider has any. Every non-empty path is expanded at
-		// construction (rpcsmartrouter.go), so a provider that CAN serve the
-		// path has a matching endpoint and the filter bites. A provider that
-		// has none — a provider-relay pool, where the path lives on the
-		// provider's side and not in our url, or an endpoint list built some
-		// way we haven't modelled — keeps every endpoint it had, rather than
-		// losing the whole pool to a filter that was never populated for it.
+		// when this provider has any. Every path is expanded at construction
+		// (rpcsmartrouter.go), so a provider that CAN serve the path has a
+		// matching endpoint and the filter bites. A provider that has none — a
+		// provider-relay pool, where the path lives on the provider's side and
+		// not in our url, or an endpoint list built some way we haven't
+		// modelled — keeps every endpoint it had, rather than losing the whole
+		// pool to a filter that was never populated for it.
+		//
+		// Matching is exact in both directions. On a chain with no internal
+		// paths every endpoint is "" and every relay asks for "", so this is a
+		// no-op. On a chain that enables a root collection ALONGSIDE versioned
+		// ones (STRK: "" carries the whole method set, /rpc/v0_8 … carry it
+		// again per version), a root relay must stay on the root url rather
+		// than drift onto a versioned door the expansion generated.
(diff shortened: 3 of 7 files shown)


nimrod-teich:
_(Folded into the PR description — see **Every internal path in the spec repo** and **Future-proofing** there, which also cover the sweep across all 247 specs.)_

avitenzer (commented):
Reviewed this end-to-end — build, vet, and the full `lavasession` / `rpcsmartrouter` / `chainlib` suites pass on `27d2f49`, and the sweep of internal paths across the spec repo checks out. One finding worth resolving before merge; everything else I found was minor.

## `expandInternalPaths` drops the transport guard that `autoGenerateMissingInternalPaths` has

`rpcsmartrouter.go` — the doc comment states the function "mirrors `chainRouterImpl.BatchNodeUrlsByServices` **exactly**, because the two have to agree on which url answers a given api collection." That invariant is the right one, but the two don't currently agree. `autoGenerateMissingInternalPaths` (`chain_router.go:145-152`) has a guard with no counterpart in the new code:

```go
if isWs && !subscriptionTagFound {
	continue // this is ws, don't auto generate http paths
} else if !isWs && subscriptionTagFound {
	continue // this is http, don't auto generate ws paths
}
```

The chain router only generates a path url on a transport matching the collection's subscription-ness. `expandInternalPaths` generates every path on every url unconditionally.

### Reproduction

Running the shipped function against STRK's real enabled path set from `lava-specs/starknet.json`:

```
https://strk.example  ->  https://strk.example/ws/rpc/v0_8   internalPath="/ws/rpc/v0_8"   <- ws collection, http scheme
wss://strk.example    ->  wss://strk.example/rpc/v0_9        internalPath="/rpc/v0_9"      <- http collection, ws scheme
```

Both are urls `autoGenerateMissingInternalPaths` explicitly refuses to build.

### Why the new filter turns this into a misroute

STRK's spec has subscriptions, so the ws enforcement at `chain_router.go:328` requires a ws node-url — a real STRK config carries both `https://` and `wss://`. Each expands to 8 endpoints, so **every internal path ends up matching two endpoints with different schemes**:

```
path "/rpc/v0_9"     ->  https://strk.example/rpc/v0_9
path "/rpc/v0_9"     ->  wss://strk.example/rpc/v0_9        <- same path, ws transport
path "/ws/rpc/v0_9"  ->  https://strk.example/ws/rpc/v0_9   <- same path, http transport
path "/ws/rpc/v0_9"  ->  wss://strk.example/ws/rpc/v0_9
```

`ServesInternalPath` compares only the path string, and `DetectProtocol` derives transport purely from the url scheme — so `wss://.../rpc/v0_9` becomes a `WebSocketDirectRPCConnection` that a plain HTTP JSON-RPC relay can be routed onto by latency sort, with nothing downstream to catch it.

This is a regression rather than a pre-existing gap. Before this PR there was no expansion, so a `/rpc/v0_9` relay went to the configured `https://strk.example` — the wrong *path* (the bug being fixed here), but a working upstream. After, the relay is pinned to the `/rpc/v0_9` set, half of which is a scheme the chain router deems invalid for that collection.

The PR description already names the hazard — *"least of all onto a `/ws/...` door generated over an http url"* — but the `nil` / `""` distinction it credits only protects **root** traffic. `/ws/rpc/v0_8` traffic is now pinned to exactly that door.

### Suggested fix

Port the `subscriptionTagFound` check into `expandInternalPaths`. It needs `chainParser.IsTagInCollection(FUNCTION_TAG_SUBSCRIBE, ...)` plus the url's scheme, both available at the call site in `CreateSmartRouterEndpoint`.

### Scope

STRK/STRKS are the only affected specs today — the other five families in the table have no ws internal paths, so their expansion is correct. The 12-spec sweep is accurate; it just doesn't separate ws paths from http ones. Worth noting the live STRK run reported 8 healthy trackers, but trackers come from the chain router's own list, so a green tracker board wouldn't have surfaced this.

To be explicit about what's verified vs. inferred: the divergence between the two functions and the expansion output above are both directly reproduced. That `/ws/rpc/v0_8` resolves to a SUBSCRIBE-carrying collection after inheritance is inferred from the spec structure plus the description's own account of it (`/rpc/v0_8` + `WS-ONLY`).

---

**Minor, same root cause, not blocking:** `nodeUrl.Url + internalPath` is a raw string concat, so a base url carrying a query string or trailing slash yields `https://host/api?key=X/v2` or `https://host/api//v2`. The chain router does the same concat, so trackers break identically and visibly at boot — which is why I'd rank it below the above. A URL-aware join in the same change would cover both.

Nice work on the rest of this — the `nil` / `""` sentinel split and the "provider with no matching endpoint isn't filtered to nothing" fallback are both the right calls, and the A/B on the tracker double-prefix is a genuinely convincing piece of evidence.

nimrod-teich:
Good catch — fixed in `1f87a35`.

The `/ws/rpc/v0_x` half you flagged as inferred is confirmed, straight out of `lava-specs/starknet.json`:

```
path='WS-ONLY'         enabled=False  tags=[SUBSCRIBE, UNSUBSCRIBE]
path='/rpc/v0_8'       enabled=True   inherit=['HTTP-ONLY']
path='/ws/rpc/v0_8'    enabled=True   inherit=['/rpc/v0_8', 'WS-ONLY']   ← SUBSCRIBE
```

So `subscriptionTagFound` is true for exactly the three `/ws/...` paths and false for the four http ones. For a config carrying both schemes the chain router builds **9** proxies and the expansion built **16** — the 7 extras being the crossed pair.

## The fix

`subscriptionCollectionProbe` reads the `SUBSCRIBE` tag the same way `autoGenerateMissingInternalPaths` does — `SeparateAddonsExtensions`, then both `POST` and `""` connection types — and `expandInternalPaths` skips a path whose subscription-ness disagrees with the url's scheme:

```go
isWs, err := chainlib.IsUrlWebSocket(nodeUrl.Url)
if err != nil { isWs = false }   // bare gRPC host:port, as the chain router reads it
...
if isWs != servesSubscriptions(internalPath, nodeUrl.Addons) { continue }
```

The guard sits ahead of both the append branch and the already-ends-in-the-path branch, so a `wss://host/rpc/v0_9` an operator pinned by hand is no longer registered as that path's endpoint either — matching the chain router, which builds no proxy for it.

`TestExpandInternalPaths` gains a case using STRK's real enabled path set over both schemes, asserting the exact 9-url output. I confirmed it's load-bearing: stub the guard out and it fails. A second case covers the bare `host:port` gRPC shape so the scheme fallback can't silently stop expanding. Build, vet, and the `lavasession` / `rpcsmartrouter` / `chainlib` suites are clean.

## Two places I'd push back on the framing

**"A regression rather than a pre-existing gap"** — the transport mixing isn't new. There's no scheme-based endpoint filtering anywhere in the smart-router path: no `WebSocketExtension` or `IsUrlWebSocket` in `protocol/lavasession/` or `protocol/rpcsmartrouter/` at all, and `CheckSupportForServices` only sees operator-declared addons. Pre-PR a `/rpc/v0_9` relay was not reliably sent to `https://strk.example` — both schemes sat unfiltered in the same pool. Post-PR the candidate set was `{https://host/rpc/v0_9, wss://host/rpc/v0_9}`: the same coin flip, aimed at different urls. What the expansion genuinely added was 7 endpoints no upstream answers, probed every sweep and registered in the metrics, plus a false "mirrors exactly" invariant. Worth fixing on those grounds, which is why it is fixed — but it wasn't a new misroute.

**"Nothing downstream to catch it"** — `WebSocketDirectRPCConnection` doesn't implement `HTTPDirectRPCDoer` (`direct_rpc_connection.go:132-140`), so a JSON-RPC or REST relay landing on a ws connection fails at `direct_rpc_relay.go:365` with `connection does not support HTTP requests`. A failed relay, but a loud one — not the silent wrong-door class this PR is about.

## On the url join

Agreed it's real, and agreed it's not blocking — but I'd scope it tighter than a URL-aware join, in its own PR.

Query-param API keys don't live in `nodeUrl.Url` in this codebase: they go through `AuthConfig.AuthQuery` and are appended at request time by `AddAuthPath`, which already handles `?` vs `&`. The `https://host/api?key=X/v2` shape needs an operator to hand-write the key inline instead of using `auth-query`, and I found no instance of that across the helm chart, standalone, growth-deployments, or infrastructure repos.

The trailing slash is the half that occurs — `https://arbitrum-one-rpc.publicnode.com/`, `https://eth1.lava.build:443/`, `https://rpc.mainnet.fastnear.com/` are all real config entries, just not on an internal-path chain today. There's also a second consequence: `strings.HasSuffix(nodeUrl.Url, internalPath)` misses on a trailing slash, so `https://host/rpc/v0_8/` expands instead of being recognized as that path's endpoint. One `TrimSuffix` closes both.

A full parse-and-reassemble join would reorder query and fragment relative to path for every Lava provider, not just the smart router, on configs that work today — more blast radius than the case warrants. And it has to land in `chain_router.go` and here together, or the two functions diverge again on a second axis, which is the bug class this PR exists to close. Separate ticket, off `main`.

## One thing to re-check before merge

You're right that the green tracker board wouldn't have surfaced this, and it cuts at the description too: the live STRK section's endpoint enumeration and `failed=0 success=8` line predate the guard. I've marked it in the description and it wants a re-run — the per-url dial table below it is unaffected, since every path in it is an http collection.

avitenzer (approved):
Approving — the transport guard in `96f6b1b` resolves what I flagged, and I verified the mirroring against `BatchNodeUrlsByServices` rather than reading the commit message:

- `if isWs != servesSubscriptions(...)` is logically identical to the chain router's `if isWs && !sub / else if !isWs && sub`, just tighter.
- `subscriptionCollectionProbe` reads the tag the same way `autoGenerateMissingInternalPaths` does — `SeparateAddonsExtensions`, the `len(addons)==0 -> [""]` fallback, and both `POST` and `""` connection types.
- Guard placement is right, and for a subtler reason than the description gives: it sits inside the auto-generation loop only, so a url declaring `internal-path` is added by `add(nodeUrl)` before the guard runs. That matches the chain router, where `setRouterKeyInBatch` runs unconditionally for every node-url and `autoGenerateMissingInternalPaths` — which carries the guard — runs only for root urls.
- The new STRK case pins exactly the 9 urls across both schemes. That's the assertion that would have caught the original bug.
- `IsUrlWebSocket` on a bare `host:port` returns `(false, nil)` rather than erroring — Go parses it as `scheme:opaque` — so the `err -> isWs = false` fallback is belt-and-braces. Covered either way by the gRPC case.

Build, vet, and the full `rpcsmartrouter` / `lavasession` / `chainlib` suites pass on `96f6b1b`.

Two non-blocking notes for follow-up.

## 1. AVALANCHEC's `/C/rpc` is no longer reached by this fix

The guard is faithful to the chain router, and it inherits a chain-router quirk along with the rule. AVALANCHEC's `/C/rpc` collection (`add_on=''`, `type=POST`, enabled) carries a `SUBSCRIBE` parse directive **alongside** `GET_BLOCKNUM` / `GET_BLOCK_BY_NUM` / `GET_EARLIEST_BLOCK`:

```
add_on='debug'   type='POST'  SUBSCRIBE=False
add_on=''        type='POST'  SUBSCRIBE=True   <- tags: GET_BLOCKNUM, GET_BLOCK_BY_NUM, GET_EARLIEST_BLOCK, SUBSCRIBE, UNSUBSCRIBE
add_on='warp'    type='POST'  SUBSCRIBE=False
```

So `servesSubscriptions("/C/rpc")` is true, and on an https url the guard now skips it — no `https://host/ext/bc/C/rpc` endpoint is generated. Two outcomes, neither a break:

- **http-only config** — `/C/rpc` relays match no endpoint, `restrictToPath` stays false, and selection falls back to the root url. That is the pre-PR behaviour, so AVALANCHEC's `/C/rpc` is silently not fixed by this PR.
- **config with a wss url** — `/C/rpc` is generated only on wss, so ordinary HTTP JSON-RPC `/C/rpc` calls get pinned to the WebSocket endpoint. Functional, since ws direct connections serve request/response, but not what you would expect from the config.

Both match the chain router, so the invariant this PR establishes genuinely holds — the root cause sits upstream of it: the chain router treats "collection has SUBSCRIBE" as "ws-only" even for a mixed collection that also carries http methods. Worth a follow-up ticket, and worth softening the table in the description, which currently lists AVALANCHEC `/C/rpc` among the fixed paths.

**AVAX itself is unaffected** — none of `avalanche.json`'s `/C/rpc` collections carry SUBSCRIBE, so the live AVAX verification still holds. This is AVALANCHEC only; AVALANCHECT does not carry the tag either.

## 2. One sentence in the fix comment over-claims

> "The guard sits ahead of both the append branch and the already-ends-in-the-path branch, so a `wss://host/rpc/v0_9` an operator pinned by hand is no longer registered as that path's endpoint either — matching the chain router, which builds no proxy for it."

That holds for a url which *ends in* the path without declaring `internal-path` (the suffix branch). A url that declares `internal-path: /rpc/v0_9` explicitly is added before the guard runs — and the chain router does build a proxy for that case, via the unconditional `setRouterKeyInBatch`.

The code is right either way: honouring an explicit operator declaration on both sides is correct mirroring. It is only the sentence that describes behaviour the code does not have. Flagging it so nobody later "fixes" the code to match the comment. The in-code comment on the guard is accurate, so there is nothing to change in the merged artifact.

---

One thing to weigh, not a blocker: the fix narrowed STRK's expansion from 16 endpoints to 9, so the live STRK run in the description was against the pre-fix build. The unit test pins the exact 9 urls and the change is a pure narrowing, so I am comfortable — but if STRK is heading to production soon, a post-fix boot log is a cheap confirmation.

PR #298, opened 2026-08-19
fix(smart-router): resolve a config argument that is a path as a path

avitenzer:
# Description

`smartrouter /abs/path/config.yml` always failed, even with the file present and readable, and the error named search directories the operator never mentioned:

```
Config File "/abs/path/config.yml" Not Found in "[/repo /repo/config ~/.lava]"
```

The argument was handed to viper's `SetConfigName`, which treats its value as a **name** joined onto each search path — and `filepath.Join` swallows the leading separator, so `Join(".", "/abs/path/config.yml")` is `"abs/path/config.yml"`. Every absolute path was quietly demoted to a relative one and looked for *inside* each search directory, where it could never exist. I reproduced this against viper v1.21.0 in isolation before changing anything:

```
what viper actually looks for under ".": "var/folders/.../router.yml"
OLD -> Config File "/var/folders/.../router.yml" Not Found in "[. ./config ~/.lava]"
```

Relative paths worked only by coincidence, on two of them: `.` happens to be a search path, and `searchInPath` only falls back to the bare name — without appending an extension — because `SetConfigType("yml")` happens to be set. That is why every script under `scripts/pre_setups/` has to `cd` to the project root first, and why the one that did not (the Sui script, on MAG-2643) read as a broken router until the cause was traced.

## What changed

Resolution now lives in `protocol/rpcsmartrouter/config_source.go` (**the file to review**), shared by the two commands that take a config so they cannot drift:

- an argument that **names a directory component** — absolute, or containing a separator — is an explicit path and goes to `SetConfigFile`
- **anything else is a name** and keeps the name-plus-search-path lookup untouched, which is what lets a bare `smartrouter myconfig` still find `./config/myconfig.yml`

The classification is **lexical and does not consult the filesystem**. What is being decided is what the operator *meant*, and that must not vary with the contents of the working directory. An earlier revision of this PR keyed on `os.Stat`; review caught that it made a bare `smartrouter router` load an extension-less `./router` in preference to the `./router.yml` the search path has always returned, and it also sent a missing `nested/missing.yml` down the name branch, reproducing the very search-directory error this ticket exists to remove. Both are fixed in c28d095 and both have regression tests.

The lexical split is only safe because extension resolution moved with it. `smartrouter config/akash` has always resolved to `config/akash.yml` via `searchInPath`'s extension loop; `resolveConfigFilePath` now does the same for an explicit path, trying viper's `SupportedExts` in viper's own order and preferring an extension to the bare path exactly as `searchInPath` does. **The two branches resolve an argument identically** — the only thing the branch changes is which error message you get when nothing is there. (This was *not* true of `c28d095`, which dropped the search paths for relative paths; see the review fixes below.)

## Two consequences of `SetConfigFile` that had to be handled

Without these the fix would have traded a confusing message for a worse one.

**It changes the error type.** A missing searched name is `viper.ConfigFileNotFoundError`; a missing explicit file is a plain `fs.ErrNotExist` from the open. `rpcsmartrouter.go`'s clean-error branch keyed only on the former, so a typo'd path would have fallen through to `LavaFormatFatal` and dumped a stack trace instead of saying the file is not there. `isConfigNotFound` covers both.

**It disables extension-appending**, so an extension-less path needs an explicit `configType` or viper rejects it as `UnsupportedConfigError` before opening the file at all. A recognized extension now declares the format; everything else assumes YAML, which every smartrouter config is.

Messages report where the config was looked for in the terms the operator used — an exact path names that path, its resolved absolute form and the working directory; a bare name names the search paths:

```
ERR config file not found at the given path: /etc/smartrouter/typo.yml
    error="open /etc/smartrouter/typo.yml: no such file or directory"
    config_file=/etc/smartrouter/typo.yml resolved_path=/etc/smartrouter/typo.yml
    working_directory=/repo
```

`health_cmd.go` resolved its config the same way and had the same behaviour, as the ticket suspected — it now shares the helper, and reports a missing config inside its JSON envelope rather than as a generic read failure. Its `Long` now documents how its config argument resolves, which it never did.

The router's own `Long` claimed configs "should be located in the local running directory /config or `$HOME/.lava`"; it now documents that a path works, and that a relative path is still resolved against those directories.

## Review fixes (c28d095)

- **The first commit did not compile.** Dropping the now-unused `errors` import took `fmt` with it, and `fmt` is used at 22 sites in `rpcsmartrouter.go`. The build and test results quoted on `ab53600` were not run against its content. Restored, and the committed tree is now verified from a clean worktree checkout rather than from the working directory.
- **Path/name classification** reworked from `os.Stat` to lexical, as described above.

## Review fixes (062e9e6)

Review found that the lexical split fixed absolute paths but **narrowed** relative ones, and I reproduced both findings before changing anything.

**Relative paths lost two of the three search paths.** `filepath.Join` swallows only a *leading* separator, so it destroyed absolute paths while handling relative ones perfectly — `SetConfigName` + `AddConfigPath` already resolved `smartrouter_examples/smartrouter_eth.yml` to `config/smartrouter_examples/smartrouter_eth.yml`. Sending every path-shaped argument to `SetConfigFile` resolved it against the working directory alone, so that invocation stopped loading, as did anything under `$HOME/.lava`. Measured against binaries built from each commit, from the repo root:

| | `c28d095` | `062e9e6` |
|---|---|---|
| `smartrouter smartrouter_examples/smartrouter_eth.yml` | `config file not found at the given path` | loads |

`resolveConfigFilePath` now probes a relative path against `configSearchPaths` in the same precedence order a bare name uses, each candidate through the extension loop it already had. Absolute paths are still never joined onto a search path — that join *is* MAG-2861 — and a test now pins it as an invariant rather than as a symptom, by planting a decoy at the exact location the demotion would have found. `lavaDefaultNodeHome` is the literal `"$HOME/.lava"` and viper expands it via `absPathify`, so the path branch expands it too; without that it would resolve against a directory named `$HOME`.

This restores the "strictly additive" property the checklist claims. It was not true of `c28d095`.

**A path argument naming a directory took the fatal path.** Viper fails with `EISDIR`, not `fs.ErrNotExist`, so `smartrouter ./config` classified as "not a not-found" and fell through to `LavaFormatFatal` — the exact stack-dump failure mode this PR exists to remove, in a narrow case. `isConfigNotFound` covers `EISDIR`, and the message says what is actually wrong:

```
ERR the given config path is a directory, not a config file: ./config
    error="read ./config: is a directory"
```

Stack-dump lines for that invocation: **1** on `c28d095`, **0** on `062e9e6`.

**Help text.** Both commands now state that a relative path is resolved against the search paths too — the previous wording attached them to bare names only, which is the documentation gap the first finding hid behind. `TestConfigSearchPathsMatchDocumentedHelp` now checks both commands and asserts the relative-path claim is present. The router's `Example` named `smartrouter_examples/full_smartrouter_example.yml`, which does not exist in the repo; it now names one that does.

## Review fixes (88d3074)

Two compatibility gaps in `062e9e6`, both reproduced before changing anything.

**Supported-extension paths did not follow viper's suffix order.** `searchInPath` appends every supported extension to `v.configName` *unconditionally* — it never checks whether the name already ends in one — so `sub/router.yml` has always been able to resolve to `sub/router.yml.json`. `resolveWithSupportedExt` short-circuited on a recognized extension, which looks obviously right and was the last way the two branches could still disagree:

| arg | on disk | `062e9e6` | `88d3074` |
|---|---|---|---|
| `sub/router.yml` | `sub/router.yml.json` | `open sub/router.yml: no such file or directory` | loads `sub/router.yml.json` |

The layout is pathological, but "strictly additive" has to cover it. The short-circuit is gone and appended forms are tried first for every candidate. Two tests pin it: one on the path branch, one on the name branch — which never stopped doing this — so the pair asserts an equivalence rather than a quirk.

**The not-found diagnostics described a search that no longer matched the one that ran.** Once relative paths were probed across `configSearchPaths` again, naming only the path and the working directory understated it, and a test asserted the search paths "were never consulted" — true of an absolute path, false of a relative one. The message now says so, and the attributes carry the exact candidate list rather than a restatement of `configSearchPaths`:

```
ERR config file not found: nested/missing.yml — a relative path is resolved against each of the search paths
    error="open nested/missing.yml: no such file or directory"
    searched_paths="nested/missing.yml config/nested/missing.yml /Users/x/.lava/nested/missing.yml"
```

Absolute paths are unchanged — they still name nothing but themselves, and a test asserts no candidate list is attached to them, since that would re-introduce the MAG-2861 confusion from the other direction.

## Testing

28 subtests pin the resolution matrix (absolute, extension-less absolute, relative, `./`-prefixed, extension-less path, bare name with and without extension, default, inline endpoint args, cwd-beats-`./config` precedence, and extension-beats-extension-less in both branches), all three not-found shapes, and that a malformed config stays on the loud path rather than being mistaken for a missing one.

Each of the three review regressions was verified to **fail against a faithful replica of `ab53600`'s logic**. Worth noting for anyone reproducing: restoring the `os.Stat` classification alone is not enough to reproduce the `router`/`router.yml` shadowing, because the new extension resolution masks it — the replica also has to drop `resolveConfigFilePath`.

Verified end to end against a binary built from the committed tree in an isolated worktree:

| Case | Result |
|---|---|
| Absolute path, run from `/tmp` | loads (was the bug) |
| `config/smartrouter_examples/smartrouter_eth.yml` from repo root — every `pre_setups` launcher | still loads |
| Bare name via `./config` | still loads |
| Missing absolute path | clean error naming the path, no stack dump |
| Missing relative path `nested/missing.yml` | `config file not found at the given path: nested/missing.yml` |
| `smartrouter router` with both `router` and `router.yml` present | loads `router.yml` |
| `smartrouter config/akash` with only `config/akash.yml` present | loads `config/akash.yml` |
| No argument, no config | original search-path message preserved |
| `smartrouter_examples/smartrouter_eth.yml` from repo root | loads via `./config` (regressed by `c28d095`) |
| `smartrouter_examples/smartrouter_eth` (no extension) from repo root | loads via `./config` |
| Relative path present under `$HOME/.lava` | loads |
| Absolute path with a decoy at the demotion location | loads the real file, not the decoy |
| `smartrouter ./config` (a directory) | `the given config path is a directory, not a config file`, no stack dump |
| `sub/router.yml` with only `sub/router.yml.json` present | loads the `.json`, matching `searchInPath` |
| Missing relative path | names the path *and* the three candidates probed |
| Missing absolute path | names that path only, no candidate list |
| `health <abs path>` | resolves, probes ran |
| `health <missing abs path>` | clean error inside the JSON envelope, exit 0 contract intact |

`go build ./...`, `go vet` and the full `protocol/rpcsmartrouter` suite pass; `gofmt` clean.

**CI note, so the green checks are not read for more than they are:** this repo has no unit-test lane. The workflows run lint, `govulncheck`, cross-platform builds, the Docker image, GoReleaser config, Jira validation and the labeler — but nothing invokes `go test`. Every subtest cited above is local evidence only, reproducible with:

```
go test ./protocol/rpcsmartrouter/ -run 'TestPointViperAtConfig|TestIsConfigFilePath|TestConfigSearchPaths' -v
```

Out of scope: the `scripts/pre_setups/` launchers still `cd` and pass relative paths. They work and changing them is churn this fix does not need.

## Jira ticket

Jira ticket: MAG-2861

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change — not applicable, strictly additive: every argument form that resolved before resolves to the same file
* [x] provided a link to the relevant issue or specification — MAG-2861
* [x] updated the relevant documentation or specification, including comments for documenting Go code — command `Long` help and package comments in `config_source.go`
* [x] confirmed all CI checks have passed — all green on `88d3074` (see the CI note under Testing: none of them is a unit-test lane)

### DIFF
--- a/protocol/rpcsmartrouter/config_source.go
+++ b/protocol/rpcsmartrouter/config_source.go
@@ -141,13 +141,17 @@ func configFilePathCandidates(arg string) []string {
 	return candidates
 }
 
-// resolveWithSupportedExt reports the file a single candidate resolves to, trying each
-// supported extension before the bare path. A candidate that already declares its format
-// has nothing to append and must exist as given.
+// resolveWithSupportedExt reports the file a single candidate resolves to, mirroring
+// searchInPath: every supported extension is appended and tried first, in viper's own
+// order, and only then the candidate as given.
+//
+// The appended forms are tried even when the candidate already ends in a supported
+// extension, because searchInPath appends to v.configName unconditionally — it never checks
+// whether the name already carries one. `sub/router.yml` has therefore always been able to
+// resolve to sub/router.yml.json, and short-circuiting on a recognized extension (which
+// looks obviously right) is the one remaining way the two branches could disagree. Pathological
+// as that layout is, "strictly additive" has to mean it.
 func resolveWithSupportedExt(candidate string) (resolved string, found bool) {
-	if ext := strings.TrimPrefix(filepath.Ext(candidate), "."); slices.Contains(viper.SupportedExts, ext) {
-		return candidate, isExistingFile(candidate)
-	}
 	for _, ext := range viper.SupportedExts {
 		if withExt := candidate + "." + ext; isExistingFile(withExt) {
 			return withExt, true
@@ -179,12 +183,18 @@ func isConfigNotFound(err error) bool {
 }
 
 // configNotFoundMessage explains where the config was looked for, in the terms the
-// operator used: the one path they named, or the search paths their bare name was
-// resolved against.
+// operator used — and, importantly, describes the search that actually happened. An
+// absolute path is resolved against nothing but itself, so naming search paths for it is
+// the MAG-2861 confusion; a relative path *is* probed across configSearchPaths, so claiming
+// otherwise is the same confusion pointed the other way.
 func configNotFoundMessage(target string, isFile bool) string {
 	if isFile {
 		if info, err := os.Stat(target); err == nil && info.IsDir() {
 			return "the given config path is a directory, not a config file: " + target
+		}
+		if !filepath.IsAbs(target) {
+			return "config file not found: " + target +
+				" — a relative path is resolved against each of the search paths"
 		}
 		return "config file not found at the given path: " + target
 	}
@@ -194,7 +204,9 @@ func configNotFoundMessage(target string, isFile bool) string {
 
 // configLocationAttributes describe where the config was looked for, for attachment to a
 // not-found error. A relative target is reported alongside the working directory it was
-// resolved against, since that is the piece an operator cannot see from the log line.
+// resolved against and the full list of paths actually probed, since neither is visible
+// from the log line — and reporting only the working directory would describe a narrower
+// search than the one that ran.
 func configLocationAttributes(target string, isFile bool) []utils.Attribute {
 	cwd, err := os.Getwd()
 	if err != nil {
@@ -205,11 +217,20 @@ func configLocationAttributes(target string, isFile bool) []utils.Attribute {
 		if abs, err := filepath.Abs(target); err == nil {
 			resolved = abs
 		}
-		return []utils.Attribute{
+		attributes := []utils.Attribute{
 			{Key: "config_file", Value: target},
 			{Key: "resolved_path", Value: resolved},
 			{Key: "working_directory", Value: cwd},
 		}
+		if !filepath.IsAbs(target) {
+			// The exact candidate list, not a restatement of configSearchPaths: it is what
+			// was probed, and it saves the operator from having to perform the join in
+			// their head to see where the file was expected.
+			attributes = append(attributes, utils.Attribute{
+				Key: "searched_paths", Value: strings.Join(configFilePathCandidates(target), " "),
+			})
+		}
+		return attributes
 	}
 	return []utils.Attribute{
 		{Key: "config_name", Value: target},
--- a/protocol/rpcsmartrouter/config_source.go
+++ b/protocol/rpcsmartrouter/config_source.go
@@ -1,0 +1,138 @@
+package rpcsmartrouter
+
+import (
+	"errors"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/magma-Devs/smart-router/utils"
+	"github.com/spf13/viper"
+)
+
+// Config resolution for the commands that take one — `smartrouter [config]` and
+// `smartrouter health [config]` — lives here so the two cannot drift apart.
+//
+// Viper locates a config in one of two mutually exclusive ways, and the argument used to
+// be handed to the wrong one for path-shaped values. SetConfigName treats its value as a
+// *name* joined onto every search path, and filepath.Join swallows the leading separator
+// of an absolute path — Join(".", "/etc/smartrouter/config.yml") is
+// "etc/smartrouter/config.yml". An absolute path was therefore looked for *inside* each
+// search directory, could never exist there, and failed against directories the operator
+// never named (MAG-2861). Relative paths worked only by accident, because "." happens to
+// be a search path.
+//
+// So a path-shaped argument now goes to SetConfigFile and is loaded verbatim, while a
+// bare name keeps the search-path lookup it has always had.
+
+// configSearchPaths are the directories a bare config *name* is resolved against, in
+// precedence order.
+var configSearchPaths = []string{".", "./config", lavaDefaultNodeHome}
+
+// defaultConfigType is the format assumed for a config whose extension does not declare
+// one — including the extension-less names the search-path lookup accepts. Every
+// smartrouter config is YAML.
+const defaultConfigType = "yml"
+
+// pointViperAtConfig aims the global viper at the config selected by the command's
+// arguments. Exactly one argument names the config; anything else (none, or inline
+// endpoint triplets) selects the default config name.
+//
+// It reports the target it settled on, for use in messages, and whether that target is an
+// exact file rather than a name searched for across configSearchPaths — which decides how
+// a failure to read it should be explained.
+func pointViperAtConfig(args []string) (target string, isFile bool) {
+	target = DefaultRPCSmartRouterFileName
+	if len(args) == 1 {
+		target = args[0]
+	}
+
+	if isConfigFilePath(target) {
+		viper.SetConfigFile(target)
+		// Let a recognized extension declare the format, and fall back to YAML for
+		// anything else. Set it either way: without an explicit type an extension-less
+		// path fails as UnsupportedConfigError instead of being read at all.
+		if ext := strings.TrimPrefix(filepath.Ext(target), "."); slices.Contains(viper.SupportedExts, ext) {
+			viper.SetConfigType(ext)
+		} else {
+			viper.SetConfigType(defaultConfigType)
+		}
+		return target, true
+	}
+
+	viper.SetConfigName(target)
+	viper.SetConfigType(defaultConfigType)
+	for _, path := range configSearchPaths {
+		viper.AddConfigPath(path)
+	}
+	return target, false
+}
+
+// isConfigFilePath reports whether the argument names a config file directly, rather than
+// a name to look up across the search paths.
+//
+// An absolute path always qualifies: the search-path lookup provably cannot resolve one.
+// A relative path qualifies once it actually resolves to a file from the working
+// directory — which is the very file the "." search path would have returned, so every
+// invocation that worked before still loads exactly what it loaded before. Anything that
+// does not resolve to a file stays on the search-path lookup, which is what lets a bare
+// `smartrouter myconfig` still find ./config/myconfig.yml.
+func isConfigFilePath(arg string) bool {
+	if arg == "" {
+		return false
+	}
+	if filepath.IsAbs(arg) {
+		return true
+	}
+	info, err := os.Stat(arg)
+	return err == nil && !info.IsDir()
+}
+
+// isConfigNotFound reports whether err means "no config file was there", which viper
+// phrases two different ways: ConfigFileNotFoundError when it searched and came up empty,
+// and a plain fs.ErrNotExist when it was pointed at an exact file. Both are operator
+// mistakes worth a clean message; anything else (malformed YAML, a permissions problem) is
+// not.
+func isConfigNotFound(err error) bool {
+	var notFound viper.ConfigFileNotFoundError
+	return errors.As(err, &notFound) || errors.Is(err, fs.ErrNotExist)
+}
+
+// configNotFoundMessage explains where the config was looked for, in the terms the
+// operator used: the one path they named, or the search paths their bare name was
+// resolved against.
+func configNotFoundMessage(target string, isFile bool) string {
+	if isFile {
+		return "config file not found at the given path: " + target
+	}
+	return "no config file found — pass a config file as an argument (e.g. `smartrouter <config-file>.yml`, " +
+		"absolute or relative), or place " + DefaultRPCSmartRouterFileName + " in one of the search paths"
+}
+
+// configLocationAttributes describe where the config was looked for, for attachment to a
+// not-found error. A relative target is reported alongside the working directory it was
+// resolved against, since that is the piece an operator cannot see from the log line.
+func configLocationAttributes(target string, isFile bool) []utils.Attribute {
+	cwd, err := os.Getwd()
+	if err != nil {
+		cwd = "<unknown>"
+	}
+	if isFile {
+		resolved := target
+		if abs, err := filepath.Abs(target); err == nil {
+			resolved = abs
+		}
+		return []utils.Attribute{
+			{Key: "config_file", Value: target},
+			{Key: "resolved_path", Value: resolved},
+			{Key: "working_directory", Value: cwd},
+		}
+	}
+	return []utils.Attribute{
+		{Key: "config_name", Value: target},
+		{Key: "search_paths", Value: strings.Join(configSearchPaths, " ")},
+		{Key: "working_directory", Value: cwd},
+	}
+}
--- a/protocol/rpcsmartrouter/config_source.go
+++ b/protocol/rpcsmartrouter/config_source.go
@@ -50,6 +50,7 @@ func pointViperAtConfig(args []string) (target string, isFile bool) {
 	}
 
 	if isConfigFilePath(target) {
+		target = resolveConfigFilePath(target)
 		viper.SetConfigFile(target)
 		// Let a recognized extension declare the format, and fall back to YAML for
 		// anything else. Set it either way: without an explicit type an extension-less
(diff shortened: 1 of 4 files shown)


AnnaR-prog:
## Review: request changes — one real regression, otherwise a good fix

Verified locally at `c28d095`. `go build ./...`, `go vet`, `gofmt` clean, and the full `protocol/rpcsmartrouter` suite passes (87s). The root-cause analysis is right — `filepath.Join(".", "/abs/x.yml")` really does swallow the separator — and the absolute-path fix is correct and well covered.

But the lexical split narrows resolution for **relative** path-shaped arguments. I proved it with in-process old-vs-new probes rather than by reading.

---

### 1. Path-shaped relative args lose two of the three search paths

Same argument, same file on disk, `main`'s logic vs `pointViperAtConfig`:

| arg | file on disk | `main` | this PR |
|---|---|---|---|
| `smartrouter_examples/eth.yml` | `config/smartrouter_examples/eth.yml` | loads | `open smartrouter_examples/eth.yml: no such file or directory` |
| `sub/router.yml` | `$HOME/.lava/sub/router.yml` | loads | not found |

Anything containing a separator now goes to `SetConfigFile` and resolves against the working directory only — `./config` and `$HOME/.lava` are no longer consulted for it.

The probe, if you want to reproduce (drop it in `protocol/rpcsmartrouter`, it passes either way and just logs):

```go
func TestProbe(t *testing.T) {
	dir := t.TempDir()
	t.Chdir(dir)
	os.MkdirAll(filepath.Join(dir, "config", "smartrouter_examples"), 0o755)
	os.WriteFile(filepath.Join(dir, "config", "smartrouter_examples", "eth.yml"), []byte("marker: found\n"), 0o644)

	arg := "smartrouter_examples/eth.yml"

	viper.Reset() // main's logic
	viper.SetConfigName(arg)
	viper.SetConfigType("yml")
	viper.AddConfigPath(".")
	viper.AddConfigPath("./config")
	viper.AddConfigPath(lavaDefaultNodeHome)
	t.Logf("OLD: err=%v used=%q", viper.ReadInConfig(), viper.ConfigFileUsed())

	viper.Reset() // this PR
	target, isFile := pointViperAtConfig([]string{arg})
	t.Logf("NEW: target=%q isFile=%v err=%v", target, isFile, viper.ReadInConfig())
}
```

The router command's own `Example` documents exactly this shape (`rpcsmartrouter.go:2586`):

```
rpcsmartrouter smartrouter_examples/full_smartrouter_example.yml --cache-be "127.0.0.1:7778"
```

So this checklist line doesn't hold as written:

> confirmed `!` in the type prefix if API or client breaking change — not applicable, strictly additive: every argument form that resolved before resolves to the same file

**Scope, in fairness:** arguments starting with `config/` resolve from the working directory and are completely unaffected. The health command's Example and every `scripts/pre_setups/` launcher pass `config/smartrouter_examples/...`, so all of them still work — which is why the manual matrix in the description didn't surface this. The break is specifically arguments whose first segment is not `config/`.

It also contradicts the invariant the description rests on:

> **The two branches resolve an argument identically** — the only thing the branch changes is which error message you get when nothing is there.

They don't. The name branch tries `{., ./config, $HOME/.lava} × {exts, bare}`; the path branch tries `{cwd} × {exts, bare}`.

**Suggested fix**, which restores the invariant rather than reverting the split: have `resolveConfigFilePath` probe `arg`, then `filepath.Join("./config", arg)`, then `filepath.Join(os.ExpandEnv(lavaDefaultNodeHome), arg)` — each through the extension loop it already has — and fall back to `arg` so a miss still names what the operator typed. Absolute paths gain nothing from search paths, so this is purely additive to the relative branch and does not touch what MAG-2861 fixed.

(Note `lavaDefaultNodeHome` is the literal string `"$HOME/.lava"`. Viper's `absPathify` expands it; a manual `filepath.Join` will not, hence the `os.ExpandEnv`.)

### 2. A path argument that is a directory now takes the fatal path

`smartrouter ./config`, and this repo has a `config/` directory:

```
OLD: Config File "./config" Not Found in "[... ...]"   isConfigNotFound=true
NEW: read ./config: is a directory                     isConfigNotFound=false
```

`isConfigNotFound=false` means it falls through to `LavaFormatFatal` and dumps a stack trace — the exact failure mode this PR set out to remove, in a narrow case.

`isExistingFile` correctly excludes directories in the candidate loop, but the `return arg` fallback hands one straight to viper. One line either way: don't fall back to a path that stats as a directory, or fold it into `isConfigNotFound`.

### 3. `TestConfigSearchPathsMatchDocumentedHelp` now guards a half-true claim

It asserts the `Long` text names every entry in `configSearchPaths`. After this change those paths apply only to bare names — the help says a bare name is looked up in the three directories, but nothing says a relative *path* is excluded from them. That is the documentation gap finding 1 hides behind, and worth fixing in the same pass whichever way finding 1 goes.

### 4. Nit

The description attributes a `Long` update to the `health_cmd.go` paragraph, but that file's `Long` is unchanged — its diff is a single hunk in `collectHealthProviders`. Only `rpcsmartrouter.go`'s `Long` changed, so the health command's help still doesn't document how its config argument resolves.

- `resolveConfigFilePath` faithfully mirrors viper's `searchInPath`: extensions in `SupportedExts` order first, bare path last, and `isExistingFile` matches viper's `exists` in excluding directories.
- `isConfigNotFound` covering both `ConfigFileNotFoundError` and `fs.ErrNotExist` is the right call, and permission errors correctly stay on the loud path.
- Choosing lexical classification over `os.Stat` is right, and the `router` / `router.yml` shadowing case pins it properly — good catch in the earlier review round.
- `$HOME/.lava` does expand correctly through viper's `absPathify` — verified live, not assumed.

avitenzer:
All four accepted and fixed in `062e9e6`. Findings 1 and 2 were real regressions — thank you, both would have shipped.

I reproduced each one before touching the code, then verified the fix against binaries built from `c28d095` and `062e9e6` in clean worktrees.

## 1. Relative paths losing the search paths — fixed

Confirmed exactly as you described, and worth stating the mechanism precisely because it is the same `filepath.Join` quirk cutting the other way: `Join` swallows only a **leading** separator, so it destroyed absolute paths while handling relative ones perfectly. `SetConfigName` + `AddConfigPath` really was resolving `smartrouter_examples/eth.yml` through `./config`. My root-cause write-up treated the whole `SetConfigName` route as broken when only half of it was.

Live, from the repo root:

| | `c28d095` | `062e9e6` |
|---|---|---|
| `smartrouter smartrouter_examples/smartrouter_eth.yml` | `config file not found at the given path` | loads |

I took your suggested fix. `resolveConfigFilePath` now probes a relative path against `configSearchPaths` in the same precedence order a bare name uses, each candidate through the extension loop it already had. Two notes on the implementation:

- Your `os.ExpandEnv` point was load-bearing and I would have missed it — `lavaDefaultNodeHome` is the literal `"$HOME/.lava"`, so without expansion the join resolves against a directory named `$HOME`. There is a test for that path specifically.
- I guarded the join on `filepath.IsAbs`. It is harmless without the guard, since candidates are stat-checked and `config/etc/foo.yml` never matches — but re-creating the demotion shape MAG-2861 removed is not something I want sitting in the code unremarked. The new test plants a decoy at exactly the location the demotion would have found, so MAG-2861 is now pinned as an invariant rather than as a symptom.

Your scope analysis was right and saved me from over-correcting: every `scripts/pre_setups/` launcher passes `$PROJECT_ROOT/config/...` or `config/...`, so nothing in-repo was broken — which is precisely why my manual matrix missed it. The checklist's "strictly additive" line is now true; it was not true of `c28d095`, and I have said so in the description rather than quietly repairing it.

**One correction to the finding, in the interest of the record:** the `Example` you cite names `smartrouter_examples/full_smartrouter_example.yml`, and that file does not exist in `config/smartrouter_examples/` — so nobody was running that literal command, and the citation is weaker than it reads. It does not rescue anything: twelve real configs sit at that path and I reproduced the break with `smartrouter_eth.yml`, which does exist. I have pointed the `Example` at a file that is actually in the repo.

## 2. Directory argument taking the fatal path — fixed

Reproduced, including the consequence you inferred:

```
c28d095:  smartrouter ./config  ->  stack-dump lines: 1
062e9e6:  smartrouter ./config  ->  stack-dump lines: 0
```

Viper returns `*fs.PathError` wrapping `syscall.EISDIR`, so `errors.Is(err, syscall.EISDIR)` classifies it cleanly; `isConfigNotFound` now covers all three shapes. I went slightly past your suggestion on the message — "not found at the given path" would have been misleading when something *is* there:

```
ERR the given config path is a directory, not a config file: ./config
    error="read ./config: is a directory"
```

I kept the `return arg` fallback rather than suppressing directories in it, so a miss still names what the operator typed. The directory case is handled where it is actually observable, in the classification.

## 3. Half-true documented claim — fixed

Right, and it was the gap finding 1 hid behind. Both `Long` texts now say a relative path is resolved against the search paths too, and `TestConfigSearchPathsMatchDocumentedHelp` covers **both** commands and asserts the relative-path claim is present — it previously guarded only `rpcsmartrouter`, which is part of why nothing caught this.

## 4. `health_cmd.go` `Long` misattribution — correct, fixed

You are right, and I checked `main` rather than take it on trust: the text I quoted (`all configs should be located in the local running directory /config or $HOME/.lava`) is `rpcsmartrouter.go:2578` on `main`, not health's. Health's `Long` was untouched and its config argument was undocumented. It is documented now, and the description no longer claims otherwise.

## Verification

Four of the six new subtests fail against a faithful revert of `c28d095`'s resolution logic — the three search-path cases and the directory case — so they are regression tests rather than tests that pass either way. The other two (cwd-beats-`./config` for a path, absolute-not-joined) pass in both directions by design; they are invariant guards, and I would rather say that than imply a coverage number I did not earn.

25 subtests, full `protocol/rpcsmartrouter` suite green (86s), `go build ./...`, `go vet`, `gofmt` clean on the touched files. Verified from a clean worktree checkout of the committed sha, not the working directory.

Ready for another look.

avitenzer:
Both fixed in `88d3074`. P2 was a real compatibility break and I would not have found it — the short-circuit reads as obviously correct.

## P2 — supported-extension paths now follow viper's suffix order

Confirmed at the source, not just by the repro. `searchInPath` (viper v1.21.0, `file.go:75`) appends every supported extension to `v.configName` **unconditionally** — it never checks whether the name already ends in one:

```go
for _, ext := range SupportedExts {
    if b, _ := exists(v.fs, filepath.Join(in, v.configName+"."+ext)); b {
        return filepath.Join(in, v.configName+"."+ext)
    }
}
if v.configType != "" { /* bare name last */ }
```

So `SetConfigName("sub/router.yml")` really did try `sub/router.yml.json` before `sub/router.yml`. Your reproduction, measured:

| | `062e9e6` | `88d3074` |
|---|---|---|
| `sub/router.yml` with `sub/router.yml.json` on disk | `open sub/router.yml: no such file or directory` | loads `sub/router.yml.json` |

I took the mirror option rather than documenting the break. The layout is pathological and I would not design it this way, but the PR's entire thesis is "strictly additive" and "the branches resolve identically", and a two-line short-circuit is not worth qualifying that with a footnote. If we later want *explicit path means exactly that file*, that is a deliberate semantic change and deserves its own ticket rather than arriving as a side effect of a bug fix.

Two tests, deliberately paired: one on the path branch, one on the **name** branch — which never stopped appending. Alone the first reads as a quirk; together they assert the equivalence the description claims.

I also checked that `findConfigFileOld` is the live path — the locafero `Finder` is behind the opt-in `ExperimentalFinder()` option, which we do not set — so `searchInPath` is what we are actually mirroring, not a superseded code path.

## P3 — diagnostics now describe the search that ran

You are right, and the test comment was the worse half: it asserted the search paths "were never consulted", which was load-bearing justification for the *absolute* case that I left attached to the *relative* one when relative paths went back through the search paths. It documented an intent that had stopped being true.

Split on `filepath.IsAbs`, and I took your "ideally report the candidates" over the roots — the roots would still make the operator perform the join themselves:

```
ERR config file not found: nested/missing.yml — a relative path is resolved against each of the search paths
    searched_paths="nested/missing.yml config/nested/missing.yml /Users/x/.lava/nested/missing.yml"
```

Absolute paths are untouched: still `config file not found at the given path`, no candidate list, with a test asserting none is attached — listing search paths for an absolute path is the MAG-2861 confusion pointed the other way, and I did not want to fix one direction by breaking the other.

## Verification

Both new subtests fail against a faithful revert of `062e9e6`'s logic, so they are regression tests. The name-branch companion passes in both directions by design — it is the equivalence anchor, not a regression test, and I would rather name that than let it pad a count.

28 subtests; full `protocol/rpcsmartrouter` suite green (90s); `go build ./...`, `go vet ./protocol/rpcsmartrouter/...`, `gofmt` clean on touched files. Live matrix re-run against a binary built from a clean worktree checkout of `88d3074`: all four real-world argument forms still load, the relative miss names its three candidates, the absolute miss does not.

On **BEHIND**: I have deliberately not rebased. Your synthetic-merge check against `4f409137` is the same signal, and force-pushing mid-review would invalidate the shas cited throughout this thread for no gain. Happy to update the branch whenever you want it done — say the word and I will.

AnnaR-prog (approved):

PR #303, opened 2026-08-19
fix(smart-router): cap response size on the gRPC block-extraction path (MAG-2557)

avitenzer:
# Description

`extractBlockHeightFromGRPCResponse` parsed whatever the upstream sent, with no size
limit. Its JSON-RPC sibling `extractBlockHeightFromJSONResponse` has capped at 1 MB since
it was written; this path never got the same guard.

This PR adds the guard on the gRPC path, and — following review from @nimrod-teich —
lands it as a **single cap shared by both transports**, with a value derived from the
largest block a chain can legally produce rather than picked.

Two earlier revisions were wrong and are worth recording, because the corrections are the
substance of the change:

1. The first reused the JSON-RPC 1 MB constant. That copies the number and inverts its
   meaning — on gRPC the largest response *is* the block source.
2. The second split it into a separate 32 MB gRPC constant. That fixed the value but split
   on the wrong axis: what differs between JSON-RPC and gRPC is the encoding, not the size
   of the block we want to read.

## The cap, and why it is that number

`maxResponseSizeForBlockExtraction` = **32 MB**, used by both extractors.

A block-extraction cap has to sit **above** the largest response we legitimately want to
parse, because on both transports the biggest response is the block source:

* **gRPC** — `GET_BLOCKNUM` resolves to
  `cosmos.base.tendermint.v1beta1.Service/GetLatestBlock` for AKASH, BABYLON, COSMOSSDK,
  DYDX, KAVA and SEI, and `GET_BLOCK_BY_NUM` to `GetBlockByHeight`: the entire block,
  every tx included.
* **JSON-RPC** — `GET_BLOCK_BY_NUM` is `eth_getBlockByNumber`, equally a
  parse-directive-tagged block source.

Measured live over recent blocks (JSON sizes; the gRPC guard measures proto, which is
smaller):

| chain | median | max observed | ceiling |
|---|---|---|---|
| Ethereum mainnet (`eth_getBlockByNumber`, full txs) | 463 KB | 668 KB | gas-bound |
| Base (same) | 385 KB | 577 KB | gas-bound |
| Arbitrum (same) | 13 KB | 52 KB | gas-bound |
| dYdX (`GetLatestBlock`) | 27 KB | 74 KB | **4.0 MB** `block.max_bytes` |
| Osmosis (same) | 35 KB | 138 KB | **2.9 MB** |
| Sei (same) | 11 KB | 13 KB | n/a |
| Cosmos Hub | — | — | **1.9 MB** |

The ceiling to clear is the chain's consensus `block.max_bytes` — and Tendermint's own
default is 21 MB. **32 MB is the next power of two above that default.** A cap below it
fires only on full blocks: during congestion, silently, exactly when tip accuracy matters
most. A cap belongs above the range of legitimate responses, not inside it.

### Why not the transport's own limit

The natural simplification is to reuse `chainproxy.MaxCallRecvMsgSize` (512 MB) and keep
one number in the process. It does not work: gRPC refuses to decode a message above that
limit before extraction is ever reached (`direct_rpc_connection.go:906`; `Data` is then a
re-marshal of what already decoded), so `len(responseData) > 512MB` is a branch no input
can take. It would read as a guard and behave as none, with a test that can only ever
exercise the negative arm. `cap stays below the transport receive limit` pins this so the
collapse fails loudly instead of silently disarming the guard.

## The review also surfaced a live bug on the JSON-RPC side

The 1 MB JSON-RPC cap was assumed correct and is not. Per the table above,
`eth_getBlockByNumber` with full transaction objects runs 463 KB median / 668 KB peak on
Ethereum mainnet across ten recent blocks — a quiet sample already inside 1.5x of the cap.
**That guard has been dropping tip observations on busy Ethereum and Base blocks in
production.** Unifying at 32 MB fixes that instance too; it is not merely a tidy-up to
close the review thread.

One correction to the old comment while in here: it claimed the 1 MB cap protected against
`debug_traceTransaction` / `trace_replayBlockTransactions`. It never did —
`extractBlockHeightFromEVMResponse` is a switch over named methods with no default branch,
so those return without unmarshalling at any size.

## Scope: what this guard does and does not buy

Stated precisely, because an earlier write-up of this PR overstated it:

* **It does not prevent the large allocation.** By the time extraction runs,
  `response.Data` is already fully materialized — gRPC received and allocated it under
  `MaxCallRecvMsgSize`. The guard removes the *downstream* expansion (`proto.Unmarshal`
  into a `dynamic.Message`, `MarshalJSON`, then a `[]byte` copy — roughly a 3–4x
  multiplier), not the base allocation. A hostile 512 MB response still lands 512 MB in the
  heap with this merged. If heap-driven pod death is the concern, the lever is
  `MaxCallRecvMsgSize` (`chainproxy/connector.go:36`), not this cap. Worth filing
  separately rather than closing MAG-2557 believing that vector is shut.
* **On gRPC it is defense in depth, not the primary filter.** Extraction returns early
  unless the method carries a spec `parse_directive`, and only ~17 of those exist across
  all 55 gRPC collections in the catalog. An untagged multi-MB response never reaches the
  expansion at any size — the nil-check already gates it.
* **The one method it genuinely bounds is `eth_getLogs`**, which unmarshals its full array
  to read a block number off the first element. Raising the JSON-RPC cap from 1 MB to 32 MB
  widens that. Flagged rather than buried: the real fix is not unmarshalling a whole log
  array for one field, which is follow-up work, not this PR.

## Skipping extraction is safe — verified, not assumed

When the guard fires, `Reply.LatestBlock` is 0 and the relay contributes no tip
observation. The MAG-2159 traffic gate (`freshRelayTip`, `endpointstate/observation.go:275`)
suppresses a dedicated poll **only** when the freshest observation is `SourceRelay` and
fresh. A skipped extraction records nothing, so the gate opens and the ChainTracker poll
runs — tip tracking degrades to poll cadence rather than freezing.

Two smaller consequences of `LatestBlock == 0`, for completeness: the
`Provider-Latest-Block` response header is omitted (`rpcsmartrouter_server.go:4518`), and
for a `LATEST_BLOCK` request the cache write falls back to `SeenBlock` and is skipped
entirely if that is also 0 (`rpcsmartrouter_server.go:3272-3284`). Both degrade gracefully,
and at 32 MB the guard should effectively never fire on a legitimate block.

## Also in this PR

`upstream_grpc_pool.go` used a bare `512*1024*1024` rather than
`chainproxy.MaxCallRecvMsgSize`. It dials the same upstreams as `chainproxy.GRPCConnector`,
so a divergence between them would mean the same endpoint accepting a response on one path
and rejecting it on the other. Now uses the shared constant.

Two further copies of that limit remain out of scope here: `grpcproxy/grpcproxy.go:22`
declares its own const (server-side listener, arguably a legitimately separate knob) and
`performance/connection/connection_cmd.go:49` is another literal.

## Most critical to review

* `protocol/rpcsmartrouter/direct_rpc_relay.go` — the constant and the two guard sites.

## Testing

`TestBlockExtractionResponseSizeGuard` drives **both extractors through the same
constant** — that shared drive is the assertion, so a future re-split by transport fails
here rather than passing quietly. Each gets an over-cap and an exactly-at-cap case. It
asserts on a `GetParseDirective` call count rather than the returned height: extraction
returns 0 both when it skips and when it parses and finds nothing, so only the call count
separates a guard that fired from one that was never there. The exactly-at-cap arm pins the
`>` boundary so a later refactor can't silently turn it into `>=`.

Two sizing assertions bound the constant from both directions, because the regression to
defend against is not "the guard is missing" but "the guard is set wrong":

* `cap clears the consensus block ceiling` — the cap exceeds Tendermint's default
  `block.max_bytes` (21 MB). Catches a re-unification down to the old 1 MB.
* `cap stays below the transport receive limit` — the cap is strictly under
  `chainproxy.MaxCallRecvMsgSize`. Catches a collapse to the transport number, which would
  make the branch unreachable.

Verified by disabling the guard: both over-cap arms fail on the call count, and both pass
again restored. Full `protocol/rpcsmartrouter` package suite passes (90.0s); `go build ./...`
and `go vet` clean. Build and tests re-run from a clean `git worktree` checkout of the
pushed sha, not the working directory.

Two caveats, stated plainly:

* this is a Go unit test, so per our standing rule it is not nightly regression coverage;
* a live wire reproduction still needs the gRPC streaming simulator support that the ticket
  itself flags as pending. The bug was filed from source review, and the fix is verified the
  same way plus the revert check above. The sizing evidence, though, is live-measured
  against public Ethereum/Base/Arbitrum and dYdX/Osmosis/Sei gRPC endpoints.

Note the ticket locates the code at `protocol/lavasession/direct_rpc_relay.go:215`. The file
actually lives at `protocol/rpcsmartrouter/direct_rpc_relay.go`; the line number is exact.

## Jira ticket

Jira ticket: MAG-2557

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change — n/a, no API or client change
* [x] updated the relevant documentation or specification, including comments for documenting Go code — the constant carries its own derivation and both guard sites point at it
* [x] confirmed all CI checks have passed

## Reviewers Checklist

I have...

### DIFF
--- a/protocol/rpcsmartrouter/direct_rpc_integration_test.go
+++ b/protocol/rpcsmartrouter/direct_rpc_integration_test.go
@@ -868,7 +868,9 @@ func TestExtractBlockHeightFromJSONResponse_EVMFallback(t *testing.T) {
 // to JSON and copied — several full-size allocations off one upstream reply.
 //
 // Both paths are asserted together because the fix is a parity claim: the guard has to
-// hold on gRPC *and* stay on JSON-RPC, at the same threshold and the same boundary.
+// hold on gRPC *and* stay on JSON-RPC. Parity of protection, NOT parity of number — each
+// extractor is driven by its own cap, because the transports guard inverted situations
+// (see maxGRPCResponseSizeForBlockExtraction).
 //
 // The assertion is parseDirectiveCalls, not the returned height. Extraction returns 0
 // whether it skipped the response or parsed it and found nothing, so the return value
@@ -886,21 +888,22 @@ func TestBlockExtractionResponseSizeGuard(t *testing.T) {
 
 	extractors := []struct {
 		name    string
+		limit   int
 		extract func([]byte, chainlib.ChainMessage) int64
 	}{
-		{"grpc", extractBlockHeightFromGRPCResponse},
-		{"jsonrpc", extractBlockHeightFromJSONResponse},
+		{"grpc", maxGRPCResponseSizeForBlockExtraction, extractBlockHeightFromGRPCResponse},
+		{"jsonrpc", maxResponseSizeForBlockExtraction, extractBlockHeightFromJSONResponse},
 	}
 
 	for _, extractor := range extractors {
 		t.Run(extractor.name+"/over cap is skipped without parsing", func(t *testing.T) {
 			msg := newMsg()
-			oversized := make([]byte, maxResponseSizeForBlockExtraction+1)
+			oversized := make([]byte, extractor.limit+1)
 
 			assert.Equal(t, int64(0), extractor.extract(oversized, msg))
 			assert.Zero(t, msg.parseDirectiveCalls,
 				"response over the %d byte cap must be skipped before any parsing work",
-				maxResponseSizeForBlockExtraction)
+				extractor.limit)
 		})
 
 		// Boundary: the guard is `>`, so a response of exactly the cap still parses.
@@ -908,7 +911,7 @@ func TestBlockExtractionResponseSizeGuard(t *testing.T) {
 		// dropping block tracking for responses that are within budget.
 		t.Run(extractor.name+"/exactly at cap still parses", func(t *testing.T) {
 			msg := newMsg()
-			atCap := make([]byte, maxResponseSizeForBlockExtraction)
+			atCap := make([]byte, extractor.limit)
 
 			// The payload is zero bytes, so parsing yields no height — the point is
 			// that extraction was attempted at all.
@@ -917,4 +920,17 @@ func TestBlockExtractionResponseSizeGuard(t *testing.T) {
 				"response exactly at the cap must not be skipped")
 		})
 	}
+
+	// The regression this whole test exists to prevent is not "the guard is missing" but
+	// "the guard is set too low". On gRPC the biggest response IS the block source:
+	// GET_BLOCKNUM is cosmos.base.tendermint.v1beta1.Service/GetLatestBlock, which returns
+	// the entire block. A cap below the chain's consensus block.max_bytes fires only on
+	// full blocks — during congestion, silently, exactly when tip accuracy matters most.
+	// Tendermint's default max_bytes is the ceiling to clear; re-unifying this constant
+	// with the 1 MB JSON-RPC number would land far under it.
+	t.Run("grpc cap clears the consensus block ceiling", func(t *testing.T) {
+		const tendermintDefaultMaxBytes = 22020096 // 21 MB, Tendermint's default block.max_bytes
+		assert.Greater(t, maxGRPCResponseSizeForBlockExtraction, tendermintDefaultMaxBytes,
+			"gRPC block-extraction cap must exceed the largest block a chain can legally produce")
+	})
 }
--- a/protocol/rpcsmartrouter/direct_rpc_integration_test.go
+++ b/protocol/rpcsmartrouter/direct_rpc_integration_test.go
@@ -677,6 +677,13 @@ type mockChainMessage struct {
 	// spectypes.LATEST_BLOCK for a latest-requesting method, or a concrete height for a
 	// historical request (used by the MAG-2159 tip-eligibility tests).
 	requestedBlock int64
+	// parseDirective is returned by GetParseDirective(); the nil default preserves
+	// the "no spec parsing" behaviour older tests rely on.
+	parseDirective *spectypes.ParseDirective
+	// parseDirectiveCalls counts GetParseDirective() calls. The MAG-2557 size-guard
+	// tests read it to prove extraction bailed out before touching the parse
+	// machinery — the return value alone can't tell a skip from a failed parse.
+	parseDirectiveCalls int
 }
 
 func (m *mockChainMessage) GetRPCMessage() rpcInterfaceMessages.GenericMessage {
@@ -734,8 +741,11 @@ func (m *mockChainMessage) GetRequestedBlocksHashes() []string                  
 func (m *mockChainMessage) UpdateEarliestInMessage(incomingEarliest int64) bool { return false }
 func (m *mockChainMessage) SetExtension(extension *spectypes.Extension)         {}
 func (m *mockChainMessage) GetUsedDefaultValue() bool                           { return false }
-func (m *mockChainMessage) GetParseDirective() *spectypes.ParseDirective        { return nil }
-func (m *mockChainMessage) IsBatch() bool                                       { return false }
+func (m *mockChainMessage) GetParseDirective() *spectypes.ParseDirective {
+	m.parseDirectiveCalls++
+	return m.parseDirective
+}
+func (m *mockChainMessage) IsBatch() bool { return false }
 
 type mockGenericMessage struct {
 	data []byte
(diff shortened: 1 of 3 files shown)


nimrod-teich:
Is the 1MB limit is enough?

avitenzer:
Good catch — no, 1 MB was not enough, and the reason is sharper than a margin call. Fixed in aec0e33.

The problem was that I reused the JSON-RPC constant, which copies the number but inverts its meaning. The two transports guard opposite situations:

- On JSON-RPC the oversized responses are `debug_traceTransaction` / `trace_replayBlockTransactions` — never block sources. "Too big to parse" and "nothing to extract anyway" overlap almost perfectly, so 1 MB sits above everything we actually want to read.
- On gRPC the biggest response **is** the block source. `GET_BLOCKNUM` resolves to `cosmos.base.tendermint.v1beta1.Service/GetLatestBlock` for AKASH, BABYLON, COSMOSSDK, DYDX, KAVA and SEI, and `GET_BLOCK_BY_NUM` to `GetBlockByHeight` — methods returning the whole block, every tx included.

I measured 30 recent blocks per chain against public gRPC endpoints (JSON sizes; the guard measures proto, which is smaller):

| chain | median | max observed | consensus `block.max_bytes` |
|---|---|---|---|
| dYdX | 27 KB | 74 KB | **4.0 MB** |
| Osmosis | 35 KB | 138 KB | **2.9 MB** |
| Sei | 11 KB | 13 KB | n/a |
| Cosmos Hub | — | — | **1.9 MB** |

So the typical case had 7–100x headroom, but every one of these chains permits a block 2–4x *above* the cap, and Tendermint's default `max_bytes` is 21 MB. A 1 MB cap fires only on full blocks — during congestion, silently, exactly when tip accuracy matters most. A cap belongs above the range of legitimate responses, not inside it.

Split the constant: JSON-RPC keeps 1 MB, gRPC gets 32 MB — clears every legal block with margin, still 16x under `MaxCallRecvMsgSize` (512 MB). Added a test pin asserting the gRPC cap exceeds Tendermint's default `max_bytes`, since the regression to defend against is "set too low", not "missing".

Two things I corrected in the PR body while in here:

1. **The original memory claim was overstated.** By the time extraction runs, `response.Data` is already fully materialized under `MaxCallRecvMsgSize`. This guard removes the downstream expansion (~3–4x), not the base allocation — a hostile 512 MB response still lands 512 MB in the heap with this merged. If that vector matters, the lever is `MaxCallRecvMsgSize`, and I'd rather file it separately than close MAG-2557 implying it's shut.
2. **The guard is defense in depth, not the primary filter.** Extraction returns early unless the method carries a spec `parse_directive` — only ~17 exist across all 55 gRPC collections — so an untagged multi-MB response never reaches the expansion at any size regardless of the cap.

One unrelated thing worth a look separately: `EndpointPoller.FetchLatestBlockNum` (`endpoint_poller.go:168-179`) parses the same `GetLatestBlock` response through `FormatResponseForParsing` + `ParseBlockFromReply` with **no** cap, on a cadence. Whatever we decide about capping the relay path, the poll path is currently unbounded.

nimrod-teich:
I still think this is an arbitrary choice of size. why don't you just use the overall limit instead of maintaining multiple limits?

avitenzer:
Both halves of that land — fixed in `4f7db28`.

**One cap.** Collapsed to a single `maxResponseSizeForBlockExtraction` used by both extractors. You're right that transport was the wrong axis to split on: what differs between JSON-RPC and gRPC is the encoding, not the size of the block we want to read. Two knobs for one decision.

**On "just use the overall limit"** — reading that as `MaxCallRecvMsgSize` (512 MB), that specific collapse doesn't work, and it's worth writing down because it's the natural next suggestion. gRPC refuses to decode a message above that limit before extraction is ever reached (`direct_rpc_connection.go:906`; `Data` is then a re-marshal of what already decoded), so `len(responseData) > 512MB` is a branch no input can take. It would read as a guard and behave as none, with a test that can only ever exercise the negative arm. I added `cap stays below the transport receive limit` to pin that, so the collapse fails loudly rather than silently disarming things.

**The number is now derived rather than picked**, which I think was the real objection. A block-extraction cap has to sit above the largest response we legitimately want to parse. That ceiling is the chain's consensus `block.max_bytes` — 4 MB dYdX, 3 MB Osmosis, 2 MB Cosmos Hub, 21 MB Tendermint default. 32 MB is the next power of two above the default. Not "a big round number", and if someone raises the ceiling the test tells them.

**Pushing on this turned up a live bug I'd missed.** I'd framed 1 MB as correct-for-JSON-RPC. It isn't. `GET_BLOCK_BY_NUM` is `eth_getBlockByNumber`, a parse-directive-tagged block source exactly like `GetLatestBlock` is on gRPC. Measured over ten recent blocks with full transaction objects:

| chain | median | max |
|---|---|---|
| Ethereum mainnet | 463 KB | 668 KB |
| Base | 385 KB | 577 KB |
| Arbitrum | 13 KB | 52 KB |

That's a quiet sample and it's already inside 1.5x of the cap. The 1 MB guard has been silently dropping tip observations on busy Ethereum and Base blocks in production. So unifying wasn't a concession to close the thread — it fixed a second instance of the bug.

**One honest cost.** `eth_getLogs` is in the EVM fallback switch and unmarshals its full array to read a block number off the first element. Raising the JSON-RPC cap widens that from 1 MB to 32 MB. I don't think it's a blocker — the real fix is not unmarshalling a whole log array for one field — but it's the one place a single number genuinely serves two jobs imperfectly, and I'd rather flag it than have you find it. Happy to file it.

Worth correcting the old comment while I'm here: it claimed the 1 MB cap protected against `debug_traceTransaction` / `trace_replayBlockTransactions`. It never did — that fallback is a switch over named methods with no default branch, so those return without unmarshalling at any size.

**And you're right about duplication generally.** I swapped `upstream_grpc_pool.go`'s bare `512*1024*1024` for `chainproxy.MaxCallRecvMsgSize`; it dials the same upstreams as `chainproxy.GRPCConnector`, so a divergence there would mean the same endpoint accepting a response on one path and rejecting it on the other. Two copies of the overall limit still remain — `grpcproxy/grpcproxy.go:22` declares its own const (server-side listener, so arguably a legitimately separate knob) and `performance/connection/connection_cmd.go:49` is another literal. Both are outside this PR; say the word and I'll take them in a follow-up.

PR body rewritten to match — the old "Why the two caps differ" section argued for a design that no longer exists.

nimrod-teich (approved):

PR #306, opened 2026-08-19
feat(smart-router): capture Retry-After from a rate-limited upstream

nimrod-teich:
#Closes MAG-2912

Jira ticket: MAG-2912

**Independent of the other rate-limit PRs.** Branches off `main`, merges in any order.

## Why

When an upstream sends `Retry-After` it is telling us when to come back, and the router throws it away — the header is not read anywhere in the codebase. Any caller wanting to slow down has to guess, and a guess fails both ways: too eager and it takes another 429, too cautious and a provider sits unverified far longer than the vendor asked.

## The part that looked expensive, and wasn't

`ValidateStatusCodes(statusCode int, strict bool)` mints the typed 429 and takes no headers — which reads like honouring `Retry-After` means changing a shared helper's signature and rippling through its callers.

It's avoidable. Every site that calls it already has the response in scope:

| site | response |
|---|---|
| `chainproxy/rpcclient/http.go:242` | `resp` |
| `chainlib/rest.go:536` | `res` |
| `chainlib/tendermintRPC.go:738` | `res` |

The header is attached at the call site; `ValidateStatusCodes` is untouched.

## What changed

**`common.RateLimitedError`** carries the duration and **wraps** the error it enriches, so it keeps unwrapping to the existing sentinel:

```go
func (e *RateLimitedError) Unwrap() error { return e.underlying() } // the enriched error, or StatusCodeError429
```

So every existing `errors.Is(err, StatusCodeError429)` keeps matching, whatever context was already wrapped around the sentinel stays reachable, and only callers that want the duration need to know the type exists.

**Both wire forms** RFC 9110 permits are parsed: delta-seconds (`120`) and HTTP-date (`Wed, 21 Oct 2015 07:28:00 GMT`). Anything unparseable, zero, negative, or already past reads as **absent** — a malformed header pushing a caller to retry immediately is the one outcome worse than having no header at all.

**A delay is clamped to `MaxRetryAfter` (1h).** RFC 9110 sets no ceiling, so an upstream — or a proxy in front of it — can name one measured in years, and a caller honouring it parks a provider for that long. The same clamp keeps delta-seconds clear of int64 nanosecond overflow, where a large enough value wrapped to a *negative* duration and reported itself as usable. The ceiling lives at the capture rather than in each backoff, since `RetryAfterFrom` is what every consumer reads.

**The HTTP-date form is measured against the upstream's `Date` header** when it sends one. The date is absolute, so a skewed upstream clock otherwise inflates the wait or backdates its own header into "already past"; `Date` states the clock that wrote the deadline, so the gap between the two is skew-free. Local time stays the fallback.

**The direct-RPC path captures it too.** `HTTPDirectRPCConnection` mints its own 429 from a response it holds, and `recovery_probe.go` branches on that status to call a provider merely busy — a second "come back later" decision made with no upstream input. `HTTPStatusError` now carries the duration and unwraps to the shared rate-limit error, which leaves the existing type assertions working and makes `common.RetryAfterFrom` read the same on both paths.

**A one-line propagation fix** in `rest.go` and `tendermintRPC.go`: both passed the status error as a LavaFormat *attribute* rather than its `err` parameter, which severs the chain and makes the enriched error impossible to unwrap. The same fix appears in PR #304; where they overlap the resolution is identical, so whichever merges second takes a trivial conflict.

## Tests

Unit coverage for both wire forms, every malformed shape, the clamp (including the overflow value), clock skew in both directions, and sentinel survival through further wrapping.

Beyond that, `protocol/chainlib/retry_after_propagation_test.go` drives all three status-code call sites end to end (REST, JSON-RPC, Tendermint URI) against a 429 test server. That is what pins the propagation fix — with the status error passed as an attribute instead of the cause, the log line still reads correctly and nothing unwraps, and each of the three subtests fails.

## Scope

This **captures** the value. Acting on it belongs to whichever backoff wants it, via `common.RetryAfterFrom(err)`.

PR #305 adds an exponential backoff for re-verification; wiring it to prefer `Retry-After` where it is longer is a small follow-up once both are in. I have that change written and tested (a longer request wins, a 1-second one does not shorten a minutes-long interval, an absurd one clamps to the cap) — it is held back only because it cannot compile until both this and #305 exist.

`Retry-After` is also defined on 503, which the router maps to `StatusCodeErrorStrict` and disables endpoints on. Left alone deliberately: a 503 is not a rate limit, so carrying it would mean a second error type with no consumer yet.

## How to verify

```
go build ./... && go vet ./protocol/common/ ./protocol/chainlib/ ./protocol/lavasession/ ./protocol/endpointstate/
go test ./protocol/common/ ./protocol/chainlib/... ./protocol/lavasession/ ./protocol/endpointstate/
go test ./protocol/common/ -run RetryAfter -v
go test ./protocol/chainlib/ -run RetryAfterPropagates -v
```

### DIFF
--- a/protocol/common/retry_after.go
+++ b/protocol/common/retry_after.go
@@ -7,6 +7,14 @@ import (
 	"time"
 )
 
+// MaxRetryAfter is the longest delay an upstream can ask for. RFC 9110 puts no ceiling on
+// the header, so an upstream — or a proxy in front of it — can name one measured in years,
+// and a caller honouring it verbatim parks a provider for that long. Anything above this
+// clamps to it, which also keeps delta-seconds well clear of int64 nanosecond overflow.
+const MaxRetryAfter = time.Hour
+
+const maxRetryAfterSeconds = int(MaxRetryAfter / time.Second)
+
 // RateLimitedError carries what the upstream told us about when to come back.
 //
 // It wraps StatusCodeError429 rather than replacing it, so every existing
@@ -16,16 +24,29 @@ type RateLimitedError struct {
 	// RetryAfter is what the upstream asked for, or 0 when it said nothing. Zero means
 	// "no opinion", never "retry immediately" — callers fall back to their own backoff.
 	RetryAfter time.Duration
+
+	// cause is the rate-limit error being enriched. Whatever context was already wrapped
+	// around the sentinel stays reachable; nil means the bare sentinel.
+	cause error
+}
+
+// underlying is what this error stands in for: the enriched error, or the sentinel when it
+// was built directly from a duration.
+func (e *RateLimitedError) underlying() error {
+	if e.cause != nil {
+		return e.cause
+	}
+	return StatusCodeError429
 }
 
 func (e *RateLimitedError) Error() string {
 	if e.RetryAfter > 0 {
-		return StatusCodeError429.Error() + ", retry after " + e.RetryAfter.String()
+		return e.underlying().Error() + ", retry after " + e.RetryAfter.String()
 	}
-	return StatusCodeError429.Error()
+	return e.underlying().Error()
 }
 
-func (e *RateLimitedError) Unwrap() error { return StatusCodeError429 }
+func (e *RateLimitedError) Unwrap() error { return e.underlying() }
 
 // RetryAfterFrom extracts an upstream-requested delay from an error chain. ok is false when
 // the error is not a rate-limit, or is one that carried no usable Retry-After.
@@ -38,9 +59,10 @@ func RetryAfterFrom(err error) (d time.Duration, ok bool) {
 }
 
 // ParseRetryAfter reads the header in either wire form RFC 9110 permits: delta-seconds
-// ("120") or an HTTP-date ("Wed, 21 Oct 2015 07:28:00 GMT"). Anything unparseable, negative,
-// or already in the past is treated as absent — a malformed header must leave the caller on
-// its own backoff, not push it to retry immediately.
+// ("120") or an HTTP-date ("Wed, 21 Oct 2015 07:28:00 GMT"). Anything unparseable, zero,
+// negative, or already in the past is treated as absent — a malformed header must leave the
+// caller on its own backoff, not push it to retry immediately. A delay longer than
+// MaxRetryAfter clamps to it.
 func ParseRetryAfter(h http.Header, now time.Time) (time.Duration, bool) {
 	if h == nil {
 		return 0, false
@@ -53,14 +75,31 @@ func ParseRetryAfter(h http.Header, now time.Time) (time.Duration, bool) {
 		if secs <= 0 {
 			return 0, false
 		}
+		if secs > maxRetryAfterSeconds {
+			return MaxRetryAfter, true
+		}
 		return time.Duration(secs) * time.Second, true
 	}
 	if t, err := http.ParseTime(raw); err == nil {
-		if d := t.Sub(now); d > 0 {
+		if d := t.Sub(retryAfterReferenceTime(h, now)); d > 0 {
+			if d > MaxRetryAfter {
+				return MaxRetryAfter, true
+			}
 			return d, true
 		}
 	}
 	return 0, false
+}
+
+// retryAfterReferenceTime picks the clock an HTTP-date is measured against. The date is
+// absolute, so a skewed upstream clock either inflates the wait or backdates the header into
+// "already past"; its own Date header states the same clock that wrote the deadline, which
+// cancels the skew. Local time is the fallback when it sends none.
+func retryAfterReferenceTime(h http.Header, now time.Time) time.Time {
+	if date, err := http.ParseTime(h.Get("Date")); err == nil {
+		return date
+	}
+	return now
 }
 
 // WithRetryAfter enriches a rate-limit error with the upstream's Retry-After, if it sent one.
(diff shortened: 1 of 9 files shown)


protocol/common/retry_after.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (e *RateLimitedError) Unwrap() error { return e.underlying() }
+
+// RetryAfterFrom extracts an upstream-requested delay from an error chain. ok is false when
+// the error is not a rate-limit, or is one that carried no usable Retry-After.
+func RetryAfterFrom(err error) (d time.Duration, ok bool) {
+	var rle *RateLimitedError
+	if errors.As(err, &rle) && rle.RetryAfter > 0 {
+		return rle.RetryAfter, true
```
avitenzer, line 56: **Overflow → negative duration returned as `ok=true`.** The `secs <= 0` guard runs before the multiply, so it can't catch the overflow in `time.Duration(secs) * time.Second`:

```
"10000000000" -> -2346317h47m53s  ok=true
"9223372037"  -> -2562047h47m16s  ok=true   (safe max is 9223372036)
```

`RetryAfterFrom` happens to guard `> 0`, but `RetryAfter` is an exported field — `errors.As` + `time.Sleep(rle.RetryAfter)` fires instantly, which is the retry-immediately outcome this file exists to prevent.

```go
const maxRetryAfter = 24 * time.Hour
if secs <= 0 || secs > int(maxRetryAfter/time.Second) {
    return 0, false
}
```

protocol/common/retry_after.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	raw := h.Get("Retry-After")
+	if raw == "" {
+		return 0, false
+	}
+	if secs, err := strconv.Atoi(raw); err == nil {
+		if secs <= 0 {
+			return 0, false
+		}
+		if secs > maxRetryAfterSeconds {
```
avitenzer, line 78: **`err` is dropped, not wrapped.** `Unwrap()` hardcodes the sentinel, so the error passed in is discarded. Harmless today (every call site passes the bare sentinel from `ValidateStatusCodes`), but the doc says "enriches" and invites unconditional use — the first `fmt.Errorf("upstream %s: %w", url, StatusCodeError429)` loses the URL.

```go
type RateLimitedError struct {
    Err        error
    RetryAfter time.Duration
}
func (e *RateLimitedError) Unwrap() error { return e.Err }
```

`errors.Is(…, StatusCodeError429)` still holds by the function's own precondition.

avitenzer (commented):
Two fixes in `retry_after.go`, details inline. Build/vet/tests all pass locally.

avitenzer (approved):
## Approving — both findings fixed, and the tests pin them

Reviewed `0c101c5`. `go build ./...`, `go vet` on all four touched packages, and `go test` on `common`, `chainlib`, `lavasession`, `endpointstate` and `rpcsmartrouter` (88.9s — the package most exposed to the new `Unwrap`) all pass.

**Overflow** — clamping to `MaxRetryAfter` puts the ceiling three orders of magnitude below the overflow threshold, so the class of bug is gone rather than patched. Tests cover the exact repro input, `MaxInt64`, over-cap delta-seconds and over-cap HTTP-date.

**Dropped `err`** — `cause` + `underlying()` is the right shape, and the test uses precisely the `fmt.Errorf("upstream %s: %w", url, StatusCodeError429)` case and asserts the context survives into `Error()`.

**Call-site tests** — checked by mutation rather than taking on faith:

| revert | result |
|---|---|
| `err` → `nil` in `rest.go` / `tendermintRPC.go` | `rest` + `tendermint_uri` fail on "the sentinel must survive" |
| drop `WithRetryAfter` in `http.go` | `jsonrpc` fails on "Retry-After must reach the caller" |

Real regression coverage. (Small note: the commit message says all three fail against the attribute form — only two do, since the JSON-RPC path never had that bug. All three pin something real.)

**Two corrections to my earlier comment**, now that I've checked them against the fix:

- `direct_rpc_relay.go:409` was **not** a gap — it only builds an `HTTPStatusError` for `>= 500` and never sees a 429. Skipping it was right; listing it as an uncovered producer was my error.
- The substring-classifier hazard stays theoretical: `ExtractNodeErrorDetails` reads the head error's `Error()` (`node_error_handler.go:74`), and `HTTPStatusError.Error()` is unchanged, so the duration never reaches `httpStatusMatcher`.

Also verified `http.Header(resp.Headers)` in the poller is safe — `Headers` comes straight from a real `http.Header` (`direct_rpc_connection.go:563`), so keys are canonical and `Get` hits.

### Notes, none blocking

- `MaxRetryAfter = time.Hour` clamps *down*, so a 2h request reads as 1h and we return before the upstream asked — clearly the better trade against parking a provider for years, and `MaxRetryAfter` is exported so a consumer can tell. Just worth knowing `ok=true` no longer means "this is what the upstream said".
- `HTTPStatusError.Unwrap()` allocates a fresh `*RateLimitedError` per call, and `errors.Is`/`errors.As` invoke it on every traversal — bounded to the error path, but a 429 storm is exactly when it fires. Building it once at construction would avoid that.
- The `Date`-header skew fix is beyond what was asked but sound, and tested three ways. Side benefit: with `Date` present both timestamps are second-granular, so durations come out whole.

### Carried forward

- The `nil` → `err` change still widens identity for 504 and strict as well as 429 — worth confirming that's intended with #304/#305.
- `chain_fetcher.go:142-147` still swallows non-fatal verification errors, so a 429 there never reaches a backoff.
- `resp.Body` is still never closed on the 429 return (`http.go:242-247`) — pre-existing, out of scope.

PR #312, opened 2026-08-20
fix(chainlib): match REST apis regardless of a trailing slash, and deterministically

nimrod-teich:
## Why

REST spec api names are turned into regexes at spec load ([`base_chain_parser.go`](https://github.com/Magma-Devs/smart-router/blob/main/protocol/chainlib/base_chain_parser.go)) and matched anchored `^...$`, so a **trailing slash alone decides the match**:

```
spec name : /chains/main/blocks/{block_id}/header
compiled  : ^/chains/main/blocks/[^\/\s]+/header$

/chains/main/blocks/9427283/header      → match
/chains/main/blocks/9427283/header/     → NO match
/chains/main/blocks/head/header/        → NO match
```

A miss falls through to `defaultApiContainer`, which mints a synthetic api named `Default-<raw path>` with a flat **20 compute units** and `PARSER_FUNC_DEFAULT` — block parsing pinned to latest instead of the block actually requested. So this is a routing/billing error, not only a metrics one.

Measured in production (2026-08-20):

- **43% of all live TEZOS traffic** rides the `Default-` fallthrough — 0.256 req/s vs 0.337 req/s matched.
- The `method` label holds **157,695 distinct values**, 157,532 of them of the shape `Default-/chains/main/blocks/<n>/header/` — one series per block number, across the **25 metric families** that carry a `method` label. Still growing ~1,100 new values/hour.

The reverse direction is broken too: the STACKS spec names **11 apis with a trailing slash** (`/extended/v1/tx/`, `/v1/names/`, …), so a client that omits it misses those.

Adding trailing-slash duplicates to the specs is not a fix — TEZOS alone would go from 219 apis to 438, and every REST spec has the same hole.

### A second defect, found while measuring the first

More than one api can cover the same path. Specs pair a literal name with the `{placeholder}` sibling that swallows it all over the catalog:

```
ARWEAVE   /info                 (10 CU)   vs  /{id}                        (20 CU)
ARWEAVE   /block_index          (1000 CU) vs  /{id}                        (20 CU)
CARDANO   /blocks/latest        (10 CU)   vs  /blocks/{hash_or_number}     (20 CU)
STACKS    /v3/tenures/info      (10 CU)   vs  /v3/tenures/{block_id}       (20 CU)
```

`matchSpecApiByName` returned the first match it found while ranging over `serverApis` — a Go map, so iteration order is randomised per call. Confirmed against this repo's own `specs/cosmossdk.json`, which names `/cosmos/base/tendermint/v1beta1/blocks/latest` next to `/blocks/{height}`: 60 identical `ParseMsg` calls resolved 9 times to one api and 51 times to the other.

Swept over [lava-specs](https://github.com/Magma-Devs/lava-specs) (129 files, 255 specs, 103 with a REST collection, 3,231 enabled REST apis):

| | shapes |
|---|---|
| request shapes that only resolve thanks to the slash fallback | **2,996** |
| shapes with several candidates on the exact path — order-dependent before this PR | **105** |
| shapes with several candidates through the fallback — order-dependent if the fallback shipped unranked | **94** |
| of those, shapes whose candidates carry **different compute units** (7 specs) | **79** |

Shipping the slash fallback without fixing this would have roughly doubled the ambiguous surface, so both go in together.

## What changed

**Trailing slash is optional on both sides.** `trimOptionalTrailingSlash` leaves a bare `/` alone — ARWEAVE, APT1 and XLM each name a real api exactly `/`. A match on the path exactly as sent always outranks a slash-insensitive one, so the change can only widen the accepted set: a request that resolves today can never be re-routed.

**The best candidate wins instead of the first one found.** Ranking, in order:

1. a match on the path exactly as sent beats a slash-insensitive one;
2. fewer `{placeholder}` segments — a literal describes a path better than a placeholder that happens to swallow it;
3. more literal characters;
4. the api name, as a stable last resort.

No shape in the catalog reaches step 4 — two names that rank equal also compile to the same pattern (see the note below). It is there to keep the ordering total, and is unit-tested directly.

**Patterns are compiled at spec load, not per lookup.** Ranking means scanning every candidate rather than returning on the first hit, which is only affordable without `regexp.Compile` in the loop — the standing `TODO` on this function. `ApiContainer` gains a `restMatcher` built by the new `newRestApiContainer`, and the scan still stops early on a literal name matching the path as sent, which nothing can outrank. Against a 200-api spec (TEZOS carries 219):

| | main | this PR |
|---|---|---|
| hit | 1.99 ms | **0.09 ms** |
| hit, trailing slash | 4.18 ms | **0.09 ms** |
| miss | 3.87 ms | **0.09 ms** |

`restApiNameToRegex` is the name→pattern transform lifted out of the `getServiceApis` loop verbatim, so tests build `serverApis` through the same constructor the loader uses.

## Behaviour changes to expect

- TEZOS `Default-/chains/main/blocks/<n>/header/` traffic resolves to `/chains/main/blocks/{block_id}/header` — 10 CU instead of 20, and the height in the path becomes the requested block again.
- The 79 compute-unit coin flips settle on the literal api, which is the correct one. For some ARWEAVE paths that is *higher* than what a coin flip sometimes charged (`/block_index` settles at 1000 CU, not 20). Billing becomes correct and stable rather than cheaper.
- `method` label cardinality will not *drop* — Prometheus keeps old label values for the retention window — it stops growing, and stops flapping between two values for identical traffic.

## How to verify

```bash
go test ./protocol/chainlib/... -count=1
go test ./protocol/chainlib/ -run XXX -bench BenchmarkMatchSpecApiByName
```

Map-order-dependent cases run **32×** — a two-candidate case decided by iteration order passes at random, so a single run proves nothing.

| test | asserts |
|---|---|
| `TestMatchSpecApiByNameTrailingSlash` | both directions (TEZOS omits the slash, STACKS carries it, incl. a placeholder before it), the root `/` api, an exact match winning over the fallback in both directions, an unspecced path still missing, a sibling path not swallowed |
| `TestMatchSpecApiByNameRanksMostSpecific` | the real ambiguous pairs above resolve to the literal, with and without the slash; fewer placeholders wins; a templated path still resolves when no literal covers it; the list-path limitation below is pinned |
| `TestRestApiMatcherOrdering` | the ordering is total, including the name tie-break real specs cannot reach |
| `TestRestApiKeyCollapsesNamesWithTheSameShape` | two names differing only in placeholder identifiers key one `ApiKey` |
| `TestMatchSpecApiByNameTrailingSlashConnectionType` | a GET api does not answer a POST |
| `TestRegexParsing` | end-to-end through `ParseMsg` on the real LAVA spec: `/blocks/244590/` resolves to `{height}` at 10 CU **with requested block 244590**, `/blocks/latest[/]` resolves the same way 32× running, unspecced paths still fall through |
| `TestRestMessageGetParamsTrailingSlash` | `GetParams` zips spec and request paths by segment index and the slash adds an empty segment — covered in both directions, with several placeholders and with a query string |

Non-vacuous: with the trim neutered exactly the trailing-slash cases fail; with the ranking neutered exactly the ambiguous cases fail.

After deploy:

```
sum(rate(smartrouter_requests_total{spec="TEZOS",method=~"Default-.*"}[5m]))
```

should fall to ~0, with the traffic reappearing under `method="/chains/main/blocks/{block_id}/header"`.

## Known limitations, deliberately not fixed here

- **A list path with a trailing slash still resolves to its item api.** A trailing `{placeholder}` compiles to `[^\/\s]*`, which also matches empty, so `/cosmos/gov/v1/proposals/` is an *exact* match for `/cosmos/gov/v1/proposals/{proposal_id}` and never reaches the fallback that would prefer `/cosmos/gov/v1/proposals`. 197 shapes across 36 specs, 27 of them with a divergent CU. Fixing it means re-routing requests that resolve this way today, which is a different change with a different risk profile. Pinned by a test so it is written down rather than assumed.
- **Two spec names differing only in placeholder identifiers collapse to one api.** `{address_bytes}` and `{address_string}` compile to the same pattern, so they key the same `ApiKey` and the later replaces the earlier at load — 8 names catalog-wide (COSMOSSDK, COSMOSHUB, CANTO, ELYS). A spec-authoring issue, not a lookup one; covered by a test.
- Bounding the `method` label for paths that genuinely miss (#269), and the missing APT1 `/v1/*` paths found in the same sweep (lava-specs).

#Closes MAG-2943

Jira ticket: MAG-2943

### DIFF
--- a/protocol/chainlib/base_chain_parser.go
+++ b/protocol/chainlib/base_chain_parser.go
@@ -821,11 +821,19 @@ func buildRestApiMatcher(apiName, apiPattern string) (*restApiMatcher, error) {
 // can never re-route a request that resolves today. Beyond that the most specific name
 // wins: a literal segment describes a path better than a {placeholder} that happens to
 // swallow it, which is what keeps ARWEAVE /info on the 10-CU /info instead of the 20-CU
-// /{id}. Names that rank equal are indistinguishable — the spec describes one path twice,
-// as COSMOSSDK does with {address_bytes} and {address_string} — so the last resort is
-// lexicographic, chosen only because it is stable across restarts. Without a total order
-// the winner is Go map iteration order: a coin flip per call, between apis that can carry
-// different compute units and different block parsing.
+// /{id}.
+//
+// Two distinct names can still rank equal: /a/{x}/c and /a/b/{y} both cover /a/b/c with one
+// placeholder and five literal characters, and they compile to different patterns, so they
+// do not collapse into one ApiKey the way {address_bytes} and {address_string} do. No shape
+// in the catalog reaches that today, but it is reachable, so the last resort is the api name
+// — without a total order the winner is Go map iteration order: a coin flip per call between
+// apis that can carry different compute units and different block parsing.
+//
+// Comparing names lexicographically is also the right answer rather than merely a stable one:
+// '{' sorts above every character a path segment holds, so it reads as "literal beats
+// placeholder at the first segment where the two names differ" — the precedence an HTTP
+// router applies.
 func (m *restApiMatcher) moreSpecificThan(exact bool, other *restApiMatcher, otherExact bool) bool {
 	if exact != otherExact {
 		return exact
@@ -837,15 +845,6 @@ func (m *restApiMatcher) moreSpecificThan(exact bool, other *restApiMatcher, oth
 		return m.literalLen > other.literalLen
 	}
 	return m.name < other.name
-}
-
-// matcher returns the api's precompiled matcher, compiling one on demand for a serverApis
-// map that was not built by getServiceApis (only tests build one by hand).
-func (apiCont ApiContainer) matcher(apiKey ApiKey) (*restApiMatcher, error) {
-	if apiCont.restMatcher != nil {
-		return apiCont.restMatcher, nil
-	}
-	return buildRestApiMatcher(apiKey.Name, apiKey.Name)
 }
 
 // matchSpecApiByName returns the service api that best matches the given name.
@@ -869,9 +868,13 @@ func matchSpecApiByName(name, connectionType string, serverApis map[ApiKey]ApiCo
 	bestExact := false
 
 	for apiKey, apiCont := range serverApis {
-		matcher, err := apiCont.matcher(apiKey)
-		if err != nil {
-			utils.LavaFormatError("regex Compile api", err, utils.Attribute{Key: "apiName", Value: apiKey})
+		matcher := apiCont.restMatcher
+		if matcher == nil {
+			// Every REST api is keyed through newRestApiContainer, which always compiles a
+			// matcher, so this is unreachable. Matching without one would mean compiling in
+			// the loop — the per-lookup cost the precompilation exists to remove — so an
+			// entry that arrived some other way is dropped loudly rather than absorbed.
+			utils.LavaFormatError("REST api container has no compiled matcher", nil, utils.Attribute{Key: "apiKey", Value: apiKey})
 			continue
 		}
 		exact := matcher.pattern.MatchString(name)
--- a/protocol/chainlib/base_chain_parser.go
+++ b/protocol/chainlib/base_chain_parser.go
@@ -639,13 +639,8 @@ func getServiceApis(
 
 				// TODO: find a better spot for this (more optimized, precompile regex, etc)
 				if rpcInterface == spectypes.APIInterfaceRest {
-					re := regexp.MustCompile(`{[^}]+}`)
-					processedName := string(re.ReplaceAll([]byte(api.Name), []byte("replace-me-with-regex")))
-					processedName = regexp.QuoteMeta(processedName)
-					processedName = strings.ReplaceAll(processedName, "replace-me-with-regex/", `[^\/\s]+/`)
-					processedName = strings.ReplaceAll(processedName, "replace-me-with-regex", `[^\/\s]*`)
 					serverApis[ApiKey{
-						Name:           processedName,
+						Name:           restApiNameToRegex(api.Name),
 						ConnectionType: collectionKey.ConnectionType,
 					}] = ApiContainer{
 						api:           api,
@@ -736,10 +731,45 @@ func (bcp *BaseChainParser) ExtensionsParser() *extensionslib.ExtensionParser {
 	return &bcp.extensionParser
 }
 
+// restApiNameToRegex turns a REST spec api name into the pattern stored on its ApiKey:
+// each {placeholder} becomes a single path segment, everything else is literal. A trailing
+// placeholder may match empty, an inner one may not.
+func restApiNameToRegex(apiName string) string {
+	re := regexp.MustCompile(`{[^}]+}`)
+	processedName := string(re.ReplaceAll([]byte(apiName), []byte("replace-me-with-regex")))
+	processedName = regexp.QuoteMeta(processedName)
+	processedName = strings.ReplaceAll(processedName, "replace-me-with-regex/", `[^\/\s]+/`)
+	processedName = strings.ReplaceAll(processedName, "replace-me-with-regex", `[^\/\s]*`)
+	return processedName
+}
+
+// trimOptionalTrailingSlash drops one trailing slash, leaving a bare "/" untouched —
+// several specs (ARWEAVE, APT1, XLM) name a real api exactly "/".
+func trimOptionalTrailingSlash(path string) string {
+	if len(path) > 1 && strings.HasSuffix(path, "/") {
+		return path[:len(path)-1]
+	}
+	return path
+}
+
 // matchSpecApiByName returns service api which match given name
+//
+// A trailing slash is optional on both sides: specs name apis both ways (TEZOS omits it,
+// STACKS carries it) and clients send either form, while the compiled name is anchored
+// "^...$" so the slash alone decides the match. A path that misses here falls through to
+// defaultApiContainer, which bills a flat 20 compute units and pins block parsing to
+// latest, so the miss is a routing error and not only a metrics one.
+//
+// The relaxed comparison is a FALLBACK: every api is first tried against the path exactly
+// as sent, and a slash-insensitive hit is returned only when nothing matched. That way the
+// change can only widen the accepted set — it can never re-route a request that matches an
+// api today.
 func matchSpecApiByName(name, connectionType string, serverApis map[ApiKey]ApiContainer) (*ApiContainer, bool) {
 	// TODO: make it faster and better by not doing a regex instead using a better algorithm
 	foundNameOnDifferentConnectionType := ""
+	trimmedName := trimOptionalTrailingSlash(name)
+	var slashInsensitiveMatch *ApiContainer
+
 	for apiName, api := range serverApis {
 		re, err := regexp.Compile("^" + apiName.Name + "$")
 		if err != nil {
@@ -752,7 +782,35 @@ func matchSpecApiByName(name, connectionType string, serverApis map[ApiKey]ApiCo
 			} else {
 				foundNameOnDifferentConnectionType = apiName.ConnectionType
 			}
+			continue
 		}
+		if slashInsensitiveMatch != nil || apiName.ConnectionType != connectionType {
+			continue
+		}
+		trimmedApiName := trimOptionalTrailingSlash(apiName.Name)
+		if trimmedApiName == apiName.Name {
+			// The spec name carries no trailing slash, so re is already the pattern for
+			// its trimmed form — the relaxation costs nothing beyond one more match.
+			if trimmedName != name && re.MatchString(trimmedName) {
+				matched := api
+				slashInsensitiveMatch = &matched
+			}
+			continue
+		}
+		// Only a spec name that itself ends in a slash needs a second pattern.
+		trimmedRe, err := regexp.Compile("^" + trimmedApiName + "$")
+		if err != nil {
+			utils.LavaFormatError("regex Compile api", err, utils.Attribute{Key: "apiName", Value: trimmedApiName})
+			continue
+		}
+		if trimmedRe.MatchString(trimmedName) {
+			matched := api
+			slashInsensitiveMatch = &matched
+		}
+	}
+
+	if slashInsensitiveMatch != nil {
+		return slashInsensitiveMatch, true
 	}
 	if foundNameOnDifferentConnectionType != "" { // its hard to notice when we have an API on only one connection type.
 		utils.LavaFormatWarning("API was found on a different connection type", nil,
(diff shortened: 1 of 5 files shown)


protocol/chainlib/base_chain_parser.go
```diff
@@ -736,23 +723,176 @@ func (bcp *BaseChainParser) ExtensionsParser() *extensionslib.ExtensionParser {
...
+		exact := matcher.pattern.MatchString(name)
+		// When the spec name carries no trailing slash, trimmed is pattern, so this second
+		// match only ever succeeds for a request that carried one.
+		if !exact && !matcher.trimmed.MatchString(trimmedName) {
+			continue
 		}
+		if apiKey.ConnectionType != connectionType {
+			// its hard to notice when we have an API on only one connection type.
+			foundNameOnDifferentConnectionType = apiKey.ConnectionType
+			continue
```
avitenzer, line 889: **Redundant regex match on the per-request hot path.**

When the request has no trailing slash *and* the spec name has none, `matcher.trimmed == matcher.pattern` and `trimmedName == name` — so this second `MatchString` provably re-runs the match that just failed one line above. That is the dominant case: 57% of TEZOS traffic by your own measurement, and 219/219 TEZOS api names.

Measured on this PR's own benchmark, 3 runs each:

| | as written | with the guard below |
|---|---|---|
| miss | 35–40 µs | **19–20 µs** |
| hit | 51–61 µs | **33–41 µs** |
| hit, trailing slash | 45–49 µs | 45–50 µs (unchanged, correctly) |

Roughly halves the miss path. `go test ./protocol/chainlib/... -count=1` stays green with it:

```go
exact := matcher.pattern.MatchString(name)
if !exact {
    if matcher.trimmed == matcher.pattern && trimmedName == name {
        continue // same pattern, same string — the match above already answered this
    }
    if !matcher.trimmed.MatchString(trimmedName) {
        continue
    }
}
```

protocol/chainlib/base_chain_parser.go
```diff
@@ -736,23 +723,176 @@ func (bcp *BaseChainParser) ExtensionsParser() *extensionslib.ExtensionParser {
...
+
+// matcher returns the api's precompiled matcher, compiling one on demand for a serverApis
+// map that was not built by getServiceApis (only tests build one by hand).
+func (apiCont ApiContainer) matcher(apiKey ApiKey) (*restApiMatcher, error) {
+	if apiCont.restMatcher != nil {
+		return apiCont.restMatcher, nil
+	}
+	return buildRestApiMatcher(apiKey.Name, apiKey.Name)
+}
+
```
avitenzer: **This fallback is test-only scaffolding sitting in the hot loop.**

In production every REST container comes from `newRestApiContainer`, so `restMatcher` is never nil. The nil branch exists only for `TestMatchSpecApiByName` (`common_test.go:41`), which hand-builds `serverApis` with literal patterns.

The cost isn't cycles, it's the silent failure mode: if anyone later adds a REST entry to `serverApis` without going through the constructor, the lookup quietly reverts to `regexp.Compile`-per-api-per-request — the exact TODO this PR just removed — with no error and no log.

Migrating that one legacy test to the new `restApis()` helper (every new test already uses it) would let `matcher()` be deleted and `restMatcher != nil` be treated as an invariant.

protocol/chainlib/base_chain_parser.go
```diff
@@ -736,23 +723,176 @@ func (bcp *BaseChainParser) ExtensionsParser() *extensionslib.ExtensionParser {
...
+// can never re-route a request that resolves today. Beyond that the most specific name
+// wins: a literal segment describes a path better than a {placeholder} that happens to
+// swallow it, which is what keeps ARWEAVE /info on the 10-CU /info instead of the 20-CU
+// /{id}. Names that rank equal are indistinguishable — the spec describes one path twice,
+// as COSMOSSDK does with {address_bytes} and {address_string} — so the last resort is
+// lexicographic, chosen only because it is stable across restarts. Without a total order
+// the winner is Go map iteration order: a coin flip per call, between apis that can carry
+// different compute units and different block parsing.
+func (m *restApiMatcher) moreSpecificThan(exact bool, other *restApiMatcher, otherExact bool) bool {
+	if exact != otherExact {
```
avitenzer, line 838: **"Names that rank equal are indistinguishable" isn't true in general** — worth correcting, because this sentence is precisely the argument a future maintainer would use to delete the tie-break as dead code.

Counterexample, verified by running it:

```
/a/{x}/c  →  /a/[^\/\s]+/c    placeholders=1  literalLen=5
/a/b/{y}  →  /a/b/[^\/\s]*    placeholders=1  literalLen=5
```

Both match `/a/b/c`, both rank equal, and they compile to *different* patterns — so they do not collapse to one `ApiKey`, and the lexicographic step is genuinely reached.

The *empirical* claim is fine: my independent catalog sweep also found **0** shapes reaching the tie-break with distinct patterns, so nothing is broken today. Only the justification needs a rewording.

Also worth stating, because it makes the tie-break look principled rather than arbitrary: `{` is `0x7B`, above every letter, digit, `/`, `-`, `_` and `.`. So comparing api names as strings *is* left-to-right "literal beats placeholder at the first differing segment" — the precedence rule HTTP routers converge on. On the pair above it independently picks `/a/b/{y}`, the same answer segment-wise ranking gives.

Minor and related: `literalLen` is measured on the `QuoteMeta`'d pattern, so `.` counts as 2 characters. Zero catalog impact, same tier.

protocol/chainlib/common_test.go
```diff
@@ -1039,3 +1039,342 @@ func TestConstructFiberCallback_NoOriginWhenMetricsDisabled(t *testing.T) {
...
+
+	for i := 0; i < 32; i++ {
+		api, ok := matchSpecApiByName("/cosmos/auth/v1beta1/bech32/lava@1abc/", "GET", serverApis)
+		require.True(t, ok)
+		require.Equal(t, "/cosmos/auth/v1beta1/bech32/{address_bytes}", api.api.Name)
+	}
+}
+
+// TestRestApiMatcherOrdering exercises moreSpecificThan directly, including the
+// lexicographic last resort that real specs cannot reach (two names that rank equal also
```
avitenzer: Same claim as in the `moreSpecificThan` doc comment — "(two names that rank equal also compile to the same pattern, so they collapse to one `ApiKey`)" does not hold in general.

`/a/{x}/c` and `/a/b/{y}` compile to `/a/[^\/\s]+/c` and `/a/b/[^\/\s]*`: distinct patterns, distinct `ApiKey`s, both match `/a/b/c`, both rank `placeholders=1 literalLen=5`. So `TestRestApiMatcherOrdering` is covering a reachable case, not just a theoretical one — which is a better reason to keep it than the one given.

avitenzer (commented):
## Review

Well-built change. The core logic is correct and the central safety claim holds up under independent verification — comments below are one efficiency fix worth making, plus notes for the deploy.

### What I verified independently

Rather than take the numbers on faith, I re-implemented the name→pattern transform and the ranking separately and swept the `lava-specs` catalog:

| claim | independent sweep |
|---|---|
| 129 files, 255 specs, 3,231 enabled REST apis | exact match |
| STACKS names 11 apis with a trailing slash | exact match |
| ARWEAVE `/info` 10 CU, `/block_index` 1000 CU, `/{id}` 20 CU | exact match |
| TEZOS 219 apis, `…/{block_id}/header` 10 CU | exact match |
| ~2,996 shapes only resolve via the slash fallback | 2,993 |
| ~105 shapes order-dependent on the exact path | 118 (my sweep does not resolve spec `imports`) |

The claim that matters most — *"a request that resolves today can never be re-routed"* — I tested directly: for every generated request shape where `main` had exactly **one** exact match, does the ranking pick a different api? **0 re-routes across the whole catalog.**

Putting `exact != otherExact` as the first tier in `moreSpecificThan` is what buys that, and it is load-bearing: had specificity been ranked first, a more-specific fallback match could out-rank a less-specific exact one and silently re-route live traffic. Worth keeping in mind against a future "simplification" of that ordering.

I also mutation-tested the suite in a clean worktree at `1d84ab8`: neutering `trimOptionalTrailingSlash` kills exactly the trailing-slash cases, neutering `moreSpecificThan` kills exactly the ranking cases. Non-vacuous as advertised.

`GetParams` bounds also hold — I swept spec/request slash permutations including `/`, `//`, `/blocks/`, `/blocks//`, a spec name with a trailing slash after a placeholder, and two placeholders with a trailing slash. No index-out-of-range: the extra empty segment always lands last and is never a placeholder.

### Two deploy consequences missing from "Behaviour changes to expect"

Neither is a defect — both are the spec's intent finally being honoured — but whoever watches the rollout should know to look past the `Default-` metric.

**Archive routing.** `tezos.json` defines `archive` with `rule.block = 28000`. Today `Default-` pins the requested block to latest (negative), so `ArchiveParserRule.isPassingRule` returns false and TEZOS traffic *never* goes archive. After this change the concrete height is parsed, so any request more than 28,000 blocks behind the tip acquires the archive extension and is restricted to archive-capable providers — the same 43% of traffic changing provider pool. It fails soft (`rpcsmartrouter_server.go:3686` retries with no extensions on `PairingListEmptyError` when session degradation is allowed), but it is a latency and pool-utilisation change worth a dashboard.

**Relay timeout halves.** `GetTimePerCu` is `cu × 100ms` with a 1s floor. `Default-` at 20 CU gives a 2s per-relay budget; `/chains/main/blocks/{block_id}/header` at 10 CU gives 1s. That traffic's budget is cut in half, which could surface as new timeouts against slow Tezos nodes. In the other direction ARWEAVE `/block_index` goes 20 → 1000 CU, i.e. 2s → 100s (capped by the 30s default).

### Repo-level note

There is no `go test` step in `.github/workflows/` — `smartrouter.yml` is named "Build and Test" but only builds. Pre-existing and not this PR's problem, but directly load-bearing here: this change's safety argument rests on a suite CI never executes, and the 32× map-order loops only pay off if something runs them.

avitenzer (approved):
Approving.

Findings 2 and 3 from my earlier review are addressed in 37d53ee (pushed to this branch): the on-demand `matcher()` fallback is gone so a missing precompiled matcher is now an error rather than a silent return to per-lookup `regexp.Compile`, and the `moreSpecificThan` note no longer claims the lexicographic last resort is unreachable — `/a/{x}/c` and `/a/b/{y}` reach it — with `TestRestApiMatcherOrdering` pinning that pair. Full `./protocol/chainlib/...` suite green, and both changes are mutation-checked (flipping the tie-break and forcing the nil-matcher path each fail a test).

Finding 1, the provably-redundant second `MatchString` when neither side carries a trailing slash, is left for a follow-up rather than held against this PR — it's a further ~2x on the miss path, on top of the 50–100x this branch already delivers.

The two deploy-watch items from the review body still stand and aren't code changes: TEZOS traffic acquires the archive extension once concrete heights are parsed (fails soft — session degradation is allowed by default and retries without extensions), and its per-relay budget goes from 2s to 1s as those requests move from 20 CU to 10 CU.

PR #313, opened 2026-08-20
fix(smart-router): keep typed 429 errors readable through every wrap

nimrod-teich:
#Closes MAG-2946

Jira ticket: MAG-2946

First PR of the 429/Retry-After stack (umbrella investigation on MAG-2910). Propagation only — no scoring, cadence, or membership change. The follow-ups (hold-off registry MAG-2947, re-verify consumer MAG-2910, hot-path selection MAG-2948, ws/gRPC + tracker MAG-2949) all depend on this signal surviving.

## Why

The `Retry-After` capture (`protocol/common/retry_after.go`) mints a typed rate-limit error on every HTTP transport — and then four call sites flatten it into a log string before any consumer can `errors.Is` / `common.RetryAfterFrom` it:

- `protocol/chainlib/chain_fetcher.go:518` and `:616` passed the send error as a `LavaFormatDebug` **attribute**. `LavaFormatDebug` has no err parameter, so `LavaFormatLog` gets `err=nil` and returns a plain `fmt.Errorf` — no `Unwrap`, chain severed. These two lines sit under `Validate`'s latest-block pre-fetch (startup verification AND every epoch re-verification), the ChainTracker poll/init, and fork-detection hash fetches.
- `protocol/endpointstate/endpoint_poller.go:146` and `:281` — same idiom, one frame **above** the site that parses `Retry-After` into `HTTPStatusError.RetryAfter` (`endpoint_poller.go:460`). The value was parsed and then destroyed before the ChainTracker's backoff could ever read it.

On the hot relay path the sever was one step later: `relayInnerDirect` returned on a 429/5xx **before** the block that copies the sender's result into `relayResult`, so the classifier's `IsRateLimited` was computed and then discarded, and the upstream's `Retry-After` died with the discarded reply.

## What changed

- `utils/lavalog.go` — new `LavaFormatDebugErr`: debug severity, err as the returned error's cause. The missing variant that made the sever idiom attractive in the first place.
- The four sites above now use it — log level unchanged, chain intact.
- `protocol/rpcsmartrouter/rpcsmartrouter_server.go` — the 429/5xx early-return now copies the fault-axis flags (`IsRateLimited`, `IsNodeError`, `IsNonRetryable`, `IsUnsupportedMethod`, `IsDataScope`) + `ProviderInfo` into `relayResult`, and fails the relay with `httpStatusRelayError`: for a 429 that is the typed sentinel with the reply's `Retry-After` attached, so the relay path reads the same as every other transport.
- `StatusCode` is deliberately **not** copied on the failure path: listeners surface a non-zero `relayResult.StatusCode` to the client, and what the client sees on a failed relay is MAG-2948's decision, not this fix's.

**Scope of the `relayResult` copy.** It makes the verdict *present*; it does not yet make anything act on it, and no current consumer reads it on this path. `shouldFailSessionForResult` short-circuits on `if err != nil` before it reaches the rate-limit carve-out, and this path still fails the relay — so the carve-out stays unreachable for a real HTTP 429. The relay processor is the same story: a response with `Err != nil` is bucketed into `protocolResponseErrors`, and `GetResultsSummary` reads the fault-axis flags only from node-error results. Teaching the gate to honour the carve-out is MAG-2948's, and it now has the signal to do it with. That dormancy is also why this PR carries no retry- or scoring-behaviour risk.

## How to verify

```
go build ./...
go test ./utils/... ./protocol/chainlib/... ./protocol/endpointstate/ ./protocol/rpcsmartrouter/ ./protocol/common/
go test ./protocol/chainlib/ -run 'TestChainFetcher_Fetch' -v
go test ./protocol/endpointstate/ -run 'TestEndpointPoller_FetchLatestBlockNum' -v
go test ./protocol/rpcsmartrouter/ -run 'TestHTTPStatusRelayError' -v
```

The new tests drive each fixed path through the wrapping production actually builds (stub router / stub connection failing with the real error shapes), and pin the negatives: a plain transport failure and a 503 wrapped identically must stay unclassified.

### DIFF
--- a/protocol/chainlib/chain_fetcher.go
+++ b/protocol/chainlib/chain_fetcher.go
@@ -515,7 +515,7 @@ func (cf *ChainFetcher) FetchLatestBlockNum(ctx context.Context) (int64, error) 
 	}
 	reply, _, _, proxyUrl, chainId, err := cf.chainRouter.SendNodeMsg(ctx, nil, chainMessage, nil)
 	if err != nil {
-		return spectypes.NOT_APPLICABLE, utils.LavaFormatDebug(tagName+" failed sending chainMessage", []utils.Attribute{{Key: "chainID", Value: cf.endpoint.ChainID}, {Key: "APIInterface", Value: cf.endpoint.ApiInterface}, {Key: "error", Value: err}}...)
+		return spectypes.NOT_APPLICABLE, utils.LavaFormatDebugErr(tagName+" failed sending chainMessage", err, []utils.Attribute{{Key: "chainID", Value: cf.endpoint.ChainID}, {Key: "APIInterface", Value: cf.endpoint.ApiInterface}}...)
 	}
 	parserInput, err := FormatResponseForParsing(reply.RelayReply, chainMessage)
 	if err != nil {
@@ -613,7 +613,7 @@ func (cf *ChainFetcher) fetchSingleBlockHashByNum(ctx context.Context, blockNum 
 	reply, _, _, proxyUrl, chainId, err := cf.chainRouter.SendNodeMsg(ctx, nil, chainMessage, nil)
 	if err != nil {
 		timeTaken := time.Since(start)
-		return "", nil, utils.LavaFormatDebug(tagName+" failed sending chainMessage", []utils.Attribute{{Key: "sendTime", Value: timeTaken}, {Key: "error", Value: err}, {Key: "chainID", Value: cf.endpoint.ChainID}, {Key: "APIInterface", Value: cf.endpoint.ApiInterface}}...)
+		return "", nil, utils.LavaFormatDebugErr(tagName+" failed sending chainMessage", err, []utils.Attribute{{Key: "sendTime", Value: timeTaken}, {Key: "chainID", Value: cf.endpoint.ChainID}, {Key: "APIInterface", Value: cf.endpoint.ApiInterface}}...)
 	}
 
 	responseData := reply.RelayReply.Data
--- a/protocol/chainlib/chain_fetcher_retry_after_test.go
+++ b/protocol/chainlib/chain_fetcher_retry_after_test.go
@@ -1,0 +1,91 @@
+package chainlib
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"testing"
+	"time"
+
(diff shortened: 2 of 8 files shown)


avitenzer:
**The block-hash half of this fix is re-severed one frame up, in the ChainTracker.**

`protocol/chaintracker/chain_tracker.go:533` still uses the exact idiom this PR removes:

```go
forked, err := cs.forkChanged(ctx, newLatestBlock)
if err != nil {
    return false, utils.LavaFormatDebug("could not fetchLatestBlock Hash in ChainTracker",
        utils.Attribute{Key: "error", Value: err}, ...)   // <- chain severed again
}
```

Full path from the two block-hash sites this PR fixes:

`fetchAllPreviousBlocksIfNecessary:531` → `forkChanged:429` → `DefaultChainTrackerFetcher.FetchBlockHashByNum:116` → `ChainFetcher.fetchSingleBlockHashByNum` (`chain_fetcher.go:616`, fixed here) / `EndpointPoller.fetchSingleBlockHash` (`endpoint_poller.go:280`, fixed here).

`forkChanged` propagates verbatim (`return false, err` at `:434` and `:447`), and then line 533 flattens it before the poll loop at `:663` can see it. So fork detection — one of the three consumers the description names for that site — still ends up holding a plain `fmt.Errorf` with no `Unwrap`.

For contrast, the latest-block path *is* clean end to end: `fetchAllPreviousBlocksIfNecessary:488` returns `err` verbatim at `:502`. And `fetchAllPreviousBlocks` → `readHashes` (`:378`) is clean too. It is specifically `forkChanged`'s return.

Both new tests pass because each stops at `ChainFetcher` / `EndpointPoller` and never crosses the tracker frame — which is the general shape of the risk here: chain integrity is a whole-path property, but per-frame tests assert it exactly where it is guaranteed to hold. A single assertion of `common.RetryAfterFrom` at the final consumer would cover this site and keep covering it as MAG-2947/2948/2949 land.

Suggestion: it is a one-line `LavaFormatDebugErr` swap, so either fold it in here, or state in the description that the tracker frame is MAG-2949's scope. Left unmentioned, MAG-2949 inherits a fix that reads as complete.

avitenzer (approved):

PR #314, opened 2026-08-20
feat(smart-router): shared tiered rate-limit hold-off registry

nimrod-teich:
#Closes MAG-2947

Jira ticket: MAG-2947

Second PR of the 429/Retry-After stack (umbrella: MAG-2910). Sibling of PR #313 in smart-router — keep typed 429 errors readable through every wrap. The shared registry is now consumed on `main` by spec re-verification, relay selection, and recovery probing; WS/gRPC recognition and tracker polling remain under MAG-2949.

## Why

The audit on MAG-2910 found that `Retry-After` was captured on every HTTP transport and consumed by nothing: no path in the router waited longer after being told to slow down. Each path growing its own backoff would triple the tuning surface and still miss the account-wide nature of vendor caps, so the backoff state lives in one place.

## What changed

- New package `protocol/holdoff` — the shared rate-limit hold-off registry ([holdoff.go](protocol/holdoff/holdoff.go)):
  - **Tiered keying.** A 429 holds off the endpoint URL that returned it; when 2 distinct URLs of one provider are held off at once, the hold-off escalates to the provider name. The provider hold-off ends when its longest-held member does.
  - **Retry-After as the floor.** Otherwise the registry uses 30s doubling per consecutive strike, capped at 30m before jitter. One-sided jitter extends by up to +20% and never shortens; the final result is strictly bounded by `common.MaxRetryAfter` (1h), including jitter.
  - **Any answer clears.** Success or genuine failure clears the URL strikes and provider escalation. Expired strike memory becomes eligible after 1h of quiet and is reclaimed lazily on the next recorded 429.
  - **Explicit APIs.** Production uses process-wide `Shared`; composition roots and tests can inject `NewRegistry` or jitter-free `NewRegistryWithClock`. Consumers query `HeldOff`/`ReadyAt` per endpoint and `ProviderReadyAt` for provider-wide selection. `ReadyAt` returns zero for both absent and expired state.
  - The registry never sleeps, never synthesizes a customer-facing 429, and never decides consumer policy.
- Review hardening:
  - sample the clock under the registry lock and prevent stale non-zero `ReadyAt` results;
  - enforce the 1h bound after jitter and share the parser's `common.MaxRetryAfter` constant;
  - exercise `NewRegistryWithClock` in the package tests and cover endpoint/provider expiry plus cap-and-jitter interaction.
- `docs/RATE-LIMIT-HOLDOFF.md` documents the APIs, lazy reclamation, and one row per consumer path.

## How to verify

```
go build ./...
go vet ./protocol/holdoff/
go test ./protocol/holdoff/ ./protocol/common/ -count=1 -race
go test ./protocol/lavasession/...
go test ./protocol/rpcsmartrouter/...
```

12 holdoff tests pin the semantics: strike growth to the cap, Retry-After floor in both directions, the jitter-inclusive 1h bound, one-sided jitter, clear-on-answer and strike reset, escalation and its clearing, endpoint/provider expiry, strike-memory aging, the unknown-endpoint zero value, and provider-level readiness.

### DIFF
--- a/docs/RATE-LIMIT-HOLDOFF.md
+++ b/docs/RATE-LIMIT-HOLDOFF.md
@@ -3,6 +3,8 @@
 `protocol/holdoff` is the shared answer to "an upstream said 429 — when may we ask it
 again?". Every path that sends requests upstream records rate limits into one
 `holdoff.Registry` and consults it before sending, instead of growing a private backoff.
+Production consumers use the process-wide `holdoff.Shared`; composition roots may inject
+another registry, and tests use `NewRegistry` or the jitter-free `NewRegistryWithClock`.
 
 ## Semantics
 
@@ -13,15 +15,27 @@ again?". Every path that sends requests upstream records rate limits into one
   when its longest-held member does.
 - **Retry-After is the floor.** When the upstream said how long to wait
   (`common.RetryAfterFrom` on the typed error), the hold-off is at least that long,
-  bounded at 1h. Otherwise a per-URL exponential: 30s doubling per consecutive strike,
-  capped at 30m. Jitter extends each hold-off by up to +20% and never shortens it, so a
-  fleet held off by one vendor-wide limit does not return in a synchronized burst.
+  bounded at 1h including jitter. Otherwise a per-URL exponential: 30s doubling per
+  consecutive strike, capped at 30m before jitter. Jitter extends each hold-off by up to
+  +20% and never shortens it, so a fleet held off by one vendor-wide limit does not return
+  in a synchronized burst.
 - **Any answer clears.** An upstream that answered a request — success or a genuine
   failure — is no longer refusing us for load: the URL's strikes and the provider
   escalation are dropped, and later failures reach their normal handling at the normal
-  cadence. Strike memory also ages out after 1h of quiet.
+  cadence. Strike memory becomes eligible for reclamation after 1h of quiet and is
+  reclaimed lazily on the next recorded 429 anywhere in the registry.
 - **Internal only.** The registry and the Retry-After values drive internal retry and
   scheduling decisions. Nothing here is surfaced to the customer.
+
+## Query APIs
+
+- `HeldOff(provider, url)` answers the endpoint-level yes/no question.
+- `ReadyAt(provider, url)` returns the later of the URL and provider-wide deadlines for
+  soonest-to-expire selection. It returns the zero time for absent **and expired** state.
+- `ProviderReadyAt(provider)` answers whether the provider is held as a whole and when it
+  next has a recorded URL worth trying. The registry only knows URLs that have answered
+  429; if a provider has additional never-limited URLs, this query can conservatively
+  over-hold until the earliest recorded deadline expires.
 
 ## Consumer map
 
--- a/protocol/holdoff/holdoff.go
+++ b/protocol/holdoff/holdoff.go
@@ -31,6 +31,8 @@ import (
 	"math/rand"
 	"sync"
 	"time"
+
+	"github.com/magma-Devs/smart-router/protocol/common"
 )
 
 const (
@@ -42,10 +44,9 @@ const (
 	// maxUpstreamHoldoff — the vendor knows its own reset schedule better than we do.
 	MaxHoldoff = 30 * time.Minute
 
-	// maxUpstreamHoldoff bounds what a Retry-After header can demand, mirroring
-	// common.MaxRetryAfter so a hostile or broken upstream cannot park an endpoint
-	// indefinitely.
-	maxUpstreamHoldoff = time.Hour
+	// maxUpstreamHoldoff is the parser's Retry-After bound, shared here so capture and
+	// consumption cannot silently drift apart.
+	maxUpstreamHoldoff = common.MaxRetryAfter
 
 	// JitterFactor extends each hold-off by up to +20%, spreading expirations.
 	JitterFactor = 0.2
@@ -136,6 +137,9 @@ func (r *Registry) RecordRateLimit(provider, url string, retryAfter time.Duratio
 		d = maxUpstreamHoldoff
 	}
 	d += time.Duration(r.randFloat() * JitterFactor * float64(d))
+	if d > maxUpstreamHoldoff {
+		d = maxUpstreamHoldoff
+	}
 	e.readyAt = now.Add(d)
 	e.lastSeen = now
 
@@ -178,8 +182,7 @@ func (r *Registry) RecordAnswer(provider, url string) {
 // HeldOff reports whether requests to url should be avoided right now, either because
 // the URL itself is held off or its provider escalated.
 func (r *Registry) HeldOff(provider, url string) bool {
-	readyAt := r.ReadyAt(provider, url)
-	return readyAt.After(r.now())
+	return !r.ReadyAt(provider, url).IsZero()
 }
 
 // ReadyAt returns when url stops being held off — the later of its own hold-off and its
@@ -192,9 +195,13 @@ func (r *Registry) ReadyAt(provider, url string) time.Time {
 	if ps == nil {
 		return time.Time{}
 	}
+	now := r.now()
 	ready := ps.escalatedUntil
 	if e := ps.urls[url]; e != nil && e.readyAt.After(ready) {
 		ready = e.readyAt
+	}
+	if !ready.After(now) {
+		return time.Time{}
 	}
 	return ready
 }
--- a/protocol/holdoff/holdoff_test.go
+++ b/protocol/holdoff/holdoff_test.go
@@ -7,13 +7,11 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// newTestRegistry pins the clock and disables jitter so durations are exact. Tests that
-// exercise jitter override randFloat explicitly.
+// newTestRegistry pins the clock through the public consumer-test constructor. Tests
+// that exercise jitter override randFloat explicitly.
 func newTestRegistry(start time.Time) (*Registry, *time.Time) {
 	now := start
-	r := NewRegistry()
-	r.now = func() time.Time { return now }
-	r.randFloat = func() float64 { return 0 }
+	r := NewRegistryWithClock(func() time.Time { return now })
 	return r, &now
 }
 
@@ -54,8 +52,9 @@ func TestRetryAfterIsTheFloor(t *testing.T) {
 
 func TestRetryAfterIsBounded(t *testing.T) {
 	r, _ := newTestRegistry(t0)
+	r.randFloat = func() float64 { return 1 } // worst case must not exceed the hard bound
 	d := r.RecordRateLimit("vendor", "u", 24*time.Hour)
-	require.Equal(t, maxUpstreamHoldoff, d)
+	require.Equal(t, maxUpstreamHoldoff, d, "the bound includes jitter")
 }
 
 func TestJitterOnlyExtends(t *testing.T) {
@@ -82,7 +81,7 @@ func TestAnswerClearsStrikesAndHoldoff(t *testing.T) {
 }
 
 func TestEscalationCoversTheWholeProvider(t *testing.T) {
-	r, _ := newTestRegistry(t0)
+	r, now := newTestRegistry(t0)
 
 	r.RecordRateLimit("vendor", "https://a.example/eth", 5*time.Minute)
 	require.False(t, r.HeldOff("vendor", "https://b.example/arb"),
@@ -96,6 +95,11 @@ func TestEscalationCoversTheWholeProvider(t *testing.T) {
 
 	// Another provider is untouched.
 	require.False(t, r.HeldOff("other", "https://d.example/eth"))
+
+	// An expired escalation must not leave an unrelated URL with a stale non-zero
+	// ReadyAt value.
+	*now = t0.Add(10*time.Minute + time.Second)
+	require.True(t, r.ReadyAt("vendor", "https://never-limited.example/base").IsZero())
 }
 
 func TestAnswerDropsTheEscalation(t *testing.T) {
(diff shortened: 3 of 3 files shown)


protocol/holdoff/holdoff.go
```diff
@@ -0,0 +1,269 @@
...
+	defer r.mu.Unlock()
+	ps := r.providers[provider]
+	if ps == nil {
+		return time.Time{}
+	}
+	ready := ps.escalatedUntil
+	if e := ps.urls[url]; e != nil && e.readyAt.After(ready) {
+		ready = e.readyAt
+	}
+	return ready
```
avitenzer, line 206: **Finding 1 — the contract here doesn't hold, and MAG-2948 is the one that will pay for it.**

> The zero time means it is not held off at all.

That's only true for a provider the registry has never seen. `ReadyAt` never compares against `now`, and `gcLocked` runs **only** inside `RecordRateLimit` (line 117) — so in a process that has stopped getting 429s, nothing is ever reclaimed:

```
6h after a 1m hold-off, with no further 429s anywhere: ReadyAt=12:01:00 IsZero=false
```

Worse via the escalation path: line 195 returns `ps.escalatedUntil` for **any** URL of the provider, including ones the registry has never recorded — and `escalatedUntil` goes stale-high because line 155 only recomputes it when `held >= 2`, never lowering it when the count drops back to 1:

```
ReadyAt for a URL never rate-limited, 11m after a long-expired escalation:
  12:10:00 (IsZero=false, now=12:11:00)
```

`docs/RATE-LIMIT-HOLDOFF.md` hands `ReadyAt` to exactly the consumer most likely to misread it — hot-path selection, *"fall back to the soonest-to-expire one"*. A selection loop written as `if !reg.ReadyAt(p, u).IsZero() { deprioritize }` — the reading this comment invites — permanently deprioritizes every healthy endpoint of any provider that was ever escalated. `HeldOff` is correct; only `ReadyAt`'s contract is sharp.

Cheapest fix, since `ReadyAt` already holds the lock and `HeldOff` needs `now` anyway:

```go
func (r *Registry) ReadyAt(provider, url string) time.Time {
	r.mu.Lock()
	defer r.mu.Unlock()
	ps := r.providers[provider]
	if ps == nil {
		return time.Time{}
	}
	now := r.now()
	ready := ps.escalatedUntil
	if e := ps.urls[url]; e != nil && e.readyAt.After(ready) {
		ready = e.readyAt
	}
	if !ready.After(now) {
		return time.Time{} // expired is not held off
	}
	return ready
}
```

That makes `IsZero()` mean what the comment says, and folds finding 5 in as a side effect. If you'd rather keep `ReadyAt` a pure accessor, the comment needs to say *"a zero **or past** time means not held off — always compare against `now`"*. Given the consumers are three separate future PRs, I'd rather the type enforced it than the doc described it.

protocol/holdoff/holdoff.go
```diff
@@ -0,0 +1,269 @@
...
+
+	e.strikes++
+	d := exponentialFor(e.strikes)
+	if retryAfter > d {
+		d = retryAfter
+	}
+	if d > maxUpstreamHoldoff {
+		d = maxUpstreamHoldoff
+	}
+	d += time.Duration(r.randFloat() * JitterFactor * float64(d))
```
avitenzer, line 139: **Finding 2 — jitter lands after both clamps, so neither documented bound is the actual bound.**

Confirmed with worst-case jitter:

| Bound | Documented | Actual |
| --- | --- | --- |
| `maxUpstreamHoldoff` (hostile-upstream guard) | 1h | **1h12m** |
| `MaxHoldoff` (exponential cap) | 30m | **36m** |

30m → 36m is arguably fine — jitter spreading a soft cap is the point. The 1h → 1h12m one isn't the same kind of thing: that bound exists *"so a hostile or broken upstream cannot park an endpoint indefinitely"* (line 46), and `docs/RATE-LIMIT-HOLDOFF.md:16` states it flatly as "bounded at 1h". A ceiling an attacker-controlled header can push 20% past isn't a ceiling. Swapping the order fixes it:

```go
d += time.Duration(r.randFloat() * JitterFactor * float64(d))
if d > maxUpstreamHoldoff {
	d = maxUpstreamHoldoff
}
```

The existing tests miss this by construction — `TestRetryAfterIsBounded` pins `randFloat` to 0, and `TestJitterOnlyExtends` uses 100s, nowhere near a cap. Neither exercises *bound and jitter together*, which is the only place they interact. Worth a test that does.

protocol/holdoff/holdoff.go
```diff
@@ -0,0 +1,269 @@
...
+	InitialHoldoff = 30 * time.Second
+
+	// MaxHoldoff caps the exponential. An upstream-sent Retry-After may exceed it, up to
+	// maxUpstreamHoldoff — the vendor knows its own reset schedule better than we do.
+	MaxHoldoff = 30 * time.Minute
+
+	// maxUpstreamHoldoff bounds what a Retry-After header can demand, mirroring
+	// common.MaxRetryAfter so a hostile or broken upstream cannot park an endpoint
+	// indefinitely.
+	maxUpstreamHoldoff = time.Hour
```
avitenzer: **Finding 3 — "mirroring `common.MaxRetryAfter`" is an unenforced mirror.**

Two files that exist to be wired together, restating the same bound as a literal, where a silent divergence produces a clamp that quietly stops matching what the parser already clamped to. `common.MaxRetryAfter` is exported (`protocol/common/retry_after.go:14`) — I checked the substitution:

```go
maxUpstreamHoldoff = common.MaxRetryAfter
```

**compiles, no import cycle** (`common` doesn't reach `holdoff`).

Of all the constants in this block, this is the one that has a real upstream bound available rather than a chosen number — so it's the one that shouldn't be a literal.

docs/RATE-LIMIT-HOLDOFF.md
```diff
@@ -0,0 +1,44 @@
...
+  when its longest-held member does.
+- **Retry-After is the floor.** When the upstream said how long to wait
+  (`common.RetryAfterFrom` on the typed error), the hold-off is at least that long,
+  bounded at 1h. Otherwise a per-URL exponential: 30s doubling per consecutive strike,
+  capped at 30m. Jitter extends each hold-off by up to +20% and never shortens it, so a
+  fleet held off by one vendor-wide limit does not return in a synchronized burst.
+- **Any answer clears.** An upstream that answered a request — success or a genuine
+  failure — is no longer refusing us for load: the URL's strikes and the provider
+  escalation are dropped, and later failures reach their normal handling at the normal
+  cadence. Strike memory also ages out after 1h of quiet.
```
avitenzer: **Finding 4 — this is conditional, not automatic.** Same root cause as finding 1: `gcLocked` is only reachable from `RecordRateLimit`, so a process that goes quiet keeps its entries forever. `entryRetention` really means "1h of quiet, **plus** the next 429 from anywhere in the process".

Bounded by the endpoint-set size, so not a leak — but the doc overstates it, and the accuracy matters if a consumer ever relies on the aging. Either say "reclaimed lazily on the next recorded 429", or also gc on the read paths.

protocol/holdoff/holdoff.go
```diff
@@ -0,0 +1,269 @@
...
+		delete(r.providers, provider)
+	}
+}
+
+// HeldOff reports whether requests to url should be avoided right now, either because
+// the URL itself is held off or its provider escalated.
+func (r *Registry) HeldOff(provider, url string) bool {
+	readyAt := r.ReadyAt(provider, url)
+	return readyAt.After(r.now())
+}
```
avitenzer, line 186: **Finding 5 (nit) — the clock is sampled outside the lock.** `ReadyAt` takes and releases `mu`, then `r.now` is read unlocked on line 182.

Benign in production (`now` is never reassigned after construction), but it's a bare field read racing any future clock swap — and `newTestRegistry` in the test file reassigns `r.now` post-construction, so the pattern the code permits is the pattern the tests already use. Folded in for free if you take finding 1's fix.

protocol/holdoff/holdoff.go
```diff
@@ -0,0 +1,269 @@
...
+
+// NewRegistryWithClock is the deterministic variant for consumer tests: time comes from
+// the given clock and jitter is disabled, so applied hold-offs are exact.
+func NewRegistryWithClock(now func() time.Time) *Registry {
+	return &Registry{
+		providers: map[string]*providerState{},
+		now:       now,
+		randFloat: func() float64 { return 0 },
+	}
+}
```
avitenzer, line 109: **Finding 6 (nit) — exported, and used by nothing, including this PR's own tests.**

The helper added so consumer tests don't poke unexported fields is bypassed by `newTestRegistry`, which pokes them anyway. Understandable — this variant hard-disables jitter and `TestJitterOnlyExtends` needs it — but it means the seam ships untested and unexercised.

Either build `newTestRegistry` on top of it (overriding `randFloat` only in the one test that needs jitter), or hold the export until MAG-2948 actually needs it.

avitenzer (commented):
## Review

Design decisions here are the right ones and the reasoning is on the record rather than in someone's head. Three things I want to call out as good before the findings:

- **One-sided jitter.** Repo precedent (`protocol/rpcsmartrouter/websocket_backoff.go:91`) uses symmetric `delay*(1±jitter)`. Diverging to `+jitter` only is correct here and correctly justified — coming back *before* the vendor's ask defeats the purpose, coming back *spread out* is the purpose.
- **"Any answer clears"** is genuinely self-selecting. Under a real account-wide cap the sibling URLs 429 too, so the escalation survives; under a per-endpoint limit they answer and it correctly collapses. The policy distinguishes the two cases without having to classify them.
- `randFloat` is correctly confined to under-`mu`, which matters because `rand.Rand` isn't concurrency-safe.

**Verified** from a clean worktree at `c30b1a5` (not a working dir): `go build ./...` ✅ · `go vet ./protocol/holdoff/` ✅ · `go test ./protocol/holdoff/ ./protocol/common/ -count=1 -race` ✅ · `gofmt -l` clean. `.golangci.yml` enables neither `gochecknoglobals` nor `mnd`, so nothing below is a lint failure.

### Two things worth fixing before this lands

**1. `ReadyAt` returns a stale, long-past, non-zero timestamp indefinitely.** Its doc-comment says "The zero time means it is not held off at all", which is only true for a provider the registry has never seen. `gcLocked` runs only inside `RecordRateLimit`, so a quiet process never reclaims anything. Confirmed by probe:

```
6h after a 1m hold-off, with no further 429s anywhere: ReadyAt=12:01:00 IsZero=false

ReadyAt for a URL never rate-limited, 11m after a long-expired escalation:
  12:10:00 (IsZero=false, now=12:11:00)
```

This is a trap aimed squarely at MAG-2948 — the doc file hands `ReadyAt` to the hot path for "fall back to the soonest-to-expire one", and the `IsZero()` reading the comment invites would permanently deprioritize every healthy endpoint of any provider that was ever escalated. Inline comment has a fix.

**2. Jitter is applied after both caps**, so the documented bounds are off by up to +20%: `maxUpstreamHoldoff` 1h → **1h12m**, `MaxHoldoff` 30m → **36m**. The soft cap overshooting is arguably fine; the 1h one isn't the same kind of thing, since it exists so "a hostile or broken upstream cannot park an endpoint indefinitely". Inline comment has details.

Findings 3–6 (inline) are cleanups worth folding into the same push. The two design notes at the bottom are for the consumer PRs, not blocking here.

### Two notes on the PR body / docs

The body describes `HeldOff`/`ReadyAt` and "11 tests" — it's 12 now, and the branch is 4 commits, not 1. Neither the body nor `docs/RATE-LIMIT-HOLDOFF.md` mentions `Shared`, `NewRegistryWithClock`, or `ProviderReadyAt`. `ProviderReadyAt` in particular is a distinct semantic ("is this provider held *as a whole*", with its own documented over-hold caveat), and the consumer catalog is its natural home. For a doc that's explicitly the spec of record for a five-consumer stack, that gap will cost someone.

---

*One framing that ties findings 1 and 4 together: they're the same bug wearing two hats — **lazy GC leaking into a read contract**. Reclaiming on write is the right call for a hot-path registry. But it means "expired" and "absent" stop being the same state, and any read API returning a value rather than a verdict exposes that difference to callers. `HeldOff` is immune because it returns a verdict computed against `now`; `ReadyAt` isn't, because it hands back a raw timestamp and lets the caller decide what it means.*

protocol/holdoff/holdoff.go
```diff
@@ -0,0 +1,269 @@
...
+	ps := r.providers[provider]
+	if ps == nil {
+		return
+	}
+	delete(ps.urls, url)
+	ps.escalatedUntil = time.Time{}
+	if len(ps.urls) == 0 {
+		delete(r.providers, provider)
+	}
+}
```
avitenzer, line 180: **Design note for MAG-2948, not blocking here — an in-flight request can tear down an escalation that was set after it left.**

`RecordAnswer` clears `escalatedUntil` for *any* URL of the provider, including ones the registry has never recorded (line 171 is a no-op for those, line 172 is not — verified). So: `a` and `b` 429 → escalate; twenty requests to `c` that were already in flight return 200 → escalation gone.

Self-correcting under a true account-wide cap, since `c` would 429 too. But selection is where this either matters or doesn't, and it's better decided in MAG-2948 than discovered in it.

protocol/holdoff/holdoff.go
```diff
@@ -0,0 +1,269 @@
...
+	urls           map[string]*entry
+	escalatedUntil time.Time
+}
+
+// Shared is the process-wide registry every production consumer records into and
+// consults. One instance across chains and paths is what makes the provider-name
+// escalation tier meaningful: a vendor cap is account-wide, and a process probing
+// several chains through one vendor must slow all of them together. Tests build their
+// own via NewRegistry/NewRegistryWithClock instead of touching this one.
+var Shared = NewRegistry()
```
avitenzer, line 79: **Design note, not blocking — this is the only package-level mutable registry under `protocol/`.** I checked `main`; there's no precedent for it.

The rationale is sound: the escalation tier is only meaningful process-wide. But this repo has been bitten specifically by process-global state standing in for per-instance state (MAG-2333), and the failure modes here are test cross-contamination and multi-router-in-one-process surprises.

The comment already tells tests to build their own. Consider going one step further and having consumers take a `*Registry` parameter that defaults to `Shared` at the wiring site — so the global is a composition-root choice rather than a reference baked into each consumer.

avitenzer (approved):

PR #316, opened 2026-08-20
fix(smart-router): a rate-limited re-verify is inconclusive, and held off

nimrod-teich:
#Closes MAG-2910

Jira ticket: MAG-2910

> **Stacked** on PR #314 in smart-router (hold-off registry, MAG-2947) — PR #313 already merged and this branch is rebased past it. Until #314 merges, this diff shows its four commits too; the reviewable change here is the top commit (`spec_reverifier.go` + three test files). Gets a final rebase when #314 lands.

Third PR of the 429/Retry-After stack. Supersedes the closed "a rate-limited probe is inconclusive" PR (its classification logic and test suite are salvaged here) and the closed "back off a rate-limited provider" PR (its hold-off semantics land here on the shared registry instead of a private backoff).

## Why

Re-verification treats a 429 like any other `Validate` failure: the demote streak advances and two consecutive rate-limited epochs drop the provider out of the pairing. A rate-limit says we asked too fast — nothing about whether the node serves what it declares. In production this demoted providers that were serving traffic fine, because the fleet's own synchronized verification burst blew a shared vendor cap. And nothing backed off: `Validate` retries 3× with zero delay and the epoch re-asks on the same schedule.

## What changed

- `isRateLimitFailure` — structural detection (`errors.Is` on `common.StatusCodeError429`, kept readable by the propagation PR), plus the one text signature gRPC forces (no status code exists there; MAG-2949 adds the typed path for gRPC).
- A rate-limited re-verify is **inconclusive**: streak untouched, membership unchanged, an inactive provider not promoted on no evidence. Genuine capability failures demote exactly as before.
- The pass **holds off** a rate-limited provider via the shared registry (`docs/RATE-LIMIT-HOLDOFF.md`): while held, the probe is skipped and the skip returns the rate-limit error, so reconciliation reaches the same inconclusive verdict without spending a request. Retry-After floors the hold-off; any other outcome clears it.

## How to verify

```
go build ./...
go test ./protocol/rpcsmartrouter/ -run 'TestIsRateLimitFailure|TestApplyReverification_RateLimit|TestApplyReverification_Real|TestApplyReverification_Mixed|TestReverify_' -v
go test ./protocol/rpcsmartrouter/ ./protocol/holdoff/
```

The suite covers: never-demotes across 10 rate-limited cycles, real failures still demote, no promotion on a 429, mixed sets, probe-skip-while-held (probe call counts), clear-on-answer with strike reset, and the Retry-After floor.

### DIFF
--- a/docs/RATE-LIMIT-HOLDOFF.md
+++ b/docs/RATE-LIMIT-HOLDOFF.md
@@ -30,7 +30,7 @@ off" means on its path. This table is the catalog; add a row when wiring a new c
 
 | Path | What "held off" means there | Records into the registry | Status |
 | --- | --- | --- | --- |
-| Spec re-verification (`spec_reverifier.go`) | Skip the probe, return the rate-limit error so reconciliation stays inconclusive — membership unchanged, streak untouched — without spending a request | 429 from `Validate` | lands with MAG-2910 |
+| Spec re-verification (`spec_reverifier.go`) | Skip the probe, return the rate-limit error so reconciliation stays inconclusive — membership unchanged, streak untouched — without spending a request | 429 from `Validate` | live |
 | Hot relay path (endpoint selection) | Prefer endpoints that are not held off; when every endpoint is held off, fall back to the soonest-to-expire one — the customer is never answered with a synthesized 429 | HTTP 429 relay results | lands with MAG-2948 |
 | Recovery probe (`recovery_probe.go`) | Skip the replay while held off (verdict stays inconclusive), instead of burning replay-attempt budget on a vendor that said stop | 429 probe responses | lands with MAG-2948 |
 | Chain tracker / endpoint poller | The poll backoff takes the upstream's Retry-After as a floor instead of guessing with the fail-count doubling | 429 poll responses | lands with MAG-2949 |
--- a/docs/RATE-LIMIT-HOLDOFF.md
+++ b/docs/RATE-LIMIT-HOLDOFF.md
@@ -1,0 +1,44 @@
+# Rate-limit hold-off registry
+
+`protocol/holdoff` is the shared answer to "an upstream said 429 — when may we ask it
+again?". Every path that sends requests upstream records rate limits into one
+`holdoff.Registry` and consults it before sending, instead of growing a private backoff.
+
+## Semantics
+
+- **Tiered keying.** A 429 holds off the endpoint URL that returned it. When 2 distinct
+  URLs of one provider name are held off at once, the hold-off escalates to the provider
+  name — a vendor cap is account-wide, and per-URL keying alone would let the process
+  keep hammering the account through its other chains. The provider-wide hold-off ends
+  when its longest-held member does.
+- **Retry-After is the floor.** When the upstream said how long to wait
+  (`common.RetryAfterFrom` on the typed error), the hold-off is at least that long,
+  bounded at 1h. Otherwise a per-URL exponential: 30s doubling per consecutive strike,
+  capped at 30m. Jitter extends each hold-off by up to +20% and never shortens it, so a
+  fleet held off by one vendor-wide limit does not return in a synchronized burst.
+- **Any answer clears.** An upstream that answered a request — success or a genuine
+  failure — is no longer refusing us for load: the URL's strikes and the provider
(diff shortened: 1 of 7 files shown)


avitenzer (approved):

avitenzer:
## Review — "a rate-limited re-verify is inconclusive, and held off"

**Verdict: approve.** Reviewed the top commit (`7432b4f`) as the reviewable change per the stacking note — the other four commits belong to #314. Checked out the branch, built it, and ran `./protocol/rpcsmartrouter/` and `./protocol/holdoff/` including with `-race`; everything passes. One design question worth settling before merge, plus minor notes.

### The change is sound

- The classification (`isRateLimitFailure`) is structural-first via `errors.Is(err, common.StatusCodeError429)`, with exactly one text signature for gRPC, where no status code survives. The test suite covers real production shapes on both sides, including the "block number containing 429" false-positive trap and a 504 wrapped identically.
- The reconciliation branch is placed correctly: after the success branch, before both failure branches, so an active rate-limited provider stays paired, an inactive one isn't promoted, and the streak is untouched. The synthesized-429 skip path funnels into the same branch, so held-off epochs are indistinguishable from fresh-429 epochs — a nice invariant.
- Concurrency is fine: the registry is mutex-guarded, the `inputs.rateLimitHoldoff = holdoff.Shared` fallback happens before goroutines spawn and under `RPCSmartRouter.mu`, and the static→backup tier ordering means a 429 recorded by the static pass correctly holds off the same provider in the backup pass of the same epoch.
- Production wiring leaves `rateLimitHoldoff` nil → falls back to `holdoff.Shared`, matching the cross-chain escalation intent.

### One design question: the streak survives long inconclusive gaps

Rate-limited epochs leave `demoteFailStreak` untouched — neither advanced nor cleared. So this sequence demotes:

1. Epoch N: genuine failure → streak = 1 (grace, stays paired)
2. Epochs N+1 … N+k: rate-limited / held off — streak frozen at 1, possibly for 30+ minutes (`MaxHoldoff`)
3. Epoch N+k+1: genuine failure → streak = 2 → **demoted**

MAG-2445's rationale ("a blip must persist across two epoch boundaries") assumed the two failures are in *consecutive* epochs. Here they can be an hour apart with no evidence in between, and the second failure demotes immediately. That's defensible (two real failures with zero successes between them), but the comment on the inconclusive branch in `spec_reverifier.go` only argues for not *advancing* the streak, not for *preserving* it. If the intent is "inconclusive epochs are invisible to the streak," say so in the comment; if not, clearing the streak on a rate-limited epoch would be the conservative choice. Worth one sentence either way.

- **`RecordAnswer` fires on non-answers.** The wrapper clears strikes on any non-429 outcome, including `connection refused` and probe timeouts — but the holdoff package's contract is "an upstream that *answered* … is no longer refusing us for load." A vendor flapping 429 → dial-timeout → 429 never escalates its exponential backoff. Practically cheap (the next 429 re-arms a 30s hold), but the comment claims a stronger justification than the code implements. Fine to ship as-is; consider a note or excluding transport-level errors later.
- **`holdoffURLKey` uses only `NodeUrls[0]`,** while `validateProvider` probes all of a provider's URLs (plus addon expansions). A 429 from the second URL is recorded against the first URL's key. Harmless for this consumer, but when MAG-2948 wires the relay path (which will record the *actual* URL), the same physical endpoint can appear under two keys. A one-line comment acknowledging the approximation would help the next consumer.
- **Boot validation stays 429-blind.** `validateProviderTier` still excludes a provider on a boot-time 429 and records nothing into the registry — and the synchronized fleet burst that motivated this PR is exactly a boot-shaped burst. The docs table doesn't list the boot path in the catalog. Probably deliberate scope (re-verify promotes it back once it passes), but worth a row in `RATE-LIMIT-HOLDOFF.md` marked "not wired" so it reads as a decision rather than an omission.
- **Log volume:** a held-off provider emits a warning per epoch per tier ("treating as inconclusive"). At 15-minute epochs this is negligible; just noting the skip-path debug + reconcile-warning pair is slightly redundant.

The PR description accurately matches the diff, and the doc table flip to "live" is correct. Once #314 lands and this gets its final rebase, it's good to go.

PR #317, opened 2026-08-20
fix(smart-router): a rate-limited relay is neither failure nor success

nimrod-teich:
#Closes MAG-2948

Jira ticket: MAG-2948

> **Stacked** on PR #314 in smart-router (hold-off registry, MAG-2947) — PR #313 already merged and this branch is rebased past it. Until #314 merges, this diff shows its four commits too; the reviewable change here is the top commit. Gets a final rebase when #314 lands.

Fourth PR of the 429/Retry-After stack: the hot path.

## Why

The relay path scores the two shapes of rate-limit in inverted proportion to their severity:

- An explicit HTTP 429 becomes `err != nil`, and `shouldFailSessionForResult` short-circuits on err before its `IsRateLimited` carve-out — a full availability failure. Fifteen consecutive ones hard-block a healthy-but-busy endpoint, the outcome the rate-limit contract (`error_registry.go`) explicitly forbids.
- A 2xx-body or gRPC-text rate limit passes the gate and is reported as a full success: availability 1 plus a latency sample so fast it ranks the limiter above endpoints doing real work.
- No cooldown exists anywhere: the retry excludes the endpoint within the request, and the very next request re-picks it immediately.

## What changed

- **Session accounting**: both shapes now release the session via `OnSessionRateLimited` (the `releaseWithoutPenalty` primitive the client-cancel path already uses) — no QoS sample in either direction, no health verdict. The response the client receives is untouched.
- **Recording**: the attempt lands in the shared hold-off registry keyed (provider, node URL), Retry-After as the floor. An answered relay clears it.
- **Selection**: `getValidProviderAddresses` filters held-off providers before the optimizer chooses, so the in-request retry and subsequent requests prefer somewhere else. When every candidate is held off, the **soonest-to-expire** one stays in — the request is still served; Retry-After stays internal and a customer is never answered with a synthesized 429. A header-pinned provider and existing sticky sessions bypass the filter: an explicit ask outranks the hold-off.
- **Recovery probe**: a 429 replay records the endpoint (honouring `HTTPStatusError.RetryAfter`, which it previously held and ignored) and further replays are skipped while held — inconclusive without spending the request — instead of re-asking on the 5s probe cadence. An answered replay clears.

## How to verify

```
go build ./...
go test ./protocol/lavasession/ -run TestFilterRateLimited -v
go test ./protocol/rpcsmartrouter/ -run 'TestIsRateLimitedRelayOutcome|TestReplayFailingRelay' -v
go test ./protocol/lavasession/ ./protocol/rpcsmartrouter/ ./protocol/holdoff/
```

### DIFF
--- a/docs/RATE-LIMIT-HOLDOFF.md
+++ b/docs/RATE-LIMIT-HOLDOFF.md
@@ -31,8 +31,8 @@ off" means on its path. This table is the catalog; add a row when wiring a new c
 | Path | What "held off" means there | Records into the registry | Status |
 | --- | --- | --- | --- |
 | Spec re-verification (`spec_reverifier.go`) | Skip the probe, return the rate-limit error so reconciliation stays inconclusive — membership unchanged, streak untouched — without spending a request | 429 from `Validate` | live |
-| Hot relay path (endpoint selection) | Prefer endpoints that are not held off; when every endpoint is held off, fall back to the soonest-to-expire one — the customer is never answered with a synthesized 429 | HTTP 429 relay results | lands with MAG-2948 |
-| Recovery probe (`recovery_probe.go`) | Skip the replay while held off (verdict stays inconclusive), instead of burning replay-attempt budget on a vendor that said stop | 429 probe responses | lands with MAG-2948 |
+| Hot relay path (endpoint selection) | A rate-limited relay releases its session with no QoS sample in either direction, and selection prefers providers that are not held off; when every candidate is held off, the soonest-to-expire one stays in — the customer is never answered with a synthesized 429 | 429 relay results, both shapes (typed HTTP error, 2xx-body/gRPC classification) | live |
+| Recovery probe (`recovery_probe.go`) | Skip the replay while held off (verdict stays inconclusive), instead of burning replay-attempt budget on a vendor that said stop | 429 probe responses (Retry-After floor honoured) | live |
 | Chain tracker / endpoint poller | The poll backoff takes the upstream's Retry-After as a floor instead of guessing with the fail-count doubling | 429 poll responses | lands with MAG-2949 |
 | WS / gRPC transports | Recognition first: a 429 on the WS upgrade or a corroborated gRPC rate limit produces the typed sentinel these consumers key on | handshake / metadata 429s | lands with MAG-2949 |
 
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/magma-Devs/smart-router/protocol/common"
 	"github.com/magma-Devs/smart-router/protocol/endpointtip"
+	"github.com/magma-Devs/smart-router/protocol/holdoff"
 	metrics "github.com/magma-Devs/smart-router/protocol/metrics"
 	"github.com/magma-Devs/smart-router/protocol/provideroptimizer"
 	"github.com/magma-Devs/smart-router/protocol/qos"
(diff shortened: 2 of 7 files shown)


avitenzer:
## Review — "a rate-limited relay is neither failure nor success"

**Verdict: approve.** Reviewed the top commit (`84ebc0a`) as the reviewable change per the stacking note — the other four commits belong to #314. Checked out the branch, built it, and ran `./protocol/lavasession/`, `./protocol/rpcsmartrouter/`, and `./protocol/holdoff/`, including the new tests under `-race`. Everything passes, and the three claims in the description all hold in the code. Nothing blocking; minor notes below.

### What I verified

- **Session accounting is correct.** The new `isRateLimitedRelayOutcome` branch sits after `isClientCancel` and before both scoring arms, so a rate-limit in either shape (typed 429 error, or `IsRateLimited` flag on a 2xx-body/gRPC classification) reaches neither `OnSessionDone` nor `OnSessionFailure`. `OnSessionRateLimited` is a thin alias over the existing `releaseWithoutPenalty` primitive — CU returned, session freed, no QoS sample, no consecutive-error increment. A side effect that looks intentional and good: a provider on second-chance probation can no longer clear probation with a rate-limited "success", and tip harvesting is skipped for rate-limited responses.
- **The typed-429 path is airtight.** `relayInnerDirect` converts every HTTP 429 into `httpStatusRelayError` (the typed sentinel with Retry-After attached, from #313) before the gate ever sees it, so the `err != nil` shape is always caught. The client-visible response is genuinely untouched — only the session-manager callback choice changed.
- **Selection filter placement is right.** Header-pinned providers and sticky sessions return before `filterRateLimitedProviders` runs, matching the "explicit ask outranks the hold-off" claim. The fallback logic correctly excludes held providers already in the request's ignored list (they've been tried), and the all-held → soonest-to-expire path guarantees the request is still served. The filter test suite covers the branch combinations including the tricky ignored-interaction cases.
- **Recovery probe wiring is consistent**: skip-while-held returns inconclusive without touching the connection (verified by the `gotPayload` assertion), 429 replays record with `HTTPStatusError.RetryAfter` honoured — fixing that field being previously held-and-ignored — and recovery clears strikes.

- **Stale comment at the gate call site.** `rpcsmartrouter_server.go` (just above the `shouldFailSessionForResult` call in `sendRelayToDirectEndpoints`) still says "For REST 5xx/429, err == nil" — since #313, both arrive as typed errors and the `StatusCode >= 500 || == 429` clause in `shouldFailSessionForResult` is defense in depth. Worth updating so a reader doesn't conclude the 429-with-nil-err shape is live.
- **Symmetric defense in depth**: `isRateLimitedRelayOutcome` could also check `relayResult.StatusCode == 429`, mirroring the gate's own fallback clause. Today it's unreachable (429 always arrives typed on this path), but if any future transport sets the bare status without the typed error or the flag, it would silently land in `OnSessionFailure` — the exact bug this PR fixes.
- **Three consumers, three clearing policies.** The registry contract says "any answer clears," but: reverify (#316) clears on any non-429 outcome including timeouts; the hot path clears only on the success arm (a 5xx answer does *not* clear); the probe clears only on `relayProbeRecovered` (a still-failing *answer* does not clear). All individually defensible and the differences are conservative, but the holdoff package doc overpromises relative to two of its three consumers. One clarifying sentence in `holdoff.go` or the catalog doc would prevent the next consumer from guessing.
- **Cross-PR key alignment with #316**: this PR records `(PublicLavaAddress, targetEndpoint.NetworkAddress)` — the URL actually dialed — while reverify records `(Name, NodeUrls[0].Url)`. Names align by existing convention, but for multi-URL providers the URL keys can differ for the same physical endpoint, so hold-offs from the two paths won't always merge (they do still meet at the provider-escalation tier). Same note I left on #316; a comment on `holdoffURLKey` is enough.
- **Probe edge case**: when `ep.Endpoint` is nil, the probe records under URL `""`. Harmless (consistent within the probe), but an explicit fallback to `ep.ProviderAddress` like reverify's would keep keys meaningful.
- **Coverage gap, acknowledged**: the hot-path branch itself (the `sendRelayToDirectEndpoints` goroutine wiring — `RecordRateLimit` + `OnSessionRateLimited` + `RecordAnswer` on the success arm) has no direct test; the tests cover the classifier, the filter, and the probe. Given how entangled that closure is, a reasonable trade — but it's the one piece where a regression wouldn't be caught by this suite.

The doc table updates (both rows flipped to "live", with the accounting semantics added to the hot-path row) match the diff. Same merge condition as #316: final rebase once #314 lands.

avitenzer (approved):

PR #318, opened 2026-08-20
feat(smart-router): recognize 429s on ws + gRPC, floor the tracker poll

nimrod-teich:
#Closes MAG-2949

Jira ticket: MAG-2949

Fifth PR of the 429/Retry-After stack: the transports where a 429 was invisible, and the poller that ignored what it had already parsed.

## What changed

- **WS handshake** (`chainproxy/rpcclient/websocket.go`) — the dial kept only `resp.Status` as a string and dropped the headers, so a 429 upgrade rejection looked like any dial failure to all four WS consumers (jsonRPC ws, tendermint ws, upstream ws pool, direct WS connection). `wsHandshakeError` now records the status code and parsed Retry-After, and `Unwrap` presents a 429 as the typed rate-limit error.
- **gRPC** — no HTTP status exists, so recognition needs corroboration. `common.RateLimitFromGRPC` keys on the known rate-limit texts (including the `429 (Too Many Requests)` a vendor's HTTP edge leaves inside `codes.Unavailable`), or on `RESOURCE_EXHAUSTED` backed by a retry delay in the response metadata (`grpc-retry-pushback-ms` / `retry-after`). Code 8 alone never suffices — grpc-go mints it for an oversized message too. Both gRPC paths consult it: `GRPCStatusError.Unwrap` scopes to the typed error on the direct connection, and the chainlib proxy fails the send typed instead of converting the rejection into a node-error reply. `common.RateLimited` is the constructor for these no-HTTP-status transports.
- **Chain tracker floor** (`chaintracker`) — the poll backoff was `base<<fails` capped at 60s regardless of what the upstream asked, while the parsed Retry-After sat one call frame away. The poll loop now notes each real poll's outcome and `computePollInterval` floors the next interval with it, so a 429 with `Retry-After: 300` is not re-asked at the ceiling. Any answered poll clears the floor; a traffic-gate skip (which contacts nobody) neither sets nor clears it. The fork-hash fetch also stops severing the typed error.

## How to verify

```
go build ./...
go test ./protocol/common/ -run 'TestRateLimitFromGRPC|TestRateLimitedConstructor' -v
go test ./protocol/chainlib/chainproxy/rpcclient/ -run 'TestDialWebsocket_' -v
go test ./protocol/chaintracker/ -run TestComputePollInterval_RetryAfterFloors -v
go test ./protocol/lavasession/ -run 'TestHandleGRPCError_' -v
```

Negative cases pinned: oversized-message code 8 stays unclassified, a 503 upgrade stays untyped, a short Retry-After never shortens the poll interval.

### DIFF
--- a/protocol/chainlib/chainproxy/rpcclient/websocket.go
+++ b/protocol/chainlib/chainproxy/rpcclient/websocket.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/gorilla/websocket"
+	"github.com/magma-Devs/smart-router/protocol/common"
 )
 
 const (
@@ -40,8 +41,10 @@ const (
 var wsBufferPool = new(sync.Pool)
 
 type wsHandshakeError struct {
-	err    error
-	status string
+	err        error
+	status     string
+	statusCode int
+	retryAfter time.Duration
 }
 
 func (e wsHandshakeError) Error() string {
@@ -50,6 +53,17 @@ func (e wsHandshakeError) Error() string {
 		s += " (HTTP status " + e.status + ")"
 	}
 	return s
+}
+
+// Unwrap presents a 429 upgrade rejection as the typed rate-limit error, so
+// errors.Is(err, common.StatusCodeError429) and common.RetryAfterFrom read the same on a
+// refused WS handshake as on every HTTP transport. Any other status unwraps to the dial
+// error itself.
+func (e wsHandshakeError) Unwrap() error {
+	if e.statusCode == http.StatusTooManyRequests {
+		return &common.RateLimitedError{RetryAfter: e.retryAfter}
+	}
+	return e.err
 }
 
 // DialWebsocketWithDialer creates a new RPC client that communicates with a JSON-RPC server
@@ -64,7 +78,11 @@ func DialWebsocketWithDialer(ctx context.Context, endpoint string, dialer websoc
 		if err != nil {
 			hErr := wsHandshakeError{err: err}
 			if resp != nil {
+				// Capture the rejection while the response is in scope — the status code
+				// types a 429 (see Unwrap), and Retry-After is what any backoff needs.
 				hErr.status = resp.Status
+				hErr.statusCode = resp.StatusCode
+				hErr.retryAfter, _ = common.ParseRetryAfter(resp.Header, time.Now())
 			}
 			return nil, hErr
 		}
--- a/protocol/chainlib/chainproxy/rpcclient/ws_handshake_429_test.go
+++ b/protocol/chainlib/chainproxy/rpcclient/ws_handshake_429_test.go
@@ -1,0 +1,54 @@
+package rpcclient
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/magma-Devs/smart-router/protocol/common"
+	"github.com/stretchr/testify/require"
+)
+
+// A 429 on the WS upgrade must surface as the typed rate-limit error with the upstream's
+// Retry-After attached — the handshake used to keep only the status string and drop the
+// headers, leaving the rejection invisible to every consumer.
+func TestDialWebsocket_429UpgradeIsTypedRateLimit(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Retry-After", "120")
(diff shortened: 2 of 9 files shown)


avitenzer:
## Review — "recognize 429s on ws + gRPC, floor the tracker poll"

**Verdict: approve.** Unlike #316/#317 this PR is based directly on main (single commit `c06592d`), so the whole diff is reviewable. Checked out the branch, built it, and ran the five affected packages plus the new suites under `-race` — all green. The design is careful in exactly the places that needed care; nothing blocking.

### What I verified

- **The gRPC corroboration rule is the right call and correctly implemented.** `RateLimitFromGRPC` accepts a known rate-limit text unconditionally, but `RESOURCE_EXHAUSTED` (code 8) only when the metadata carries a real delay (`grpc-retry-pushback-ms` or `retry-after`) — and the oversized-message false positive, the one production shape that shares code 8, is pinned by a test. Both gRPC paths consult it: the chainlib proxy fails the send typed instead of synthesizing a node-error reply, and the direct connection scopes `GRPCStatusError.Unwrap`.
- **The `GRPCStatusError.Unwrap` change doesn't break its existing consumer.** `direct_rpc_relay.go` reaches it via `errors.As`, which matches the wrapper itself before unwrapping, so `.Code`/`.Message` stay visible; only chain traversal beyond the wrapper changes. The rate-limit texts can't collide with the `SessionOutOfSyncGRPCCode` (677) contract the old comment pins, since 677 is minted internally with a non-matching message.
- **The tracker floor is reachable end-to-end.** The HTTP-family proxies already attach `RateLimitedError` via `WithRetryAfter` (from #313, on main), `FetchLatestBlockNum` returns the fetch error unmodified, and the one site that severed the typed error (the fork-hash fetch's `LavaFormatDebug`-attribute idiom) is fixed here. The ordering in the poll loop is right — `noteFetchOutcome` runs before `updateTimer` → `computePollInterval` reads the floor — and the traffic-gate skip correctly neither sets nor clears it. The forced-poll path (`PollNow`) also feeds the floor without touching the schedule.
- **The WS handshake fix is minimal and tested against a real HTTP server**, including the negative 503 case, with Retry-After parsed while the response is still in scope.
- **`RateLimited` (the constructor) composes properly**: double-`%w` keeps both the sentinel and the original cause reachable, retryAfter is clamped, and a zero delay degrades to the bare typed sentinel so `RetryAfterFrom` correctly reports "no opinion".

- **Terminal-classification severing, twice.** Both new `Unwrap`s (`wsHandshakeError` on a 429, `GRPCStatusError` when rate-limited) replace the original cause with a fresh `RateLimitedError` — so `status.Code(err)` on a corroborated gRPC rate limit now resolves to `Unknown`, and a 429 WS handshake no longer unwraps to the underlying dial error (e.g. for `net.Error` introspection). Both are deliberate ("rate-limit is the terminal classification") and I found no consumer that breaks, but the `GRPCStatusError` doc comment still opens with the old "makes status.Code(err) resolve to the underlying gRPC code" claim, which is now only true for the non-rate-limited arm — worth a one-line amendment.
- **Signature-list philosophy diverges from #316.** #316 keeps its text list minimal on principle ("a new entry is a signal some path is discarding a typed error"); this PR's gRPC list is deliberately broader (`"rate limit"`, `"ratelimit"`, etc., lowercased `Contains`). Defensible — gRPC status messages are a much narrower channel than arbitrary wrapped error strings — but a sentence saying why the philosophies differ would prevent someone "harmonizing" them later. Pleasant interlock: once both PRs merge, gRPC 429s arrive typed at the source, and #316's text signature becomes a true fallback rather than the primary detection.
- **The docs catalog can't be flipped here — don't forget it later.** `docs/RATE-LIMIT-HOLDOFF.md` lands with #314, which this branch (based on main) doesn't have, so the two rows this PR implements ("Chain tracker / endpoint poller" and "WS / gRPC transports") will still read "lands with MAG-2949" after everything merges. Needs a trivial follow-up flip to "live" once the stack settles — worth a note in the PR description so it isn't lost.
- **The clearing wrinkle from the sibling PRs appears here too**: `noteFetchOutcome` clears the floor on any non-rate-limit outcome, including a local fetch timeout where the upstream never actually answered. Same conservative trade as #316/#317, same one-line-of-doc fix if it ever matters.
- **Recognition vs. consumption on WS**: the four WS consumers now *can* see the typed 429, but nothing backs off on it yet — consistent with the catalog's "Recognition first" phrasing for this row; just noting the scope boundary is intentional.

The verification commands in the PR description all pass as written. No stacking dependency — mergeable independent of #314's timing.

avitenzer (approved):

PR #319, opened 2026-08-20
fix(smart-router): a rate-limited Validate attempt ends the retry burst

nimrod-teich:
#Closes MAG-2976

Jira ticket: MAG-2976

Closes the last finding of the MAG-2910 audit that no other PR in the 429/Retry-After stack covers. Depends only on merged PR #313 in smart-router (typed-429 propagation) — single commit on main, independently mergeable.

## Why

`Validate` retries its latest-block fetch and each verification 3 times back-to-back with zero delay ("we give several chances for starting up") — a budget written for a cold-booting node. Against a rate limiter it is counterproductive: all three attempts land inside the same limiter window, fail identically, and the two extra requests only add to the load that produced the 429. This burst runs on every startup verification and every epoch re-verification, for every configured provider, five in parallel — it is the fleet-side amplifier of the shared-vendor-cap incident that started this stack.

The re-verify hold-off (PR #316) stops the pass from re-probing across epochs; this closes the burst inside a single `Validate` call.

## What changed

- `stopValidateRetries` — the retry loops break early when the attempt failed with the typed rate-limit error. Applied to all four loops: `Validate`'s latest-block + verification loops, and `ValidateCollect`'s (the `smartrouter health` command) same pair.
- The error propagates unchanged, so downstream classification (inconclusive in re-verify) is untouched. Every other failure keeps the full 3-attempt budget.

## How to verify

```
go build ./...
go test ./protocol/chainlib/ -run 'TestValidate_' -v
go test ./protocol/chainlib/...
```

The two tests drive `Validate` through a real ETH1 parser against a counting router: a rate-limited upstream is asked exactly once per probe, a plainly-failing one still gets its 3 startup attempts.

### DIFF
--- a/protocol/chainlib/chain_fetcher.go
+++ b/protocol/chainlib/chain_fetcher.go
@@ -3,6 +3,7 @@ package chainlib
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strconv"
 	"sync/atomic"
@@ -121,6 +122,14 @@ func needsLatestBlock(url common.NodeUrl, verifications []VerificationContainer)
 	return false
 }
 
+// stopValidateRetries reports whether a failed attempt should end Validate's zero-delay
+// retry budget early. The budget exists for a cold-booting node; a rate-limited attempt
+// is different in kind — the extra requests land inside the same limiter window and only
+// add to the load that produced the 429.
+func stopValidateRetries(err error) bool {
+	return err == nil || errors.Is(err, common.StatusCodeError429)
+}
+
 func (cf *ChainFetcher) Validate(ctx context.Context) error {
 	for _, url := range cf.endpoint.NodeUrls {
 		utils.LavaFormatInfo("starting validation for url", utils.LogAttr("url", url.String()))
@@ -137,7 +146,7 @@ func (cf *ChainFetcher) Validate(ctx context.Context) error {
 		if needsLatestBlock(url, verifications) {
 			for attempts := 0; attempts < 3; attempts++ {
 				latestBlock, err = cf.FetchLatestBlockNum(ctx)
-				if err == nil {
+				if stopValidateRetries(err) {
 					break
 				}
 			}
@@ -157,7 +166,7 @@ func (cf *ChainFetcher) Validate(ctx context.Context) error {
 			var err error
 			for attempts := 0; attempts < 3; attempts++ {
 				err = cf.Verify(ctx, verification, uint64(latestBlock))
-				if err == nil {
+				if stopValidateRetries(err) {
 					break
 				}
 			}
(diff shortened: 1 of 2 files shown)


avitenzer:
## Review — "a rate-limited Validate attempt ends the retry burst"

**Verdict: approve.** The smallest and cleanest PR of the stack: one commit (`3809a7e`) on main, one 8-line helper plus four one-line substitutions, and a focused test pair. Checked out the branch, built it, ran the full `./protocol/chainlib/...` tree, and ran the two new tests under `-race` — everything passes. Nothing blocking.

### What I verified

- **The change is behavior-preserving except for exactly the intended case.** `stopValidateRetries(err)` returns true for `err == nil`, so the success break is untouched, and each substitution is a strict widening of the break condition. I read the post-loop handling at all four sites: the error propagates identically whether the loop broke early or exhausted its budget — `Validate` returns the latest-block error / applies severity handling on verification errors, `ValidateCollect` records the failure into the structured result. So the re-verify pass (#316) receives the same typed 429 and classifies it inconclusive exactly as before.
- **Coverage is complete.** Grepped for other zero-delay `attempts < 3` retry loops in `protocol/` — the four sites this PR touches are the only ones.
- **The tests prove the right thing at the right level.** They drive the real `Validate` through a real ETH1 spec parser against a counting router, so the assertion "asked exactly once" also implicitly proves that the error survives `Verify`'s wrapping with `errors.Is` intact — which is the load-bearing assumption. The negative test pins the 3-attempt budget for a plain failure, so the cold-boot behavior can't silently regress. (The `rateLimited429` helper is reused from #313's existing test file, which is why it isn't in this diff.)
- **Dependency claim checks out**: only #313 (merged) is needed; the branch is independently mergeable regardless of where #317/#318 stand.

- **Interim gRPC gap, closes itself.** Detection is structural only (`errors.Is` on the sentinel) — deliberately no text-signature fallback like #316's. Until #318 merges, a gRPC 429 (which today survives only as text inside `codes.Unavailable`) still burns all 3 attempts. Once #318 lands, the chainlib gRPC proxy types the error at the source and this helper starts catching it with no change here. That's the right layering — a text match here would be the exact anti-pattern #316's comment warns against — but a line in the helper's comment noting "gRPC coverage arrives with the typed gRPC errors (MAG-2949)" would stop someone from 'fixing' the gap with a string match.
- **Per-verification scope is intentional but worth knowing**: a rate-limited `Validate` still sends one request per verification (plus one for the latest block) — the burst collapses 3→1 per probe, not to 1 per call. That matches the PR's "asked exactly once per probe" phrasing; the cross-verification suppression is #316's hold-off job, not this one's.
- **Context cancellation still burns the budget**: a `Validate` whose ctx expired retries its remaining attempts against a dead context (each fails instantly). Pre-existing behavior, cheap, and out of scope — just noting the helper could grow a `ctx.Err()` arm someday if anyone cares.

The verification commands in the PR description pass as written.

avitenzer (approved):

PR #320, opened 2026-08-20
fix(smart-router): ws subscriptions hold off a rate-limited endpoint

nimrod-teich:
#Closes MAG-2977

Jira ticket: MAG-2977

> **Stacked** on PR #317 in smart-router (hot-path rate-limit handling, MAG-2948) and PR #318 (ws/gRPC recognition, MAG-2949) — their commits show in this diff until they merge; the reviewable change here is the top commit. Gets a final rebase when they land.

Follow-up to the 429/Retry-After stack: the last consumer path where a rate limit was still punished.

## Why

PR #318 makes a 429 on the WS upgrade a typed error — but nothing on the subscription path consumed it. `direct_ws_subscription_manager.go` called `AppendRelayFailure` unconditionally on any connect/subscribe failure (an availability sample of 0 for an endpoint that is busy, not broken — the same inversion #317 fixed on the HTTP hot path), and the pool's reconnect ladder re-dialed a rate-limited upstream at its 100ms–30s cadence regardless of what the handshake asked.

## What changed

- `noteSubscribeFailure` — a rate-limited connect/subscribe records the endpoint into the shared hold-off registry (Retry-After as the floor) instead of scoring it; every other failure penalizes exactly as before.
- `selectFromTier` skips held-off endpoints while something ready remains; when everything is held, the full tier stays in — a subscription must still be served.
- A successful subscribe clears the endpoint's hold-off.
- `reconnectDelay` floors the reconnect ladder with the dial error's Retry-After.
- Keying is per WS URL on both registry tiers — this path works on bare node URLs with no provider name in scope, so vendor-tier escalation deliberately does not apply (documented in the consumer table).

## How to verify

```
go build ./...
go test ./protocol/rpcsmartrouter/ -run 'TestWSNoteSubscribeFailure|TestWSSelectFromTier|TestReconnectDelay' -v
go test ./protocol/rpcsmartrouter/
```

The reconnect test drives the exact production wrapping: a real `rpcclient.DialWebsocket` against a 429+Retry-After test server, wrapped by `connect`'s `%w`.

### DIFF
--- a/docs/RATE-LIMIT-HOLDOFF.md
+++ b/docs/RATE-LIMIT-HOLDOFF.md
@@ -48,6 +48,7 @@ off" means on its path. This table is the catalog; add a row when wiring a new c
 | Hot relay path (endpoint selection) | A rate-limited relay releases its session with no QoS sample in either direction, and selection prefers providers that are not held off; when every candidate is held off, the soonest-to-expire one stays in — the customer is never answered with a synthesized 429 | 429 relay results, both shapes (typed HTTP error, 2xx-body/gRPC classification) | live |
 | Recovery probe (`recovery_probe.go`) | Skip the replay while held off (verdict stays inconclusive), instead of burning replay-attempt budget on a vendor that said stop | 429 probe responses (Retry-After floor honoured) | live |
 | Chain tracker / endpoint poller | The poll backoff takes the upstream's Retry-After as a floor instead of guessing with the fail-count doubling | 429 poll responses | lands with MAG-2949 |
+| WS subscriptions (`direct_ws_subscription_manager.go`) | A rate-limited connect/subscribe is held off instead of scored (no availability sample), selection skips held endpoints while something ready remains, and the pool's reconnect ladder is floored by the dial's Retry-After. Keyed per WS URL — no provider name exists on this path, so vendor-tier escalation does not apply | 429 handshakes and subscribe failures | live |
 | WS / gRPC transports | Recognition first: a 429 on the WS upgrade or a corroborated gRPC rate limit produces the typed sentinel these consumers key on | handshake / metadata 429s | lands with MAG-2949 |
 
 ## Relation to `common/retry_after.go`
--- a/protocol/rpcsmartrouter/direct_ws_subscription_manager.go
+++ b/protocol/rpcsmartrouter/direct_ws_subscription_manager.go
@@ -3,6 +3,7 @@ package rpcsmartrouter
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"sync"
 	"sync/atomic"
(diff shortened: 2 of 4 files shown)


avitenzer:
## Review — "ws subscriptions hold off a rate-limited endpoint"

**Verdict: approve — with one precondition: the final rebase before merging.** The reviewable change is the top commit (`c74137f`; the rest is the now-merged #317/#318 stack). Built and tested it on its own base, then replayed just the top commit onto current main (which now also has #319 and the holdoff-bounds fix `ddd0245`) — it applies with zero conflicts, including the doc table row, and everything passes including the new tests under `-race` against the tightened registry.

### What I verified

- **`noteSubscribeFailure` is the right shape.** A typed 429 records `(url, url)` into the shared registry with Retry-After as the floor and produces no availability sample; every other failure calls `AppendRelayFailure` exactly as before. Both failure sites in `StartSubscription` (connection acquisition and upstream subscribe) are converted, and a successful subscribe clears via `RecordAnswer`.
- **The error chain is genuinely `%w` all the way**: `DialWebsocket` (typed by #318) → `connect`'s `"failed to dial WebSocket %s: %w"` → `GetConnection`'s `"failed to create connection: %w"` → `noteSubscribeFailure`. The reconnect test drives a real dial against a real 429+Retry-After server through that exact wrapping — the load-bearing assumption is tested, not assumed.
- **`selectFromTier` narrowing is correct by construction**: it narrows only when `0 < ready < all`, so the all-held case keeps the full tier and a subscription is always attempted — the same "must still be served" doctrine as #317. The sticky-session path bypasses the filter, consistent with #317's "explicit ask outranks."
- **`reconnectDelay` sits in the right spot** — after the ladder picks its delay, before the sleep, flooring with the current dial error's ask and never shortening.
- **The per-URL keying decision is sound and documented**: no provider name exists on this path, so `(url, url)` keys both tiers and vendor escalation intentionally can't fire; no namespace collision with the provider-keyed consumers.

### Findings (none blocking)

- **Latent gap: the hold-off narrowing ignores `ignoredEndpoints`.** With tier = [A held-off + not ignored, B ready + ignored], `notHeldOff` narrows to [B], and selection then fails with "all endpoints ignored" — even though A was the legitimate fallback. This is precisely the case #317's HTTP filter handles with its `readyChoosable` counting. It's **currently unreachable**: the WS `selectEndpoint` is only ever called with a nil ignored set in production. But the machinery exists, so either handle ignored in the narrowing now (cheap) or leave a comment that this must align with #317's semantics when the ignored set gets wired.
- **The reconnect ladder floors its delay but doesn't record into the registry** — a 429 during background reconnection is invisible to selection until a subscribe-path failure records it. Conservative asymmetry, fine as-is, worth a sentence in the code.
- **Recognition boundary**: only typed handshake 429s are held off; a post-handshake JSON-RPC "rate limit" node error on subscribe still scores. Consistent with the stack's typed-only philosophy — just noting the boundary is intentional.
- **The merge-conflict situation is #317's rerun**: merging the branch as-is conflicts (add/add on `holdoff_test.go`) because it still carries pre-rebase copies of the merged stack. The top commit alone replays cleanly onto main — same final-rebase procedure as #317.

The verification commands in the PR description pass as written, on both the branch base and the rebased-onto-main combination.

avitenzer (approved):

PR #321, opened 2026-08-20
feat(smart-router): expose rate-limit hold-off events as metrics

nimrod-teich:
#Closes MAG-2978

Jira ticket: MAG-2978

Follow-up to the 429/Retry-After stack: the hold-off registry decides when the router stops asking a rate-limited upstream, and until now that was visible only in debug logs — a vendor cap tripping in production had no dashboard signal.

## What changed

The registry emits its own events ([protocol/holdoff/metrics.go](protocol/holdoff/metrics.go)), so every consumer path — re-verify, hot path, recovery probe, ws subscriptions — is covered uniformly and a new consumer cannot forget to wire it:

- `smartrouter_rate_limit_holdoffs_total{provider, event}` — `recorded` (a 429 held an endpoint off), `escalated` (counted on the transition to a provider-wide hold-off, not per refresh), `cleared` (only when an answer actually dropped a standing penalty).
- `smartrouter_rate_limit_holdoff_seconds{provider}` — histogram of applied hold-offs, showing upstream Retry-After magnitudes against the exponential default.

Registration follows the repo's best-effort idiom (`error_metrics.go`): reuse on `AlreadyRegisteredError`, never panic.

**Docs:** [docs/METRICS.md](docs/METRICS.md) gains a *Rate-limit hold-off* section (closed event set, seconds-not-milliseconds buckets, label sanitization, starter PromQL), and its intro no longer claims every metric lives under `protocol/metrics/`. The public docs site gets the matching update in a separate PR in docs.

**Label safety:** the ws path keys the registry by node URL, and node URLs can embed API keys in path or query. URL-shaped keys are reduced to `scheme://host` before they become a label, so a credential never reaches a Prometheus series — pinned by test.

## How to verify

```
go build ./...
go test ./protocol/holdoff/ -count=1 -v
```

`TestMetrics_RecordEscalateClear` pins the counting semantics (escalate-on-transition, clear-only-on-real-penalty); `TestMetrics_ProviderLabelSanitizesURLs` pins the credential guard.

### DIFF
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -1,7 +1,9 @@
 # Smart Router — Metrics Reference
 
 Every metric the Smart Router exposes over Prometheus, with its type, labels, and
-meaning. All metrics are defined under [`protocol/metrics/`](../protocol/metrics/).
+meaning. Metrics are defined under [`protocol/metrics/`](../protocol/metrics/), except the
+[rate-limit hold-off](#rate-limit-hold-off) pair, which the registry that owns the events
+emits itself from [`protocol/holdoff/metrics.go`](../protocol/holdoff/metrics.go).
 
 ## Exposition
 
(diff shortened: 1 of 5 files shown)


avitenzer:
## Review — "expose rate-limit hold-off events as metrics"

**Verdict: approve — same precondition as its siblings: the final rebase before merge.** The reviewable change is the top commit (`84f54d9`); the branch still carries the stale pre-rebase #314 commits. Built and tested it on its own base, then replayed the top commit onto current main — conflict-free, including the doc's new Metrics section — and verified the result against the holdoff bounds fix (`ddd0245`): `./protocol/holdoff/` under `-race`, plus `./protocol/rpcsmartrouter/` and `./protocol/lavasession/` all pass.

### What I verified

- **Instrumenting at the registry is the right layer.** All four consumer paths get coverage for free, and a future consumer can't forget to wire it — the same argument the registry itself was built on.
- **The counting semantics are correct and pinned.** `escalated` fires only on the inactive→active transition (`!ps.escalatedUntil.After(now)` guard), so refreshes don't inflate it, while a re-escalation after natural expiry correctly counts as a new episode. `cleared` fires only when the answer dropped real state (`hadEntry || escalation active`), and the test pins both the transition behavior and the no-op clear.
- **It composes correctly with `ddd0245`, which it predates.** This was the main risk — the bounds fix reworked `RecordRateLimit` (post-jitter re-cap) and touched the same test file whose helpers (`newTestRegistry`, `t0`) the new test uses. After the replay, `metricRecorded(provider, d)` sits after the final capped-and-jittered `d`, so the histogram observes the actually-applied hold-off; the helpers exist on both sides; everything passes.
- **The credential guard works and covers more than it claims**: `scheme://host` reduction strips path and query API keys (pinned by test), and because `url.URL.Host` excludes userinfo, a `wss://APIKEY@host/ws` credential is stripped too — that case just isn't in the test.
- **Cardinality is bounded** (provider names from config; URLs collapse to scheme://host), and the best-effort registration matches the `error_metrics.go` idiom — reuse on `AlreadyRegisteredError`, never panic.

- **Add the userinfo case to the sanitizer test** (`wss://KEY@host/ws` → `wss://host`) — it works today, but it's exactly the kind of guarantee that deserves a pin next to the path/query ones.
- **`cleared` also counts answers to already-expired entries**: `hadEntry` is true while an expired entry's strike memory is still retained, so an answer then increments `cleared` even though no active hold-off was dropped. Defensible — strike memory *is* a standing penalty (the next 429 would double) — but it's a shade looser than the help text's "an answer dropped the penalty," and unpinned. One sentence in the comment or one test case settles it.
- **Metric calls run under the registry mutex.** Prometheus counter/histogram ops are atomic and never call back, so no deadlock and negligible cost at these call rates (events fire on 429s, not per relay). Fine as-is; hoisting them outside the lock would be trivial if the registry ever gets hotter.
- **Same stale-stack state as #317/#320**: merging as-is would conflict; the top commit alone replays cleanly onto main. Needs the final rebase.

The verification commands in the PR description pass as written, on both the branch base and the rebased-onto-main combination.

avitenzer (approved):

PR #325, opened 2026-08-23
fix(smart-router): retry stateful/batch relays on a rate limit, 503 when fully capped

nimrod-teich:
#Closes MAG-2987

Jira ticket: MAG-2987

Follow-up to the 429/Retry-After stack (umbrella MAG-2910): the three places a rate limit was still handled as a generic failure.

## Why

A 429 means the upstream refused *before executing anything*. That property was not being used:

1. **Stateful relays and batches failed outright.** `relaypolicy` stops retrying them because a possibly-executed write must not run twice — the one concern a 429 does not raise. A tx broadcast or a JSON-RPC batch landing on a capped vendor failed to the customer, even though a second endpoint would have served it and selection already knew to avoid the first.
2. **A fully capped chain surfaced as 500.** `buildFailureResult` returned the best error's own result, rendered as 500 — "the router broke" — when the truth is "temporarily unservable".
3. **gRPC streaming subscriptions ignored the hold-off.** Selection ran over every URL; the ws manager got the filter in PR #320 (ws subscriptions hold off a rate-limited endpoint).

## What changed

- `ResultsSummary.OnlyRateLimited` — true when at least one attempt failed and every failure was a typed 429 protocol error or a node error the classifier flagged `IsRateLimited`. Computed by `onlyRateLimited`, a pure predicate over the error slices the caller already holds: it takes no lock of its own, and both callers (`GetResultsSummary` and `buildFailureResult`) pass results snapshotted under the results manager's `rm.lock` — a different mutex from `rp.lock`.
- `relaypolicy.Decide`: stateful and batch relays **retry** when `OnlyRateLimited` — but only on the `gotResults` path. `OnlyRateLimited` describes attempts that have *completed*; a ticker hedge fires on a timer with a relay still in flight, one that may already have broadcast, so `rateLimitRetrySafe = OnlyRateLimited && !IsTickerHedge` and a hedged stateful/batch relay stops exactly as before. Any other failure stops as before, and the limit checks (`MaxRetries`, `RelayRetryLimit`) still bound the retry.
- `buildFailureResult`: all failures rate-limited → **503**, stamped on a *copy* of the result (the local aliases the stored `RelayResult`, and this runs under a read lock that holds nothing of `rm.lock`). No Retry-After is surfaced (unchanged decision — the registry owns it internally). Any other failure mix keeps its shape.
- `DirectGRPCSubscriptionManager.selectFromTier` reuses `notHeldOff`: skips held-off endpoints while something ready remains; the full tier stays when nothing is.
- `docs/RATE-LIMIT-HOLDOFF.md`: two new semantics bullets + the gRPC-subscriptions consumer row.

## How to verify

```
go build ./...
go test ./protocol/relaypolicy/ -run TestDecide_ -v
go test ./protocol/relaycore/ -run 'TestProcessingResult_' -v
go test ./protocol/rpcsmartrouter/ -run TestGRPCSelectFromTier -v
go test ./protocol/relaycore/ ./protocol/relaypolicy/ ./protocol/rpcsmartrouter/
```

The relaycore tests drive a real `RelayProcessor` (LAVA REST mocks). Two providers both answering typed 429s → summary `OnlyRateLimited`, `HasPermanentProtocolError` false (a typed 429 stays retryable), `ProcessingResult` 503; a 429 + a connection-refused → not only-rate-limited, status untouched; and a two-round case in the `sendRelayWithRetries` shape — all-429 round asserts 503, then a connection-refused arrives in a second batch and `ProcessingResult` must report `0`, not the earlier 503.

The relaypolicy tests cover both directions of the carve-out: stateful/batch retry when every failure was a rate limit, and stateful/batch stop on a ticker hedge under the same summary (with a stateless-non-batch hedge as the control).

---

*Description refreshed after the review round (2026-08-24): the `OnlyRateLimited`, `Decide` and `buildFailureResult` bullets described the pre-review implementation. Known follow-up, deliberately not in this PR: **MAG-3066** — the hold-off narrowing in `selectFromTier` can strand a tier whose ready endpoints are all ignored, in the WS manager as well as this gRPC copy, so it wants fixing in both at once.*

### DIFF
--- a/protocol/relaycore/rate_limit_failure_test.go
+++ b/protocol/relaycore/rate_limit_failure_test.go
@@ -1,0 +1,80 @@
+package relaycore
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/magma-Devs/smart-router/protocol/chainlib"
+	"github.com/magma-Devs/smart-router/protocol/chainlib/extensionslib"
+	"github.com/magma-Devs/smart-router/protocol/common"
+	"github.com/magma-Devs/smart-router/protocol/lavasession"
+	spectypes "github.com/magma-Devs/smart-router/types/spec"
+	"github.com/stretchr/testify/require"
+)
+
+func newTwoProviderProcessor(t *testing.T) (*RelayProcessor, *lavasession.UsedProviders, func()) {
+	t.Helper()
+	ctx := context.Background()
+	serverHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(http.StatusOK) })
+	chainParser, _, _, closeServer, _, err := chainlib.CreateChainLibMocks(ctx, "LAVA", spectypes.APIInterfaceRest, serverHandler, nil, "../../", nil)
+	require.NoError(t, err)
+	chainMsg, err := chainParser.ParseMsg("/cosmos/base/tendermint/v1beta1/blocks/17", nil, http.MethodGet, nil, extensionslib.ExtensionInfo{LatestBlock: 0})
+	require.NoError(t, err)
+	protocolMessage := chainlib.NewProtocolMessage(chainMsg, nil, nil, "dapp", "123.11")
+	usedProviders := lavasession.NewUsedProviders(nil)
+	rp := NewRelayProcessor(ctx, nil, RelayProcessorMetrics, RelayProcessorMetrics, RelayRetriesManagerInstance, newMockRelayStateMachine(protocolMessage, usedProviders))
+
+	lockCtx, cancel := context.WithTimeout(ctx, 10*time.Millisecond)
+	defer cancel()
+	require.Nil(t, usedProviders.TryLockSelection(lockCtx))
+	usedProviders.AddUsed(lavasession.ConsumerSessionsMap{"lava@a": &lavasession.SessionInfo{}, "lava@b": &lavasession.SessionInfo{}}, nil)
+	closer := func() {
+		if closeServer != nil {
+			closeServer()
+		}
+	}
+	return rp, usedProviders, closer
+}
+
+// Every attempt refused for rate: the chain is temporarily unservable, not broken — the
+// client gets 503, and the policy summary says the failures were only rate limits.
+func TestProcessingResult_AllRateLimitedIs503(t *testing.T) {
+	rp, _, closer := newTwoProviderProcessor(t)
+	defer closer()
+
+	go SendProtocolError(rp, "lava@a", time.Millisecond, common.RateLimited(errors.New("HTTP 429"), 30*time.Second))
+	go SendProtocolError(rp, "lava@b", 2*time.Millisecond, common.RateLimited(errors.New("HTTP 429"), 0))
+	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
+	defer cancel()
+	_ = rp.WaitForResults(ctx)
+
+	summary := rp.GetResultsSummary()
+	require.True(t, summary.OnlyRateLimited)
+	require.False(t, summary.HasPermanentProtocolError, "a typed 429 must stay retryable")
+
+	result, err := rp.ProcessingResult()
+	require.Error(t, err)
+	require.Equal(t, http.StatusServiceUnavailable, result.StatusCode)
+}
+
+// Mixed failures keep today's shape — the best error's own result, status untouched (the
+// listener applies its default) — and the summary does not claim only-rate-limited.
+func TestProcessingResult_MixedFailuresUnchanged(t *testing.T) {
+	rp, _, closer := newTwoProviderProcessor(t)
+	defer closer()
+
+	go SendProtocolError(rp, "lava@a", time.Millisecond, common.RateLimited(errors.New("HTTP 429"), 0))
+	go SendProtocolError(rp, "lava@b", 2*time.Millisecond, errors.New("connection refused"))
+	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
+	defer cancel()
+	_ = rp.WaitForResults(ctx)
+
+	require.False(t, rp.GetResultsSummary().OnlyRateLimited)
+	result, err := rp.ProcessingResult()
+	require.Error(t, err)
+	require.NotEqual(t, http.StatusServiceUnavailable, result.StatusCode, "only an all-rate-limited failure is 503")
+	require.Equal(t, 0, result.StatusCode, "the protocol error's own result is returned as before")
+}
--- a/protocol/relaycore/rate_limit_failure_test.go
+++ b/protocol/relaycore/rate_limit_failure_test.go
@@ -39,6 +39,20 @@ func newTwoProviderProcessor(t *testing.T) (*RelayProcessor, *lavasession.UsedPr
 	return rp, usedProviders, closer
 }
 
+// startBatch opens a new selection batch of the given providers, so a follow-up
+// WaitForResults expects exactly their responses — the shape of a retry round.
+func startBatch(t *testing.T, usedProviders *lavasession.UsedProviders, providers ...string) {
+	t.Helper()
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
+	defer cancel()
+	require.Nil(t, usedProviders.TryLockSelection(ctx))
+	sessions := lavasession.ConsumerSessionsMap{}
+	for _, provider := range providers {
+		sessions[provider] = &lavasession.SessionInfo{}
+	}
+	usedProviders.AddUsed(sessions, nil)
+}
+
 // Every attempt refused for rate: the chain is temporarily unservable, not broken — the
 // client gets 503, and the policy summary says the failures were only rate limits.
 func TestProcessingResult_AllRateLimitedIs503(t *testing.T) {
@@ -77,4 +91,36 @@ func TestProcessingResult_MixedFailuresUnchanged(t *testing.T) {
 	require.Error(t, err)
 	require.NotEqual(t, http.StatusServiceUnavailable, result.StatusCode, "only an all-rate-limited failure is 503")
 	require.Equal(t, 0, result.StatusCode, "the protocol error's own result is returned as before")
+}
+
+// The 503 is stamped on a copy, never on the stored result. sendRelayWithRetries reuses one
+// processor and calls ProcessingResult() per iteration, so an all-429 iteration must not
+// leave a 503 behind for a later, mixed one to inherit.
+func TestProcessingResult_RateLimited503DoesNotStick(t *testing.T) {
+	rp, usedProviders, closer := newTwoProviderProcessor(t)
+	defer closer()
+
+	// Iteration 1: every failure is a rate limit.
+	go SendProtocolError(rp, "lava@a", time.Millisecond, common.RateLimited(errors.New("HTTP 429"), 0))
+	go SendProtocolError(rp, "lava@b", 2*time.Millisecond, common.RateLimited(errors.New("HTTP 429"), 0))
+	firstCtx, cancelFirst := context.WithTimeout(context.Background(), 200*time.Millisecond)
+	defer cancelFirst()
+	_ = rp.WaitForResults(firstCtx)
+
+	result, err := rp.ProcessingResult()
+	require.Error(t, err)
+	require.Equal(t, http.StatusServiceUnavailable, result.StatusCode)
+
+	// Iteration 2: the retry lands on an endpoint that is down rather than capped, so the
+	// chain is no longer merely unservable and the 429 result must report its own status.
+	startBatch(t, usedProviders, "lava@c")
+	go SendProtocolError(rp, "lava@c", time.Millisecond, errors.New("connection refused"))
+	secondCtx, cancelSecond := context.WithTimeout(context.Background(), 200*time.Millisecond)
+	defer cancelSecond()
+	_ = rp.WaitForResults(secondCtx)
+
+	require.False(t, rp.GetResultsSummary().OnlyRateLimited)
+	result, err = rp.ProcessingResult()
+	require.Error(t, err)
+	require.Equal(t, 0, result.StatusCode, "the earlier 503 must not have been stamped onto the stored result")
 }
--- a/protocol/relaycore/relay_processor.go
+++ b/protocol/relaycore/relay_processor.go
@@ -455,6 +455,7 @@ func (rp *RelayProcessor) GetResultsSummary() ResultsSummary {
 		HasUnsupportedMethod:      hasUnsupportedMethod,
 		HasPermanentProtocolError: hasPermanentProtocolError,
 		HasEpochMismatch:          hasEpochMismatch,
+		OnlyRateLimited:           rp.onlyRateLimitedLocked(),
 		HashErr:                   hashErr,
 	}
 }
@@ -1229,6 +1230,27 @@ func (rp *RelayProcessor) processNonCrossValidationResult(
 	return rp.buildFailureResult(nodeErrorCount, protocolErrorCount)
 }
 
+// onlyRateLimitedLocked reports whether at least one attempt failed and every failure was an
+// upstream rate limit — a typed 429 protocol error, or a node error the classifier flagged
+// IsRateLimited. Caller must hold rp.lock.
+func (rp *RelayProcessor) onlyRateLimitedLocked() bool {
+	_, nodeErrorResults, protocolErrorResults := rp.GetResultsData()
+	failures := 0
+	for _, result := range nodeErrorResults {
+		failures++
+		if !result.IsRateLimited {
+			return false
+		}
+	}
+	for _, protocolError := range protocolErrorResults {
+		failures++
+		if !errors.Is(protocolError.GetError(), common.StatusCodeError429) {
+			return false
+		}
+	}
+	return failures > 0
+}
+
 // buildFailureResult constructs an error result when no consensus can be reached. It returns the best
 // node/protocol error's own RelayResult, whose ProviderInfo.ProviderAddress is a SINGLE provider — the
 // source of the error body being returned. It must never be a comma-joined list: packing the whole
@@ -1257,6 +1279,13 @@ func (rp *RelayProcessor) buildFailureResult(
 		if protocolErr.Response != nil {
 			returnedResult = &protocolErr.Response.RelayResult
 		}
+	}
+
+	// Every attempt was refused for rate: the chain is temporarily unservable, not broken.
+	// 503 is the honest status and what well-behaved clients back off on; no Retry-After
+	// is surfaced — the hold-off registry owns that, internally.
+	if rp.onlyRateLimitedLocked() {
+		returnedResult.StatusCode = http.StatusServiceUnavailable
 	}
 
 	// Log with classified error code for metrics/observability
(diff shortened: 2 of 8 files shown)


protocol/relaypolicy/policy.go
```diff
@@ -23,7 +23,11 @@ func (p *Policy) Decide(input DecisionInput) DecisionOutput {
 	if input.Selection == relaycore.CrossValidation {
 		return DecisionOutput{Action: Stop, Reason: "CrossValidation"}
 	}
-	if input.Selection == relaycore.Stateful {
+	// A stateful relay is not retried because a possibly-executed write must not run
+	// twice. A rate limit is the one failure that concern does not cover: the upstream
+	// refused before executing anything, so the retry lands on a different endpoint
+	// with nothing at risk. The limit checks below still bound it.
+	if input.Selection == relaycore.Stateful && !input.Summary.OnlyRateLimited {
```
avitenzer: **Finding 1** — this carve-out is consulted by the ticker-hedge path too (`unified_relay_state_machine.go:413`), where a relay is deliberately still in flight. `OnlyRateLimited` is a predicate over *completed* failures, so there it can be true while the attempt that may already have broadcast the tx is still running.

Gating on `!IsTickerHedge` is self-reinforcing rather than merely defensive: the ticker is the only thing that starts a batch without waiting for the previous one to drain (the `batchUpdate` → `SendRetry` path at `:361` only fires when *dispatch* failed, and stateful dispatches `NumOfProviders = 1`, all-or-nothing). Close it, and stateful batches never overlap — which is exactly the precondition the carve-out needs. It also closes a second latent hazard: `WaitForResults` counts `responsesCount` per call but compares it against `SessionsLatestBatch()`, so an earlier batch's straggler would count toward the current batch's drain.

Naming the predicate once also makes it obvious that this carve-out and the batch one below are the same rule:

```suggestion
	// A stateful relay is not retried because a possibly-executed write must not run
	// twice. A rate limit is the one failure that concern does not cover: the upstream
	// refused before executing anything, so the retry lands on a different endpoint
	// with nothing at risk. The limit checks below still bound it.
	//
	// That is a claim about attempts that have COMPLETED, and only the gotResults path
	// guarantees it — the ticker fires on a timer with a relay still in flight, which may
	// already have broadcast. So the carve-out does not extend to a hedge.
	rateLimitRetrySafe := input.Summary.OnlyRateLimited && !input.IsTickerHedge

	if input.Selection == relaycore.Stateful && !rateLimitRetrySafe {
```

Worth a test case pinning `Stateful + IsTickerHedge: true + OnlyRateLimited: true → Stop`.

protocol/relaypolicy/policy.go
```diff
@@ -41,7 +45,8 @@ func (p *Policy) Decide(input DecisionInput) DecisionOutput {
 	if input.AttemptNumber >= p.config.MaxRetries {
 		return DecisionOutput{Action: Stop, Reason: "MaxRetriesReached"}
 	}
-	if input.IsBatch && p.config.DisableBatchRetry {
+	// Same reasoning for batches: a rate-limited batch executed nothing.
+	if input.IsBatch && p.config.DisableBatchRetry && !input.Summary.OnlyRateLimited {
```
avitenzer: Same exposure here, and it needs the same guard. This check sits in step 3, which runs *before* the `if !input.IsTickerHedge` block at line 61 — so a batch is hedgeable on the ticker for exactly the same reason a stateful relay is. And `DisableBatchRetry` exists for exactly the same concern: a JSON-RPC batch can carry an `eth_sendRawTransaction`.

```suggestion
	// Same reasoning for batches: a rate-limited batch executed nothing — and the same
	// in-flight caveat, since a batch can carry an eth_sendRawTransaction.
	if input.IsBatch && p.config.DisableBatchRetry && !rateLimitRetrySafe {
```

protocol/relaycore/relay_processor.go
```diff
@@ -1259,6 +1281,13 @@ func (rp *RelayProcessor) buildFailureResult(
 		}
 	}
 
+	// Every attempt was refused for rate: the chain is temporarily unservable, not broken.
+	// 503 is the honest status and what well-behaved clients back off on; no Retry-After
+	// is surfaced — the hold-off registry owns that, internally.
+	if rp.onlyRateLimitedLocked() {
+		returnedResult.StatusCode = http.StatusServiceUnavailable
+	}
```
avitenzer: **Finding 2** — `returnedResult` is not a local here. Both assignments above take the address of the *stored* response (`&nodeErr.Response.RelayResult` at :1273, `&protocolErr.Response.RelayResult` at :1280), so this writes into results-manager memory. Two consequences:

1. **The 503 is permanent.** A later `ProcessingResult()` on the same processor returns 503 even once the failure set is mixed and `onlyRateLimitedLocked()` is false — nothing resets it. `sendRelayWithRetries` (`rpcsmartrouter_server.go:791-842`) is exactly that shape, and the probe in the review body reproduces it.
2. **It's a write under a read lock.** `ProcessingResult` holds `rp.lock.RLock()` (:1122), which permits concurrent readers, and holds nothing of `rm.lock` — the separate mutex in `results_manager.go:285-301` that actually guards this memory.

Copying matches what the sibling branches in this same file already do (`processNonCrossValidationResult` returns `result := nodeErrors[0]; return &result`):

```suggestion
	if rp.onlyRateLimitedLocked() {
		// Copy before stamping: returnedResult aliases the STORED RelayResult (the
		// assignments above take its address), and this runs under a read lock. Mutating
		// in place writes shared state without the write lock and makes the 503 permanent
		// for every later ProcessingResult call.
		stamped := *returnedResult
		stamped.StatusCode = http.StatusServiceUnavailable
		returnedResult = &stamped
	}
```

I applied this locally: the probe passes and both new relaycore tests stay green.

FWIW `relay_errors.go:101` already does the same thing (`Response.RelayResult.CrossValidation = errorCount` under an RLock), so this is a pre-existing pattern — just not one worth extending.

protocol/relaycore/relay_processor.go
```diff
@@ -1229,6 +1230,27 @@ func (rp *RelayProcessor) processNonCrossValidationResult(
 	return rp.buildFailureResult(nodeErrorCount, protocolErrorCount)
 }
 
+// onlyRateLimitedLocked reports whether at least one attempt failed and every failure was an
+// upstream rate limit — a typed 429 protocol error, or a node error the classifier flagged
+// IsRateLimited. Caller must hold rp.lock.
```
avitenzer: Two small things on this helper:

**Naming.** The `Locked` suffix and "Caller must hold `rp.lock`" point at a lock this function doesn't actually depend on. It calls `GetResultsData()`, which takes `rm.lock.RLock()` internally — a different mutex. So it is neither lock-free (as the PR description says) nor reliant on `rp.lock`. `onlyRateLimited`, with a note that it reads through the results manager's own lock, would describe it accurately — and matters here because Finding 2 above is precisely a case of the two locks being confused.

**Redundant traversal.** `GetResultsSummary` already loops both error slices; this walks them again, and `NodeErrors()` allocates a fresh slice of `RelayResult` copies on every call. That's once per `Decide`, i.e. per retry. Folding the two checks into the existing loops in `GetResultsSummary` would drop both the second pass and the allocation.

protocol/rpcsmartrouter/direct_grpc_subscription_manager.go
```diff
@@ -1102,6 +1102,13 @@ func (dgm *DirectGRPCSubscriptionManager) selectFromTier(ctx context.Context, ti
 		return nil, fmt.Errorf("tier is empty")
 	}
 
+	// Rate-limit hold-off: prefer endpoints that are not currently held off after a 429.
+	// Only narrows the tier when something ready remains — a subscription must still be
+	// served when every endpoint is held off, so the full tier stays in that case.
+	if ready := notHeldOff(tier); len(ready) > 0 && len(ready) < len(tier) {
+		tier = ready
+	}
```
avitenzer, line 1110: Not a blocker — this is a faithful, byte-identical mirror of the shipped WS code (`direct_ws_subscription_manager.go:485`), and parity was the stated goal. Flagging it because the copy propagates a gap worth a follow-up on both.

The narrowing happens before `ignoredEndpoints` is consulted, so when the ready set and the ignored set overlap it can leave nothing selectable:

```
tier = [a, b];  a ignored;  b held off
→ ready = [a]  →  tier = [a]  →  "all endpoints in tier are ignored/unavailable"
without the filter: b is returned
```

Confirmed by probe. The comment says "the full tier stays when nothing is [ready]", but the guard only tests hold-off, not "nothing *usable* remains". Impact is softened by the cascade falling through to backup (`selectEndpoint:1080-1090`), so a follow-up fixing both managers together seems right:

```go
if ready := notHeldOff(tier); len(ready) > 0 && hasSelectable(ready, ignoredEndpoints) {
    tier = ready
}
```

(Minor: the `len(ready) < len(tier)` half of the condition is a no-op — assigning an identical slice changes nothing.)

protocol/rpcsmartrouter/grpc_rate_limit_select_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+	reg.RecordRateLimit("grpc-a.example:443", "grpc-a.example:443", time.Minute)
+	ep, err := dgm.selectFromTier(context.Background(), tier, nil, nil)
+	require.NoError(t, err)
+	require.Equal(t, "grpc-b.example:443", ep.Url)
+
+	reg.RecordRateLimit("grpc-b.example:443", "grpc-b.example:443", time.Minute)
+	ep, err = dgm.selectFromTier(context.Background(), tier, nil, nil)
+	require.NoError(t, err)
+	require.NotNil(t, ep, "with everything held off the tier still serves")
```
avitenzer: This assertion can't fail: `selectFromTier` never returns `(nil, nil)`, so after `require.NoError` on the line above, `ep` is necessarily non-nil. The "full tier stays" branch — the interesting half of the change — isn't actually covered.

```suggestion
	require.Equal(t, "grpc-a.example:443", ep.Url, "with everything held off the full tier stays, so the first endpoint serves")
```

avitenzer (commented):
Reviewed at `df34d19` in a clean worktree (note: the branch was rebased, so a local `git diff` shows a different tree). `go build ./...` passes, `./protocol/relaycore ./protocol/relaypolicy ./protocol/rpcsmartrouter` pass, `gofmt` clean on every file touched. Each finding below was confirmed with a throwaway probe test rather than by reading alone.

The core design is right, and unusually well argued — "the upstream refused before executing anything" is a real property, using it to unlock a retry follows, and 503 over 500 for a fully capped chain is the honest status. Two things I'd want fixed before merge.

### 1. The stateful carve-out also opens the ticker hedge, which the safety argument doesn't cover

`Decide` is called from two places with different in-flight invariants:

| Call site | Is a relay in flight? |
|---|---|
| `unified_relay_state_machine.go:386` (`gotResults`) | No — `WaitForResults` returned, the latest batch drained |
| `unified_relay_state_machine.go:413` (`startNewBatchTicker.C`) | **Yes, by design** — that is what a hedge is |

`OnlyRateLimited` describes failures already *recorded*. On the hedge path it says nothing about the attempt currently executing:

1. Attempt 1 → 429, recorded.
2. `gotResults` → retry → attempt 2 dispatched to endpoint B.
3. B accepts the tx but is slow. `relayTimeout` elapses.
4. Ticker fires. Summary is still "1 protocol error, all 429" → `OnlyRateLimited == true` → **Retry**.
5. The same signed tx goes to endpoint C while attempt 2 is live.

Probe: `{Selection: Stateful, IsTickerHedge: true, Summary: {ProtocolErrors: 1, OnlyRateLimited: true}}` returns `action=Retry, reason="Default"`.

The bound is looser than the description implies, too. The hedge path skips the error-tolerance check (`if !input.IsTickerHedge`), so `RelayRetryLimit = 2` never applies — only `MaxRetries = MaximumNumberOfTickerRelayRetries = 10`. So up to 10 hedged sends, not the ~3 the `gotResults` path allows.

Most chains dedupe a re-broadcast by tx hash, so this is unlikely to be catastrophic — but "never retry a stateful relay" was a deliberate invariant, and the stated justification is simply not true of an in-flight attempt. Suggested fix inline.

### 2. The 503 is stamped onto shared stored state and never cleared

`returnedResult` aliases the stored response (`&nodeErr.Response.RelayResult` / `&protocolErr.Response.RelayResult`), so the stamp writes into results-manager memory — under `rp.lock.RLock()`, and holding nothing of `rm.lock`, which is the mutex that actually guards it.

Probe against the `sendRelayWithRetries` shape (`rpcsmartrouter_server.go:791-842` reuses one processor and calls `ProcessingResult()` per iteration):

```
iteration 1: two identical 429s        → OnlyRateLimited=true,  StatusCode=503  ✓
iteration 2: + one connection-refused  → OnlyRateLimited=false, StatusCode=503  ✗ (expected 0)
```

The 429 stays the best error and now carries 503 permanently. That contradicts the contract `TestProcessingResult_MixedFailuresUnchanged` asserts — it passes only because it calls `ProcessingResult()` once. Today's blast radius is small (init relays only log), but it's a live trap for whoever adds a second call. One-line fix inline; I verified it makes the probe pass with both new relaycore tests still green.

---

Everything else below is follow-up material or nits — the docs row is correct and the tests are meaningful rather than tautological.

protocol/relaycore/rate_limit_failure_test.go
```diff
@@ -0,0 +1,80 @@
...
+	go SendProtocolError(rp, "lava@b", 2*time.Millisecond, errors.New("connection refused"))
+	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
+	defer cancel()
+	_ = rp.WaitForResults(ctx)
+
+	require.False(t, rp.GetResultsSummary().OnlyRateLimited)
+	result, err := rp.ProcessingResult()
+	require.Error(t, err)
+	require.NotEqual(t, http.StatusServiceUnavailable, result.StatusCode, "only an all-rate-limited failure is 503")
+	require.Equal(t, 0, result.StatusCode, "the protocol error's own result is returned as before")
```
avitenzer, line 93: This is the contract Finding 2 breaks — it holds only because the test calls `ProcessingResult()` exactly once. Call it a second time with a non-429 arriving in between (the `sendRelayWithRetries` shape) and this assertion fails: the stored result still carries the 503 stamped by the first call.

Worth adding that case here once the copy fix lands, since it's the regression test for it.

protocol/relaypolicy/policy.go
avitenzer: Applied verbatim, including the `rateLimitRetrySafe` naming — you're right that this and the batch check below are one rule and should read as one.

The self-reinforcing part is the argument that convinced me: the carve-out needs "no stateful attempt overlaps another", the ticker is the only thing that can break that, so gating on `!IsTickerHedge` makes the precondition hold rather than just guarding against its absence. The `SessionsLatestBatch()` straggler hazard you note is a real second reason to want the ticker out of this path.

Test added: `TestDecide_TickerHedgeKeepsTheStatefulAndBatchStops` pins `Stateful + IsTickerHedge + OnlyRateLimited → Stop/"Stateful"`, the batch equivalent → `Stop/"BatchDisabled"`, plus a stateless-non-batch hedge control that still retries. It fails on the pre-fix code.

Docs bullet updated too — it said "safe to retry, always", which is now the in-flight caveat spelled out.

protocol/relaypolicy/policy.go
avitenzer: Applied — same guard, same comment framing. The `eth_sendRawTransaction`-in-a-batch point is exactly why `DisableBatchRetry` exists, so it would have been incoherent to gate one and not the other.

protocol/relaycore/relay_processor.go
avitenzer: Fixed as suggested. The aliasing was the part I had wrong — I read `returnedResult` as the local it starts out as and missed that both branches reassign it to the address of the stored result.

Both consequences confirmed. The stamped copy now carries the 503 and the stored 429 result keeps its own status.

On `relay_errors.go:101` — agreed, noted and not extended.

protocol/relaycore/relay_processor.go
avitenzer: Both fixed, though the second one slightly differently than you proposed.

**Naming.** Renamed to `onlyRateLimited` and rewritten as a package-level pure function over the slices, with the doc saying what actually holds: it takes no lock of its own and does not depend on `rp.lock`; callers pass results already snapshotted under the results manager's `rm.lock`. Your point that this mattered is well taken — the `Locked` suffix is what made Finding 2 easy to write.

**Redundant traversal.** Dropped the re-fetch rather than inlining the checks into the two loops. `GetResultsSummary` now passes the `nodeErrorResults`/`protocolErrorResults` it already read, so the extra `GetResultsData()` call and the `NodeErrors()` allocation are gone from the per-`Decide` path — which was the actual cost. `buildFailureResult` gets the same slices threaded down from `ProcessingResult` (`processNonCrossValidationResult` now takes slices instead of counts), so its re-fetch is gone too.

I kept it as one predicate rather than folding into the loops so the rule has a single definition — `buildFailureResult` needs the same answer, and duplicating "429 protocol error OR IsRateLimited node error" in two places seemed the worse trade for one extra pass over a 1–3 element slice. Happy to fold if you disagree.

Side benefit: `buildFailureResult` now evaluates the predicate over the same snapshot the result was selected from, instead of re-reading results that can have moved under `rp.lock` in between.

protocol/rpcsmartrouter/grpc_rate_limit_select_test.go
avitenzer: Applied. You're right that it was vacuous — `selectFromTier` returns either an endpoint or an error, so `require.NoError` had already done the work. Now asserts `grpc-a` specifically, which is what "the full tier stays" actually means here, and it exercises the branch that was uncovered.

protocol/relaycore/rate_limit_failure_test.go
avitenzer: Added: `TestProcessingResult_RateLimited503DoesNotStick`. It drives two rounds against one processor the way `sendRelayWithRetries` does — round one is two 429s and asserts 503, round two opens a new selection batch, feeds a connection-refused, and asserts `OnlyRateLimited` is false and the result reports `0` rather than the earlier 503.

Worth noting it needs the real second round: pushing the third response without a new batch is silently dropped, since `WaitForResults` has already stopped draining `rp.responses`. Extracted a small `startBatch` helper for that.

Verified it fails on the pre-fix code (in-place stamp) and passes with the copy, and `-race` is clean.

protocol/rpcsmartrouter/direct_grpc_subscription_manager.go
avitenzer: Agreed on the gap and agreed on the disposition — leaving both managers alone here.

The reason: the fix has to land in `direct_ws_subscription_manager.go:485` too, and that is shipped code from #320 outside this PR's scope. Changing only the gRPC copy would trade a real gap for a worse one, two managers that look identical but no longer are. Your `hasSelectable` shape is the right fix and it should go in once, to both.

Same for the `len(ready) < len(tier)` no-op — it is genuinely dead, but removing it here would break the byte-identical parity that made this copy reviewable, so it belongs in the same follow-up.

Want me to file the ticket, or will you?

avitenzer: Filed as **MAG-3066** — https://magmadevs.atlassian.net/browse/MAG-3066

Carries your repro verbatim (the `[a ignored, b held off]` case), both file refs, the `hasSelectable` fix, the note that it has to land in the WS manager and this copy together, and the dead `len(ready) < len(tier)` half-condition as part of the same change. Low priority, 1 point, unassigned.

Leaving the code in this PR as the byte-identical mirror, as agreed.

avitenzer (approved):

PR #328, opened 2026-08-24
fix(lavasession): release the blocked provider list only after backup fails (MAG-2599)

Tomelia1999:
Jira ticket: MAG-2599

Implements [FAILOVER-TASKS](../blob/main/agent_docs/bug-reports/failover/FAILOVER-TASKS.md) **section 5**: the last-resort "release every blocked provider" must fire only when primary *and* backup are both exhausted.

> **Note for the reviewer:** commit `d8373be`'s message opens with "WORK IN PROGRESS / Do not merge as-is". **That warning is stale.** It described a defect that the following commit, `7d2911d`, fixes and pins with a test. The branch carries no known defect. I left the sequence intact rather than rewrite history — happy to squash the two if you'd prefer a single commit.

---

## The bug

The router already had the right last resort *and* the right fallback chain. They were wired in the wrong order.

`getSessionWithProviderOrError` walks primary → backup → blocked-provider list, which is what sections 4 and 5 describe. But `GetSessions` called `validatePairingListNotEmpty` at the **top of every request**, and that refilled `validAddresses` the moment the primary pool went empty. So:

- by the time the chain ran, the pool was never empty
- `PairingListEmptyError` never fired from a standing-empty pool
- **the backup tier was unreachable through that path**
- "every primary is blocked" silently meant "serve a primary we just blocked"

And because it ran per request rather than once, the block never survived a single relay — the next request released it again. That is what makes `bench-after` and `cooldown` (sections 2 and 3) impossible to build today: any block they set is wiped by the next relay. **This change is their prerequisite.**

`validatePairingListNotEmpty` has not been touched since `22cbf32`, the initial implementation.

## The change

`validatePairingListNotEmpty` did two unrelated jobs. Job 1 — warming the addon address cache — stays at the top of `GetSessions`. Job 2 — the release — moves into the chain, after backup:

```
1. a healthy primary
2. no healthy primary            -> serve backup, block still standing
3. nothing healthy anywhere      -> release the block and keep serving
```

Traffic returns to primary by itself once one recovers; there is no state to unwind. `numberOfResets` is now read *after* the chain, since a release inside it has to reach the blocklisted-session allowance that scales with it.

### Two guards, both load-bearing

`releaseBlockedProvidersIfPoolEmpty` releases only when:

1. **the pool is genuinely empty** — `PairingListEmptyError` also covers "every valid address is already ignored by this request", which is ordinary retry exhaustion. Releasing on that would let one retried relay wipe the block for every other relay in the process.
2. **the release could help the request making it** (`releaseCouldServeThisRequest`) — `GetSessions` runs this chain more than once per relay, so a request whose providers are blocked one-by-one as they fail arrives a second time having already tried them all. Releasing then cannot rescue it and only wipes the standing block, leaving the pool reading healthy while nothing in it can serve.

Guard 2 is the fix in `7d2911d`. Without it, against providers whose endpoints are down — the realistic outage — this branch ended each relay with `valid=0 blocked=2` turned into `valid=2 blocked=0`, and grew `numberOfResets` twice as fast as `main`.

### Pinned providers

`getValidProviderAddresses` resolves a `lava-select-provider` pin **before** the empty-pool check and fails with `SelectedProviderUnavailableError`, not `PairingListEmptyError`. An earlier version of this change caught only the latter, which broke pinning whenever every provider was blocked. The chain now catches both.

The pinned path releases but never falls through to backup or the blocked-provider walk — a pin exists so the request is never served by someone the caller did not ask for. `SelectedProviderAlreadyFailedError` (added by `2debe75` on main) correctly does **not** release: the request has already tried that provider.

## Verification

- `go build ./...`, `go vet ./...` clean; `gofmt` clean on touched files
- full `go test ./...` green, multiple consecutive runs
- **Differential matrices against `main`**, since almost nothing here should change behaviour:
  - **pin matrix** — 9 `lava-select-provider` scenarios (healthy pin, blocked pin with others healthy, blocked pin with empty pool, unknown name, case-folded name, already-failed-this-request, no pin) × with and without a backup tier
  - **release-persistence matrix** — endpoints up and down, three relays each
  - Every row is **identical to `main`** except the one intended row: all-primaries-blocked with a backup configured, which now leaves the block standing instead of releasing it.

Two of the new tests fail on the pre-fix code: `GetSessions` hands back a just-blocked primary instead of the backup. `TestProvidersBlockedMidRelay_DoNotReleaseTheBlockAgain` fails without guard 2.

**Integration-tested against a live router** — full write-up in [this comment](https://github.com/Magma-Devs/smart-router/pull/328#issuecomment-5397521529). Headline: with both primaries blocked and backup healthy, **20 steady-state relays produce 20 releases on `main` and 0 on this branch**. Seven scenarios checked (backup serves, block holds, recovery, last-resort ordering, retry exhaustion, guard 2, and the pin hole below), each against `main` and against this branch under an identical rig.

## ⚠️ Behaviour change reviewers should weigh

**Backup providers now receive traffic in a case where they previously never did.** For a customer whose backup is the metered or paid tier, that is a cost change, not just a routing change. It is exactly what section 4 intends — but it lands before `spill-on` exists, so there is no way yet for a customer to say which failures may cross.

Operationally: `NO VALID PROVIDERS - TRIGGERING RESET` becomes much rarer and now appears *after* `[BackupProviders] Static providers exhausted`.

**What this does NOT change, contrary to an earlier version of this description:** in a direct-RPC deployment `main` does not actually route customer traffic to a dead primary. `GetSessions` hands one back, but the relay to it fails immediately (its endpoints are disabled) and the request reaches backup through the per-request ignored list. Measured: **both builds serve `backup-node`**. The defect this fixes is the churn — a wasted attempt on a known-bad provider per relay, `numberOfResets` incrementing per relay (which inflates `maximumBlockedSessionsAllowed`, since that scales with `numberOfResets+1`), and a standing block that never survives long enough for §2 `bench-after` / §3 `cooldown` to be built on it.

I also expected the blocked-provider metric to start reflecting reality during an outage. It does not: `smartrouter_csm_blocked_providers` reads `0` while `/debug/provider-routing` correctly shows two blocked providers. That is a pre-existing metric bug, unrelated to this PR, and is being tracked separately.

## Known hole in guard 2 (pre-existing behaviour, reproduced)

`releaseCouldServeThisRequest` asks *"is there any provider this request has not tried?"*. When `lava-select-provider` is set, only that one provider can ever serve the request, so the guard answers a question that does not apply. A pin naming a provider that does not exist passes the guard, releases **every** blocked provider, and then fails anyway.

Reproduced on a live router — same router, nothing changed between the two halves:

```
23 steady-state relays   ->  backup-node x23,  releases delta = 0
                             blocked=[primary-a, primary-b]   valid=[]

one relay with:
  -H 'lava-select-provider: does-not-exist'
                         ->  HTTP 500,        releases delta = 1
                             blocked=[]       valid=[primary-a, primary-b]
```

**This behaves identically on `main`, so it is not a regression from this PR** — but the guard added here is meant to prevent exactly this class of useless release and does not catch it. Tracked separately; happy to fold the fix in here instead if a reviewer prefers.

## Not fixed here, deliberately

- **The release does not re-enable endpoints.** Providers return to `validAddresses` with endpoints still disabled, get picked, and are re-blocked. Pre-existing, documented in `HEALTH-STORE-PLAN.md`. This change makes the release rarer, so it bites less often.
- **Pin semantics are asymmetric.** A pinned-but-blocked provider resolves only when *every* provider is blocked, because that is the only case where the release happens to put it back. Identical on `main`.

## Commits

| | |
|---|---|
| `f30ab22` | pre-existing test flake fix — unrelated to the feature, split out so it can be taken separately |
| `d8373be` | the section 5 change + tests (stale WIP warning in its message, see the note at top) |
| `7d2911d` | guard 2 — the double-release fix |

`f30ab22` addresses a race that reproduces on `main` roughly one full `go test ./...` run in three: `TestCheckAndUnblock_BackupUnblockedWhenHealthy` asserted state that `UpdateAllProviders`' own background probe is free to change first.

### DIFF
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -911,17 +911,54 @@ func (csm *ConsumerSessionManager) cacheAddonAddresses(addon string, extensions 
 	return result
 }
 
+// releaseCouldServeThisRequest reports whether releasing the blocked list could hand THIS request a
+// provider it has not already tried. A release refills validAddresses from the whole pairing pool,
+// so the question is whether any of those providers is both absent from this request's ignored set
+// and able to serve the addon and extensions asked for.
+//
+// When the answer is no, releasing cannot rescue the request and only destroys standing state that
+// every other relay depends on. That happens whenever the pool empties *during* a request rather
+// than before it: each provider is blocked as it fails, and by the time the last one goes the
+// request has already tried them all.
+func (csm *ConsumerSessionManager) releaseCouldServeThisRequest(ignored map[string]struct{}, addon string, extensions []string, ctx context.Context) bool {
+	csm.lock.RLock()
+	defer csm.lock.RUnlock()
+
+	for _, address := range csm.pairingAddresses {
+		if _, alreadyTried := ignored[address]; alreadyTried {
+			continue
+		}
+		provider, ok := csm.pairing[address]
+		if !ok || provider == nil {
+			continue
+		}
+		if provider.IsSupportingAddon(addon) && provider.IsSupportingExtensions(extensions, ctx) {
+			return true
+		}
+	}
+	return false
+}
+
 // releaseBlockedProvidersIfPoolEmpty releases the standing blocked list and retries selection once,
 // as the last resort of the failover cascade.
 //
-// It is a no-op unless the valid pool is genuinely empty. That guard is the whole subtlety: the
-// errors that bring us here also cover "every valid address is already ignored by this request",
-// which is ordinary retry exhaustion, and releasing on that would let one retried relay wipe the
-// blocked list for every other relay in the process.
+// Two guards, and both matter. The pool must be genuinely empty: the errors that bring us here also
+// cover "every valid address is already ignored by this request", which is ordinary retry
+// exhaustion, and releasing on that would let one retried relay wipe the blocked list for every
+// other relay in the process. And the release must be able to help the request making it — see
+// releaseCouldServeThisRequest — because GetSessions runs this chain more than once per relay, so a
+// request whose providers are blocked one by one arrives here a second time with nothing left to
+// try. Releasing then leaves the pool looking healthy while nothing in it can serve.
 //
 // Returns ok=false when nothing was released, or when the retry after a release still found nothing.
 func (csm *ConsumerSessionManager) releaseBlockedProvidersIfPoolEmpty(ctx context.Context, wantedProviderNumber int, tempIgnoredProviders *ignoredProviders, cuNeededForSession uint64, requestedBlock int64, addon string, extensionNames []string, stateful uint32, virtualEpoch uint64, stickiness string, selectedProvider string, minGroups, perGroupTarget int) (SessionWithProviderMap, bool) {
 	if len(csm.cacheAddonAddresses(addon, extensionNames, ctx)) != 0 {
+		return nil, false
+	}
+
+	if !csm.releaseCouldServeThisRequest(tempIgnoredProviders.providers, addon, extensionNames, ctx) {
+		utils.LavaFormatDebug("every provider has already been tried by this request, leaving the blocked list standing",
+			utils.LogAttr("addon", addon), utils.LogAttr("extensions", extensionNames), utils.LogAttr("GUID", ctx))
 		return nil, false
 	}
 
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -911,27 +911,39 @@ func (csm *ConsumerSessionManager) cacheAddonAddresses(addon string, extensions 
 	return result
 }
 
-// validating we still have providers, otherwise reset valid addresses list
-// also caches validAddresses for an addon to save on compute
-func (csm *ConsumerSessionManager) validatePairingListNotEmpty(addon string, extensions []string, ctx context.Context) uint64 {
-	numberOfResets := csm.atomicReadNumberOfResets()
-	validAddresses := csm.cacheAddonAddresses(addon, extensions, ctx)
-	// currentlyBlockedProviderAddresses is intentionally not read here: this method holds no csm.lock,
-	// so reading the shared slice for a log attr races a concurrent writer (blockProvider / restore).
-	utils.LavaFormatInfo("VALIDATING PROVIDERS", utils.LogAttr("addon", addon), utils.LogAttr("extensions", extensions), utils.LogAttr("validAddressesCount", len(validAddresses)), utils.LogAttr("validAddresses", validAddresses), utils.LogAttr("GUID", ctx))
-	if len(validAddresses) == 0 {
-		utils.LavaFormatWarning("NO VALID PROVIDERS - TRIGGERING RESET", nil, utils.LogAttr("addon", addon), utils.LogAttr("extensions", extensions), utils.LogAttr("GUID", ctx))
-		numberOfResets = csm.resetValidAddresses(addon, extensions)
-		utils.LavaFormatInfo("RESET COMPLETED", utils.LogAttr("addon", addon), utils.LogAttr("extensions", extensions), utils.LogAttr("newValidAddressesCount", len(csm.cacheAddonAddresses(addon, extensions, ctx))), utils.LogAttr("GUID", ctx))
+// releaseBlockedProvidersIfPoolEmpty releases the standing blocked list and retries selection once,
+// as the last resort of the failover cascade.
+//
+// It is a no-op unless the valid pool is genuinely empty. That guard is the whole subtlety: the
+// errors that bring us here also cover "every valid address is already ignored by this request",
+// which is ordinary retry exhaustion, and releasing on that would let one retried relay wipe the
+// blocked list for every other relay in the process.
+//
+// Returns ok=false when nothing was released, or when the retry after a release still found nothing.
+func (csm *ConsumerSessionManager) releaseBlockedProvidersIfPoolEmpty(ctx context.Context, wantedProviderNumber int, tempIgnoredProviders *ignoredProviders, cuNeededForSession uint64, requestedBlock int64, addon string, extensionNames []string, stateful uint32, virtualEpoch uint64, stickiness string, selectedProvider string, minGroups, perGroupTarget int) (SessionWithProviderMap, bool) {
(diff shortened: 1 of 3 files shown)


Tomelia1999:
## Local integration testing — results, evidence, and three corrections

Ran this branch against a live router rather than relying on the unit tests. Posting in full because it produced findings that change what this PR claims, including two that are **not** section 5's fault but should not be lost.

### Rig

Router in Docker, three switchable counting upstreams on the host in front of a real local lava node. Two primaries (`primary-a`, `primary-b`) and one `backup-node`. Images built from git worktrees at this branch's head (`7d2911d`), at `origin/main` (`498c298`), and at this branch's first commit `d8373be` (before the guard-2 fix), so all three could be compared under an identical rig.

Instrument: the `Lava-Provider-Address` response header names the provider that served, and `/debug/provider-routing` reports `ValidAddresses` / `CurrentlyBlockedProviderAddresses` directly. Releases are counted from `NO VALID PROVIDERS - TRIGGERING RESET` log lines.

### The headline number

Identical rig, identical config, identical relay sequence, both primaries blocked and backup healthy:

| | `origin/main` | this branch |
|---|---|---|
| releases during 20 steady-state relays | **20 — one per relay** | **0** |
| releases while reaching the blocked state (~140 relays) | 92 | 0 |
| serves | `backup-node` | `backup-node` |

**20 relays, 20 releases on main.** That is the spec's *"the bench never gets a chance to hold"* as a measurement, and this branch takes it to zero.

### All seven checks

| | claim | result |
|---|---|---|
| P1 | backup serves when every primary is blocked | pass |
| P2 | the block holds across relays | pass — 23 relays, 0 releases |
| P3 | recovery returns traffic to primary on its own | pass |
| P4 | last resort still fires, **and only after backup** | pass |
| P5 | ordinary retry exhaustion releases nothing | pass |
| P6 | guard 2 keeps the standing state truthful | pass |
| P7 | the known pin hole | **reproduced** |

P4's ordering is worth showing, since it is the whole point of the change — one request's trace:

```
Pairing list empty                                   IgnoredProviderList=map[]
[BackupProviders] Static providers exhausted — entering backup fallback
[BackupProviders] No eligible backup providers after filtering
Backup providers failed, releasing the blocked provider list
NO VALID PROVIDERS - TRIGGERING RESET
RESET COMPLETED newValidAddressesCount=2
```

Backup is consulted first, ruled out, and only then is the block released.

P6, same scenario on both images, one relay measured:

| | `d8373be` (guard 2 absent) | `7d2911d` (guard 2 present) |
|---|---|---|
| end state | **`valid=2  blocked=0`** — reads healthy while nothing can serve | **`valid=0  blocked=2`** — the truth |

### P7 — a bad `lava-select-provider` header wipes the blocked list

One capture, one router, nothing changed between the two halves:

```
23 steady-state relays   ->  backup-node x23,  releases delta = 0
                             blocked=[primary-a, primary-b]   valid=[]

one relay with:
  -H 'lava-select-provider: does-not-exist'
                         ->  HTTP 500,        releases delta = 1
                             blocked=[]       valid=[primary-a, primary-b]
```

23 ordinary relays could not move that state. One header did.

Measured severity, rather than assumed:

- The **next ordinary relay still went to `backup-node`**, because the primaries' endpoints stay disabled independently of the provider block. Immediate traffic damage here: none. Where a provider is blocked while its endpoints are healthy (the dead-sessions-cap path), it would be real.
- The header is **client-supplied and unauthenticated** — any caller can clear the blocked list during an outage.
- It returns **HTTP 500** for what is a client error.

**This behaves identically on `origin/main`, so it is pre-existing and not a regression from this PR.** But the PR adds a guard whose stated purpose is preventing useless releases, and this one walks straight past it, because the guard asks *"is there any provider this request has not tried?"* when a pin means only one provider can ever serve.

---

## Three corrections

**1. This PR's description overstates the impact — I wrote it, and it is wrong.**

It implies `main` routes traffic to a just-blocked primary. It does not. In a direct-RPC deployment **both builds serve `backup-node`**: `GetSessions` on main does hand back a primary (which is what the unit test shows), but at the relay layer that primary immediately fails and the request reaches backup through the per-request ignored list.

The real defect is **churn** — a wasted attempt on a known-bad provider plus a release per relay, and a standing block that never survives long enough for `bench-after` / `cooldown` to be built on it. 20-per-20 is the number to quote, not "serves a dead provider". Happy to edit the description.

**2. `smartrouter_csm_blocked_providers` appears broken — separate ticket.**

It reported `0` for the entire test while `/debug/provider-routing` correctly showed `["primary-a","primary-b"]` blocked. This PR's description claims dashboards will start showing the real state during an outage; on this evidence they will not. Unrelated to section 5, but worth its own issue.

**3. Reaching a standing provider block is much harder than the spec implies.**

Three rig attempts failed before one worked, and each failure is a real fact about the product:

- **503 everything** → the endpoints go stale and selection skips the provider **per-request**. `csm_blocked_providers` stays 0 and the section 5 path is never reached at all — backup is reached by the old per-request failover instead.
- **Fail relays, serve polls** → endpoints hit the 50-failure threshold and the health prober immediately re-enables them: `probe re-enabled recovered endpoint ... reenable_probe_flaps=2, re_enable_after_k=12`. It flaps indefinitely and never holds a block.
- **Only** with `--probe-loop-interval 2h` (prober parked), relay-only failures, and a relay method that is not the tracker's poll method does a provider actually reach the blocked state.

So section 5 fixes a genuine ordering bug, but **the state it governs may be rarer in production than the spec assumes**, because the prober actively works against it. Worth weighing when prioritising sections 2 and 3.

---

### Method notes for anyone reproducing this

- The **epoch is 15 minutes** and `UpdateAllProviders` re-enables every endpoint, wiping the setup. Each run must finish inside one.
- **Start every measurement from a fresh container.** I published a wrong number mid-run — "5 relays, 5 releases" on the fixed build, contradicting the clean 0 — because an earlier check had left the backup endpoint disabled, so selection failed before backup could serve. Discarded and re-run from clean.
- One client relay can produce more than one release, because the relay layer retries and each attempt is a fresh `GetSessions` with an empty ignored set. Guard 2 bounds releases **within one selection attempt**, not across a client request. Both builds behave this way; "one release per relay" is the wrong way to state it.

The plan, full write-up and raw captures live in `agent_docs/bug-reports/failover/unbanch-order-test/` (gitignored, so local only) — happy to paste any specific capture here.

nimrod-teich (approved):

PR #336, opened 2026-08-26
feat(observability): record and surface why a provider is blocked (MAG-2599)

Tomelia1999:
Jira ticket: MAG-2599

## The problem

Every provider block looks identical today. All five block paths emit the same line —
`🔒 BLOCKING PROVIDER address=X` — and the only durable state is an address in a `[]string`. An
operator looking at an idle provider can tell **that** it is out but never **why**.

Four of the eight release paths emit nothing at all, so a provider can leave and re-enter rotation
without a trace.

The distinction is operational, not cosmetic. A provider blocked for the dead-session cap is never
reported, so the 30-second reconnect loop can never release it. One blocked for
all-endpoints-disabled is. Same line, different recovery.

## What this adds

A reason per block, held for as long as the block stands, surfaced three ways.

```
INF provider blocked  address=primary-a block_reason=never-served-successfully
    detail=consecutiveErrors=16 reported=true second_chance=true scope=primary
    blocked_count=1 valid_remaining=0
INF provider released address=primary-a block_reason=all-endpoints-disabled
    blocked_for=41.0s released_by=health-probe scope=primary
```

**`valid_remaining` is the field to read first in an incident** — it says whether this block is the
one that emptied the pool.

**`released_by` is what makes recovery diagnosable.** The health probe can release a provider on
cheap-poll evidence moments after real relay traffic blocked it, and on the primary path that was
previously silent — it logged only for backups.

`/debug/provider-routing` gains `Blocked[]` (the reason) and `HeldOff[]`. The three existing address
lists are untouched — the MAG-2202 suite reads them by name.

`smartrouter_csm_blocked_providers_by_reason{reason}` gives the split.

## It removes more log lines than it adds

Measured against `origin/main` on an identical rig, 250 relays with nothing able to serve,
`--log-level info`:

| | `origin/main` | this branch |
|---|---|---|
| `🔒 BLOCKING PROVIDER` | 801 | 0 — now DEBUG |
| `➕ ADDED TO BLOCKED LIST` | 800 | 0 — removed |
| `blocking backup provider` | 1 | 0 — folded in |
| `Resetting blocked provider list` | 399 | 0 — removed |
| `Blocked provider list reset complete` | 399 | 0 — removed |
| **`provider blocked`** | — | **801** |
| **`blocked provider list released`** | — | **399** |
| **total** | **2400** · 9.60/relay | **1200** · 4.80/relay |

Half the lines, and every one of them now says why.

**On a healthy router it contributes nothing:** 180 relays over 3 minutes produced **zero** new
lines. That bar is set by `2a01cd8`, which had to revert `Pairing list empty` from INFO at 3.8 lines
per relay.

The `801` vs `800` gap on main is a defect this replaces: `🔒 BLOCKING PROVIDER` fires *before* the
epoch check and the lock, so it logged one block that never happened.

## Two design decisions worth a reviewer's opinion

**The per-reason gauge deliberately carries no provider label.** `smartrouter_csm_provider_blocked`
is level-triggered — `publishStateSizes` republishes the whole pairing every tick. Adding `reason`
to *that* gauge would leave the previous reason's series stuck at 1 when a provider is re-blocked
for a different reason. A separate provider-free gauge republishes every reason each tick, zeros
included, so a reason that stops applying returns to 0. It counts **both pools**, so the identity is `sum()` == `smartrouter_csm_blocked_providers` + `smartrouter_csm_blocked_backup_providers`. Cardinality is bounded by
the enum, not the provider count.

**`HeldOff[]` exposes a mechanism we intend to delete.** A provider skipped after a 429 stays in
`ValidAddresses` and stays healthy, so today it appears eligible and idle at once — which reads as a
bug that is not there. This is a stopgap: it is derived at read time, holds no state, and should
disappear when hold-off folds into the same gate as everything else. It is deliberately a separate
field — held off is **not** blocked, and merging them would be worse than omitting it.

## Verification

`go build ./...`, `go vet ./...` clean · `gofmt` clean on changed files · full `go test ./...` green.
11 new tests, driven through the product path (a probe that finds no usable endpoint, a session out
of error budget, a sentinel) rather than by seeding the map — seeding would prove only that the map
exists and would pass against a call site that forgot to name a reason.

Live-router run: **9 of 10 scenarios pass**, the tenth partially and for a product reason (the epoch
unblocks everything within ~500ms of re-blocking it, so a carried record has no observable
lifetime; the unit test covers it). Plan, results and raw captures are local under
`agent_docs/bug-reports/failover/store-block-reason/` (gitignored) — happy to paste any capture here.

`golangci-lint` could not run locally: it is built against go1.24 and this module targets 1.26.6.

## A finding this surfaced — not fixed here

**The epoch gives every blocked regular provider a clean slate, no evidence required.**

`UpdateAllProviders` re-blocks the previous epoch's blocked providers to *"prevent known-bad
providers from getting a clean slate."* The release pass then treats "not in `reportedProviders`" as
"this provider is fine" — but `reportedProviders.Reset()` runs at `consumer_session_manager.go:502`,
while that pass is a **deferred goroutine that sleeps up to 500ms first** (`:440-448`). The register
is always already empty when it is read, so every non-backup takes the immediate-unblock branch and
the comprehensive-probe branch is unreachable for them.

Observed live: both providers blocked with `reported=true`, both released with
`released_by=epoch-not-reported`.

Pre-existing and out of scope here, but it matters for FAILOVER-TASKS §2 and §3 — any `bench-after`
count or `cooldown` state would be wiped every epoch regardless of behaviour. Separate ticket to
follow.

---

## Review round — six defects fixed (`e6f2498`)

A review of `221bca5` found six defects, four sharing one root cause: **the record map is a third store
alongside the two existing blocked stores, and four of the six mutation sites did not keep it in step.**
All six were reproduced before fixing, and each new test was verified to fail against the previous behaviour.

| | Defect | Fix |
|---|---|---|
| F1 | a backup dropped from config left a record nothing deleted — a **permanent** phantom, since the gauge is level-triggered | scope-aware release |
| F2 | the bulk regular release stripped the record off a provider whose **backup block still stood** (and the mirror in `ResetTransientFailureState`) | same |
| F3 | no already-blocked guard on the backup branch: re-fired the INFO, reset `Since`, overwrote the reason | guard added |
| F4 | the documented `sum()` identity was wrong — it counts **both** pools | doc + `Help` corrected, pinned by a test |
| F5 | `withCarryOver` grew `Detail` unbounded, once per epoch | `Carries` counter |
| F6 | `epoch-rebuild` logged a release for providers it immediately re-blocks | suppressed for that route |

`releaseBlockRecordLocked` replaces `takeBlockRecordLocked`: it asks whether the **other** store still
holds the address, drops the record only when neither does, and otherwise re-points it at the block that
still stands. Its second return now means *"genuinely back in rotation"*, so a partial release no longer
announces a recovery that has not happened — and a missing record on a real release warns instead of
silently logging `unspecified`.

**The test gap was systematic:** every earlier test used a single pool, and all three store bugs lived in
the paths that never got the dual-pool treatment. Adds a dual-pool fixture covering F1, F2 in both
directions, F3 and the F4 identity, plus a **drift guard** — the old assertion compared the published map
against the list that produced it, so adding a reason and forgetting `AllBlockReasons` would have passed,
and the consequence is a series that never returns to zero.

**Two review claims I checked and did not carry over:** the `sum()` claim was **not** in the Prometheus
`Help` string (only in `docs/METRICS.md` and this description), so F4 was a doc fix rather than a shipped
artifact; and the measured rig **did** have a backup configured — one backup block on each build — so the
4.80-lines-per-relay figure stands as measured.

**Deferred deliberately:** folding the three stores into one owning type. That is stage 1 of the
block/unblock consolidation this work is scoping, and doing it here and then again there is work done
twice. The scope-aware helper plus the dual-pool test matrix is what prevents recurrence in the meantime.


protocol/lavasession/block_reason_test.go
```diff
@@ -0,0 +1,674 @@
...
+	// MaximumNumberOfFailuresAllowedPerConsumerSession. Sessions are reused, so the consecutive
+	// errors accumulate on one of them and it is retired — and because the provider has never
+	// completed a relay, retiring it blocks the provider.
+	for i := 0; i <= MaximumNumberOfFailuresAllowedPerConsumerSession; i++ {
+		sessions, err := csm.GetSessions(ctx, 1, cuForFirstRequest, NewUsedProviders(nil), servicedBlockNumber, "", nil, common.NO_STATE, 0, "", "")
+		if err != nil {
+			break // already blocked; the loop has done its job
+		}
+		for _, session := range sessions {
+			require.NoError(t, csm.OnSessionFailure(session.Session, fmt.Errorf("upstream refused the relay")))
```
avitenzer, line 113: **F1 (blocker) — this test makes `go test -race ./protocol/lavasession/` fail deterministically.**

`origin/main` is race-clean on this package. This branch is not:

```
WARNING: DATA RACE
Write at 0x000106354870 by goroutine 906:
  TestSecondChanceRecoveryFlow()
      consumer_session_manager_test.go:658      // retrySecondChanceAfter = time.Second * 2
Previous read at 0x000106354870 by goroutine 324:
  (*ConsumerSessionManager).blockProvider.func1.1()
      consumer_session_manager.go:2264          // <-time.After(retrySecondChanceAfter)
```

This line drives the `allowSecondChance` path, which spawns the three-minute goroutine at `consumer_session_manager.go:2263`. That goroutine reads the package-level `retrySecondChanceAfter` with no happens-before edge to the write `TestSecondChanceRecoveryFlow` performs later, so the detector is right to flag it. Reproduced 3/3 with just the two tests:

```
go test -race -run 'TestBlockReason_NeverServedSuccessfully_ViaSessionErrors|TestSecondChanceRecoveryFlow' ./protocol/lavasession/
```

It matters more than a test-only race normally would: this PR's central claims are about concurrent state kept in step across three stores, and `-race` is how a reviewer checks them. Leaving the suite unusable under `-race` removes that instrument at exactly the moment it is most useful — and CI will never tell anyone, because the gate does not run `go test`.

I would fix it in the product rather than the test, since it also removes a real (if currently benign) cross-goroutine read:

```go
retryAfter := retrySecondChanceAfter   // read on the calling goroutine
go func() {
    <-time.After(retryAfter)
    ...
}()
```

That creates the edge the existing test pattern already assumes, and no test has to know about it.

avitenzer (commented):
Reviewed at `225adc1` from a clean `git worktree`, so this is the committed tree rather than a working dir. Build, vet and the full unit suite were run locally; each finding below was reproduced with a throwaway test before being written up, and those probe files were removed afterwards.

The design reasoning holds up. `releaseBlockRecordLocked` asking *"does the other store still hold this address"* is the right shape for the round-2 defects, the `AllBlockReasons()` drift guard is the highest-value test in the set (an omission there produces a series that never returns to zero, which is unfalsifiable from a dashboard), and keeping the provider label off the per-reason gauge is correct for the level-triggered reason given.

What is left is concentrated in one place, and the PR description already names it: **the release half of the record map is centralised and the block half is not.** Four sites write the map directly, and three of the four findings below are a state transition that changes a store or the report register without re-syncing the record. The fourth is a test-only race.

I would not merge as-is — F1 breaks `-race` on this package, and F2/F3 make the new operator-facing fields state things that are provably false, which is the specific failure this work exists to remove.

### Verification

| | Result |
|---|---|
| `go build ./...` | clean |
| `go vet` (lavasession, metrics, rpcsmartrouter) | clean |
| `go test ./protocol/lavasession/...` | pass, 42s |
| `go test -race ./protocol/lavasession/` | **FAIL** — F1. `origin/main` is clean |
| `go test -race ./protocol/rpcsmartrouter/` | pass |
| `go test -race ./protocol/metrics/` | fails on `TestHappyFlow` — **pre-existing on `origin/main`**, unrelated to this PR |

Worth restating because the checks on this PR are all green: the gate is builds / lint / govulncheck / Jira. It never runs `go test`, so it cannot see F1.

### Findings

| | Finding | Severity |
|---|---|---|
| F1 | `go test -race ./protocol/lavasession/` now fails deterministically | blocker |
| F2 | `Reported` goes stale on the call that actually reports | medium |
| F3 | `second_chance=true` on a backup block promises a recovery that cannot happen | medium |
| F4 | carried `Backup` flag mislabels `Scope` on `/debug/provider-routing` after an epoch | low |

Details are inline. Three smaller notes that did not warrant their own thread:

- `block_reason_test.go:458` — `blockInBothPools` writes `csm.blockedBackupProviders` directly, which sits a little against this file's own stated discipline of never seeding state. Defensible for a dual-pool fixture; worth one line in the comment acknowledging the exception and why.
- `setValidAddressesToDefaultValue` with a non-empty `addon` names providers in `blocked provider list released` that the addon branch does not actually return to `validAddresses`. Pre-existing limbo, but the new line overstates it. Not blocking.
- `endpointApproachingDisable` checks out — `wasEnabled && !transitioned` does imply `refusals < 50`, so `MaxConsecutiveConnectionAttempts - refusals` cannot wrap, and the bound really is 10 lines per disable episode. Lock ordering in `heldOffProviderInfoLocked` also checks out: `protocol/holdoff` has no dependency on `lavasession`, so there is no cycle.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -2088,12 +2338,32 @@ func (csm *ConsumerSessionManager) blockProvider(ctx context.Context, address st
...
 		} else {
 			csm.reportedProviders.ReportProvider(address, consecutiveErrors, disconnections, reconnectCallback)
+			reportedNow = true
 		}
 	}
 
+	// The second-chance decision is only known now, and it is the difference between a provider that
+	// returns on its own in three minutes and one that does not — so the stored record and the log
+	// both have to wait for it.
+	if blocked != nil {
```
avitenzer, line 2364: **F2 (medium) — `Reported` goes stale on precisely the call that makes it true.**

This write-back is gated on `blocked != nil`, so it only runs on the *transition into* the blocked state. But the call that genuinely reports a provider is the **second** `blockProvider` for an address that is already blocked: `secondChanceGivenToAddresses` is hit, `ReportProvider` runs, `reportedNow = true` — and `blocked` is nil on that call, so the update is dropped.

Reproduced (first offence, then a repeat block while still out of rotation):

```
record after the real report: Reported=false SecondChanceGranted=true Reason="never-served-successfully"
csm.reportedProviders.IsReported(address) == true
```

Reachable without contrivance: a session handed out before the block fails afterwards and lands in `OnSessionFailure`, which is ordinary during the outage this reporting is built for. Once it happens the record never self-corrects — the map is edge-triggered, and the next edge is the release.

The consequence is the documented contract inverted. `BlockRecord.Reported` says `Reported == false → the 30-second reconnect loop will never look at it`, so an operator reading `/debug/provider-routing` concludes the provider needs manual intervention when the reconnect loop is in fact already working on it. `6cefd16` set out to make `Reported` mean *actually reported* rather than *requested*; this is the other half of that.

Fix is local: when `blocked == nil && reportedNow`, patch the stored record in place instead of skipping the write.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -2050,26 +2276,50 @@ func (csm *ConsumerSessionManager) blockProvider(ctx context.Context, address st
...
 			if _, isBackup := csm.backupProviders[address]; isBackup {
-				csm.blockedBackupProviders[address] = struct{}{}
-				utils.LavaFormatInfo("blocking backup provider", utils.LogAttr("address", address), utils.LogAttr("GUID", ctx))
+				// Already out — leave the first reason and Since standing, and emit nothing. The
+				// regular path gets this for free (removeAddressFromValidAddresses returns
+				// AddressIndexWasNotFoundError once the address has left validAddresses); backups
+				// had no equivalent, so a failing backup re-fired the INFO on every block-triggering
+				// failure, reset its Since so blocked_for never grew, and overwrote the reason that
+				// actually took it out with whatever failed last.
+				if _, alreadyBlocked := csm.blockedBackupProviders[address]; !alreadyBlocked {
```
avitenzer, line 2316: **F3 (medium) — a backup block records `second_chance=true`, but the timer can never release a backup.**

`runSecondChance` is decided below this branch and applies to backup-scope blocks too, but the timer calls `validateAndReturnBlockedProviderToValidAddressesList`, which only walks `currentlyBlockedProviderAddresses`. A backup lives in `blockedBackupProviders`, so the timer finds nothing and returns.

Reproduced with a backup-only provider (`retrySecondChanceAfter` shortened to 200ms):

```
record: scope=backup second_chance=true reported=false
after the timer: stillBlockedAsBackup=true stillHasRecord=true
```

Read against the two meanings this PR documents, that record asserts both *"the reconnect loop will never look at it"* (`Reported == false`) and *"it comes back on its own in `retrySecondChanceAfter`"* (`SecondChanceGranted == true`). Neither holds — the backup has no route back until the epoch.

That backups were never released by the timer is pre-existing. What is new is that the record now advertises the recovery, so a stuck provider reads as self-healing. The previous behaviour was wrong and silent; this is wrong and confident, which is worse for the operator this work is aimed at.

Two ways out. In scope here: do not set `runSecondChance` on the backup branch, and record `SecondChanceGranted=false`. Out of scope: make the timer release backups as well — a product change, better as its own ticket alongside the epoch clean-slate finding already in the description.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -402,18 +543,22 @@ func (csm *ConsumerSessionManager) UpdateAllProviders(epoch uint64, pairingList
...
 	}
-	csm.setValidAddressesToDefaultValue("", nil, context.Background()) // the starting point is that valid addresses are equal to pairing addresses.
+	csm.setValidAddressesToDefaultValue("", nil, context.Background(), ReleaseEpochRebuild) // the starting point is that valid addresses are equal to pairing addresses.
 
 	// Re-block providers that were blocked in previous epoch and still exist in new pairing
 	// This prevents users from hitting known-bad providers at epoch transition
-	for blockedAddr := range csm.previousEpochBlockedProviders {
+	for blockedAddr, carried := range csm.previousEpochBlockedProviders {
 		if _, exists := csm.pairing[blockedAddr]; exists {
+			record := carried.withCarryOver()
```
avitenzer, line 552: **F4 (low) — the carried record keeps `Backup: true`, so `/debug/provider-routing` mislabels the scope.**

`carried` comes from `previousEpochBlockedProviders`, which stores whatever `blockRecordOrUnknownLocked` returned — including `Backup: true` for a block that was in the backup pool. This loop is the *regular* re-block path and never resets it, so a provider blocked as a backup in epoch N that turns up in the regular pairing in epoch N+1 (and not in the new backup list) is stored describing a pool it is no longer blocked in.

Reproduced, reading the snapshot immediately after `UpdateAllProviders` returns:

```
/debug row: Address=moves-pools Scope="backup" Carries=1
blockedBackupProviders is empty — the only standing block is the primary one
```

Low severity on two counts: the window closes once the deferred epoch release pass wakes ~500ms later, and the release log is unaffected because `releaseBlockRecordLocked` sets `ended.Backup` from its own argument rather than from the record. But it is one line, and the same masking-by-epoch-timing you note for `withCarryOver` applies — if the epoch clean-slate defect is ever fixed, this window widens to a full epoch.

```go
record := carried.withCarryOver()
record.Backup = false // this loop is the regular pool
```

protocol/lavasession/consumer_session_manager_test.go
```diff
@@ -2819,6 +2820,27 @@ func (r *stateSizeRecorder) SetBlockedProvider(_, _, providerAddress, _ string,
 }
 
 // providerBlockedSnapshot returns the last value published for each provider.
+func (r *stateSizeRecorder) SetCSMBlockedProvidersByReason(_, _ string, countsByReason map[string]int) {
```
avitenzer, line 2823: Nit — this landed directly beneath `// providerBlockedSnapshot returns the last value published for each provider.`, which now documents the wrong function. Moving the new method one line down (or below `providerBlockedSnapshot`) reunites them.

The recorder itself is right, and the reason given for replacing the map wholesale rather than merging is the correct one.

protocol/lavasession/block_reason_test.go
Tomelia1999: Fixed in `f6c5b7d`, in the product as you suggested.

I confirmed the comparative claim before touching anything — `origin/main` is race-clean on this package, this branch was not, so it is a regression I introduced rather than something I exposed. `go test -race ./protocol/lavasession/` is now **0 races**, and so is your exact pair at `-count=3`.

One thing worth stating so nobody downgrades this later: it is **not a production bug** — nothing rewrites `retrySecondChanceAfter` after startup, so the only writer is a test. Your severity argument still holds and is the reason I treated it as a blocker: this PR is a claim about concurrent state kept in step across three stores, and `-race` is how a reviewer checks that claim. Leaving it unusable removes the instrument at the moment it matters most.

I also verified the CI point. No workflow under `.github/workflows/` runs `go test` at all — every check here is a build, a Docker image or a secret scan. So green means *it compiles*, and none of the 14 tests on this branch are ever run by CI. That is worth raising with the team separately; it is much bigger than this PR.

protocol/lavasession/consumer_session_manager.go
Tomelia1999: Fixed in `f6c5b7d`, and you found a real filing error on my part.

The fix already existed — I wrote `refreshBlockOutcomeLocked` in the previous round, but I put it in **#337**, because that is where the staleness becomes *dangerous* (it bypasses the epoch probe gate). Your framing is the better one: the harm you describe is a **#336-only** harm. `BlockRecord.Reported` documents `false → the reconnect loop will never look at it`, so an operator reads `/debug/provider-routing` and concludes manual intervention is needed while the loop is already working on the provider — and that is true whether or not #337 exists.

So it has moved down here, with `TestBlockRecord_RepeatBlockRefreshesTheOutcomeButNotTheIdentity`. #337 keeps only the epoch consequence: a provider reported by a repeat block must still face the probe.

If #336 had merged alone, the bug would have shipped. Good catch on a defect whose fix already existed.

protocol/lavasession/consumer_session_manager.go
Tomelia1999: Fixed in `f6c5b7d`, but **not the way you proposed** — and I think the difference matters.

Your diagnosis is right and I reproduced it: `scope=backup second_chance=true reported=false`, and after the timer `stillBlockedAsBackup=true`. The record asserts both "the reconnect loop will never look at it" and "it comes back on its own", and neither holds. "Wrong and silent" → "wrong and confident" is exactly right, and confident-and-wrong is the failure this work exists to remove.

Where I diverged: **not setting `runSecondChance` on the backup branch is a behaviour change.** It does not just change bookkeeping — it falls through to `ReportProvider`, so a backup would be reported on its *first* offence instead of its second. That is arguably an improvement (reporting gives a backup the 30-second reconnect loop, which the timer cannot), but it is a routing change riding in a PR that is otherwise observability-only.

So this records `SecondChanceGranted = runSecondChance && !blocked.Backup` — the record stops advertising a recovery that cannot happen, and the behaviour is untouched. The same carve-out is in `refreshBlockOutcomeLocked` for the repeat-block path.

**The underlying defect is unfixed and needs a ticket**: should the timer release backups, or should backups skip the chance and be reported immediately? I have written that into the code comment and the commit message so it cannot get lost, and it belongs beside the epoch clean-slate finding. Happy to file it — or say the word if you would rather it ride here after all.

protocol/lavasession/consumer_session_manager.go
Tomelia1999: Fixed in `f6c5b7d`, one line as you wrote it.

Reproduced first: `/debug row: Address=moves-pools Scope="backup"` while `blockedBackupProviders` was empty. Pinned by `TestBlockRecord_CarriedRecordTakesTheScopeItIsReBlockedIn`, which asserts the snapshot reads `primary` after a backup-pool block is carried into the regular pairing.

Your point about the window widening is the reason I did not defer it — same masking-by-epoch-timing as `withCarryOver`, and if the epoch clean-slate defect (#337) lands, the window goes from ~500ms to a full epoch.

protocol/lavasession/consumer_session_manager_test.go
Tomelia1999: Fixed. The comment is back with `providerBlockedSnapshot`, and both new recorder methods have their own.

You were right about the cause — my insertion anchored on that function name and pushed the new method above its comment.

Also took your other two notes: `blockInBothPools` now says why it is the one fixture that seeds state directly (`blockProvider` cannot produce a dual-pool block — it takes one branch or the other — while the epoch re-block pass can), and `setValidAddressesToDefaultValue` documents that with a non-empty addon its release line names providers the addon branch does not return to `validAddresses`. Pre-existing limbo, but the new line does overstate it, so it is written down.

Tomelia1999:
Thanks — all four reproduced before I touched anything, and every fix is in `f6c5b7d`. Details are on the individual threads; this is the shape of it.

| | Finding | Outcome |
|---|---|---|
| F1 | `-race` fails deterministically | fixed in the product, as you suggested. **0 races** on the package, and on your exact pair at `-count=3` |
| F2 | `Reported` stale on the call that reports | fixed — the fix existed but I had filed it in **#337**; your framing is right that the harm is #336-only, so it moved down here |
| F3 | backup claims a second chance it cannot get | fixed **differently** — see below |
| F4 | carried `Backup` mislabels `Scope` | fixed, one line, pinned by a test |

Also took all three smaller notes, and the branch is rebased onto `main` so it is no longer `BEHIND`.

### Where I did not follow the suggestion

**F3.** Not setting `runSecondChance` on the backup branch is not only bookkeeping — it falls through to `ReportProvider`, so a backup would be reported on its *first* offence rather than its second. Defensible as an improvement, since reporting gives a backup the 30-second reconnect loop that the timer cannot, but it is a routing change inside a PR that is otherwise observability-only.

So the record stops advertising the recovery and the behaviour is untouched. **The underlying defect stands and wants a ticket** — should the timer release backups, or should backups skip the chance and be reported immediately? Written into the code comment and the commit so it does not get lost. Tell me if you would rather it ride here.

### One correction to the write-up

F1 is **not a production bug** — nothing rewrites `retrySecondChanceAfter` after startup, so the only writer is a test. That does not change the severity, and your reasoning for it is the right one: this PR's central claim is about concurrent state kept in step across three stores, and `-race` is how a reviewer checks that claim.

### The CI point deserves its own thread somewhere

I verified it. **No workflow under `.github/workflows/` runs `go test` at all** — every check on this PR is a build, a Docker image, or a secret scan. So a green PR here means *it compiles*; it says nothing about tests and nothing about races. None of the 14 tests on this branch are ever executed by CI.

That is much bigger than this PR and I would rather not bury it in a review thread. Happy to raise it separately unless you already have.

### Acceptance

`go build ./...`, `go vet ./...`, `gofmt` clean · full `go test ./...` green · `go test -race ./protocol/lavasession/` clean. Each new test verified to fail against the previous behaviour by reverting the fix and re-running.

#337 is rebased on top and keeps only the epoch consequences.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -439,12 +588,16 @@ func (csm *ConsumerSessionManager) UpdateAllProviders(epoch uint64, pairingList
 	}
 
 	// Re-block backup providers that were blocked in previous epoch and still exist in new backup list
-	for blockedAddr := range csm.previousEpochBlockedProviders {
+	for blockedAddr, carried := range csm.previousEpochBlockedProviders {
 		if _, exists := csm.backupProviders[blockedAddr]; exists {
 			csm.blockedBackupProviders[blockedAddr] = struct{}{}
+			record := carried.withCarryOver()
+			record.Backup = true
+			csm.blockedProviderRecords[blockedAddr] = record
```
avitenzer, line 596: **F5 (low) — F3's carve-out doesn't cover the carry path, so this loop can still record `second_chance=true`.**

The regular re-block loop now clears `Backup` (F4). This one sets `record.Backup = true` but leaves `SecondChanceGranted` as `withCarryOver()` found it, so the state F3 deletes on the fresh-block path is reachable one loop over.

Reproduced — a regular block that takes its second chance, then demotion out of `pairing` into `backupProviders` at the next epoch:

```
epoch N   record: Backup=false SecondChanceGranted=true Reported=false
epoch N+1 record: Backup=true  SecondChanceGranted=true blockedAsBackup=true Carries=1
```

Read against the two meanings this PR documents, that record asserts both *"the reconnect loop will never look at it"* (`Reported == false`) and *"it comes back on its own in `retrySecondChanceAfter`"* (`SecondChanceGranted == true`) — the same pair F3 removes, and neither holds.

Low, because it needs a demotion **and** a probe that fails. But it is not the ~500ms window F4 had: a backup always takes the comprehensive-probe branch, so one that cannot pass holds this record for the whole epoch rather than until the release pass wakes. Your own reason for not deferring F4 applies here with more force — #337 is approved and queued directly behind this, and it widens the window on every path that skips `ResetHealth()`.

Narrow fix, symmetric with F4:

```go
record := carried.withCarryOver()
record.Backup = true
record.SecondChanceGranted = false // the timer cannot release a backup
```

The wider one is to clear it in `withCarryOver()` instead. The second-chance goroutine returns early on `sessionEpoch != csm.atomicReadCurrentEpoch()`, so a grant cannot survive an epoch boundary in *either* pool — a carried `second_chance=true` is stale by construction, and the regular pool only escapes notice because the not-reported branch releases it ~500ms later.

---

Everything else from the last round verified at `f6c5b7d`, from a clean worktree: **0 races over 4 runs** (3/4 raced at `225adc1`), `go build ./...` / `go vet` / `gofmt` / `go test ./protocol/lavasession/` clean, F2 and F4 pinned by tests. I also diffed the net PR diff at `225adc1` against `72b9b82` — 32 lines of delta, all blob hashes and hunk offsets, no content — so the rebase preserved what was already reviewed and this is the only thing left from me.

On **F3** I'd withdraw my suggestion: you're right that not setting `runSecondChance` falls through to `ReportProvider` and would report a backup on its *first* offence, which is a routing change this PR shouldn't carry. `runSecondChance && !blocked.Backup` is the better shape.

avitenzer (approved):

PR #337, opened 2026-08-26
fix(lavasession): make the epoch's probe branch reachable again

Tomelia1999:
Jira ticket: MAG-3190

> 📚 **Stacked on [#336](https://github.com/Magma-Devs/smart-router/pull/336)** (MAG-2599). It reads
> `BlockRecord.Reported`, which that PR introduces. **Merge #336 first.**
>
> Three commits, each with its own section below: `e24884b` (the fix), `7f618a6` (review round 1),
> `c802a43` (review round 2).

## The bug

The epoch re-blocks the previous epoch's blocked providers on purpose. The comment says why:

> `// This prevents known-bad providers from getting a clean slate at epoch transition`

A second pass then decides who to let back in:

```go
if !isBackup && !csm.reportedProviders.IsReported(blockedAddr) {
    // "Non-backup provider whose probe succeeded (it wasn't added to reportedProviders)."
    // Unblock immediately.
} else {
    // comprehensive probe with tryReconnect=true
}
```

It reads *"not in the reported register"* as *"this provider is fine"*.

**But `UpdateAllProviders` empties that register before the check runs.** `reportedProviders.Reset()`
is in the function body (`:502`); the release pass is a **deferred goroutine that sleeps up to 500ms
before spawning** (`:440-448`). A defer runs after the body, so the register is always already empty.

So every non-backup looked unreported, every one took the immediate-unblock branch, and **the
comprehensive-probe branch was unreachable for regular providers**. The re-block was undone in the
same tick that created it.

## What that meant

**A provider that failed every single request for a full epoch was rehabilitated at the next tick,
with no health check of any kind.** Every 15 minutes, regardless of behaviour.

It also caps FAILOVER-TASKS §2 and §3: any `bench-after` count or `cooldown` state would be wiped
every epoch. #328 stopped the bench being wiped *per relay*; this wiped it *per epoch*.

## Evidence

Live router, MAG-2599 rig, `--epoch-duration 90s`, prober parked. Both primaries driven into a real
block:

```
INF provider blocked  address=primary-a block_reason=all-endpoints-disabled
    detail=disconnections=50 reported=true second_chance=false valid_remaining=0

  … 55 seconds later, at the epoch tick …

INF blocked provider list released released_by=epoch-rebuild released=primary-a,primary-b count=2
INF provider released address=primary-a block_reason=all-endpoints-disabled
    blocked_for=54.886s released_by=epoch-not-reported scope=primary
INF provider released address=primary-b … released_by=epoch-not-reported scope=primary
```

Blocked with `reported=true`, released through the **not-reported** branch.

*(Those field names come from #336. Before it this was three lines naming neither the reason nor the
route, which is why it went unnoticed.)*

## What the fix actually covers

Reported by a reviewer, and worth stating precisely rather than leaving the opening claim to imply
more than the change delivers. The gate reaches providers that were **actually reported**:

| reason | at the epoch, after this PR |
|---|---|
| `all-endpoints-disabled` | **probe** — fixed here |
| `never-served-successfully`, 2nd offence | **probe** — fixed here |
| `never-served-successfully`, 1st offence | immediate release — it took the second chance instead of reporting |
| `too-many-dead-sessions` | immediate release — unchanged |
| `explicit-block-signal` via `BlockProviderError` | immediate release — unchanged (no production producer) |

The unchanged rows are defensible rather than a gap, and the reason is worth knowing: **the epoch
resets the conditions those blocks were made on.** `updateEpoch` builds fresh
`ConsumerSessionsWithProvider` objects every tick — explicitly, to stop sessions accumulating — so
the blocklisted-session count behind `too-many-dead-sessions` returns to zero, and it calls
`ResetHealth()` on every endpoint first, so the disabled endpoints behind `all-endpoints-disabled`
are re-enabled. There is nothing left for a probe to find.

### Where that reset argument stops applying

An earlier draft of this section concluded that the gate's production effect is "close to zero"
because the epoch clears the state it inspects. That holds for the epoch tick — **and only for it.**
`UpdateAllProviders` has four production callers, and every one of them runs the full carry →
re-block → release pass:

| caller | endpoint health reset first? |
|---|---|
| initial registration (`rpcsmartrouter.go:2668`) | n/a — nothing is blocked yet, the pass is a no-op |
| `updateEpoch` (`:3567`) | **yes** — `endpoint.ResetHealth()` on every endpoint, before the call |
| `retryFailedProviders` (`:3880`) | **no** |
| reset-pairing (`:4038`) | **no** |

The last two pass `currentEpoch`, so nothing is reset and the probe runs against endpoints that are
still disabled. There a reported provider genuinely **stays blocked** where it previously sailed
back in. So the honest statement is not "close to zero" — it is that the effect is close to zero on
the epoch path and real on the other two.

That is the intended behaviour rather than a regression: those endpoints cannot serve, and releasing
the provider only feeds it relays it will fail. It also cannot wedge a provider out, by three
independent routes:

- the next epoch tick's `ResetHealth()` re-enables the endpoints and the probe then passes;
- the prober restores it directly via `RestoreRecoveredProvider`;
- `resetValidAddresses` (`consumer_session_manager.go:1156`) dumps the **entire** blocked list the
  moment a request finds zero valid providers — the gate sits above a floor, so it cannot black out
  a chain.

## The fix

Read the **carried block record** instead of the live register.

`previousEpochBlockedProviders` already carries the full `BlockRecord` across the transition, and
`record.Reported` holds what the register held at the moment of the block — exactly what it held
before the reset. **No new state, and no ordering dependency across a goroutine boundary** (which is
what produced the bug).

That is **equivalent by construction**, not merely a reasonable snapshot, which answers the obvious
"a snapshot can go stale" objection: `ReportProvider` is called from exactly two sites, both inside
`blockProvider`, and every `RemoveReport` site also unblocks the provider — so no path can change a
provider's report state while its block stands.

Backups are unchanged: `!isBackup` short-circuits ahead of the check, so a backup's report state was
never consulted and still is not.

## Tests

`TestEpochTransition_ProbeIsRequiredOnlyForAReportedProvider`, asserting **both** directions —
"keep everything blocked" would satisfy a one-sided test while being just as wrong as the bug:

| | expectation |
|---|---|
| reported provider | must face the probe → stays blocked when it cannot pass |
| unreported provider | released without one → nothing was recorded against it |

It uses a **direct-RPC** provider with no usable connection, because `probeDirectRPCEndpoints` is a
real routability gate — the gRPC probe dials lazily and reports success against a dead address, so
the two branches would otherwise have the same outcome and the test would prove nothing.

**Verified to fail on the previous condition**, with the reported provider back in `validAddresses`.

`go build ./...`, `go vet ./...` clean · `gofmt` clean · full `go test ./...` green ·
`go test -race ./protocol/lavasession/` clean (it was not, until #336 fixed the package-level
`retrySecondChanceAfter` read).

## One thing worth a reviewer's opinion

This restores the *intended* behaviour, which means **a reported provider now actually has to pass a
probe at the epoch**. That is a real behaviour change: providers that used to sail back in will now
stay blocked when the probe fails.

`BLOCKING-TODAY.md` notes that for direct-RPC providers this probe is a **structural check** — does
the provider have at least one enabled endpoint with a connection object — with no network call, and
that endpoint health was reset moments earlier, so on the epoch path it will usually pass. On the
two callers that skip the reset it will not, which is where the change actually bites; see *Where
that reset argument stops applying* above. If a stricter gate is wanted, that is a separate
conversation about what the epoch probe should actually verify.

---

## Review round 1 (`7f618a6`)

**Fail closed on an unknown record.** The record is now the source of truth for "was this reported",
so its unknown-record fallback decides whether a provider must prove itself — and it left `Reported`
at the zero value, which reads as *"never reported, no evidence, let it in"*. That silently
reproduces the bug this PR fixes, in the one case where we have lost track of why a provider is out.
It now fails closed and warns. The store paths that could strand a record were closed on the base
branch, so this is defensive — which is why it is worth pinning, since a fail-open default is silent
by definition.

The blast radius is bounded: `blockRecordOrUnknownLocked` has exactly two callers, both in
`UpdateAllProviders`' carry loops, so the `Reported: true` placeholder cannot leak into any other
decision.

**The test fixture was not what its comment said.** `IsDirectRPC()` is `len(DirectConnections) > 0`
and the fixture set none, so the endpoint was skipped a branch earlier and the probe failed with
*"no direct RPC endpoints found"* — a misconfiguration — rather than on the routability gate. It now
carries a connection object and toggles `Enabled`, failing with *"no enabled direct RPC endpoints"*.

That also made the **pass case reachable**, which it was not before: the fixture could only fail, so
nothing covered a reported provider being *released* through `ReleaseEpochProbe` — the case that
decides whether restoring the gate is a no-op or locks providers out. Now covered.

**Three tests added**, each verified to fail against the behaviour it pins: an unknown record still
faces the probe; a provider that passes its probe is released; and a backup faces the probe whatever
its report state — the short-circuit the rewritten comment leans on, previously unasserted.

**Also dropped** the `RemoveReport` in the not-reported branch: the register is reset before this
pass runs, so on the epoch path there is nothing to remove, and the branch no longer reads it.
(Strictly, a same-epoch `UpdateAllProviders` can see a fresh report land in the 500ms window, since
the epoch guard in `blockProvider` does not fire when the epoch has not advanced. The entry is
cleared by the next `Reset()` and `ReconnectCandidates` only picks up `Errors == 0` entries, so
nothing depends on it.)

---

## Review round 2 (`c802a43`) — the epoch half of a fix that moved down

This round originally carried `refreshBlockOutcomeLocked`, the fix for a record that goes stale on a
repeat block. **That has moved to #336**, on a reviewer's argument I think is right: the harm it
causes is felt there, not here. A record claiming `Reported=false` while the register says true
inverts what `BlockRecord` documents, so an operator reads `/debug/provider-routing` and concludes
manual intervention is needed while the 30-second reconnect loop is already working on the provider
— and that is true whether or not this PR exists. Had #336 merged alone, the bug would have shipped.

What stays here is the epoch consequence, which is this PR's subject. **Two tests, no production
change:**

**A provider reported by a REPEAT block must still face the probe.** Two ordinary blocks in one
epoch reach it: a `too-many-dead-sessions` block, which does not report, then an
`all-endpoints-disabled` block, which does. With a stale record that provider takes the
immediate-unblock branch — a working bypass of the gate this PR restores, reachable without anything
exotic. This test only passes on top of #336, which is the sharpest available statement of why the
stack order matters.

**The dead-session cap on its own**, previously untested in either direction. It is released without
a probe, and that is correct rather than a gap: the epoch rebuilds every session object, so the
blocklisted-session count that caused the block is already back at zero. There is nothing left for a
probe to find. This pins the `too-many-dead-sessions` row of the table above, which until now was
asserted only in prose.

### DIFF
--- a/protocol/lavasession/block_reason_test.go
+++ b/protocol/lavasession/block_reason_test.go
@@ -759,6 +759,56 @@ func TestBlockRecord_CarriedRecordTakesTheScopeItIsReBlockedIn(t *testing.T) {
 	require.Equal(t, uint32(1), row.Carries)
 }
 
+// epochPairing builds a single direct-RPC provider for the epoch-transition tests.
+//
+// usable decides whether probeDirectRPCEndpoints — the gate the epoch's release pass runs — passes
+// or fails. It is a routability check: at least one ENABLED endpoint carrying a connection object.
+//
+// Both shapes are genuinely direct-RPC (IsDirectRPC() is len(DirectConnections) > 0), which matters:
+// an endpoint with no DirectConnections at all is skipped one branch earlier, failing with "no
+// direct RPC endpoints found" — a misconfiguration, not the gate under test. Failing on the wrong
+// branch would still make these tests pass, while proving reachability through a provider shape
+// that cannot occur in a correctly configured deployment.
+func epochPairing(t *testing.T, address string, usable bool) map[uint64]*ConsumerSessionsWithProvider {
+	t.Helper()
+	// A connection object, not a live connection — NewDirectRPCConnection builds an HTTP client
+	// lazily and dials nothing, and the gate only asks whether one is present.
+	conn, err := NewDirectRPCConnection(context.Background(), common.NodeUrl{Url: "https://example.invalid:443/"}, 5, "")
+	require.NoError(t, err)
+	return map[uint64]*ConsumerSessionsWithProvider{
+		0: {
+			PublicLavaAddress: address,
+			Endpoints: []*Endpoint{{
+				NetworkAddress:    "https://example.invalid:443/",
+				Enabled:           usable, // an endpoint that is present but disabled fails the gate
+				DirectConnections: []DirectRPCConnection{conn},
+			}},
+			Sessions:        map[int64]*SingleConsumerSession{},
+			MaxComputeUnits: 200,
+			PairingEpoch:    firstEpochHeight,
+			StaticProvider:  true,
+		},
+	}
+}
+
+// awaitEpochReleasePass waits for the release pass to have run. It is spawned from a defer that
+// sleeps up to 500ms, and its own cleanup empties previousEpochBlockedProviders last.
+func awaitEpochReleasePass(t *testing.T, csm *ConsumerSessionManager) {
+	t.Helper()
+	require.Eventually(t, func() bool {
+		csm.lock.RLock()
+		defer csm.lock.RUnlock()
+		return len(csm.previousEpochBlockedProviders) == 0
+	}, 15*time.Second, 50*time.Millisecond, "the epoch release pass never ran")
+}
+
+// isValidAddress reports whether the provider is back in routing.
+func isValidAddress(csm *ConsumerSessionManager, address string) bool {
+	csm.lock.RLock()
+	defer csm.lock.RUnlock()
+	return slices.Contains(csm.validAddresses, address)
+}
+
 // The epoch re-blocks the previous epoch's blocked providers "to prevent known-bad providers from
 // getting a clean slate", then releases the ones it has no evidence against. It decided that by
 // asking csm.reportedProviders — but UpdateAllProviders empties that register in its body, while
@@ -768,28 +818,14 @@ func TestBlockRecord_CarriedRecordTakesTheScopeItIsReBlockedIn(t *testing.T) {
 // with no health check at all.
 //
 // The decision now reads the carried block record, which holds what the register held before it was
-// reset.
+// reset. That is equivalent by construction, not merely a reasonable snapshot: ReportProvider is
+// called from exactly two sites, both inside blockProvider, and every RemoveReport site also
+// unblocks the provider — so nothing can change a provider's report state while its block stands.
 //
-// Both cases are asserted, because "keep everything blocked" would pass a one-sided test while
-// being just as wrong as the bug. The provider's endpoint is deliberately unreachable so the two
-// branches have different outcomes: probe → stays blocked, no-probe → released regardless.
+// Both directions are asserted, because "keep everything blocked" would satisfy a one-sided test
+// while being just as wrong as the bug.
 func TestEpochTransition_ProbeIsRequiredOnlyForAReportedProvider(t *testing.T) {
-	// A direct-RPC provider with no usable connection. probeDirectRPCEndpoints is a routability
-	// gate, so it fails here — unlike the gRPC probe, which dials lazily and reports success
-	// against a dead address.
-	deadPairing := func() map[uint64]*ConsumerSessionsWithProvider {
-		return map[uint64]*ConsumerSessionsWithProvider{
-			0: {
-				PublicLavaAddress: "provider-unreachable",
-				Endpoints:         []*Endpoint{{NetworkAddress: "127.0.0.1:1", Enabled: true}},
-				Sessions:          map[int64]*SingleConsumerSession{},
-				MaxComputeUnits:   200,
-				PairingEpoch:      firstEpochHeight,
-				StaticProvider:    true,
-			},
-		}
-	}
-
+	const address = "provider-unreachable"
 	for _, tc := range []struct {
 		name           string
 		reportProvider bool
@@ -800,8 +836,8 @@ func TestEpochTransition_ProbeIsRequiredOnlyForAReportedProvider(t *testing.T) {
 		{
 			name: "reported provider must earn its way back", reportProvider: true, wantReported: true,
 			stillBlocked: true,
-			why: "it was reported, so the epoch owes it a probe — and the probe cannot pass against " +
-				"an unreachable endpoint. Releasing it anyway is the clean slate the re-block exists to prevent",
+			why: "it was reported, so the epoch owes it a probe — and the probe cannot pass while its " +
+				"only endpoint is disabled. Releasing it anyway is the clean slate the re-block exists to prevent",
 		},
 		{
 			name: "unreported provider is released without one", reportProvider: false, wantReported: false,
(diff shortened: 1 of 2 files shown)


avitenzer (approved):
Approving. Verified against the tree rather than the description, from a clean worktree at `9d55c63`.

## What I checked

**The diagnosis is exact.** `reportedProviders.Reset()` is in `UpdateAllProviders`' body (`consumer_session_manager.go:526`); the release pass is spawned from a `defer` that sleeps up to 500ms first (`:451-458`). The register is unconditionally empty when `checkAndUnblockHealthyReBlockedProviders` reads it, so `!IsReported(addr)` was always true and the comprehensive-probe branch was dead code for every non-backup.

**"Equivalent by construction" holds.** `ReportProvider` has exactly two production call sites (`:2321`, `:2340`), both inside `blockProvider`. `RemoveReport` has two (`:2746` in `RestoreRecoveredProvider`, `:3088` in the probe-success path), and both unblock the provider first. Nothing can flip report state while a block stands, so the carried `BlockRecord.Reported` is not a snapshot that can go stale.

**The fail-closed placeholder is safely scoped.** `blockRecordOrUnknownLocked` has exactly two callers, both in `UpdateAllProviders`' carry loops, so `Reported: true` cannot leak into any other decision.

**The tests genuinely pin the behaviour**, which I confirmed by mutation rather than by reading:

- reverting `!carried.Reported` to `!csm.reportedProviders.IsReported(blockedAddr)` fails `ProbeIsRequiredOnlyForAReportedProvider/reported_provider…`, `AMissingRecordStillFacesTheProbe`, and `AProviderReportedByARepeatBlockStillFacesTheProbe`;
- removing the `refreshBlockOutcomeLocked` call fails `RepeatBlockRefreshesTheOutcomeButNotTheIdentity` and `AProviderReportedByARepeatBlockStillFacesTheProbe`.

**`refreshBlockOutcomeLocked` is correct at its call conditions.** It is reached only from the `blocked == nil` branch — the provider was already out in one of the two stores — early-returns when no record exists, and only ever sets the two fields true. That matches reality: a report cannot be un-reported and a started timer cannot be un-started.

**Gates:** `go build ./...`, `go vet ./...`, `gofmt` all clean. `go test ./protocol/lavasession/` green on 3/3 runs; the new tests clean at `-count=8`.

## On the behaviour change

I pushed back on the original "production effect close to zero" framing and the body now covers it: `UpdateAllProviders` has four production callers and only `updateEpoch` calls `ResetHealth()` first, so on `retryFailedProviders` and reset-pairing the gate has real effect. I am satisfied that is correct rather than risky — those endpoints genuinely cannot serve — and that it cannot wedge a provider or black out a chain, since `resetValidAddresses` (`consumer_session_manager.go:1146`) dumps the entire blocked list the moment a request finds zero valid providers. The restriction sits above a floor.

## Non-blocking

- `reportedProviders.IsReported` now has zero production callers — it survives only in tests. Fine to keep, but worth knowing no decision path reads the register any more.
- `go test -race ./protocol/lavasession/` is flaky — a data race on the package-level `retrySecondChanceAfter`, written by tests at `consumer_session_manager_test.go:658`/`:747` while a second-chance goroutine leaked from an earlier test reads it at `consumer_session_manager.go:2264`. **Pre-existing — it reproduces on the base branch (3/4 runs) as readily as here.** Not this PR's to fix, but it deserves its own ticket, and since the PR gate never runs `go test` it is invisible in CI.

## Merge order

Base is `feat/mag-2599-block-reason` (#336), still open and behind `main`. This approval is to land **after** #336.

avitenzer (approved):

Tomelia1999:
Re-opened as **#348** against `main`.

This PR merged into its base (`feat/mag-2599-block-reason`) at `14:41:51Z` — **17 seconds after** that base had already merged to `main` at `14:41:34Z`. GitHub correctly reports this as MERGED, because it did merge into its base, but the commit never travelled onward: `git branch -r --contains 6cedcb3e` returns only `origin/feat/mag-2599-block-reason`, and `main` still carries the broken `csm.reportedProviders.IsReported(blockedAddr)` line at `consumer_session_manager.go:3038`.

MAG-3190 is therefore still live in `main`. Found while black-box testing §2 against a running router: a provider whose record read `Reported: true` was released at the epoch tick via `released_by=epoch-not-reported`, with no probe.

#348 cherry-picks the same three commits onto `main`. The three new tests were back-ported onto `origin/main` and **fail there**, which is independent confirmation the fix is absent.

PR #340, opened 2026-08-27
refactor(lavasession): drop the two legacy provider-block triggers

Tomelia1999:
Jira ticket: MAG-3290

Implements FAILOVER-TASKS section 2 (`bench-after`) and its row in section 6.

## The problem

The router takes an endpoint out of rotation after 50 consecutive failed requests. That counter
could not see the commonest way an endpoint fails.

When a node answers **HTTP 200 with an error inside the body**, the relay returns no Go error and a
200 status. Neither rule that raises the counter could fire, and the rule that resets it returned
false. **The counter froze** — it did not go up and did not go down. An endpoint answering every
single request with an error was never taken out.

Two other rules could block a provider, and neither fired in practice:

- **`too-many-dead-sessions`** needed 333 retired sessions at 16 consecutive failures each — about
  **5,300 failed requests**, on one address, inside one 15-minute epoch. 333 is
  `MaxSessionsAllowedPerProvider / 3`, not a number chosen to mean anything about health.
- **`never-served-successfully`** only caught an endpoint broken from its very first request. One
  success disarmed it permanently, and concurrent in-flight requests switched it off entirely.

## What this changes

**1. Errors now record whose fault they are, separately from whether to retry.**

`Retryable` answers "would asking someone else help". It was also being used to answer "is this
endpoint broken", and those are different questions:

| Answer | Retry elsewhere | Endpoint at fault |
|---|---|---|
| Node says it is broken (`-32603`, 5xx, unreachable) | yes | **yes** |
| Method disabled on this node (2002) | yes | **no** ¹ |
| "I do not hold that block / transaction" | yes | **no** |
| Rate limited | yes | **no** |
| Nonce too low, bad params, reverted | no | **no** |

`LavaError.EndpointAtFault` resolves this once. Nine codes were relabelled, plus seven more the
coverage test found — `STARKNET_BLOCK_NOT_FOUND`, `STARKNET_TX_HASH_NOT_FOUND`, `NEAR_UNKNOWN_BLOCK`,
`NEAR_UNKNOWN_CHUNK`, `SOLANA_LEDGER_JUMP`, `SOLANA_BLOCK_STATUS_UNAVAILABLE`,
`SOLANA_MIN_CONTEXT_SLOT_NOT_REACHED`. On those chains, "block not found" has been demoting healthy
nodes.

**No routing changed.** Every `Retryable` value is identical, pinned by a test over all 86 node and
chain codes.

¹ **On the endpoint-health counter only.** `shouldFailSessionForResult` — the QoS availability gate —
carves out `IsNonRetryable`, `IsRateLimited` and `IsDataScope` but not `IsNodeCapability`, so 2002
still scores against the provider's availability. That is unchanged from before this PR (the counter
already excused 2002 via a raw-code special case; QoS already punished it), so nothing regresses —
but the two gates disagree, and closing it needs a third session outcome rather than a fourth
carve-out: this gate only chooses between `OnSessionFailure` and `OnSessionDone`, so "not a failure"
resolves to a success, complete with a latency sample for an endpoint that declined without doing
work. Tracked separately; deliberately out of scope here.

**Two further consumers of `SubCategoryDataScope` shift, and they are not routing.** Adding the
label to 15 codes changes three readers, not one:

| Reader | Effect |
|---|---|
| `classifyEndpointHealth` | stops blaming the endpoint — intended, this is the fix |
| `shouldFailSessionForResult` | those 15 codes stop scoring against QoS availability |
| `recovery_probe.go:229,260` | the replay verdict flips `relayProbeStillFailing` → `relayProbeRecovered` |

The third is worth knowing: a pruned node disabled while failing an archive request now answers
"block not found" to the prober's replay and is re-enabled on that basis. The trial budget bounds
the exposure to 3 failures per flap. Consistent with the relay path no longer disabling on those
codes, but it is a behaviour change on a path this PR does not otherwise touch.

**The fix is JSON-RPC and gRPC only — it cannot fire on `rest`.** REST sets `IsNodeError` from the
HTTP status alone while the classification runs off the body, so a REST 200-with-error-body
classifies `UNKNOWN_ERROR` → not at fault → no `MarkUnhealthy`; and `IsNodeError=false` + status 200
makes `relayProvesEndpointHealthy` return true → `ResetHealth()`. On `rest` the counter is not
merely frozen, it is zeroed by the failing relays. Swept every status 200-499 × {CosmosSDK, EVM} ×
4 message shapes: zero at-fault classifications. Not a regression — pre-PR behaviour was identical —
but the freeze is not closed on Cosmos LCD / Aptos / Stellar. The real fix is a separate field
meaning "the reply carried an error", read only by the counter; deferred deliberately.

**2. The counter now has three outcomes instead of two.**

The reset rule was defined as the exact negation of the failure rule, so every relay either counted
against an endpoint or cleared it. That cannot express "this tells us nothing":

```
the endpoint was at fault          ->  +1
the relay demonstrably succeeded   ->  reset to 0
anything else                      ->  untouched
```

This fixes the freeze, and closes a hole in the other direction. Under the negation an unrecognised
4xx was not a failure, so it was a success, so it **reset the counter**. Measured against the
pre-change tree, the same was true of nonce-too-low, not-found and rate-limited: **every
caller-fault answer was wiping the endpoint's failure record.** An endpoint 49 failures into its
budget went back to zero on one `nonce too low`.

**3. The threshold is now a flag: `--bench-after`.** Default 50, unchanged. Bound to viper so it
works in `config.yml`; 0 falls back to the default with a warning.

**4. The two dead triggers are removed**, along with the 3-minute second chance that only
`never-served-successfully` produced.

## "Unrecognised" means two different things

They get opposite verdicts, and the first version of this PR got it wrong:

- an unrecognised **answer** — the node replied and we cannot interpret it → **not fault**
- an unrecognised **failure** — no answer arrived at all → **fault**

`LavaErrorUnknown` is `CategoryExternal` by construction, so excusing it wholesale silently stopped
penalising every uncatalogued transport failure, including EOF.
`TestGenuineFaults_StillPenalised` caught it — the test whose comment says it exists to stop this
fix trading a false positive for a worse false negative.

## Reviewer: three deliberate reversals

**1. `resource not found` now counts as blameless.** The comment on 2017 argued 2012 "stays scoreable
because a JSON-RPC -32001 is an error the node RAISED". A customer polling for a transaction that is
not mined yet gets not-found from **every** endpoint, and each retry lands on a different one — so
one unanswerable question walked the whole fleet toward being taken out. Comment rewritten, not left
contradicting the code.

**2. A data-scope answer no longer resets health.** It used to, on the reasoning that the endpoint's
correct answer earned the recovery. Rejecting a request without doing the work is not proof it can
serve.

**3. Three MAG-2599 values removed.** #336 merged mid-flight, so `BlockReasonTooManyDeadSessions`,
`BlockReasonNeverServed`, `ReleaseSecondChanceTimer` and `BlockRecord.SecondChanceGranted` have no
producer. These are an operator contract (metric labels, log fields); the alternative was keeping
them as permanently-zero series. **Worth a second opinion.**

## Verification

**The freeze fix is now held by a test that drives the real relay path.**
`relay_inner_direct_endpoint_health_test.go` runs `relayInnerDirect` against an `httptest` upstream
and asserts on `Endpoint.ConnectionRefusals`. Mutation-checked both ways:

| Mutation | Result |
|---|---|
| `if false && result.IsNodeAtFault` — delete the freeze fix | `TestRelayInnerDirect_NodeErrorInA200DisablesTheEndpoint` **fails** |
| `if targetEndpoint != nil` — reset arm always fires | 3 of the 4 **fail** |

Before this file, both mutations left `common`, `lavasession` and `rpcsmartrouter` fully green — the
fault axis was correct and wired to nothing.

- `go build`, `go vet`, `gofmt` clean; full `go test ./...` green across three consecutive runs
- **The fault-axis commit passes with no test changes outside `protocol/common`** — the evidence that
  it moved no routing
- **The counter commit fails on six of ten answer shapes when run against the pre-change tree** — two
  freeze cases and four reset cases. That is the evidence it changed what it should
- One earlier run reported a failure that did not reproduce in three subsequent runs and named no
  failing package; this package has a documented rare flake under full-suite load
  (`TestConnectorGrpc`). Recorded rather than claimed clean.

## Not in this PR

- **`cooldown` (section 3).** A benched endpoint has no automatic timed return. Routes back are the
  health prober, a last-resort trial, or the 15-minute epoch.
- **Whether polling should count.** Poll failures still cannot take anything out; they feed the
  score only.
- **`NODE_UNAUTHORIZED` (2016).** An endpoint whose credentials are rejected fails every request
  forever and no retry helps, which is a strong case for disabling it. Left as-is; its own decision.


nimrod-teich:
## Review

The core change is sound and the diagnosis is correct. Verified the three central claims:

- **The freeze is real.** A JSON-RPC error inside an HTTP 200 hits neither `MarkUnhealthy` (the `err != nil` arm never runs) nor `ResetHealth` (`shouldResetEndpointHealth` returned false). The new block at `rpcsmartrouter_server.go:4150` closes it, and the three `MarkUnhealthy` call sites (4081 / 4105 / 4152) are mutually exclusive — no double-counting.
- **Classification coverage is complete.** All four `IsNodeError` assignment sites in `direct_rpc_relay.go` (531, 722, 878, 950) have a paired `ApplyNodeErrorClassification`, and the gRPC one at 892 is guarded on the same `hasError`. `IsNodeAtFault` is populated on every node-error path the new gate reads.
- **The underflow guard survives the const→var change.** `reenableFromProbeLocked` (`consumer_types.go:399`) floors correctly, and `endpointApproachingDisable`'s `wasEnabled && !transitioned` precondition provably implies `refusals < threshold`, so the `Max - refusals` log attr can't wrap at `--bench-after 1`.

### On reversal 3 — removing the three MAG-2599 values

**Right call.** Grepped `smart-router-dashboard`, `smart-router-infrastructure`, `smart-router-helm-chart` and `devops-magmadevs`: nothing references `too-many-dead-sessions`, `never-served-successfully` or `second-chance-timer`. No dashboard or log query breaks, so keeping permanently-zero series would buy nothing.

---

### 1. Two stale comments describe deleted code

In a PR whose body argues a comment must not be left contradicting the code:

- **`protocol/lavasession/consumer_session_manager.go:70`** — `// provider addresses that were given a second chance instead of reporting them immediately` is now an orphan floating between `validAddresses` and `rateLimitHoldoff`; its field was deleted.
- **`protocol/lavasession/block_reason.go:110-114`** — `BlockRecord`'s doc still reads *"Reported and SecondChanceGranted are stored rather than recomputed…"* and *"SecondChanceGranted == true → it comes back on its own in `retrySecondChanceAfter`"*. Both identifiers are gone from the tree.

### 2. `docs/ERROR-REGISTRY-DESIGN.md` isn't updated

That doc mirrors the registry line-by-line — it annotates per-code subcategories (lines 58-74) and reproduces the `ErrorSubCategory` enum with its predicates (lines 407-421). This PR adds `SubCategoryNodeCapability`, re-labels 16 codes, and adds the `EndpointAtFault` predicate. The doc now describes a three-subcategory registry that no longer exists, and 2012 / 2013 / 3201-3206 / 3303 / 3309 / 3311 / 3331 / 3332 / 3360 / 3361 carry annotations that contradict the code.

### 3. A session-capped provider is now re-selected and error-logged on every request for the rest of the epoch

Dropping the block at `consumer_session_manager.go:1382` leaves the address in `validAddresses`, so each subsequent request re-selects it, calls `GetConsumerSessionInstanceFromEndpoint`, fails, and hits `utils.LavaFormatError` at line 1358 — which logs the entire `csm.validAddresses` slice, at ERROR level, per request. Previously the block took it out once.

Two knock-ons worth a sentence in the PR body:

- The allowance is `min(MaxSessions, MaxAllowed*(numberOfResets+1))` (`consumer_types.go:972`). The old block drained `validAddresses` → triggered a reset → grew the allowance. That escape hatch is now unreachable from this path; recovery waits for the epoch rebuild.
- That ERROR log reads `csm.validAddresses` without holding `csm.lock` — a pre-existing race the surrounding code explicitly avoids elsewhere (see the comment at ~1295), but this change turns it from once-per-epoch into once-per-request.

### 4. Minor

- `utils.LogAttr("threshold", uint64(MaxConsecutiveConnectionAttempts))` (`consumer_types.go:464`) is now a redundant conversion.
- `TestBlockReason_ReportedMeansActuallyReported` derives `quiet` from `csm.validAddresses[0]` *after* blocking `address` — it only picks a different provider because `blockProvider` mutated the slice underneath. Passes (10 providers), but the reader has to reconstruct that to see why.

---

### Checked, turned out fine

- The reordering in `classifyEndpointHealth` moves `CategoryInternal` behind the rate-limit arm. The only `CategoryInternal + SubCategoryRateLimit` code is 1020 `PROTOCOL_RATE_LIMITED`, which has no producer anywhere in the tree — dead-code-only.
- 2011 `NODE_LIMIT_EXCEEDED` (external, non-retryable, rate-limit subcategory) goes from `(false, false)` to `(false, true)` — it now requests backoff. Its only matcher is `CodeEquals(-32005)` on the JSON-RPC body classifier, which arrives via the node-error path and never reaches this function. Unreachable in practice.
- No `t.Parallel()` anywhere in `protocol/lavasession`, so the mutable `MaxConsecutiveConnectionAttempts` and the `withBenchAfter` helper can't race under `-race`.

Findings 1, 2 and 4 are cheap and worth doing before merge. 3 is a behaviour note that deserves a line in the PR body more than a code change.

Tomelia1999:
## Self-review findings — one blocking issue

Reviewed my own diff adversarially before asking anyone else to. **One finding is blocking: the
headline fix does not work on the REST transport, and the PR description claims it does.**

---

### 🔴 1. BLOCKING — the freeze fix does not apply to REST, and the counter is actively reset there

**`protocol/rpcsmartrouter/direct_rpc_relay.go:722`** (`sendRESTRelay`) and
**`rpcsmartrouter_server.go:1586`** (`relayProvesEndpointHealthy`)

`RelayResult.IsNodeError` means **different things on different transports**:

| Transport | Set from | Line |
|---|---|---|
| JSON-RPC | `hasError` — the **body** carried an error | `:531` |
| gRPC | `hasError` — the **body** carried an error | `:878`, `:950` |
| **REST** | `isNodeError` — a **status**-derived flag, true only for **5xx** | `:722` |

`relayProvesEndpointHealthy` was written against the JSON-RPC meaning. So for a REST node returning
**HTTP 200 with an error body** — the exact case this PR exists to fix:

| | |
|---|---|
| `IsNodeError` | `false` (status is 200) |
| `IsNodeAtFault` | `false` (see finding 2) |
| the new node-error arm | **never fires — freeze not fixed** |
| the reset gate | **fires — counter goes to 0** |

Verified by constructing the exact shape `sendRESTRelay` builds. Every message tried —
`internal error`, `node is syncing`, `service unavailable`, `bad gateway`, `no peers` — gave
`atFault=false, resets=true`.

**On Cosmos REST chains this PR's central fix is absent, and the "caller-fault answers wipe the
failure record" bug it claims to fix is fully intact.**

Suggested fix, two parts — the second is the real one:

```go
// relayProvesEndpointHealthy: key on the fault axis, not on a transport-specific flag
if rr == nil || rr.IsNodeError || rr.IsNodeAtFault { return false }

// sendRESTRelay:722 — make the field mean the same thing on all three transports
IsNodeError: isNodeError || hasError,
```

### 🟠 2. REST never extracts a body error code

**`direct_rpc_relay.go:725`**

```go
result.ApplyNodeErrorClassification(d.chainFamily, common.TransportREST, response.StatusCode, errorMessage)
```

JSON-RPC extracts the body's error code and prefers it over the HTTP status (`:538`), specifically so
code-based matchers fire. REST passes the HTTP status — which for a 200-with-error-body is literally
`200`.

**Consequence:** REST classification is message-matching only. Roughly half the registry's matchers
are code-based, so REST fault detection is structurally weaker than the other two transports. This
is also *why* `IsNodeAtFault` is false in finding 1. Probably its own ticket rather than this PR.

### 🟡 3. `StatusCode == 0` fails open

**`rpcsmartrouter_server.go:1586`**

`StatusCode == 0` is treated as gRPC `codes.OK` and **resets the counter**. Correct for gRPC — but 0
is also the zero value. Any `RelayResult` reaching this gate without `StatusCode` populated silently
certifies the endpoint healthy.

No reachable path does this today (HTTP always populates it; transport failures return earlier). But
the failure mode is silent and it fails *open*. Worth making the assumption explicit rather than
incidental.

### 🟡 4. Duplicated flag propagation — this pattern has already caused one bug here

**`rpcsmartrouter_server.go:4110-4116` and `:4136-4143`**

Both blocks copy the fault-axis flags one line at a time. The comment in this very file records that
two flags were once dropped at exactly this point — and this PR adds a fifth line to both. Extract a
`CopyFaultAxisFrom` method so a new flag cannot be added to the set and forgotten at a call site.

### 🟡 5. Test gaps

- **No REST-transport test anywhere.** This is how finding 1 got through — every case in
  `endpoint_fault_test.go` builds the JSON-RPC shape. `jsonrpcNodeError` is honestly named, which
  makes the gap visible in hindsight.
- **No gRPC coverage of the new arm.** The reset gate is covered on gRPC; `IsNodeAtFault` reaching
  `MarkUnhealthy` is not.
- **`TestEndpointFault_CounterMovesOnlyOnFault` mirrors the production gate rather than executing
  it.** It calls `MarkUnhealthy` itself under `if blameless.IsNodeAtFault`. Delete the real call site
  in `relayInnerDirect` and this test still passes.

---

### Not defects — decisions that need a second opinion

**The MAG-2599 enum removals.** `BlockReasonTooManyDeadSessions`, `BlockReasonNeverServed`,
`ReleaseSecondChanceTimer` and `BlockRecord.SecondChanceGranted` have no producer after this PR, and
are removed. These are operator-facing metric labels and log fields, added by #336 one week ago. The
alternative was keeping them as permanently-zero series. **I'd like agreement rather than silence on
this one.**

**Two deliberate reversals**, both flagged in the description and the commits: `resource not found`
now counts as blameless (reversing the documented reasoning on 2017), and a data-scope answer no
longer resets health.

---

### Verdict

**Major changes required** — not because the design is wrong, but because finding 1 makes the PR
description false on one of three transports.

Pieces A (`7a85ee2`) and C (`c30b558`) are ready as they stand. The work is confined to B
(`e4eb173`).

**Before merge:**
1. Fix the REST `IsNodeError` divergence (finding 1)
2. Add a REST-shaped test (finding 5) — same fix, and what stops the next one
3. Make the `StatusCode == 0` assumption explicit (finding 3)
4. Extract the duplicated flag propagation (finding 4)
5. Confirm the enum removals are acceptable

1 and 2 are the same change and roughly an hour. The rest can be follow-ups.

Tomelia1999:
## Correction to the review above — finding 1 was overstated

I claimed the freeze fix "does not work on the REST transport" and marked it blocking. **That is
wrong.** I reasoned from the shape of the code without checking what REST responses actually look
like. Corrected below.

### What REST actually does

`RestMessage.CheckResponseError` has three deliberate branches:

```go
if httpStatusCode >= 500 || httpStatusCode == 429 { return true, ... }  // node errors, via status
if httpStatusCode >= 200 && httpStatusCode < 300 {
    if cosmosTxErr, msg := checkCosmosTxError(data); cosmosTxErr { return true, msg }
    return false, ""
}
return false, ""                                                        // 4xx: client errors
```

**REST chains do not return node errors with HTTP 200 the way JSON-RPC does.** Node problems arrive
as **5xx or 429**, which the status arm in `relayInnerDirect` already handles correctly.

There is exactly one 2xx-with-error case, and the code names it:

> `// Cosmos returns HTTP 200 for both success and failed txs - must check tx_response.code`

A **failed Cosmos transaction**. Nothing else reaches it.

### And those are correctly judged blameless

Classified real `tx_response.raw_log` strings from failed Cosmos txs:

| raw_log | Classifies as | At fault |
|---|---|---|
| `insufficient fees; got: 100stake required: 1000stake` | CHAIN_INSUFFICIENT_FEE | no |
| `account sequence mismatch, expected 42, got 41` | CHAIN_INVALID_SEQUENCE | no |
| `out of gas in location: WriteFlat` | UNKNOWN_ERROR | no |
| `signature verification failed` | UNKNOWN_ERROR | no |
| `insufficient funds` | UNKNOWN_ERROR | no |

Every one is the caller's fault, so `IsNodeAtFault=false` is the **correct** verdict — not a missed
detection, which is what I claimed.

### Where I was wrong, specifically

| Claimed | Actually |
|---|---|
| "the freeze fix does not apply to REST" | No REST case exists where a node reports *itself* broken in a 200 body. The 5xx path works. |
| "`IsNodeAtFault` is false → detection missed" | It is false because these genuinely are not the node's fault. |
| 🔴 BLOCKING | Overstated. |

### What actually remains — 🟡 minor

**A failed Cosmos transaction resets the endpoint's failure counter instead of holding it.**
`relayProvesEndpointHealthy` reads `IsNodeError` (which REST derives from the status, so `false` at
200) plus a 2xx, and certifies the relay clean.

Inconsistent with decision 4, which says a caller-fault answer should HOLD. It is the same shape as
the `nonce too low` case — a customer retrying a failing tx on a Cosmos chain keeps wiping the
counter, which can mask a degrading endpoint.

But: **pre-existing behaviour**, one chain family, one request type, and the fault verdict itself is
already right. The fix is still the one-liner, now as a consistency cleanup rather than a repair:

```go
// sendRESTRelay:722
IsNodeError: isNodeError || hasError,
```

**Finding 2 also softens to a note.** REST passing the HTTP status as the classifier code does mean
code-based matchers never fire there — but the only body errors REST sees are Cosmos tx failures,
where the discriminating information *is* the message (`raw_log`), not a code. Message-matching is
the right mechanism for that case.

### Revised verdict: minor changes suggested

Findings 3 (`StatusCode == 0` fails open) and 4 (duplicated flag propagation) stand unchanged. The
test gap stands, now scoped as "add a Cosmos-tx-on-200 case" rather than "REST is untested".

Nothing here blocks merge. Thanks to @Tomelia1999 for pushing back on the claim rather than taking
it — the investigation is what showed it was wrong.

Tomelia1999 (commented):
Full review of the `bench-after` implementation. The diagnosis is right and the fault-axis separation is the correct abstraction — `TestFaultAxis_EveryRegisteredCodeIsAccountedFor` in particular is a genuinely good tripwire, and `bench_after_test.go` covers validation, consecutive-vs-cumulative, the probe trial budget tracking the flag, and the underflow guard.

Two things keep it from being ready: the central behaviour change is not held by any test, and two consumers move without being named.

**Top items before merge**

1. **Add one test that fails if the `relayInnerDirect` arms are deleted.** Mutation-proven that none exists: `if false && result.IsNodeAtFault` leaves all three packages green (88s / 42s / 8s, all `ok`). The freeze fix is the ticket and nothing holds it.
2. **Decide 2002 on the QoS axis.** `shouldFailSessionForResult` was not extended with `IsNodeCapability`, so 2002 still demotes — contradicting this PR's own table.
3. **Decide the `recovery_probe` flip.** 15 data-scope codes went from `relayProbeStillFailing` to `relayProbeRecovered`; a pruned node now re-enables on the strength of the "block not found" that disabled it, contradicting reversal #2.
4. **Fix the stale `KNOWN AND ACCEPTED` comment** at `rpcsmartrouter_server.go:1524-1543`, which names the six codes this PR flipped as ones "deliberately left on the scoring side".
5. **Make the two health arms mutually exclusive** (`else if` at `:4158`) — latent today, one line, and this PR actively invites the registry rows that would activate it.

**On the question flagged for a second opinion** — removing `BlockReasonTooManyDeadSessions`, `BlockReasonNeverServed`, `ReleaseSecondChanceTimer` and `BlockRecord.SecondChanceGranted`: **safe, remove them.** `git tag --contains 563161d` is empty and the newest release is v1.4.0 (2026-08-24), three days before #336 merged. Those labels never reached an operator, so there is no dashboard or log query to break — the operator-contract concern does not apply to values that never shipped.

**Not anchorable inline:** `docs/ERROR-REGISTRY-DESIGN.md` is not in this PR but documents the `ErrorSubCategory` enum (lines 407-421) and per-code subcategories in its tables (2001, 2005, 2008-2011, 2017). It has no `SubCategoryNodeCapability` and none of the 16 relabelled codes, so it drifts the moment this merges.

**Other test gaps**, in priority order after item 1: no new test uses `TransportREST`, so the transport where the two health gates can genuinely both fire is unexercised; `shouldFailSessionForResult` has no table over the newly relabelled subcategories (which would have surfaced items 2 and 3 as decisions); and `recovery_probe`'s verdict for a data-scope replay is unpinned in either direction.

<sub>Verdict: minor-to-major changes required. Findings verified against `c30b558` by reading the code and by two mutation experiments; working tree left clean.</sub>

protocol/rpcsmartrouter/endpoint_fault_test.go
```diff
@@ -0,0 +1,191 @@
...
+			"JSON-RPC %d both blames and certifies the endpoint", code)
+	}
+}
+
+// The verdict must actually reach the counter that disables a URL.
+//
+// The table above pins which answers blame; this pins that a blaming answer moves
+// Endpoint.ConnectionRefusals and crosses the threshold, and that a blameless one leaves it alone.
+// Without this the fault axis could be perfectly correct and wired to nothing.
+func TestEndpointFault_CounterMovesOnlyOnFault(t *testing.T) {
```
Tomelia1999, line 122: **This test cannot fail for a production reason.**

The blameless half is guarded by `if blameless.IsNodeAtFault {`, which the `require.False` three lines above proves is always false — so `survivor.MarkUnhealthy()` never executes, and `require.True(t, survivor.Enabled)` holds regardless of what `relayInnerDirect` does. The at-fault half calls `e.MarkUnhealthy()` directly rather than going through the relay path.

Verified by mutation — changing `rpcsmartrouter_server.go:4150` to `if false && result.IsNodeAtFault && targetEndpoint != nil`, i.e. deleting the freeze fix entirely:

```
ok  protocol/rpcsmartrouter  88.925s
ok  protocol/lavasession     42.888s
ok  protocol/common           8.097s
```

Same result neutering the reset arm at `:4158` (`if targetEndpoint != nil {`). **Both call-site arms can be deleted and the suite stays green.**

The doc comment says *"Without this the fault axis could be perfectly correct and wired to nothing"* — that is the current state. The classification is well covered; the wiring from classification to counter is not covered at all, and the wiring is the ticket.

Could this drive `relayInnerDirect` through an `httptest` server (the pattern already in `direct_rpc_integration_test.go`) returning `{"jsonrpc":"2.0","error":{"code":-32603,"message":"internal error"},"id":1}` at HTTP 200, and assert the endpoint disables after `--bench-after` responses? That single assertion is what makes the rest of this file mean something.

protocol/rpcsmartrouter/rpcsmartrouter_server.go
```diff
@@ -1571,20 +1571,34 @@ func shouldFailSessionForResult(err error, relayResult *common.RelayResult) bool
 		!relayResult.IsDataScope
```
Tomelia1999, line 1627: **2002 still demotes the endpoint via QoS — this contradicts the PR description's own table.**

`shouldFailSessionForResult` was not extended with the new axis. It carves out `IsNonRetryable`, `IsRateLimited` and `IsDataScope`, but not `IsNodeCapability`:

```
2002: IsNodeCapability=true IsNonRetryable=false IsRateLimited=false
      IsDataScope=false IsNodeAtFault=false  ->  shouldFailSessionForResult=true
```

2002 is `Retryable=true` so `IsNonRetryable` is false; it is not rate-limited; and `SubCategoryNodeCapability` is deliberately not `SubCategoryDataScope` — so it falls through every carve-out and scores against availability.

Net result: a full node asked for an archive method is **excused by the endpoint counter and blamed by QoS** — precisely the disagreement decision 4 exists to resolve. Under `--qos-selection-mode best` (merged in #262) that demotion is effectively zero traffic, so it is not academic.

```suggestion
		!relayResult.IsDataScope &&
		!relayResult.IsNodeCapability
```

If keeping 2002 on the QoS side is deliberate, could the divergence be stated here? Right now the two gates disagree with no comment saying so.

Tomelia1999, line 1627: **Stale comment — the `KNOWN AND ACCEPTED` block at `:1524-1543` now says the opposite of what the code does.** (Anchored here because those lines are outside the diff.)

Point 2 of that comment reads:

```
//  2. Errors that ARE classified retryable but are request-shaped rather than node-shaped
//     (CHAIN_BLOCK_NOT_FOUND 3201, CHAIN_TX_NOT_FOUND 3202, CHAIN_RECEIPT_NOT_FOUND 3203,
//     CHAIN_DATA_NOT_AVAILABLE 3205, NODE_RESOURCE_NOT_FOUND 2012, NODE_RESOURCE_UNAVAILABLE 2013)
//     are scored. [...] the JSON-RPC and REST codes listed here are deliberately left on the
//     scoring side for now
```

Those are exactly the six codes this PR gave `SubCategoryDataScope`, so `!relayResult.IsDataScope` on this line now excludes every one of them.

The comment even names the escape hatch — *"Tag these with a fault-axis SubCategory if that proves wrong in production"* — which is what happened, so it can become the record of the decision rather than a contradiction. The PR description makes a point of rewriting the 2017 comment rather than leaving it contradicting the code; this is its twin.

protocol/common/error_codes.go
```diff
@@ -405,26 +415,32 @@ var (
 	// State/data errors (3200-3299)
 	LavaErrorChainBlockNotFound = registerError(&LavaError{
 		Code: 3201, Name: "CHAIN_BLOCK_NOT_FOUND", Category: CategoryExternal,
+		SubCategory: SubCategoryDataScope,
 		Description: "Block not found", Retryable: true,
```
Tomelia1999, line 419: **The blast radius of the `SubCategoryDataScope` relabel is wider than the description states.**

"No routing changed" holds for `Retryable`, but `IsDataScope` has three consumers and two of them shift here:

| Consumer | Effect | In the description? |
|---|---|---|
| `classifyEndpointHealth` | stop blaming — intended | yes |
| `shouldFailSessionForResult:1571` | these 15 codes stop scoring against QoS availability | no |
| `recovery_probe.go:229,260` | `relayProbeStillFailing` → `relayProbeRecovered` | no |

The third is the one I would want a decision on. Verified against this branch:

```
CHAIN_BLOCK_NOT_FOUND  (3201) sub=data_scope -> unhealthy=false => relayProbeRecovered
CHAIN_TX_NOT_FOUND     (3202) sub=data_scope -> unhealthy=false => relayProbeRecovered
NODE_RESOURCE_NOT_FOUND(2012) sub=data_scope -> unhealthy=false => relayProbeRecovered
NODE_INTERNAL_ERROR    (2003) sub=none       -> unhealthy=true  => relayProbeStillFailing  (control)
```

Before this PR these were `CategoryExternal + Retryable` → `true, true` → `relayProbeStillFailing`.

Concretely: a pruned node is disabled, the prober replays the exact archive request that disabled it, the node answers "block not found" — and that now counts as **proof of recovery**. That is the same reasoning the PR body deliberately rejected in reversal #2 (*"Rejecting a request without doing the work is not proof it can serve"*), applied to `relayProvesEndpointHealthy` but not to `recovery_probe`.

The trial budget bounds the damage to 3 failures per flap, so this is a flap-rate regression rather than an outage — but it should be a decision, not a side effect. Either give `evaluateRelayProbe` the same positive-proof rule (`relayProbeInconclusive` for a data-scope answer, not `relayProbeRecovered`), or record why the two paths differ.

protocol/rpcsmartrouter/rpcsmartrouter_server.go
```diff
@@ -4112,14 +4126,36 @@ func (rpcss *RPCSmartRouterServer) relayInnerDirect(
...
+	// asked (common.LavaError.EndpointAtFault).
+	if result.IsNodeAtFault && targetEndpoint != nil {
+		rpcss.recordRelayProbeEvidence(targetEndpoint, chainMessage, originalRequestData, relayTimeout)
+		targetEndpoint.MarkUnhealthy()
+		rpcss.smartRouterEndpointMetrics.SetEndpointOverallHealth(rpcss.listenEndpoint.ChainID, rpcss.listenEndpoint.ApiInterface, endpointName, false)
+	}
+
+	// Reset only on POSITIVE proof the endpoint served — see relayProvesEndpointHealthy for why
+	// this is no longer the negation of the failure gate.
+	if targetEndpoint != nil && relayProvesEndpointHealthy(result) {
```
Tomelia1999: **These two arms are not mutually exclusive.**

`MarkUnhealthy()` at `:4150` and `ResetHealth()` here are separate `if`s, so nothing structurally stops both firing on one relay.

- JSON-RPC / gRPC: safe today. `IsNodeError` is true for any body error, so `relayProvesEndpointHealthy` returns false.
- **REST: not safe by construction.** `IsNodeError` is set from the HTTP status alone (`direct_rpc_relay.go:679-689`), while `ApplyNodeErrorClassification` runs off `hasError`. A 200-with-fault-body would `MarkUnhealthy()` and then immediately reset the counter to zero — wiping the endpoint's whole accumulated failure history.

I tried to reach it and **could not with today's matcher tables** — REST's generic table carries only capability codes (2009/2010), and the `HTTPStatusContains` matchers need an `HTTP N:` prefix the REST sender never adds. So this is latent, not live.

But `EndpointAtFault`'s own comment says the fix for an uncatalogued failure is *"one registry line"*, and the first REST fault matcher anyone adds switches this on with no test to catch it. One line closes it permanently:

```suggestion
	} else if targetEndpoint != nil && relayProvesEndpointHealthy(result) {
```

`TestEndpointFault_BlameAndProofAreDisjoint` is the right test for this, but it iterates JSON-RPC codes only — the one transport where disjointness is already guaranteed. Worth a `TransportREST` row.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -1373,12 +1366,18 @@ func (csm *ConsumerSessionManager) GetSessions(ctx context.Context, wantedProvid
 					// we can get a different provider, adding this provider to the list of providers to skip on.
 					tempIgnoredProviders.providers[providerAddress] = struct{}{}
 				} else if errors.Is(err, MaximumNumberOfBlockListedSessionsError) {
-					// provider has too many block listed sessions. we block it until the next epoch and ignore it so it won't pop up again when resetting the provider list.
+					// This address has hit its retired-session cap. Skip the provider for THIS
```
Tomelia1999: **Dropping the block turns a once-per-epoch cost into a per-request one.**

Removing `blockProvider` here is right, but the provider now stays in `validAddresses` — so every subsequent request re-selects it, re-fetches its endpoint, fails session creation again, and hits the `LavaFormatError` at `:1359`. Previously the block removed it from selection after one occurrence; now this repeats for up to a full 15-minute epoch.

That line also reads `csm.validAddresses` for a log attr with no lock held — the race `:1294`'s own comment warns about — and its frequency goes from once-per-epoch to once-per-request, which materially changes how likely it is to be hit.

Routing is fine: `tempIgnoredProviders` accumulates within the request and the empty-pool cascade still reaches backup. This is log volume and wasted work only. Given §7's care about log cost (2400 → 1200 lines at INFO), worth not giving it back here — suggest `LavaFormatDebug` for this branch specifically, and dropping `validAddresses` from the attrs.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -69,7 +68,6 @@ type ConsumerSessionManager struct {
 	// contains all provider addresses that are currently valid
 	validAddresses []string
 	// provider addresses that were given a second chance instead of reporting them immediately
```
Tomelia1999: **Orphaned comment** — the field it described (`secondChanceGivenToAddresses`) was removed, the comment was not. It now sits above a blank line describing nothing.

protocol/lavasession/block_reason.go
```diff
@@ -136,13 +125,9 @@ type BlockRecord struct {
 	// and for humans — never parse it.
 	Detail string
 	// Reported records whether the provider was ACTUALLY added to the reported-providers register —
```
Tomelia1999, line 131: **`BlockRecord`'s own doc comment at `:108-116` still documents the deleted field.** (Anchored here because those lines are outside the diff.)

```
// Reported and SecondChanceGranted are stored rather than recomputed because they decide WHICH
// recovery routes can ever fire [...]
//	SecondChanceGranted == true  → it comes back on its own in retrySecondChanceAfter
```

Both `SecondChanceGranted` and `retrySecondChanceAfter` are deleted in this PR. The field's own comment was removed correctly; the type-level one above it was missed.

This is exactly the leftover SECTION-2-BENCH-AFTER.md's *"Note for whoever lands #336 and §2"* warned about.

protocol/common/error_registry.go
```diff
@@ -87,6 +90,63 @@ func (sc ErrorSubCategory) IsDataScope() bool {
...
+//
+// Absence of information is not fault. An unclassified error returns false, so an endpoint failing
+// in a way the registry has never seen is demoted by QoS but never disabled by the counter. That is
+// accepted: disabling a URL is a hard action and warrants evidence, and the fix per occurrence is
+// one registry line — which TestFaultAxis_EveryRegisteredCodeIsAccountedFor forces to be a
+// deliberate choice.
+//
+// Defined here rather than at the call sites so the endpoint counter, the relay result's
+// IsNodeAtFault flag and anything added later cannot drift apart.
+func (le *LavaError) EndpointAtFault() bool {
```
Tomelia1999, line 132: **Documented as one rule, implemented as two.**

This comment says the function is defined here *"so the endpoint counter, the relay result's IsNodeAtFault flag and anything added later cannot drift apart"* — but `classifyEndpointHealth` overrides its `LavaErrorUnknown` verdict at `error_mapper.go:116`, immediately before calling it.

Both verdicts are correct (an unrecognised *answer* is not fault; an unrecognised *failure* is). The problem is that the override is invisible from the definition, so a third call site has no way to know which half of the split it belongs on — and getting it backwards is the mistake `TestGenuineFaults_StillPenalised` already caught once.

Two named entry points would make the split self-documenting: `EndpointAtFaultForAnswer()` (unknown → false) and `EndpointAtFaultForTransportFailure()` (unknown → true), the second delegating to the first. At minimum, a line in this comment pointing at the transport override.

protocol/common/error_fault_axis_test.go
```diff
@@ -0,0 +1,208 @@
...
+	{3333, "CHAIN_STARKNET_DUPLICATE_TX", SubCategoryNone, false, ""},
+	{3334, "CHAIN_STARKNET_TX_VERSION_UNSUPPORTED", SubCategoryNone, false, ""},
+	{3341, "CHAIN_BITCOIN_VERIFY_ERROR", SubCategoryNone, false, ""},
+	{3342, "CHAIN_BITCOIN_VERIFY_REJECTED", SubCategoryNone, false, ""},
+	{3343, "CHAIN_BITCOIN_ALREADY_IN_CHAIN", SubCategoryNone, false, ""},
+	{3344, "CHAIN_BITCOIN_WALLET_INSUFFICIENT_FUNDS", SubCategoryNone, false, ""},
+	{3362, "CHAIN_NEAR_INVALID_SHARD_ID", SubCategoryNone, false, ""},
+}
+
+func TestFaultAxis_RegistryContract(t *testing.T) {
```
Tomelia1999, line 135: **The contract table does not pin the fault axis.**

This asserts `Name`, `SubCategory` and `Retryable` — the inputs — but not `EndpointAtFault()`, the resolved verdict the file is named after. Changing `EndpointAtFault`'s logic (dropping the `CategoryInternal` arm, or the `Retryable` conjunct) would leave this table green.

Adding an `atFault bool` column costs one struct field and one `require`, makes the table pin the question it calls itself the contract for, and documents the answer for all 86 codes in one readable place — which is what a reviewer wants when deciding whether row 17 is right.

protocol/lavasession/common.go
```diff
@@ -28,16 +28,52 @@ func GetMaxAllowedBlockListedSessionPerProvider() int {
...
+var MaxConsecutiveConnectionAttempts uint64 = DefaultBenchAfter
+
+// SetBenchAfter validates and applies the --bench-after threshold. Must be called before any relay
+// is served.
+//
+// Zero would disable an endpoint on its very first failed request, which on a single-endpoint
+// deployment reproduces the "No pairings" symptom the default exists to avoid. Rather than abort
+// the router over one flag, an out-of-range value falls back to the default with a warning —
+// matching how MinSelectionChance and the optimizer weights are handled.
+func SetBenchAfter(value uint64) {
```
Tomelia1999, line 47: **`SetBenchAfter` rejects the value it then accepts.**

The rejection hint for 0 is *"must be > 0; 0 would disable an endpoint on its first failed request"* — but `--bench-after 1` also disables an endpoint on its first failed request, and is accepted (pinned by `TestSetBenchAfter_Validation` as `{"one is legal", 1, 1}`).

On a single-endpoint direct-RPC deployment that is the "No pairings" symptom this default exists to prevent, so the stated reason does not distinguish the rejected value from the accepted one.

Either say what actually makes 0 special (it is the zero value, so it is indistinguishable from unset), or add a warn-but-accept floor around `probeReenableTrialBudget`, which the re-enable path already assumes is smaller than the threshold.

avitenzer:
## Ship check: NOT READY

Ran a merge-readiness pass against the committed tree (`c30b558` in a clean worktree, not a working dir). The direction and the reasoning in the description are good — this is blocked on unaddressed state, not on design.

### Blocking

**1. 11 unresolved review threads, all posted after the last commit.**
Threads are timestamped `2026-08-30T13:22`; the last commit `c30b558` is `2026-08-30 11:47`. Zero replies, zero follow-up commits — every one is unaddressed. They are substantive, not nits.

**2. The headline freeze fix has no test.**
Confirming the mutation experiment in the `endpoint_fault_test.go:122` thread: setting `rpcsmartrouter_server.go:4150` to `if false && result.IsNodeAtFault && targetEndpoint != nil` — deleting the fix outright — leaves all three packages green. `TestEndpointFault_CounterMovesOnlyOnFault` calls `e.MarkUnhealthy()` directly and guards the blameless half with `if blameless.IsNodeAtFault {`, which the assertion three lines above proves is always false. Nothing exercises `relayInnerDirect`.

**3. `2002` still scores against QoS availability.**
Verified by reading `shouldFailSessionForResult` at `rpcsmartrouter_server.go:1568-1571` — it carves out `IsNonRetryable`, `IsRateLimited` and `IsDataScope`, but not `IsNodeCapability`. `2002` is `Retryable=true`, so it falls through every carve-out. This contradicts the description's own table, which lists `Method disabled on this node (2002)` → *Endpoint at fault: no*.

**4. Stale contract comment at `rpcsmartrouter_server.go:1524-1542`.**
The `KNOWN AND ACCEPTED` block states that `3201`, `3202`, `3203`, `3205`, `2012`, `2013` "are scored" and that "the JSON-RPC and REST codes listed here are deliberately left on the scoring side for now". This PR gives all six `SubCategoryDataScope`, so the code now does the opposite of what the comment promises.

**5. The description overclaims.**
"No routing changed" holds for `Retryable` specifically, but the `SubCategoryDataScope` relabel has two further consumers that shift and are not mentioned: `shouldFailSessionForResult` (15 codes stop scoring against availability) and `recovery_probe.go:229,260` (`relayProbeStillFailing` → `relayProbeRecovered`). None of the 11 self-review findings appears in the body either — which is what decides whether a reviewer goes looking for them.

### Non-blocking, but worth recording

- **The fix is JSON-RPC/gRPC only; it cannot fire on `rest`.** Independently reached the same place as the `:4158` thread, from the other direction. REST sets `IsNodeError` from the HTTP status alone (`direct_rpc_relay.go:679-689`, `<500 → false`) while `ApplyNodeErrorClassification` runs off `hasError` with `response.StatusCode` as the error code. So a REST 200-with-error-body classifies `UNKNOWN_ERROR` → not at fault → no `MarkUnhealthy`; and `IsNodeError=false` + status 200 → `relayProvesEndpointHealthy` returns **true** → `ResetHealth()`. On `rest` the counter is not merely frozen, it is zeroed by the failing relays. Swept every status 200–499 × {CosmosSDK, EVM} × 4 message shapes: **0** at-fault classifications, so the two arms at `:4150`/`:4158` cannot both fire today. Not a regression — pre-PR behaviour was identical — but the description reads as though the freeze is closed, and on Cosmos LCD / Aptos / Stellar it is not.
- Removed reason labels still documented in two operator-facing places: `docs/METRICS.md:366` and the Prometheus `Help` string at `protocol/metrics/smartrouter_metrics_manager.go:512` both still enumerate `too-many-dead-sessions` and `never-served-successfully`. (Checked the consumer repos — helm-chart, dashboard, automation, enterprise reference neither, so nothing external breaks.)
- `docs/ERROR-REGISTRY-DESIGN.md:407-421` lists the `ErrorSubCategory` enum without `SubCategoryNodeCapability`, and its per-code table still presents `2017` as the sole `SubCategoryDataScope` code.
- Moving the rate-limit case ahead of the category checks in `classifyEndpointHealth` changes two verdicts not mentioned in the body: `2011` gains `needsBackoff` (was `false,false`), `1020` loses `shouldMarkUnhealthy` (was `true,true`). Both are inert — `relayInnerDirect`'s `needsBackoff` return is discarded at its only call site, `rpcsmartrouter_server.go:1922` — but the rules list at `error_mapper.go:58-70` still leads with "CategoryInternal → unhealthy + backoff", which `1020` no longer follows.
- `consumer_session_manager.go:1369`: the `MaximumNumberOfBlockListedSessionsError` arm is now byte-identical to the `MaximumNumberOfSessionsExceededError` arm above it. Worth merging the cases and keeping the comment.

### Evidence

| Check | Result |
|---|---|
| `go build ./...`, `go vet` (common, lavasession, rpcsmartrouter) | pass |
| `go test -race -count=1 ./protocol/{common,lavasession,rpcsmartrouter}/...` | pass — 8.4s / 42.4s / 99.6s |
| `go test ./... -count=1` | pass, exit 0, zero failures (run twice) |
| `gofmt -l` | no file this PR touches is unformatted (10 pre-existing offenders, none in the diff) |
| Diff hygiene | 23/23 files scanned, positive control returned 4. No stray prints, TODOs, skips, `//nolint`, credentials, or generated-file drift |
| Branch state | 0 commits behind `main`, 0 files overlapping with what landed meanwhile |
| Tests | 6 removed, 17 added; all 6 removals traced (4 renamed + inverted, 2 mechanism-deleted) |
| Description's "Verification" claims | 3/3 reproduced at this commit |

Note that this repo's PR gate runs builds, lint, `govulncheck` and the Jira check — it never runs `go test`. All 20 green checks say nothing about the behaviour this PR changes, which is why the local runs above are the only test evidence.

### What clears it

Reply to or resolve the 11 threads; add a test that drives `relayInnerDirect` rather than `Endpoint.MarkUnhealthy` directly; extend `shouldFailSessionForResult` with the `IsNodeCapability` carve-out; rewrite the `KNOWN AND ACCEPTED` block; and fold the undisclosed `IsDataScope` consumers plus the `rest` gap into the description.

*Ship check only — the merge decision is not mine. An independent ticket-obligation pass and adversarial pass against MAG-3290 were not run; the blockers above stand without them.*

Tomelia1999:
## Addressed — rebased onto `main` and pushed (`f7ef51f` → `f773559`)

Thanks both. Four commits on top of the original four, so the diff you each reviewed is still readable underneath.

### One thing neither review could have caught

**The branch did not compile its tests once rebased onto current `main`.** #348 (MAG-3190) landed six epoch-probe tests in `block_reason_test.go`, plus two calls in `pool_inventory_test.go`, after both of your passes. They use the 9-argument `blockProvider` and `BlockReasonTooManyDeadSessions` — both removed here.

Three safety nets missed it independently:

- **git** merged without a conflict, because the new tests are appended at the end of the file and overlap no line this branch touches
- **`go build ./...`** passed, because it does not compile test files
- **CI** passed all checks, because the PR gate runs builds, lint and `govulncheck` but never `go test`

So it would have merged green and broken the test build on `main`. Avi's note that the gate never runs `go test` was not a footnote — this is a live instance of it, and it will hit the next branch that sits in review while `main` moves. Worth a separate ticket.

Fixed in `99bd59e`: nine call sites updated; two tests swapped to a surviving reason (they used the removed enum incidentally — the axis under test is report state); and `TestEpochTransition_DeadSessionCapIsReleasedWithoutAProbe` deleted, since its subject is the block trigger this PR removes. The rule underneath it is already covered by the "unreported provider is released without one" row of `ProbeIsRequiredOnlyForAReportedProvider`, and the deletion leaves a comment pointing there.

### Fixed

| | |
|---|---|
| **Freeze fix had no test** | `60d9a3e` — `relay_inner_direct_endpoint_health_test.go` drives `relayInnerDirect` against an `httptest` upstream and asserts on `Endpoint.ConnectionRefusals` |
| **Merge conflict** | rebased; `rpcsmartrouter.go` keeps both `--cache-timeout` and `--bench-after` |
| **Stale `KNOWN AND ACCEPTED` block** | `f773559` — rewritten to record the decision instead of contradicting it |
| **Metrics doc + Prometheus `Help` string** | `f773559` — both stop advertising the two dead label values |
| **`ERROR-REGISTRY-DESIGN.md`** | `f773559` — adds the fourth subcategory, its predicate, `EndpointAtFault`, and the relabelled rows |
| **`classifyEndpointHealth` rules list** | `f773559` — rewritten in switch order, documenting the `2011` / `1020` verdict shifts |
| **Two orphan comments** | `f773559` |
| **Duplicate session-cap arms** | `3758516` — merged |
| **Session-cap log at ERROR per request** | `3758516` — capacity is not an error; debug, and without the unlocked `validAddresses` read |
| **Health arms not mutually exclusive** | `3758516` — chained with `else if` |

On the test: mutation-checked both ways. Deleting the blame arm fails it; making the reset arm unconditional fails three of the five. Before this file, both mutations left `common`, `lavasession` and `rpcsmartrouter` fully green.

Worth recording — I first wrote a test asserting the two health gates can never both fire, and it failed. **They genuinely can**: they read different fields, and on REST `IsNodeError` comes from the HTTP status alone. That is exactly why the arms need chaining, so the test now pins the hazard rather than a false invariant, and says that if it ever starts failing the `else if` is safe to relax.

### Deliberately not fixed

**`2002` still scores against QoS.** Staying as-is, and the description now says so instead of claiming otherwise.

It is pre-existing rather than a regression: before this PR the endpoint counter already excused `2002` through a raw-code special case, and QoS already punished it. This PR replaced the special case with a proper label but did not carry it to the QoS gate.

**Avi — one push-back on the remedy.** *"Extend `shouldFailSessionForResult` with the `IsNodeCapability` carve-out"* was my own first suggestion too, in the `:1571` thread, and I have withdrawn it there. That gate only chooses between `OnSessionFailure` and `OnSessionDone`, so "not a failure" resolves to **a success** — an availability sample of 1 plus a latency sample. A node answering "that method is switched off here" replies in about a millisecond, so the carve-out would rank the node that refused above every node that served the request, get it picked first, and have it refuse again. It swaps a mild pre-existing problem for a worse new one that looks like a fix.

The honest fix is a third, neutral outcome next to the rate-limit branch — release with no QoS sample, the way a 429 and a race loser already are. That is a change to the shape of that ladder, not a carve-out, and it is wider than section 2's scope. Tracking separately.

### Disclosed, decision still open

**The `SubCategoryDataScope` relabel flips the probe-replay verdict.** `recovery_probe.go:229,260` moves from `relayProbeStillFailing` to `relayProbeRecovered` for those 15 codes, so a pruned node disabled while failing an archive request can now answer "block not found" to the prober's replay and be re-enabled on that basis. Consistent with the relay path no longer disabling on those codes, and the trial budget bounds the exposure to 3 failures per flap — but it is a behaviour change on a path this PR does not otherwise touch. Not changed here; flagging it rather than letting it land unnamed.

**The freeze is not closed on `rest`.** Avi's sweep and my `:4158` thread reached this from opposite directions and agree: REST sets `IsNodeError` from the HTTP status alone, so a 200-with-error-body is not at fault *and* passes the reset gate. Pre-existing, unchanged, and now stated in the description. The real fix is a separate field meaning "the reply carried an error", read only by the counter.

### Dropped as not merge-relevant

Three of my own self-review threads, closed without action — good observations, none of them worth holding a merge for, and better as follow-ups than as open threads: `EndpointAtFault` carrying two rules under one name, adding an `atFault` column to the fault-axis contract table, and `--bench-after` rejecting `0` while accepting `1`.

### Verification

`go build`, `go vet ./protocol/...`, `gofmt` clean on every file touched. Full `go test ./...` green. `go test -race` green on `common`, `lavasession` and `rpcsmartrouter`. Merges into `main` with no conflicts.

Still outstanding: the description update is written but not yet published.

protocol/rpcsmartrouter/endpoint_fault_test.go
Tomelia1999: 60d9a3e replaces it with tests that drive relayInnerDirect against a real upstream and assert on the counter. Mutation-checked: deleting the blame arm now fails them.

protocol/rpcsmartrouter/rpcsmartrouter_server.go
Tomelia1999: f773559 rewrites the block. It now records that SubCategoryDataScope — the escape hatch the old text named — was used on exactly those six codes.

protocol/rpcsmartrouter/rpcsmartrouter_server.go
Tomelia1999: 3758516 chains them with else if. Also added a test pinning that the two gates are NOT mutually exclusive, which is why the chaining is needed rather than decorative.

protocol/lavasession/consumer_session_manager.go
Tomelia1999: 3758516 drops the two capacity cases to debug and removes the unlocked validAddresses read. The ERROR line is kept for the branch that is genuinely unexpected.

protocol/lavasession/consumer_session_manager.go
Tomelia1999: f773559 removes it.

protocol/lavasession/block_reason.go
Tomelia1999: f773559 rewrites it — Reported only, plus a note that the timed return is gone until section 3's cooldown.

protocol/rpcsmartrouter/rpcsmartrouter_server.go
Tomelia1999: Not fixing this in #340 — deliberate, and recorded in the description and the summary comment.

**Withdrawing the suggestion above.** Adding `!relayResult.IsNodeCapability` to this gate is the wrong fix, not merely an incomplete one. The gate chooses only between `OnSessionFailure` and `OnSessionDone`, so "not a failure" resolves to a **success** — availability 1 plus a latency sample. A node answering "that method is switched off here" replies in about a millisecond, so the carve-out would rank the node that declined above every node that served the request, get it picked first, and have it decline again.

The real fix is a third, neutral outcome beside the rate-limit branch — release with no QoS sample, as a 429 and a race loser already do. That is a change to the shape of that ladder, not a carve-out, and wider than section 2. Tracked separately.

Note this is pre-existing, not a regression: the counter already excused 2002 through a raw-code special case and QoS already punished it. This PR replaced the special case with a proper label without carrying it to QoS.

protocol/common/error_codes.go
Tomelia1999: Accepted as-is for #340, and now stated in the description rather than landing unnamed.

The QoS effect and the `recovery_probe` verdict flip are both real and both follow from the same relabel. The probe change is consistent with the relay path no longer disabling on those codes — an endpoint that answers "I do not hold that block" is up and talking — and the trial budget bounds the exposure to 3 failures per flap.

Not changed here. If the flap rate turns out to matter in production, the fix is to make `evaluateRelayProbe` treat a data-scope answer as `relayProbeInconclusive` rather than `relayProbeRecovered`.

protocol/common/error_registry.go
Tomelia1999: Correct, but not merge-relevant — closing without action for #340. Nothing is broken: both verdicts are right today and TestEndpointFault_UnrecognisedAnswerAndUnrecognisedFailureDiffer already guards the split. This is a naming improvement, and renaming it would touch call sites across three packages — not worth a third review round on this branch.

protocol/common/error_fault_axis_test.go
Tomelia1999: Correct, but not merge-relevant — closing without action for #340. The fault logic itself is covered by the tests in protocol/rpcsmartrouter; the gap is only that this table does not assert the property it is named for. Worth doing when the table is next touched.

protocol/lavasession/common.go
Tomelia1999: Correct, but not merge-relevant — closing without action for #340. The behaviour is intentional and pinned by TestSetBenchAfter_Validation. Only the rejection message reads oddly, since it argues against a value the code accepts.

avitenzer:
@Tomelia1999 **[P1] Preserve session-cap recovery when removing the legacy provider block.**

Reviewed head `f7735591b8f99f1042747db92e0cf07fa88bd4d8` against base `0cc8789c9da98153246f3ca8e55ef1855d772721`.

The change at [`consumer_session_manager.go:1711–1728`](https://github.com/Magma-Devs/smart-router/blob/f7735591b8f99f1042747db92e0cf07fa88bd4d8/protocol/lavasession/consumer_session_manager.go#L1711-L1728) can leave an enabled direct-RPC endpoint unable to receive any more requests after its retired-session allowance is exhausted.

The capacity guard still rejects session creation at `min(MaxSessionsAllowedPerProvider, (MaxSessionsAllowedPerProvider / 3) * (numberOfResets + 1))`. Previously, the cap also removed the provider from `validAddresses`. With a single provider and no backup, the next customer request could then release the empty pool, increment `numberOfResets`, and obtain a new session.

Now the provider remains in `validAddresses`, so [`releaseBlockedProvidersIfPoolEmpty`](https://github.com/Magma-Devs/smart-router/blob/f7735591b8f99f1042747db92e0cf07fa88bd4d8/protocol/lavasession/consumer_session_manager.go#L1412-L1415) returns immediately. The allowance never grows, and each new request selects the same provider, hits the same cap, and returns `No pairings available.` Healthy endpoint probes do not reclaim these sessions. This can persist until a pairing/epoch rebuild even if the upstream has recovered.

**Reproduced on both revisions through the real direct-RPC session manager:**

| Session limit | State when the cap is reached | Base | PR head |
| --- | --- | --- | --- |
| 3 | 1 retired session; endpoint enabled; 16 refusals | Next customer request recovers | Fails: `No pairings available.` |
| 1,000 (default) | 333 retired sessions; endpoint enabled; 16 refusals | Next customer request recovers | Fails: `No pairings available.` |

The reproduction inserts a successful relay between each burst of 16 failures. The endpoint therefore never reaches the default `bench-after=50`, while retired sessions accumulate. This also shows why “~5,300 consecutive failures” in the new comment is misleading: the cap can be reached across separate bursts with successes in between.

The new `TestRetiredSessionCapDoesNotBlockTheProvider` stops after the first rejected request and checks that no block record exists. It does not check whether a subsequent request can actually obtain a session.

**Requested change:** retain a bounded way to recover session capacity when the retired-session cap is reached, while preserving the intended separation from endpoint-health blocking. Add a regression test that obtains a usable session on a subsequent request without waiting for an epoch rebuild. The default-cap case should remain covered.

The existing `common`, `lavasession`, `rpcsmartrouter`, and `metrics` package suites and `go vet` pass; the additional reproduction below passes on the base and fails on this head. It uses session-manager test fixtures and simulates relay outcomes; it does not require a live upstream.

<details>
<summary>Regression test and command</summary>

Save as `protocol/lavasession/review_session_cap_test.go` in either revision, then run:

```bash
go test ./protocol/lavasession -run '^TestReviewPR340RetiredSessionCapacityRecovers$' -count=1 -v
```

```go
package lavasession

import (
	"context"
	"errors"
	"fmt"
	"testing"
	"time"

	"github.com/magma-Devs/smart-router/protocol/common"
	"github.com/rs/zerolog"
	"github.com/stretchr/testify/require"
)

// A short transient failure retires one session while the endpoint remains
// enabled. Once the failure is over, capacity recovery must still allow a relay.
func TestReviewPR340RetiredSessionCapacityRecovers(t *testing.T) {
	logLevel := zerolog.GlobalLevel()
	zerolog.SetGlobalLevel(zerolog.Disabled)
	t.Cleanup(func() { zerolog.SetGlobalLevel(logLevel) })
	for _, cap := range []int{3, 1000} {
		t.Run(fmt.Sprintf("cap_%d", cap), func(t *testing.T) {
			reviewPR340RetiredSessionCapacityRecovers(t, cap)
		})
	}
}

func reviewPR340RetiredSessionCapacityRecovers(t *testing.T, cap int) {
	original := MaxSessionsAllowedPerProvider
	MaxSessionsAllowedPerProvider = cap
	t.Cleanup(func() { MaxSessionsAllowedPerProvider = original })

	const address = "review-direct-upstream"
	ctx := context.Background()
	csm := CreateConsumerSessionManager()
	pairing := epochPairing(t, address, true)
	pairing[0].MaxComputeUnits = 1000000000
	require.NoError(t, csm.UpdateAllProviders(firstEpochHeight, pairing, nil))
	endpoint := csm.pairing[address].Endpoints[0]
	getSession := func() (ConsumerSessionsMap, error) {
		return csm.GetSessions(ctx, 1, cuForFirstRequest, NewUsedProviders(nil),
			servicedBlockNumber, "", nil, common.NO_STATE, 0, "", "")
	}

	retiredCount := GetMaxAllowedBlockListedSessionPerProvider()
	for retired := 0; retired < retiredCount; retired++ {
		// A success between bursts keeps the endpoint below the health threshold
		// and prevents base's never-served trigger from affecting this test.
		sessions, err := getSession()
		require.NoError(t, err)
		for _, info := range sessions {
			require.True(t, info.Session.IsDirectRPC())
			require.NoError(t, csm.OnSessionDone(info.Session, servicedBlockNumber,
				cuForFirstRequest, time.Millisecond,
				info.Session.CalculateExpectedLatency(2*time.Millisecond),
				1, 1, 1, false, nil))
		}
		endpoint.ResetHealth()

		// Sixteen failures retire the session but stay below the default bench of 50.
		for i := 0; i <= MaximumNumberOfFailuresAllowedPerConsumerSession; i++ {
			sessions, err = getSession()
			require.NoError(t, err)
			endpoint.MarkUnhealthy()
			for _, info := range sessions {
				require.NoError(t, csm.OnSessionFailure(info.Session, errors.New("temporary upstream failure")))
			}
		}
		require.True(t, endpoint.IsEnabled(), "the health threshold was never reached")
	}
	require.Len(t, csm.pairing[address].Sessions, retiredCount)
	t.Logf("retired=%d refusals=%d endpoint_enabled=%t", retiredCount,
		endpoint.ConnectionRefusals, endpoint.IsEnabled())

	// Both revisions reject the first request that discovers the exhausted cap.
	_, err := getSession()
	require.Error(t, err)

	// The next customer request used to release the pool and grow the allowance.
	// No more upstream errors occur, yet the PR continues rejecting all requests.
	sessions, err := getSession()
	require.NoErrorf(t, err, "healthy endpoint stranded at session cap: resets=%d valid=%v",
		csm.atomicReadNumberOfResets(), csm.validAddresses)
	for _, info := range sessions {
		require.NoError(t, csm.OnSessionDiscarded(info.Session, nil))
	}
}
```

</details>

Tomelia1999:
@avitenzer Confirmed, and fixed — thank you, this was the right catch.

Reproduced your test on both revisions before touching anything: passes on `0cc8789`, fails on the PR head, both at `session limit 3` and at the default `1000`. `resets=0`, endpoint enabled, exactly as you described.

## What we changed, and why it is not what you asked for

You asked for *"a bounded way to recover session capacity when the retired-session cap is reached"*. We removed the cap instead.

The reasoning, so you can push back if you disagree:

**The cap belonged to the block, not to the router.** The threshold is `MaxSessionsAllowedPerProvider / 3` — a third of an unrelated limit, not a number anyone picked to mean something about an upstream. Its only consumer was the Lava-era provider block this PR deletes, and the only thing that ever raised it was the pool reset that block triggered. It is the same inheritance as the second-chance timer: a mechanism that made sense when reporting a provider had on-chain consequences, and means nothing here.

**Session growth stays bounded without it.** That was decision 1's reason for keeping it, and it does not hold: `MaxSessionsAllowedPerProvider` already caps total sessions per provider. The third only made the router hit a wall three times sooner — 5,328 failed relays instead of 16,016.

**Adding a door to a gate we do not believe in is worse than removing the gate.** A recovery path would have kept a limit whose number is arbitrary, whose purpose is served elsewhere, and which exists only because a deleted mechanism needed it.

What remains is the total-session ceiling. That one has never blocked-and-recovered on any revision — it behaves identically on `main` today — so this is not a new wall, it is the pre-existing one, three times further out. Worth its own ticket; out of scope here.

## You were also right about the "5,300 consecutive failures" line

Wrong twice over, and it was ours. The failures need not be consecutive, and the endpoint is provably healthy when the ceiling is reached — a success resets the endpoint counter but does not return a retired session, so an upstream failing in bursts retires session after session while never approaching `bench-after`'s 50. The code comment repeating that claim is gone.

## Tests

- **`TestRetiredSessionDoesNotStrandTheProvider`** replaces `TestRetiredSessionCapDoesNotBlockTheProvider`. You were right that the old one stopped one request too early — it asserted the refusal and never asked what happened next. The new one asserts a working upstream serves the *following* request, with no block, no pool reset and no epoch rebuild.
- **`TestRetiredSessionCapacityRecovers`** is your reproduction, kept at both `3` and the default `1000` as you asked.

Both mutation-checked: restoring the ceiling fails both.

`TestMaximumBlockedSessionsErrorsInPairingListEmpty` failed on this change, correctly — its premise was the removed ceiling. The property it guards still holds one session further along, so it is renamed `TestSessionCeilingErrorsInPairingListEmpty` and repointed at the total ceiling.

## One heads-up

**Your test as written no longer compiles.** It calls `GetMaxAllowedBlockListedSessionPerProvider()`, which had no callers left and went with the cap. The committed version derives the count from `MaxSessionsAllowedPerProvider / 3` directly.

`go build`, `go vet`, `gofmt` clean; full `go test ./...` green; `go test -race` green on `lavasession`.

Tomelia1999:
## Rebased onto current `main` (`9b1ea64` → `5efbf77`), and one self-reported regression

@avitenzer @nimrod-teich — the branch was a week behind; it is current now and merges clean. Two things worth your time before this goes further, and the second is a defect I introduced.

### #349 merged, and it needed a fix the rebase could not catch

#349 gives every endpoint disable a *reason*, so `Endpoint.MarkUnhealthy` now takes one. This branch adds a call site #349 never saw — the node-error arm that is the whole point of section 2.

Git merged that arm with **no conflict**: it arrives whole from this branch and touches nothing #349 edited. The result did not compile. `go build` did not say so either, because three of the five affected files are tests — only `go vet` / `go test` catch it.

Fixed in `5efbf77`. The arm records `EndpointDisableNodeError` unconditionally rather than going through `endpointDisableReasonFor`, because reaching it *requires* an answer to have arrived (`err == nil`, status neither 5xx nor 429) — so `unreachable` is unrepresentable there by construction. Verified end to end: an upstream answering HTTP 200 with `-32603` now disables with `DisableReason="node-error"`, which is the value #349's own QA matrix expects for that row. Before the fix it would have been `unspecified`, which that guide calls a bug wherever it appears.

That is now the third clean-merge-that-does-not-compile on this branch. All three were invisible to CI, because the PR gate never runs `go test`. Worth a ticket independent of this PR.

### The regression: one bad URL can starve its healthy siblings

**Found while re-reviewing my own fix for your P1, @avitenzer. It is in `6323a8b` — the commit that answered you — so it has had no reviewer but me.**

Removing the retired-session ceiling fixed the stranding you found. That ceiling was also doing a second job nobody had written down.

It counted blocked sessions **matching the current connection** — a per-address limit. The ceiling that survives counts `len(cswp.Sessions)`, which spans **every address of the provider**. So with the per-address limit gone, one bad URL can consume the provider's entire session budget.

Same probe on both trees. One provider, two URLs, `MaxSessionsAllowedPerProvider = 20`, break one:

| | bad URL stopped at | healthy URL |
|---|---|---|
| `main` | 6 sessions | **served** |
| this branch | 21 sessions | **refused** — `Provider reached maximum number of active sessions.` |

It is the mirror image of your finding. Yours was *the cap strands a healthy provider*; this is *removing the cap lets one URL starve the provider*. Before, a bad URL cost you that URL. Now it can cost you the whole provider, healthy endpoints included.

**Reachability:** at the default of 1000 the bad URL needs ~16,016 failed relays inside one 15-minute epoch — about 18/second on a single URL — and only bites a provider configured with more than one node-url. Soak-test territory, not day-to-day.

**Decision: not fixing it in this PR.** The correct fix is a per-URL bound of a different kind — count *total* sessions on that address against a fair share of the provider budget, rather than counting *retired* ones against a ceiling that can never be raised. That keeps one URL from eating the budget without reintroducing the stranding you found. It is a design change to session accounting rather than a carve-out, and it is wider than section 2.

Flagging rather than fixing so the decision is on the record instead of in the diff. If you would rather it did not merge in this state, say so — I do not think it should merge silently.

### Where your findings stand

Both of your reviews are addressed, with one deliberate exception:

- **Ship check (08-31)** — all five blockers done. The one exception is `2002` still scoring against QoS, closed as a decision; I pushed back on the suggested remedy in that thread because the gate has no neutral outcome, so the carve-out turns a punishment into a reward.
- **P1 (09-09)** — fixed and mutation-checked, with the caveat above.

Your comments are top-level rather than inline, so GitHub shows no resolved state for them. **Nothing here is cleared until you say it is** — your last verdict on the PR is still NOT READY.

`go build`, `go vet`, `gofmt` clean. Full `go test ./...` green. `go test -race` green on `common`, `lavasession` and `rpcsmartrouter`. 20/20 checks passing.

avitenzer (approved):

avitenzer (approved):

avitenzer (approved):

PR #344, opened 2026-08-30
fix(rpcsmartrouter): unsubscribe with the method the client called (MAG-3297)

AnnaR-prog:
# Description

Substrate `*_unsubscribe*` methods returned `-32601` through the router. The subscribe half worked and the unsubscribe half did not, on every Substrate spec — 12 methods on Acala, 10 of 12 on Shiden, and identically on **ASTAR, which is merged and live**.

**The ticket's hypothesis was that the router's `SUBSCRIBE`/`UNSUBSCRIBE` directive lookup keys on `function_tag` alone.** That turned out not to be the cause here. The dispatch path is correct: `GetParseDirective` (`chain_message_queries.go:23`) matches directives **by `ApiName`**, so `chain_unsubscribeNewHeads` is classified correctly and reaches the unsubscribe handler.

The actual cause is `getUnsubscribeMethod` (`direct_ws_subscription_manager.go`), which derived the upstream teardown method from the **subscribe** method by string surgery and never consulted the spec at all:

```go
if subscribeMethod == "eth_subscribe" { return "eth_unsubscribe" }
if subscribeMethod == "subscribe"     { return "unsubscribe" }
if <ends with "_subscribe">           { return <prefix> + "_unsubscribe" }
return "unsubscribe"          // ← everything Substrate lands here
```

All twelve of ASTAR's subscribe methods fall through to that last line — `chain_subscribeNewHeads`, `state_subscribeStorage`, `subscribe_newHead`, `chainHead_v1_follow` are none of them `<x>_subscribe`. So the router called the literal method `unsubscribe`, the node replied `-32601`, and that response was streamed back to the client verbatim. `eth_unsubscribe` worked only because it is one of the two hardcoded cases, which is exactly what made this look spec-shaped rather than code-shaped.

Three pairs are not derivable by any string rule: `chainHead_v1_follow` → `chainHead_v1_unfollow`, `author_submitAndWatchExtrinsic` → `author_unwatchExtrinsic`, `transactionWatch_v1_submitAndWatch` → `transactionWatch_v1_unwatch`.

**Nothing needs deriving.** The spec pairs each `SUBSCRIBE` directive with an `UNSUBSCRIBE` one carrying its own `api_name`, and `consumer_websocket_manager.go` reaches this path only after matching the incoming message against those directives by name — so the api on the message already *is* the spec's unsubscribe method. Use it. `getUnsubscribeMethod` stays as the fallback for a message with no api attached, which a spec-resolved message never is.

### Most critical to review

- `protocol/rpcsmartrouter/direct_ws_subscription_manager.go` — `resolveUnsubscribeMethod`, and the single call site it replaces (verified: `getUnsubscribeMethod` had no other callers repo-wide).

### Tests

The tests assert the method sent **upstream**, over the ASTAR base collection's full set of pairs. Asserting only that an unsubscribe does not error would pass against a node that happens to accept the generic `unsubscribe`. Each pair also carries a `NotEqual` against the old derivation, so the tests document why the fix exists and fail if anyone reinstates it. The eth and tendermint shapes are pinned as unchanged.

### Note for reviewers

This kills the falsifier proposed on the ticket — *"if a Substrate spec with exactly one unsubscribe pair works, that is strong support"*. Under this diagnosis that test **fails**: one pair or twelve, `chain_subscribeNewHeads` still hits the fallback. Not worth spending a spec on.

## Jira ticket

Jira ticket: MAG-3297

---

## Author Checklist

I have...

### DIFF
--- a/protocol/rpcsmartrouter/direct_ws_subscription_manager.go
+++ b/protocol/rpcsmartrouter/direct_ws_subscription_manager.go
@@ -912,8 +912,9 @@ func (dwsm *DirectWSSubscriptionManager) Unsubscribe(
 	// Only call upstream when we're the last client (single client on this subscription)
 	lastClient := len(activeSub.connectedClients) == 1
 
-	// Derive unsubscribe method: eth_subscribe -> eth_unsubscribe, subscribe -> unsubscribe
-	unsubscribeMethod := getUnsubscribeMethod(subscribeMethod)
+	// The method to call upstream is the one the client actually invoked, not one
+	// derived from the subscribe method. See resolveUnsubscribeMethod.
+	unsubscribeMethod := resolveUnsubscribeMethod(protocolMessage, subscribeMethod)
 	requestID := extractRequestID(protocolMessage.RelayPrivateData().Data)
 
 	var unsubscribeParams interface{}
@@ -981,8 +982,42 @@ func (dwsm *DirectWSSubscriptionManager) Unsubscribe(
 	return nodeResp, nil
 }
 
+// resolveUnsubscribeMethod returns the method name to call upstream to tear a
+// subscription down.
+//
+// It is the name the client actually invoked. The spec pairs every SUBSCRIBE
+// parse directive with an UNSUBSCRIBE one carrying its own api_name, and the
+// caller has already matched this message against those directives by name
+// (consumer_websocket_manager.go dispatches here only when the message resolved
+// to an UNSUBSCRIBE tag), so the api name on the message IS the spec's
+// unsubscribe method. Nothing needs deriving.
+//
+// MAG-3297: it used to be derived from the subscribe method by string surgery,
+// which only ever worked for the two shapes getUnsubscribeMethod hardcodes.
+// Every Substrate pair falls through those to the "unsubscribe" fallback —
+// chain_subscribeNewHeads, state_subscribeStorage, subscribe_newHead and
+// chainHead_v1_follow are none of them "<x>_subscribe" — so the router called a
+// method the node does not have and forwarded its -32601 to the client. Several
+// pairs are not derivable by any string rule at all: chainHead_v1_follow tears
+// down with chainHead_v1_unfollow, author_submitAndWatchExtrinsic with
+// author_unwatchExtrinsic.
+//
+// getUnsubscribeMethod remains the fallback for a message with no api attached,
+// which a spec-resolved message never is.
+func resolveUnsubscribeMethod(protocolMessage chainlib.ProtocolMessage, subscribeMethod string) string {
+	if protocolMessage != nil {
+		if api := protocolMessage.GetApi(); api != nil && api.Name != "" {
+			return api.Name
+		}
+	}
+	return getUnsubscribeMethod(subscribeMethod)
+}
+
 // getUnsubscribeMethod derives the unsubscribe method from the subscribe method.
 // eth_subscribe -> eth_unsubscribe, subscribe -> unsubscribe
+//
+// Only reachable when the incoming message carries no api name; prefer
+// resolveUnsubscribeMethod, which uses the method the client invoked.
 func getUnsubscribeMethod(subscribeMethod string) string {
 	if subscribeMethod == "eth_subscribe" {
 		return "eth_unsubscribe"
--- a/protocol/rpcsmartrouter/direct_ws_subscription_manager.go
+++ b/protocol/rpcsmartrouter/direct_ws_subscription_manager.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -1002,11 +1003,20 @@ func (dwsm *DirectWSSubscriptionManager) Unsubscribe(
 // down with chainHead_v1_unfollow, author_submitAndWatchExtrinsic with
 // author_unwatchExtrinsic.
 //
-// getUnsubscribeMethod remains the fallback for a message with no api attached,
-// which a spec-resolved message never is.
+// The DefaultApiName guard is the one that matters. A method the spec does not
+// declare still gets an api, synthesised with Name "Default-<method>"
+// (base_chain_parser.go defaultApiContainer), and sending that upstream would be
+// a method no node has. It is unreachable from the single current caller — a
+// "Default-" name matches no directive's ApiName, so such a message is never
+// classified UNSUBSCRIBE and never arrives here — but the guard belongs to this
+// function rather than to its caller's invariant, since the doc above invites
+// future callers.
+//
+// getUnsubscribeMethod remains the fallback for those shapes.
 func resolveUnsubscribeMethod(protocolMessage chainlib.ProtocolMessage, subscribeMethod string) string {
 	if protocolMessage != nil {
-		if api := protocolMessage.GetApi(); api != nil && api.Name != "" {
+		if api := protocolMessage.GetApi(); api != nil && api.Name != "" &&
+			!strings.HasPrefix(api.Name, chainlib.DefaultApiName) {
 			return api.Name
 		}
 	}
--- a/protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
+++ b/protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
@@ -30,6 +30,7 @@ type mockSubscriptionServer struct {
 	server          *httptest.Server
 	upgrader        websocket.Upgrader
 	subscriptions   map[string]chan struct{} // subscription ID -> close channel
+	observedMethods []string                 // every JSON-RPC method the server was actually asked for
 	lock            sync.RWMutex
 	writeMu         sync.Mutex // serializes conn.WriteMessage — gorilla requires a single concurrent writer
 	messageInterval time.Duration
@@ -78,30 +79,19 @@ func (ms *mockSubscriptionServer) handleWS(w http.ResponseWriter, r *http.Reques
 			continue
 		}
 
-		// Handle subscription requests
-		if strings.HasSuffix(req.Method, "_subscribe") || strings.HasPrefix(req.Method, "eth_subscribe") {
-			subID := ms.createSubscription()
+		ms.recordMethod(req.Method)
 
-			// Send subscription confirmation
-			response := map[string]interface{}{
-				"jsonrpc": "2.0",
-				"id":      req.ID,
-				"result":  subID,
-			}
-			respBytes, _ := json.Marshal(response)
-			ms.safeWriteMessage(conn, websocket.TextMessage, respBytes)
-
-			// Start sending subscription messages
-			go ms.sendSubscriptionMessages(conn, subID)
-		} else if strings.HasSuffix(req.Method, "_unsubscribe") || strings.HasPrefix(req.Method, "eth_unsubscribe") {
-			// Get subscription ID from params
-			params, ok := req.Params.([]interface{})
-			if ok && len(params) > 0 {
-				if subID, ok := params[0].(string); ok {
-					ms.closeSubscription(subID)
-				}
-			}
-
+		// Classify by whether the caller handed us a LIVE subscription id, never by
+		// the method name. This harness used to gate on "_subscribe"/"_unsubscribe"
+		// suffixes, which is the same assumption that produced MAG-3297: it could
+		// not serve a chain_subscribeNewHeads / chain_unsubscribeNewHeads pair at
+		// all, so no test here could have caught the bug.
+		//
+		// Everything gets a reply. Dropping an unrecognised method would make a
+		// regression hang out the router's 10s CallContext timeout instead of
+		// failing fast.
+		if subID, ok := firstStringParam(req.Params); ok && ms.hasSubscription(subID) {
+			ms.closeSubscription(subID)
 			response := map[string]interface{}{
 				"jsonrpc": "2.0",
 				"id":      req.ID,
@@ -109,8 +99,52 @@ func (ms *mockSubscriptionServer) handleWS(w http.ResponseWriter, r *http.Reques
 			}
 			respBytes, _ := json.Marshal(response)
 			ms.safeWriteMessage(conn, websocket.TextMessage, respBytes)
+			continue
 		}
+
+		subID := ms.createSubscription()
+		response := map[string]interface{}{
+			"jsonrpc": "2.0",
+			"id":      req.ID,
+			"result":  subID,
+		}
+		respBytes, _ := json.Marshal(response)
+		ms.safeWriteMessage(conn, websocket.TextMessage, respBytes)
+		go ms.sendSubscriptionMessages(conn, subID)
 	}
+}
+
+// firstStringParam returns params[0] when it is a string, which is the shape every
+// unsubscribe in every spec uses to name the subscription being torn down.
+func firstStringParam(params interface{}) (string, bool) {
+	list, ok := params.([]interface{})
+	if !ok || len(list) == 0 {
+		return "", false
+	}
+	s, ok := list[0].(string)
+	return s, ok
+}
+
+func (ms *mockSubscriptionServer) recordMethod(method string) {
+	ms.lock.Lock()
+	defer ms.lock.Unlock()
+	ms.observedMethods = append(ms.observedMethods, method)
+}
+
+// ObservedMethods returns the JSON-RPC methods the upstream was actually asked
+// for, which is the thing MAG-3297 got wrong and the only place a regression at
+// the CallContext call site is visible.
+func (ms *mockSubscriptionServer) ObservedMethods() []string {
+	ms.lock.RLock()
+	defer ms.lock.RUnlock()
+	return append([]string(nil), ms.observedMethods...)
+}
+
+func (ms *mockSubscriptionServer) hasSubscription(subID string) bool {
+	ms.lock.RLock()
+	defer ms.lock.RUnlock()
+	_, ok := ms.subscriptions[subID]
+	return ok
 }
 
 func (ms *mockSubscriptionServer) createSubscription() string {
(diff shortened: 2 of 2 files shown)


AnnaR-prog:
### Verified live against the real upstreams

Built from this branch (`docker/Dockerfile`) and run against `sdn.json` from lava-specs #129, with `ghcr.io/magma-devs/smart-router:main` — the image the spec pipeline itself boots — as the control. Same spec dir, same config, same upstreams (`rpc.shiden.astar.network` + `wss://`).

Each pair: subscribe over the router's WebSocket, take the router-issued subscription id, then call the spec's paired unsubscribe with it.

| Unsubscribe method | `:main` control | this branch |
|---|---|---|
| `chain_unsubscribeNewHeads` | `-32601` | `true` |
| `chain_unsubscribeAllHeads` | `-32601` | `true` |
| `chain_unsubscribeFinalizedHeads` | `-32601` | `true` |
| `chain_unsubscribeFinalisedHeads` | `-32601` | `true` |
| `chain_unsubscribeNewHead` | `-32601` | `true` |
| `chain_unsubscribeRuntimeVersion` | `-32601` | `true` |
| `state_unsubscribeRuntimeVersion` | `-32601` | `true` |
| `state_unsubscribeStorage` | `-32601` | `true` |
| `unsubscribe_newHead` | `-32601` | `true` |
| `chainHead_v1_unfollow` | `-32601` | `true` |

**10/10**, which is exactly the "10 of 12 on Shiden" the ticket counted — the two not listed (`author_unwatchExtrinsic`, `transactionWatch_v1_unwatch`) need a live extrinsic to subscribe to and were untestable, not failing.

The subscribe half returned a router-issued id in every case on both images, so this isolates the teardown path alone.

AnnaR-prog:
### Acala cross-check (the ticket's other affected chain)

Same method against `aca.json` (lava-specs #121), Substrate upstream `acala-rpc.aca-api.network`. **Noisier than Shiden** — that upstream is slow and several subscriptions never emitted within the 25s window, on both images and not on the same methods each run, so the timeouts are upstream flakiness rather than verdicts.

Restricting to the pairs that answered on **both** images:

| Unsubscribe method | `:main` control | this branch |
|---|---|---|
| `chain_unsubscribeAllHeads` | `-32601` | `true` |
| `chain_unsubscribeFinalizedHeads` | `-32601` | `true` |
| `chain_unsubscribeNewHead` | `-32601` | `true` |

Control answered `-32601` on 4 methods and timed out on 6; this branch answered `true` on 4 and timed out on 6. **Zero methods regressed, and every method that produced a verdict on this branch produced `true`.**

Shiden (above) is the clean run — 10/10, no timeouts — so it carries the weight here; Acala is directional confirmation on a flaky upstream.

protocol/rpcsmartrouter/direct_ws_subscription_manager.go
```diff
@@ -981,8 +982,42 @@ func (dwsm *DirectWSSubscriptionManager) Unsubscribe(
...
+// method the node does not have and forwarded its -32601 to the client. Several
+// pairs are not derivable by any string rule at all: chainHead_v1_follow tears
+// down with chainHead_v1_unfollow, author_submitAndWatchExtrinsic with
+// author_unwatchExtrinsic.
+//
+// getUnsubscribeMethod remains the fallback for a message with no api attached,
+// which a spec-resolved message never is.
+func resolveUnsubscribeMethod(protocolMessage chainlib.ProtocolMessage, subscribeMethod string) string {
+	if protocolMessage != nil {
+		if api := protocolMessage.GetApi(); api != nil && api.Name != "" {
```
avitenzer: **The guard here is inverted** — it protects the case that can't happen and not the one that produces a garbage method name.

`base_chain_parser.go:414-417`: a method *not* declared in the spec gets a synthetic api with `Name: DefaultApiName + apiKey.Name`, i.e. `"Default-chain_unsubscribeNewHeads"`. This check only tests `api.Name != ""`, so on such a message it would return `Default-…` as the method to send upstream.

That's unreachable today, and for a satisfying reason: a `Default-`-prefixed name can't match any directive's `ApiName`, so the message is never classified `UNSUBSCRIBE` and never dispatches here. But the flip side is that the conditions actually guarded — nil api, empty name — are unreachable for a real `ProtocolMessage` (`GetApi()` is never nil, `Name` is never empty after a successful parse). So the function currently guards nothing reachable, while the one input shape that would yield a bogus upstream method passes straight through.

This matters because the doc comment invites future callers to prefer `resolveUnsubscribeMethod` — at which point its safety stops being the caller's invariant and starts being its own:

```go
if api := protocolMessage.GetApi(); api != nil &&
    api.Name != "" && !strings.HasPrefix(api.Name, chainlib.DefaultApiName) {
    return api.Name
}
```

(`strings` isn't imported in this file yet, so it's a two-line change rather than a one-line one. `chainlib.DefaultApiName` is already exported and `chainlib` is already imported.)

protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
```diff
@@ -2486,3 +2486,97 @@ func TestUnsubscribe_StillRejectsForeignSubscription(t *testing.T) {
...
+	{"state_subscribeStorage", "state_unsubscribeStorage"},
+	{"subscribe_newHead", "unsubscribe_newHead"},
+	{"transactionWatch_v1_submitAndWatch", "transactionWatch_v1_unwatch"},
+}
+
+// TestResolveUnsubscribeMethod_UsesClientMethodName asserts the method the router
+// sends UPSTREAM, which is the thing MAG-3297 got wrong. Asserting only that an
+// unsubscribe does not error would pass against a node that happens to accept the
+// generic "unsubscribe".
+func TestResolveUnsubscribeMethod_UsesClientMethodName(t *testing.T) {
```
avitenzer, line 2559: **These don't assert what the PR body says they assert.** The description says "the tests assert the method sent **upstream**" — they assert the return value of `resolveUnsubscribeMethod`. Nothing exercises `Unsubscribe` → `client.CallContext`, so no test here would catch a regression at the call site (:917) — e.g. someone reverting that one line back to `getUnsubscribeMethod`. And since `mockWSProtocolMessage.GetApi()` returns `&spectypes.Api{Name: m.method}` (:193), the first assertion reduces to `m.method == m.method`; the `NotEqual` guard is what carries the real weight.

The sharper point is that this file *already has* a real WS server — `mockSubscriptionServer` (:47-114) — but its handler gates on the same string assumption that caused the bug:

```go
if strings.HasSuffix(req.Method, "_subscribe") || strings.HasPrefix(req.Method, "eth_subscribe")   // :82
} else if strings.HasSuffix(req.Method, "_unsubscribe") || ...                                     // :96
```

So the harness cannot serve a `chain_subscribeNewHeads` / `chain_unsubscribeNewHeads` pair at all — it silently drops both. Generalising it (respond to any method; record `req.Method` into a slice under `ms.lock`) unlocks the test that would actually have failed before this fix: subscribe with a Substrate pair, unsubscribe, assert the server *observed* `chain_unsubscribeNewHeads`. Roughly 20 lines, and it retires the last place in this package still encoding the assumption MAG-3297 was about.

One caveat if you write it: because the mock drops unrecognised methods rather than erroring, a naive version of that test against the *old* code would hang out the 10s `CallContext` timeout rather than failing fast — so have the generalised handler reply to everything.

protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
```diff
@@ -2486,3 +2486,97 @@ func TestUnsubscribe_StillRejectsForeignSubscription(t *testing.T) {
...
 		"unrelated client must not be able to unsubscribe via upstream-id fallback")
 }
+
+// substrateUnsubscribePairs is the full SUBSCRIBE/UNSUBSCRIBE set declared by the
+// ASTAR spec's base collection, and the same shape Acala (aca.json) and Shiden
+// (sdn.json) carry. Every Substrate spec pairs by name, and none of the pairings
+// is a "<x>_subscribe" -> "<x>_unsubscribe" rewrite: three of them
+// (chainHead_v1_follow, author_submitAndWatchExtrinsic,
+// transactionWatch_v1_submitAndWatch) do not contain "subscribe" at all.
+var substrateUnsubscribePairs = []struct {
```
avitenzer, line 2537: All twelve rows drive the identical one-line pass-through, so this is documentation rather than coverage breadth — worth not counting it as the latter. It also can't be validated in-repo: there's no Substrate spec under `specs/` (they live in lava-specs), so the table can drift silently from the spec it claims to mirror.

Either cite the lava-specs file it was transcribed from (so a reader can check it), or trim to the three non-derivable pairs — `chainHead_v1_follow`, `author_submitAndWatchExtrinsic`, `transactionWatch_v1_submitAndWatch` — which are the ones that carry the argument that no string rule could have worked. Keeping all twelve is fine too; this is a framing nit.

protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
```diff
@@ -2486,3 +2486,97 @@ func TestUnsubscribe_StillRejectsForeignSubscription(t *testing.T) {
...
+
+// mockWSProtocolMessageNoApi is a message that resolved to no spec api — the only
+// case the derivation fallback is still reachable from.
+type mockWSProtocolMessageNoApi struct {
+	mockWSProtocolMessage
+}
+
+func (m *mockWSProtocolMessageNoApi) GetApi() *spectypes.Api { return nil }
+
+func TestResolveUnsubscribeMethod_FallsBackWithoutApi(t *testing.T) {
```
avitenzer, line 2608: Follows from the `resolveUnsubscribeMethod` comment: all three subtests here pin behaviour that a real `ProtocolMessage` can't produce — `GetApi()` is never nil and `Name` is never empty after a successful parse, and the `nil` message case has no caller. Not harmful, but if you take the `Default-` prefix suggestion, a fourth row for `"Default-chain_unsubscribeNewHeads"` would be the one case in this test that corresponds to a shape the parser can actually hand you.

avitenzer (commented):
## Review

I traced the fix through the dispatch path, checked for second copies of the bug, and built/tested the committed tree at `afe0c70`.

**Verdict: correct, minimal, well-targeted.** Two non-blocking suggestions inline. The diagnosis holds up under checking — the fix's correctness rests on an invariant enforced two files away, and that invariant is real: a message only reaches `Unsubscribe` if `GetParseDirective` (`chain_message_queries.go:52`) matched `api.Name` against a directive's `ApiName`, and `serverApis` is keyed by the exact api name with no case folding or aliasing. So `api.Name` *is* the wire method verbatim. "Use the client's method" is reading a spec-validated name, not trusting client input.

### What I verified

| Claim | Result |
|---|---|
| `Unsubscribe` has one caller | ✅ only `consumer_websocket_manager.go:301`, gated on `FUNCTION_TAG_UNSUBSCRIBE` |
| `api.Name` == the client's wire method | ✅ `serverApis` keyed by exact api name |
| `getUnsubscribeMethod` has no other callers | ✅ confirmed repo-wide |
| No second copy of the derivation | ✅ `ClientSubscription.Unsubscribe()` (`subscription.go:261`) only signals `quit`; `run()` discards `forward()`'s `unsubscribeServer` flag, so this fork never sends an upstream teardown from there. `UnsubscribeAll` (:1090) likewise. The `CallContext` at :937 is the only wire teardown — so this really is a single-point fix. |
| Batch requests can't slip through | ✅ batch api names are joined with `SEP`, so `GetParseDirective` never matches and the message routes as a normal relay |
| "-32601 forwarded verbatim" | ✅ this fork's `CallContext` (`client.go:317-322`) does *not* convert a JSON-RPC error object into a Go error, unlike geth's — so the error body was marshalled into `nodeResp` and streamed on, and local cleanup still ran. Description is accurate. |
| Build / vet / tests | ✅ `go build ./...`, `go vet`, and the full `./protocol/rpcsmartrouter/` package (90s) green |

### Adjacent, pre-existing, out of scope

`extractSubscriptionIDFromUnsubscribe` (:1155) accepts only `params[0].(string)`. A numeric subscription id unmarshals to `float64` and fails *before* method resolution is reached. Substrate's usual id provider returns strings so this probably isn't biting, but it's the next thing to fail if Substrate acceptance testing goes past this fix. Not introduced here — flagging as a follow-up, not a change request.

### Security / performance

Worth stating explicitly since "send the method the client asked for" *sounds* like method injection: it isn't. Only names matching an `UNSUBSCRIBE` parse directive reach this function, so the value is drawn from a closed, spec-declared set. The residual is that a client may unsubscribe a `chain_*` subscription using, say, `state_unsubscribeStorage`; the node returns an error object, the router forwards it, and local cleanup proceeds regardless — same outcome class as before the fix, no new surface. Performance impact is one nil check per unsubscribe.

AnnaR-prog:
Thanks — this is a genuinely useful review, and I agree with all four points. All addressed in `a6ce71f`.

**The inverted guard.** You're right, and the reasoning is the part I'd missed: the conditions I guarded (`nil` api, empty name) are unreachable for a real `ProtocolMessage`, while `Default-<method>` — the one shape the parser *can* hand you — went straight through. Added the `DefaultApiName` prefix check, with the `strings` import. I kept your framing in the comment: unreachable from today's caller because a `Default-` name matches no directive's `ApiName`, but the guard belongs to the function rather than to its caller's invariant, precisely because the doc comment invites future callers.

**The tests didn't assert what I said they did.** This is the one I'm most glad you caught. You're right that the description overclaimed: I verified the upstream method against live Shiden and Acala nodes (tables above), but nothing *in the repo* did, so reverting the one line at the call site would have left CI green.

Fixed the way you suggested. `mockSubscriptionServer` now classifies by whether the caller handed it a **live subscription id**, never by the method name, records every method it was asked for, and replies to everything — including your caveat about dropping unrecognised methods causing a 10s `CallContext` hang instead of a fast failure.

That unlocked `TestEndToEnd_UnsubscribeSendsTheSpecMethodUpstream`: subscribe `chain_subscribeNewHeads`, unsubscribe `chain_unsubscribeNewHeads`, assert the server *observed* the spec's method. Verified by reverting the call site — it fails with:

```
observed=[chain_subscribeNewHeads unsubscribe]
```

naming the literal method the old derivation sent. That is exactly the regression that was previously invisible to CI.

**The twelve-row table.** Fair — it's documentation, not coverage breadth, and it can drift silently since the specs live in lava-specs. I kept all twelve but now name `astar.json` as the source so a reader can check it, and say plainly in the comment that the three non-derivable pairs are the ones carrying the argument.

**The unreachable fallback subtests.** Added the fourth row you suggested — `Default-chain_unsubscribeNewHeads` — which is the one input in that test the parser can actually produce.

**On `extractSubscriptionIDFromUnsubscribe` accepting only `params[0].(string)`:** agreed it's out of scope here and agreed it's the next thing to fail. Substrate's usual id provider returns strings, which matches what I saw live — the ids came back as strings on both Shiden and Acala — so it isn't biting today. Worth its own ticket rather than widening this PR; say the word and I'll file it.

protocol/rpcsmartrouter/direct_ws_subscription_manager.go
AnnaR-prog: Fixed in `a6ce71f`. Agreed, and the reasoning is the part I had missed: the shapes I guarded (nil api, empty name) are unreachable for a real `ProtocolMessage`, while `Default-<method>` — the one the parser can actually produce — went straight through.

```go
if api := protocolMessage.GetApi(); api != nil && api.Name != "" &&
    !strings.HasPrefix(api.Name, chainlib.DefaultApiName) {
    return api.Name
}
```

`strings` added to the imports. I carried your framing into the comment: unreachable from today's caller because a `Default-` name matches no directive's `ApiName`, but the guard belongs to this function rather than to its caller's invariant, precisely because the doc comment invites future callers.

protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
AnnaR-prog: Fixed in `a6ce71f` — this is the one I am most glad you caught. You are right that the description overclaimed: I verified the upstream method against live Shiden and Acala nodes, but nothing in the repo did, so reverting the one line at the call site would have left CI green.

Took your suggestion. `mockSubscriptionServer` now classifies by whether the caller handed it a **live subscription id**, never by the method name, records every method it was asked for (`ObservedMethods()`), and replies to everything — including your caveat that dropping an unrecognised method would hang the router's 10s `CallContext` instead of failing fast.

That unlocked `TestEndToEnd_UnsubscribeSendsTheSpecMethodUpstream`: subscribe `chain_subscribeNewHeads`, unsubscribe `chain_unsubscribeNewHeads`, assert the server observed the spec's method. Verified by reverting the call site — it fails with

```
observed=[chain_subscribeNewHeads unsubscribe]
```

naming the literal method the old derivation sent. That is exactly the regression that was previously invisible to CI.

protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
AnnaR-prog: Fixed in `a6ce71f`. Fair — it is documentation, not coverage breadth, and it can drift silently since the specs live in lava-specs rather than under `specs/` here.

Kept all twelve but the comment now names `astar.json` as the file it was transcribed from so a reader can check it, says plainly that every row drives the same one-line pass-through, and calls out the three rows that carry the argument (`chainHead_v1_follow`, `author_submitAndWatchExtrinsic`, `transactionWatch_v1_submitAndWatch`) as the ones no string rule could have handled.

protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
AnnaR-prog: Fixed in `a6ce71f`. Added the fourth row you suggested — `Default-chain_unsubscribeNewHeads` — which is the one input in this test that corresponds to a shape the parser can actually hand you. The other three stay, with the comment now saying outright that they pin shapes a real `ProtocolMessage` cannot produce.

avitenzer (approved):

avitenzer:
need to rerun WF

PR #345, opened 2026-08-30
fix(chainlib): admit a provider serving only an add-on collection (MAG-3296)

AnnaR-prog:
# Description

A provider serving only an add-on collection could not be admitted. Acala splits its two api surfaces across collections — Substrate in the base one, EVM in an `evm` add-on — and an EVM-only node was excluded outright at admission, before a single one of the verifications it *could* pass had run. Not Acala-specific: Litentry/Heima (`lit.json`) and peaq (`peaq.json`) carry the same two-collection shape.

The ticket inferred the cause from observed behaviour and asked for it to be confirmed against the source. It is confirmed, and there were **five** gates rather than one. Gates 1-3 are the same root cause in different places; gates 4 and 5 are the same modelling assumption on the verification and routing sides.

`taggedApis` holds exactly **one** parse directive per function tag — the first collection to declare it, which for a spec written base-first is the base collection. So `GetParsingByTag` answers *"the base collection's directive"* no matter who is asking.

| | Gate | Fix |
|---|---|---|
| 1 | Both head probes went through `GetParsingByTag`, so every node was asked the base collection's `GET_BLOCKNUM`. An EVM-only Acala node cannot answer `chain_getHeader`, returns `-32601`, and `Validate` returns on a failed head probe **before** reaching the verification loop. | `GetParsingByTagForCollection` resolves from the collections a node actually serves, falling back to the base answer when it declares no add-ons or none carries the tag — so nothing changes for any existing chain. Wired into `ChainFetcher`'s probe (admission + `smartrouter health`) **and** `EndpointPoller`'s (live per-endpoint tip + fork check). |
| 2 | The re-verifier validated the addon-stripped URL copies it creates for the chain router, asking an EVM-only node to be a Substrate node. `Validate` returns on the first failing url, so the copy's verdict was the provider's verdict. | The expanded list goes to the router only; the fetcher validates the urls the provider actually declared. |
| 3 | A verification that names a tag instead of carrying its own template borrowed its directive from `taggedApis`. Acala's `evm` `pruning` is declared as nothing but `{"function_tag": "GET_EARLIEST_BLOCK"}`, so the one check meant to validate an EVM node's history was rewritten into a Substrate call. | Borrow from the verification's own collection first, `taggedApis` as fallback so the ordinary case (e.g. ASTAR's `txpool`) resolves exactly as before. |
| 4 | `GetVerifications` appends the empty addon unconditionally, so every node runs the base collection's verifications too. Acala's base `chain-id` fires `chain_getBlockHash` at an EVM node. **Every `severity` in `aca.json` is omitted and an omitted severity is `Fail`** (`ParseValue_Fail = 0`), so one inherited verification excluded the provider. | New per-node-url `standalone-addons: true`. |
| 5 | Admitting the provider was only half the job. `CheckSupportForServices` filtered only on a non-empty addon and `IsSupportingAddon` returned true outright for `""`, so an EVM-only endpoint became eligible for **Substrate** traffic and answered `-32601` — which a client reads as a verdict on its request. Found by running a mixed deployment against a router built from this branch. | `Endpoint` carries the url's opt-out; both checks ask `ServesBaseCollection()`. A provider still serves base while **one** of its endpoints is ordinary. |

**Why gate 4 needed a config flag rather than existing config.** `skip-verifications` cannot express the exception: it matches on the verification **name alone**, and both Acala collections name theirs `chain-id`, so suppressing the Substrate check also suppresses the `evm` `0x313` one — the only check that actually validates the node. `standalone-addons: true` says "this url serves only the collections its `addons` name". Default false, so every existing deployment is untouched; ignored on a url with no addons. An add-on can mean *extends the base surface* (archive) or *replaces it* (Acala/Litentry/peaq) and the spec model cannot currently tell them apart — only an operator knows which a given url is. When the spec model can say it, this flag becomes redundant rather than wrong.

The gate-4 filter runs **before** `needsLatestBlock`, for the same reason the `skip-verifications` filter already does: a verification that is not going to run must not drag a head probe out to the upstream on its way past.

### The result, traced against the real `aca.json` (lava-specs #121)

A node configured `addons: ["evm"], standalone-addons: true` runs `evm` `chain-id` (`eth_chainId` → `0x313`) and `evm` `pruning` (`eth_getBlockByNumber` at head−7200), probes its head with `eth_blockNumber`, and is admitted. The base Substrate `chain-id` is dropped rather than failed.

### Most critical to review

- `protocol/chainlib/base_chain_parser.go` — `GetParsingByTagForCollection` (add-ons tried in the node's own declaration order; a keyed lookup, not a scan of `apiCollections`, which is a map and would iterate randomly), and the verification-borrow change.
- `protocol/common/endpoints.go` — the `standalone-addons` field and `ServesBaseCollection()`.
- `protocol/chainlib/chain_fetcher.go` — `verificationsForNodeUrl` and its placement relative to `needsLatestBlock`.

### Scope — what this does not cover

- **Relay routing for a mixed Substrate + EVM-only deployment is not traced.** Admission and tip polling are verified end to end; whether base-collection traffic can be routed to an admitted EVM-only endpoint is not. `GetChainRouter` constructs fine (`populateRequiredForAddon` recurses over extensions only and never requires a base-addon route), but that is a different question from relay selection.
- Admission stays all-or-nothing per provider rather than per collection.

### Verified live against the real upstreams

Built from this branch (`docker/Dockerfile`) and run against `aca.json` from lava-specs #121, with `ghcr.io/magma-devs/smart-router:main` — the image the spec pipeline itself boots — as the control. Same spec dir, same config, one variable at a time.

**Isolated EVM-only provider** (`eth-rpc-acala.aca-api.network`, `addons: ["evm"]`):

| Run | Log signature | Outcome |
|---|---|---|
| `:main` control | `chain_getHeader` ×30, `GET_BLOCKNUM Failed` ×30, `failed to fetch latest block number` ×6 | `excluding from provider list` |
| this branch, **no** flag | head probe silent; `chain_getBlockHash` ×36, `invalid Verification on provider startup` ×12 | `excluding from provider list` |
| this branch, `standalone-addons: true` | `skipping base-collection verification` ×2 | **admitted, serving** |

The middle row is the point: gates 1-3 move the failure off the head probe entirely, and what is left is exactly the inherited base `chain-id` that gate 4 addresses.

**Mixed deployment** — one router, a Substrate provider and the EVM-only provider:

| Request | `:main` control | this branch |
|---|---|---|
| `eth_chainId` | `No Providers For Addon {addon:evm}` | `0x313` |
| `eth_blockNumber` | `No Providers For Addon {addon:evm}` | `0xb571cb` |
| `chain_getHeader` | real Substrate header | real Substrate header |
| `system_chain` | `"Acala"` | `"Acala"` |

Both surfaces served from one listener, 0 exclusions. This is the run that found gate 5: before that commit, `chain_getHeader` on this branch returned `-32601` because Substrate traffic was reaching the EVM endpoint.

**One residual, non-fatal and not fixed here.** `RPCSmartRouterServer.craftRelay` builds the startup warm-up relay from the base collection too, so a disjoint-addon provider logs a handful of `-32601` node errors at boot. Measured: a one-time burst, stable across 60s, provider never excluded, tip polling correct and advancing. It is the same root cause and wants the per-collection admission work the ticket calls step 3 — worth its own ticket rather than growing this PR.

### Tests

Each fix has a guard **verified by breaking it**, not by watching it pass:
- reverting the `Validate` call site makes the probe test fail with `GetParsingByTagForCollection([GET_BLOCKNUM [] ])` where `[evm]` was expected;
- reverting the verification borrow makes the evm directive resolve to `chain_getBlockHash`.

Also covered: determinism across 200 calls and add-on declaration order (the map-iteration hazard that would otherwise have been the follow-up bug), disabled collections, internal paths, extension names mixed into `addons`, and a config round-trip through `viper.UnmarshalKey` — a mapstructure tag that did not match would decode as false and drop the provider with no error anywhere.

### Two things worth flagging

- `ServesBaseCollection()` guards on `len(Addons) == 0`, but `Addons` mixes addons and extensions — so `standalone-addons: true` on a url whose `addons` are all extension names drops every verification. Visible rather than silent: the filter sits above the existing `no verifications for url` warning.
- `chainlib_mock.go` carries `DO NOT EDIT`; a MockGen regeneration reproduces the hand-added `GetParsingByTagForCollection` verbatim now that the interface declares it.

## Jira ticket

Jira ticket: MAG-3296

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change — n/a; `ChainParser` gains a method (implementors updated, mock included), config field is additive and defaults to today's behaviour

### DIFF
--- a/protocol/chainlib/base_chain_parser.go
+++ b/protocol/chainlib/base_chain_parser.go
@@ -349,6 +349,62 @@ func (bcp *BaseChainParser) GetParsingByTag(tag spectypes.FUNCTION_TAG) (parsing
 	return val.Parsing, val.ApiCollection, ok
 }
 
+// GetParsingByTagForCollection resolves a tagged parse directive for a node that
+// serves specific add-on collections, falling back to GetParsingByTag's answer
+// when the node declares no add-ons or none of its collections carries the tag.
+//
+// MAG-3296: taggedApis holds exactly ONE directive per tag — the first collection
+// to declare it, which for a spec written base-first is the base collection. So
+// GetParsingByTag answers "the base collection's directive" no matter who is
+// asking. That is wrong for a spec whose add-on is a DISJOINT api surface rather
+// than a superset of the base one: Acala serves Substrate in the base collection
+// and EVM in an `evm` add-on, and an EVM-only node cannot answer the base
+// collection's chain_getHeader at all. Probing it with one gets -32601 and, on
+// the admission path, costs the provider its place outright.
+//
+// Add-ons are tried in the order the node declares them, so a node's own
+// preference decides between two collections that both carry the tag. The
+// connection type comes from the base answer, because the tag is served over the
+// same transport whichever collection defines it — this is a keyed lookup rather
+// than a scan of apiCollections, which is a map and would iterate randomly.
+//
+// `addons` may contain extension names too (callers pass NodeUrl.Addons, which
+// mixes both). An extension name matches no collection's AddOn, so it is skipped
+// rather than needing to be separated out first.
+func (bcp *BaseChainParser) GetParsingByTagForCollection(tag spectypes.FUNCTION_TAG, addons []string, internalPath string) (parsing *spectypes.ParseDirective, apiCollection *spectypes.ApiCollection, existed bool) {
+	bcp.rwLock.RLock()
+	defer bcp.rwLock.RUnlock()
+
+	base, baseExisted := bcp.taggedApis[tag]
+	if baseExisted {
+		parsing, apiCollection, existed = base.Parsing, base.ApiCollection, true
+	}
+	if len(addons) == 0 || !baseExisted {
+		return parsing, apiCollection, existed
+	}
+
+	for _, addon := range addons {
+		if addon == "" {
+			continue // the base collection, which is already the fallback
+		}
+		collection, ok := bcp.apiCollections[CollectionKey{
+			ConnectionType: base.ApiCollection.CollectionData.Type,
+			InternalPath:   internalPath,
+			Addon:          addon,
+		}]
+		if !ok || !collection.Enabled {
+			continue
+		}
+		for _, directive := range collection.ParseDirectives {
+			if directive.FunctionTag == tag {
+				return directive, collection, true
+			}
+		}
+	}
+
+	return parsing, apiCollection, existed
+}
+
 func (bcp *BaseChainParser) IsTagInCollection(tag spectypes.FUNCTION_TAG, collectionKey CollectionKey) bool {
 	bcp.rwLock.RLock()
 	defer bcp.rwLock.RUnlock()
--- a/protocol/chainlib/base_chain_parser.go
+++ b/protocol/chainlib/base_chain_parser.go
@@ -371,24 +371,37 @@ func (bcp *BaseChainParser) GetParsingByTag(tag spectypes.FUNCTION_TAG) (parsing
 // `addons` may contain extension names too (callers pass NodeUrl.Addons, which
 // mixes both). An extension name matches no collection's AddOn, so it is skipped
 // rather than needing to be separated out first.
-func (bcp *BaseChainParser) GetParsingByTagForCollection(tag spectypes.FUNCTION_TAG, addons []string, internalPath string) (parsing *spectypes.ParseDirective, apiCollection *spectypes.ApiCollection, existed bool) {
+//
+// allowBaseFallback is the caller's ServesBaseCollection(): false means the node
+// serves ONLY its add-on collections, and the tagged fallback is then refused
+// rather than handing back a directive that node cannot answer.
+func (bcp *BaseChainParser) GetParsingByTagForCollection(tag spectypes.FUNCTION_TAG, addons []string, internalPath string, allowBaseFallback bool) (parsing *spectypes.ParseDirective, apiCollection *spectypes.ApiCollection, existed bool) {
 	bcp.rwLock.RLock()
 	defer bcp.rwLock.RUnlock()
 
-	base, baseExisted := bcp.taggedApis[tag]
-	if baseExisted {
-		parsing, apiCollection, existed = base.Parsing, base.ApiCollection, true
+	// `tagged`, not `base`: taggedApis is first-wins in SPEC-FILE order, so for a
+	// spec that lists an add-on collection before the base one this entry is the
+	// add-on's. The name must not assert an invariant the map does not enforce.
+	tagged, taggedExisted := bcp.taggedApis[tag]
+	if taggedExisted {
+		parsing, apiCollection, existed = tagged.Parsing, tagged.ApiCollection, true
 	}
-	if len(addons) == 0 || !baseExisted {
+	if len(addons) == 0 || !taggedExisted {
+		// taggedApis is populated from EVERY enabled collection, so a tag missing
+		// from it is declared nowhere and searching the add-on collections cannot
+		// find it either.
+		if !taggedExisted {
+			return nil, nil, false
+		}
 		return parsing, apiCollection, existed
 	}
 
 	for _, addon := range addons {
 		if addon == "" {
-			continue // the base collection, which is already the fallback
+			continue // the base collection, which is the fallback below
 		}
 		collection, ok := bcp.apiCollections[CollectionKey{
-			ConnectionType: base.ApiCollection.CollectionData.Type,
+			ConnectionType: tagged.ApiCollection.CollectionData.Type,
 			InternalPath:   internalPath,
 			Addon:          addon,
 		}]
@@ -402,6 +415,16 @@ func (bcp *BaseChainParser) GetParsingByTagForCollection(tag spectypes.FUNCTION_
 		}
 	}
 
+	if !allowBaseFallback {
+		// The caller serves only its add-on collections, and none of them declares
+		// this tag. Falling back would hand it the one directive the operator said
+		// the node cannot answer — reinstating the very probe standalone-addons
+		// opted out of, quietly. The keyed lookup above misses whenever a spec
+		// declares the add-on collection only at the root and the url carries an
+		// internal path, so this is reachable the moment a spec combines a disjoint
+		// add-on with internal paths.
+		return nil, nil, false
+	}
 	return parsing, apiCollection, existed
 }
 
--- a/protocol/chainlib/base_chain_parser.go
+++ b/protocol/chainlib/base_chain_parser.go
@@ -629,6 +629,18 @@ func (apip *BaseChainParser) getApiCollection(connectionType, internalPath, addo
 	return api, nil
 }
 
+// findParseDirectiveByTag returns a collection's own directive for a tag, or nil
+// when it declares none. Used to resolve what a tag-referencing verification
+// borrows, so it borrows from the collection it belongs to.
+func findParseDirectiveByTag(apiCollection *spectypes.ApiCollection, tag spectypes.FUNCTION_TAG) *spectypes.ParseDirective {
+	for _, directive := range apiCollection.ParseDirectives {
+		if directive.FunctionTag == tag {
+			return directive
+		}
+	}
+	return nil
+}
+
 func getServiceApis(
 	spec spectypes.Spec,
 	rpcInterface string,
@@ -740,8 +752,26 @@ func getServiceApis(
 			}
 			for _, verification := range apiCollection.Verifications {
 				if verification.ParseDirective.FunctionTag != spectypes.FUNCTION_TAG_VERIFICATION {
-					if _, ok := taggedApis[verification.ParseDirective.FunctionTag]; ok {
-						verification.ParseDirective = taggedApis[verification.ParseDirective.FunctionTag].Parsing
+					// A verification that names a tag instead of carrying its own
+					// template borrows a directive. Take the one from ITS OWN
+					// collection first (MAG-3296): taggedApis holds a single
+					// directive per tag — the first collection to declare it — so
+					// borrowing from there gave every collection the base one.
+					// Acala's `evm` pruning verification names GET_EARLIEST_BLOCK
+					// and was rewritten to probe with the base collection's
+					// Substrate chain_getBlockHash, which the EVM node it was
+					// meant to check cannot answer.
+					//
+					// taggedApis stays the fallback, so a collection that borrows a
+					// tag it does not itself declare still resolves exactly as before.
+					borrowed := findParseDirectiveByTag(apiCollection, verification.ParseDirective.FunctionTag)
+					if borrowed == nil {
+						if container, ok := taggedApis[verification.ParseDirective.FunctionTag]; ok {
+							borrowed = container.Parsing
+						}
+					}
+					if borrowed != nil {
+						verification.ParseDirective = borrowed
 					} else {
 						utils.LavaFormatError("Bad verification definition", fmt.Errorf("verification function tag is not defined in the collections parse directives"), utils.LogAttr("function_tag", verification.ParseDirective.FunctionTag))
 						continue
--- a/protocol/chainlib/base_chain_parser_addon_tag_test.go
+++ b/protocol/chainlib/base_chain_parser_addon_tag_test.go
@@ -1,0 +1,226 @@
+package chainlib
+
+import (
+	"testing"
+
+	spectypes "github.com/magma-Devs/smart-router/types/spec"
+	"github.com/stretchr/testify/require"
+)
+
+// MAG-3296. A spec whose add-on is a DISJOINT api surface rather than a superset
+// of the base one: Acala serves Substrate in the base collection and EVM in an
+// `evm` add-on, and the two have no method in common. `txpool` is the ordinary
+// shape for contrast — a real add-on (ASTAR has one) that declares no tagged
+// directives of its own and is meant to inherit the base collection's.
+func acalaShapedParser(t *testing.T) (parser *BaseChainParser, base, evm *spectypes.ApiCollection) {
+	t.Helper()
+
+	collectionData := func(addon, internalPath string) spectypes.CollectionData {
+		return spectypes.CollectionData{
+			ApiInterface: spectypes.APIInterfaceJsonRPC,
+			InternalPath: internalPath,
+			Type:         "POST",
+			AddOn:        addon,
+		}
+	}
+
+	base = &spectypes.ApiCollection{
+		Enabled:        true,
+		CollectionData: collectionData("", ""),
+		ParseDirectives: []*spectypes.ParseDirective{
+			{FunctionTag: spectypes.FUNCTION_TAG_GET_BLOCKNUM, ApiName: "chain_getHeader"},
+		},
+	}
+	evm = &spectypes.ApiCollection{
+		Enabled:        true,
+		CollectionData: collectionData("evm", ""),
+		ParseDirectives: []*spectypes.ParseDirective{
+			{FunctionTag: spectypes.FUNCTION_TAG_GET_BLOCKNUM, ApiName: "eth_blockNumber"},
+			{FunctionTag: spectypes.FUNCTION_TAG_GET_BLOCK_BY_NUM, ApiName: "eth_getBlockByNumber"},
+		},
+	}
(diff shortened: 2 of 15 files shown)


protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -2459,10 +2459,14 @@ func (rpsr *RPCSmartRouter) CreateSmartRouterEndpoint(
...
-					Addons:            extensions,
-					Extensions:        extensions,
+					NetworkAddress: url.Url,
+					Enabled:        true,
+					Addons:         extensions,
+					Extensions:     extensions,
+					// Carry the url's opt-out through, so session selection keeps
+					// base-collection traffic off an endpoint that only serves its
+					// add-ons (MAG-3296).
+					StandaloneAddons:  url.StandaloneAddons,
```
avitenzer: **`standalone-addons` on an extension-only url makes the endpoint eligible for *nothing*.** The PR body flags half of this — *"`standalone-addons: true` on a url whose `addons` are all extension names drops every verification. Visible rather than silent"* — but gate 5 makes it worse than that, and not visible.

Thirty-five lines up (`:2434`), **both** `Endpoint.Addons` and `Endpoint.Extensions` are populated from the same raw `url.Addons` list. So for `addons: ["archive"], standalone-addons: true`, I ran it:

```
archive relay (addon="", ext=[archive]) supported = false
plain relay   (addon="", ext=[])        supported = false
provider IsSupportingAddon("")          = false
```

Archive traffic carries `addon=""` with the extension in the separate `extensions` slice, so `CheckSupportForServices` fails the base check before it ever reaches the extension loop. The endpoint passes admission (zero verifications, and a `no verifications for url` warning that says nothing about routing) and is then eligible for no traffic at all. A config typo yields a silently dead endpoint.

The ingredients for a guard are already in scope — `chainParser` is right here at `:2431`, and `SeparateAddonsExtensions` is on the `ChainParser` interface (`chainlib.go:78`):

```go
declaredAddons, _, err := chainParser.SeparateAddonsExtensions(ctx, url.Addons)
if url.StandaloneAddons && (err != nil || len(declaredAddons) == 0) {
    utils.LavaFormatWarning("ignoring standalone-addons: url declares no add-on collection", nil,
        utils.LogAttr("url", url.Url), utils.LogAttr("addons", url.Addons))
    // fall through with StandaloneAddons treated as false
}
```

That turns "endpoint serves nothing, no diagnostic" into a startup warning. `standalone_addons_routing_test.go` already builds its fixture with `Addons` and `Extensions` both `{"evm"}` — the production shape — so an extension-name case slots straight in.

This is the one I'd want addressed before merge; the rest below are smaller.

protocol/chainlib/base_chain_parser.go
```diff
@@ -349,6 +349,62 @@ func (bcp *BaseChainParser) GetParsingByTag(tag spectypes.FUNCTION_TAG) (parsing
...
+			continue
+		}
+		for _, directive := range collection.ParseDirectives {
+			if directive.FunctionTag == tag {
+				return directive, collection, true
+			}
+		}
+	}
+
+	return parsing, apiCollection, existed
```
avitenzer, line 428: **The fallback silently reinstates the probe the url opted out of.** Falling back to the tagged directive when none of the node's add-on collections carries the tag is exactly right for an ordinary url. For a `standalone-addons: true` one it hands back the single directive the operator declared the node cannot answer — reintroducing gate 1 for that url, quietly.

The interaction that makes this more than theoretical is the internal-path expansion: `expandInternalPaths` copies `StandaloneAddons` and `Addons` onto `/v2`-style url copies, and the keyed lookup `{Type, "/v2", "evm"}` misses whenever the spec declares the `evm` collection only at the root. `base_chain_parser_addon_tag_test.go:152` pins precisely that miss as "inherits the base one". No current spec combines a disjoint add-on with internal paths, so this isn't live — it's a trap laid for the next spec that does.

Both call sites already know `url.ServesBaseCollection()` / the endpoint equivalent. Passing it through — or having the caller compare the returned `apiCollection.CollectionData.AddOn` against what it asked for — makes this fail loudly rather than probing a method the node can't answer.

protocol/chainlib/base_chain_parser.go
```diff
@@ -349,6 +349,62 @@ func (bcp *BaseChainParser) GetParsingByTag(tag spectypes.FUNCTION_TAG) (parsing
...
+// than a scan of apiCollections, which is a map and would iterate randomly.
+//
+// `addons` may contain extension names too (callers pass NodeUrl.Addons, which
+// mixes both). An extension name matches no collection's AddOn, so it is skipped
+// rather than needing to be separated out first.
+func (bcp *BaseChainParser) GetParsingByTagForCollection(tag spectypes.FUNCTION_TAG, addons []string, internalPath string) (parsing *spectypes.ParseDirective, apiCollection *spectypes.ApiCollection, existed bool) {
+	bcp.rwLock.RLock()
+	defer bcp.rwLock.RUnlock()
+
+	base, baseExisted := bcp.taggedApis[tag]
```
avitenzer: Doc-accuracy nit: `base` here isn't necessarily the base collection. `taggedApis` is first-wins over collections in **spec-file order** (`:695`), not base-first by construction — so for a spec that lists an add-on collection before the base one, `base.ApiCollection` *is* the add-on's, and both `CollectionData.Type` (used as the lookup key below) and the returned fallback come from it.

Harmless today, and the doc comment is honest about it (*"which for a spec written base-first is the base collection"*). I'd just carry that qualifier into the identifier — `tagged` rather than `base` — so the name doesn't assert an invariant the map doesn't enforce.

protocol/chainlib/base_chain_parser_addon_tag_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+	t.Run("the resolved collection is returned alongside the directive", func(t *testing.T) {
+		_, collection, ok := parser.GetParsingByTagForCollection(
+			spectypes.FUNCTION_TAG_GET_BLOCKNUM, []string{"evm"}, "")
+		require.True(t, ok)
+		require.Same(t, evm, collection,
+			"the caller crafts the probe from CollectionData, so it must be the add-on's")
+	})
+
+	t.Run("a tag no collection declares stays absent", func(t *testing.T) {
```
avitenzer, line 176: This pins a state the constructor can't produce: the fixture's `evm` collection declares `GET_BLOCK_BY_NUM` while `taggedApis` omits it, but `getServiceApis` populates `taggedApis` from every collection (`base_chain_parser.go:695`), so `baseExisted` is always true for a tag any enabled collection declares.

More to the point, the behaviour it locks in is the one case where refusing is wrong for the node this PR is about: the add-on collection *does* declare the directive, the node *can* serve it, and the function declines to return it. The comment's reasoning — *"reporting it as found would hand callers a directive the base surface cannot serve"* — is the right instinct for an ordinary url and backwards for a standalone-addons one. Same root as the fallback comment on `base_chain_parser.go`; if that one gets an opt-out-aware path, this test's expectation is what should flip.

protocol/rpcsmartrouter/spec_reverifier.go
```diff
@@ -497,30 +497,49 @@ func validateProviderTier(
...
+	return newEndpoint(routedNodeUrls), newEndpoint(provider.NodeUrls)
+}
+
+func validateProvider(
+	ctx context.Context,
+	provider *lavasession.RPCStaticProviderEndpoint,
+	chainParser chainlib.ChainParser,
+	timeout time.Duration,
+) error {
+	routerEndpoint, verificationEndpoint := verificationEndpoints(provider)
```
avitenzer: Naming nit: `verificationEndpoints` returns `(routerEndpoint, fetcherEndpoint)` and its doc comment calls the second "the FETCHER one", but it's bound here as `verificationEndpoint`. Two names for one value within 60 lines, and the router/fetcher split is the entire point of the function — worth naming it consistently in both places.

avitenzer (commented):
## Review

Built and tested at the head commit (`3ddf79c`) in a clean worktree, and ran the edge cases rather than reasoning about them.

**Verdict: well-constructed and unusually well-evidenced.** The diagnosis holds up under checking, the fallbacks genuinely preserve today's behaviour, and the tests are strong. One finding I'd want addressed before merge (inline on `rpcsmartrouter.go`), three smaller ones.

The unifying bug is a cardinality error: `taggedApis` is `map[FUNCTION_TAG]TaggedContainer` — one directive per tag — while the real relation is tag × collection. Three of the five gates are that single map being asked a question it structurally cannot answer. Gates 4 and 5 are a different error worth naming separately: `addon == ""` was doing duty as both *"the base collection"* and *"no addon requirement"*, which coincide only while every node serves base. Separating them with a predicate (`ServesBaseCollection`) rather than another sentinel is the right call.

### What I verified

| Claim | Result |
|---|---|
| Every `lavasession.Endpoint` construction carries the flag | ✅ one site (`rpcsmartrouter.go:2461`), updated. `expandInternalPaths` (`generated := nodeUrl`) and the reverifier's `noAddonUrl := nodeUrl` are struct copies, so it propagates |
| Both `GetVerifications` callers filtered | ✅ `chain_fetcher.go:173` and `:259`, both before `needsLatestBlock` |
| `!baseExisted` early return isn't a hole | ✅ `taggedApis` is populated from **every** collection first-wins (`:695`), so no enabled collection declares a tag absent from it |
| The connection-type assumption in the keyed lookup | ✅ no spec in `specs/` has two collection `type`s within one (api_interface, internal_path), so borrowing the type from the tagged entry is sound today |
| `IsSupportingAddon("")` with zero endpoints now returns `false` (was unconditional `true`) | ✅ not reachable — `rpcsmartrouter.go:2491` skips a provider whose endpoints all failed to connect, before any CSWP is built |
| `verification.Addon` is the separated addon, not the raw mixed list | ✅ `GetVerifications` runs `SeparateAddonsExtensions` first; `VerificationContainer` embeds `VerificationKey` |
| Build / vet / tests | ✅ `go build ./...`, `go vet`, and `chainlib` + `lavasession` + `common` + `endpointstate` all green |

### On the tests

The parser table covers extension names mixed into `addons`, disabled collections, missing collections, internal paths, add-on declaration order, and 200-iteration determinism against the map-iteration hazard. The `viper.UnmarshalKey` round-trip is the right instinct — a mapstructure typo would decode as `false` and drop a provider with no error anywhere. The "verified by breaking it" note in the description is the part I'd most want other PRs to copy.

### Scope

Deferring `craftRelay`'s warm-up relay and per-collection admission to their own ticket is the right call, and the description is straight about what isn't traced (relay routing for a mixed deployment). No objection to the boundary.

AnnaR-prog:
Agreed on all five, all addressed in `9aae099`. The framing in your summary is better than mine, in particular naming gates 4/5 as a *different* error from 1–3 — `addon == ""` doing duty as both "the base collection" and "no addon requirement", which coincide only while every node serves base. I'd have kept calling it one bug.

**The blocking one — `standalone-addons` on an extension-only url.** You're right and my description was wrong about it: I wrote "visible rather than silent", and it is neither, for routing. Thank you for running it rather than reasoning about it — the three-line result you posted is unambiguous, and I'd have argued the `no verifications for url` warning was adequate without it.

Fixed with your guard, essentially verbatim: the boot path separates the declared add-ons via `chainParser.SeparateAddonsExtensions` and downgrades the flag to `false` with a warning naming the url and its addons. A config typo is now a startup diagnostic instead of a dead endpoint. Added a `lavasession` test that pins *why* — the same endpoint refuses both `addon="" ext=[archive]` and plain `addon=""` with the flag, and serves both without it.

**The fallback reinstating the opted-out probe.** Agreed, and I think this one is the more dangerous of the two precisely because it isn't live: it's a trap for the next spec that combines a disjoint add-on with internal paths, and it would present as gate 1 all over again with no obvious cause.

Went with passing the predicate through rather than post-hoc comparison — `GetParsingByTagForCollection` takes `allowBaseFallback`, and both call sites supply `ServesBaseCollection()`. When it's false and no add-on collection carries the tag, it returns not-found instead of the base directive.

**`base` → `tagged`.** Right, and the qualifier belongs in the identifier rather than only the comment. Renamed.

**The test pinning a state the constructor can't produce.** Correct on both counts. `taggedApis` covers every enabled collection, so `!taggedExisted` means "declared nowhere" and the add-on search cannot find it either — the early return is sound, my *fixture* was not. It now uses `GET_EARLIEST_BLOCK`, which nothing in the fixture declares, and the comment states the invariant. The behaviour you flagged as backwards for a standalone url is covered by the new `allowBaseFallback` subtest instead.

**`verificationEndpoint` vs `fetcherEndpoint`.** Renamed at the call site — you're right that two names for one value is worst in the function whose entire point is that split.

Full `./protocol/...` suite green. #350 (MAG-3326) is stacked on this branch and has been updated with these changes merged through.

protocol/rpcsmartrouter/rpcsmartrouter.go
AnnaR-prog: Fixed in `9aae099`. You are right and my description was wrong about it — I wrote "visible rather than silent", and it is neither, for routing. Thank you for running it rather than reasoning about it; the three-line result you posted is unambiguous, and I would have argued the `no verifications for url` warning was adequate without it.

Took your guard essentially verbatim. The boot path separates the declared add-ons and downgrades the flag with a warning naming the url, its addons and the provider:

```go
standaloneAddons := url.StandaloneAddons
if standaloneAddons {
    declaredAddons, _, sepErr := chainParser.SeparateAddonsExtensions(ctx, url.Addons)
    if sepErr != nil || len(declaredAddons) == 0 {
        utils.LavaFormatWarning("ignoring standalone-addons: url declares no add-on collection", sepErr, ...)
        standaloneAddons = false
    }
}
```

A config typo is now a startup diagnostic instead of a silently dead endpoint. Added `TestStandaloneAddonsOnAnExtensionOnlyUrlServesNothing` in `lavasession` to pin why: the same endpoint refuses both `addon="" ext=[archive]` and plain `addon=""` with the flag, and serves both without it.

protocol/chainlib/base_chain_parser.go
AnnaR-prog: Fixed in `9aae099`. Agreed — and I think this is the more dangerous of the two precisely because it is not live: it is a trap for the next spec that combines a disjoint add-on with internal paths, and it would present as gate 1 all over again with no obvious cause.

Went with passing the predicate through rather than post-hoc comparison. `GetParsingByTagForCollection` now takes `allowBaseFallback`, and both call sites supply `ServesBaseCollection()` — `ChainFetcher.FetchLatestBlockNumForCollection` from `url`, `EndpointPoller` from `ecf.endpoint`. When it is false and no add-on collection carries the tag, it returns not-found instead of the tagged directive.

New subtest covers both directions on the `/nope` internal path: an ordinary url still inherits the base directive, a standalone one is refused.

protocol/chainlib/base_chain_parser.go
AnnaR-prog: Fixed in `9aae099`. Right — the qualifier belongs in the identifier, not only in the comment. Renamed the local to `tagged`, and the doc now states outright that `taggedApis` is first-wins in spec-file order, so for a spec listing an add-on collection before the base one this entry is the add-on's — including the `CollectionData.Type` used as the lookup key.

protocol/chainlib/base_chain_parser_addon_tag_test.go
AnnaR-prog: Fixed in `9aae099`. Correct on both counts. `taggedApis` is populated from every enabled collection, so `!taggedExisted` means "declared nowhere" and the add-on search cannot find it either — the early return is sound, my fixture was not.

The test now uses `GET_EARLIEST_BLOCK`, which nothing in the fixture declares, and the comment states the invariant so the next reader does not have to rederive it. The behaviour you flagged as backwards for a standalone url is now covered by the `allowBaseFallback` subtest instead, per the fallback thread.

protocol/rpcsmartrouter/spec_reverifier.go
AnnaR-prog: Fixed in `9aae099`. Renamed the call site to `fetcherEndpoint`. You are right that two names for one value is worst in the function whose entire point is that split.

avitenzer (approved):

PR #348, opened 2026-08-31
fix(lavasession): make the epoch's probe branch reachable again (MAG-3190)

Tomelia1999:
Jira ticket: MAG-3190

## Why this exists again

**[#337](https://github.com/Magma-Devs/smart-router/pull/337) shows as MERGED, but its fix is not in
`main`.** This re-opens it against `main` directly. Same three commits, cherry-picked cleanly.

### What happened

#337 was stacked on #336, so its base was `feat/mag-2599-block-reason`, not `main`.

| | |
|---|---|
| #336 merged → `main` | `2026-08-27T14:41:34Z` |
| #337 merged → `feat/mag-2599-block-reason` | `2026-08-27T14:41:51Z` — **17s later** |

The base branch had already been merged to `main` seventeen seconds earlier, so #337 landed on a
branch nothing would read again. GitHub reports it MERGED — correctly, it *did* merge into its base
— but the commit is not an ancestor of `origin/main`.

Normally GitHub retargets a stacked PR to `main` when its base merges. Seventeen seconds was not
enough for that to take effect.

**Verified three ways:**

```
$ git show origin/main:protocol/lavasession/consumer_session_manager.go | grep carried.Reported
(nothing)

$ git show origin/main:protocol/lavasession/consumer_session_manager.go | grep "IsReported(blockedAddr)"
3038:  if !isBackup && !csm.reportedProviders.IsReported(blockedAddr) {   ← still the broken line

$ git branch -r --contains 6cedcb3e
origin/feat/mag-2599-block-reason      ← and nowhere else
```

## The bug, still live in `main`

At every epoch tick the router deliberately re-blocks the previous epoch's blocked providers, so
known-bad providers do not get a clean slate. A second pass then decides who to release: a
**reported** provider must pass a real probe first; an unreported one is released immediately.

**The reported-providers register is emptied before that check runs.** `UpdateAllProviders` calls
`reportedProviders.Reset()` in its body, while the release pass runs from a `defer` that sleeps
first. So every non-backup provider looks unreported, every one takes the immediate-unblock branch,
and the probe branch is unreachable.

**A provider that failed every single request for a full epoch is fully rehabilitated at the next
tick, with no health check of any kind.** The re-block that exists to prevent exactly that is undone
in the same tick.

## The fix

Read the report state from the **carried block record** rather than the live register. The record
holds what the register held at the moment of the block — which is the question being asked.

```diff
-   if !isBackup && !csm.reportedProviders.IsReported(blockedAddr) {
+   if !isBackup && !carried.Reported {
```

Plus, from the two follow-up commits: an unknown record **fails closed** (an unrecoverable record is
not evidence of health), and a provider reported by a *repeat* block still owes a probe — a stale
record there would be a working bypass of the fix.

## Found by black-box testing, not by reading the diff

This surfaced while validating §2 against a live router. With `--epoch-duration 40s`, a provider
whose record read `Reported: true` was released at the tick via:

```
released_by=epoch-not-reported
```

Reported true, released down the *not-reported* branch, no probe. That is the defect reproducing on
`main` today.

## Verification

- `go build ./...`, `go vet ./...` clean; **full `go test ./...` green**
- **The three new tests fail on `main` and pass here** — back-ported the test file onto `origin/main`
  and ran it:

```
--- FAIL: TestEpochTransition_ProbeIsRequiredOnlyForAReportedProvider
    it was reported, so the epoch owes it a probe — and the probe cannot pass while its only
    endpoint is disabled. Releasing it anyway is the clean slate the re-block exists to prevent
--- FAIL: TestEpochTransition_AMissingRecordStillFacesTheProbe
    an unknown block must fail closed
--- FAIL: TestEpochTransition_AProviderReportedByARepeatBlockStillFacesTheProbe
    a stale record here is a working bypass of the fix
```

## Worth checking beyond this PR

Nothing in the process caught this. CI was green, the PR was reviewed, the status said merged. It was
found only because a behavioural test looked wrong.

**If stacked PRs are used regularly, other changes may have been lost the same way** — the signature
is a PR whose merge time is within seconds of its base's merge to `main`.

### DIFF
--- a/protocol/lavasession/block_reason_test.go
+++ b/protocol/lavasession/block_reason_test.go
@@ -759,6 +759,56 @@ func TestBlockRecord_CarriedRecordTakesTheScopeItIsReBlockedIn(t *testing.T) {
 	require.Equal(t, uint32(1), row.Carries)
 }
 
+// epochPairing builds a single direct-RPC provider for the epoch-transition tests.
+//
+// usable decides whether probeDirectRPCEndpoints — the gate the epoch's release pass runs — passes
+// or fails. It is a routability check: at least one ENABLED endpoint carrying a connection object.
+//
+// Both shapes are genuinely direct-RPC (IsDirectRPC() is len(DirectConnections) > 0), which matters:
+// an endpoint with no DirectConnections at all is skipped one branch earlier, failing with "no
+// direct RPC endpoints found" — a misconfiguration, not the gate under test. Failing on the wrong
+// branch would still make these tests pass, while proving reachability through a provider shape
+// that cannot occur in a correctly configured deployment.
+func epochPairing(t *testing.T, address string, usable bool) map[uint64]*ConsumerSessionsWithProvider {
+	t.Helper()
+	// A connection object, not a live connection — NewDirectRPCConnection builds an HTTP client
+	// lazily and dials nothing, and the gate only asks whether one is present.
+	conn, err := NewDirectRPCConnection(context.Background(), common.NodeUrl{Url: "https://example.invalid:443/"}, 5, "")
+	require.NoError(t, err)
+	return map[uint64]*ConsumerSessionsWithProvider{
+		0: {
+			PublicLavaAddress: address,
+			Endpoints: []*Endpoint{{
+				NetworkAddress:    "https://example.invalid:443/",
+				Enabled:           usable, // an endpoint that is present but disabled fails the gate
+				DirectConnections: []DirectRPCConnection{conn},
+			}},
+			Sessions:        map[int64]*SingleConsumerSession{},
+			MaxComputeUnits: 200,
+			PairingEpoch:    firstEpochHeight,
+			StaticProvider:  true,
+		},
+	}
+}
+
+// awaitEpochReleasePass waits for the release pass to have run. It is spawned from a defer that
+// sleeps up to 500ms, and its own cleanup empties previousEpochBlockedProviders last.
+func awaitEpochReleasePass(t *testing.T, csm *ConsumerSessionManager) {
+	t.Helper()
+	require.Eventually(t, func() bool {
+		csm.lock.RLock()
+		defer csm.lock.RUnlock()
+		return len(csm.previousEpochBlockedProviders) == 0
+	}, 15*time.Second, 50*time.Millisecond, "the epoch release pass never ran")
+}
+
+// isValidAddress reports whether the provider is back in routing.
+func isValidAddress(csm *ConsumerSessionManager, address string) bool {
+	csm.lock.RLock()
+	defer csm.lock.RUnlock()
+	return slices.Contains(csm.validAddresses, address)
+}
+
 // The epoch re-blocks the previous epoch's blocked providers "to prevent known-bad providers from
 // getting a clean slate", then releases the ones it has no evidence against. It decided that by
 // asking csm.reportedProviders — but UpdateAllProviders empties that register in its body, while
@@ -768,28 +818,14 @@ func TestBlockRecord_CarriedRecordTakesTheScopeItIsReBlockedIn(t *testing.T) {
 // with no health check at all.
 //
 // The decision now reads the carried block record, which holds what the register held before it was
-// reset.
+// reset. That is equivalent by construction, not merely a reasonable snapshot: ReportProvider is
+// called from exactly two sites, both inside blockProvider, and every RemoveReport site also
+// unblocks the provider — so nothing can change a provider's report state while its block stands.
 //
-// Both cases are asserted, because "keep everything blocked" would pass a one-sided test while
-// being just as wrong as the bug. The provider's endpoint is deliberately unreachable so the two
-// branches have different outcomes: probe → stays blocked, no-probe → released regardless.
+// Both directions are asserted, because "keep everything blocked" would satisfy a one-sided test
+// while being just as wrong as the bug.
 func TestEpochTransition_ProbeIsRequiredOnlyForAReportedProvider(t *testing.T) {
-	// A direct-RPC provider with no usable connection. probeDirectRPCEndpoints is a routability
-	// gate, so it fails here — unlike the gRPC probe, which dials lazily and reports success
-	// against a dead address.
-	deadPairing := func() map[uint64]*ConsumerSessionsWithProvider {
-		return map[uint64]*ConsumerSessionsWithProvider{
-			0: {
-				PublicLavaAddress: "provider-unreachable",
-				Endpoints:         []*Endpoint{{NetworkAddress: "127.0.0.1:1", Enabled: true}},
-				Sessions:          map[int64]*SingleConsumerSession{},
-				MaxComputeUnits:   200,
-				PairingEpoch:      firstEpochHeight,
-				StaticProvider:    true,
-			},
-		}
-	}
-
+	const address = "provider-unreachable"
 	for _, tc := range []struct {
 		name           string
 		reportProvider bool
@@ -800,8 +836,8 @@ func TestEpochTransition_ProbeIsRequiredOnlyForAReportedProvider(t *testing.T) {
 		{
 			name: "reported provider must earn its way back", reportProvider: true, wantReported: true,
 			stillBlocked: true,
-			why: "it was reported, so the epoch owes it a probe — and the probe cannot pass against " +
-				"an unreachable endpoint. Releasing it anyway is the clean slate the re-block exists to prevent",
+			why: "it was reported, so the epoch owes it a probe — and the probe cannot pass while its " +
+				"only endpoint is disabled. Releasing it anyway is the clean slate the re-block exists to prevent",
 		},
 		{
 			name: "unreported provider is released without one", reportProvider: false, wantReported: false,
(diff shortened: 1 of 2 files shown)


avitenzer:
## Review: one correction needed in the description, then this is good to go

I ran this at `cf492c0` in a clean worktree — build, vet, and full `go test ./...` all green, and I independently reproduced both of the description's verifiable claims: the #337 forensics (merge commit `6cedcb3e` is contained only in `origin/feat/mag-2599-block-reason`) and the back-ported-test result (the three named tests fail on `main`, pass here). The diagnosis is right and the code change is sound.

**The one problem is a claim in the description, not in the diff.** The description says the bug is that *"a provider that failed every single request for a full epoch is fully rehabilitated at the next tick, with no health check of any kind"* — and presents this PR as fixing that. It does not, at the epoch tick:

- `probeProvider` short-circuits for `StaticProvider` — every smart-router provider (`rpcsmartrouter.go:2516`) — into `probeDirectRPCEndpoints` (`consumer_session_manager.go:865`), which makes **no network call**. It succeeds when ≥1 enabled endpoint carries a connection object.
- `updateEpoch` calls `endpoint.ResetHealth()` over every regular (`rpcsmartrouter.go:3448`) and backup (`:3476`) endpoint — setting `Enabled = true` unconditionally (`consumer_types.go:701`) — then hands the *same* `*Endpoint` pointers to `UpdateAllProviders` (`:3567`).

So the branch this PR makes reachable terminates in a gate that cannot fail at a tick. The reported provider is released either way; what changes is `released_by=epoch-probe` instead of `epoch-not-reported`. The real routing change lands on the mid-epoch `UpdateAllProviders` callers (`:3880` retryFailedProviders, `:4038` reset-pairing), which have no `ResetHealth` ahead of them — and that delta has no test.

This is worth correcting rather than shrugging off, because it is the same shape as the failure this PR is fixing: a reviewer reads the description instead of the diff, and merges believing the tick-time rehabilitation is closed.

Two ways to resolve it — correct the description, or order the release pass ahead of `ResetHealth` so the gate means what the description says. The second is a different change and probably its own ticket. Suggested text for the first below; everything factual in the original is preserved, the forensics section is untouched.

<details>
<summary>Suggested replacement description</summary>

````markdown
Jira ticket: MAG-3190

## Why this exists again

**[#337](https://github.com/Magma-Devs/smart-router/pull/337) shows as MERGED, but its fix is not in
`main`.** This re-opens it against `main` directly. Same three commits, cherry-picked cleanly.

### What happened

#337 was stacked on #336, so its base was `feat/mag-2599-block-reason`, not `main`.

| | |
|---|---|
| #336 merged → `main` | `2026-08-27T14:41:34Z` |
| #337 merged → `feat/mag-2599-block-reason` | `2026-08-27T14:41:51Z` — **17s later** |

The base branch had already been merged to `main` seventeen seconds earlier, so #337 landed on a
branch nothing would read again. GitHub reports it MERGED — correctly, it *did* merge into its base
— but the commit is not an ancestor of `origin/main`.

Normally GitHub retargets a stacked PR to `main` when its base merges. Seventeen seconds was not
enough for that to take effect.

**Verified three ways:**

```
$ git show origin/main:protocol/lavasession/consumer_session_manager.go | grep carried.Reported
(nothing)

$ git show origin/main:protocol/lavasession/consumer_session_manager.go | grep "IsReported(blockedAddr)"
3038:  if !isBackup && !csm.reportedProviders.IsReported(blockedAddr) {   ← still the broken line

$ git branch -r --contains 6cedcb3e
origin/feat/mag-2599-block-reason      ← and nowhere else
```

## The bug, still live in `main`

At every epoch tick the router re-blocks the previous epoch's blocked providers, so known-bad
providers do not get a clean slate. A second pass then decides who to release: a **reported**
provider is routed to a probe first; an unreported one is released immediately.

**The reported-providers register is emptied before that check runs.** `UpdateAllProviders` calls
`reportedProviders.Reset()` in its body, while the release pass runs from a `defer` that sleeps
first. So every non-backup provider looks unreported, every one takes the immediate-unblock branch,
and the probe branch is dead code for regular providers.

The consequence is the one the ticket names: **the block record's report state does not survive the
epoch.** A provider blocked *and* reported is released down the not-reported route — its record
reads `Reported: true` while the log reads `released_by=epoch-not-reported`.

## The fix

Read the report state from the **carried block record** rather than the live register. The record
holds what the register held at the moment of the block — which is the question being asked.

```diff
-   if !isBackup && !csm.reportedProviders.IsReported(blockedAddr) {
+   if !isBackup && !carried.Reported {
```

Plus, from the two follow-up commits: an unknown record **fails closed** (an unrecoverable record is
not evidence of health), and a provider reported by a *repeat* block still owes a probe — a stale
record there would be a working bypass of the fix.

## What this changes, and what it does not

Worth being exact, because the branch this restores does not mean at an epoch tick what it looks
like it means.

`probeProvider` short-circuits for `StaticProvider` — every smart-router provider
(`rpcsmartrouter.go:2516`) — into `probeDirectRPCEndpoints` (`consumer_session_manager.go:865`),
which makes **no network call**: it succeeds when at least one enabled endpoint carries a connection
object. And `updateEpoch` calls `endpoint.ResetHealth()` over every regular
(`rpcsmartrouter.go:3448`) and backup (`:3476`) endpoint — setting `Enabled = true` unconditionally
— before handing the same `*Endpoint` pointers to `UpdateAllProviders` (`:3567`).

| | Before | After |
|---|---|---|
| epoch tick, reported provider | released, `released_by=epoch-not-reported` | released, `released_by=epoch-probe` |
| epoch tick, unreported provider | released immediately | unchanged |
| mid-epoch `UpdateAllProviders` (`:3880`, `:4038`) — no `ResetHealth` ahead of it | released with no check | routed to the gate, and can stay blocked |
| block record with no stored reason | read as "never reported" → released | fails closed → routed to the gate |

**At the epoch tick this changes the release reason, not the release.** The routing change lands on
the mid-epoch paths. Making the tick-time gate meaningful is a different change — ordering the
release pass ahead of `ResetHealth`, or giving the gate a real liveness signal — and belongs in its
own ticket.

What this fixes unconditionally: the report state now survives the epoch (the ticket's subject), the
probe branch is reachable rather than dead code, and an unrecoverable record no longer reads as
evidence of health.

## Found by black-box testing, not by reading the diff

This surfaced while validating §2 against a live router. With `--epoch-duration 40s`, a provider
whose record read `Reported: true` was released at the tick via:

```
released_by=epoch-not-reported
```

Reported true, released down the *not-reported* branch. That is the defect reproducing on `main`
today.

## Verification

- `go build ./...`, `go vet ./...` clean; **full `go test ./...` green**
- Six tests added. **Three of them fail on `main` and pass here** — back-ported the test file onto
  `origin/main` and ran it:

```
--- FAIL: TestEpochTransition_ProbeIsRequiredOnlyForAReportedProvider
--- FAIL: TestEpochTransition_AMissingRecordStillFacesTheProbe
--- FAIL: TestEpochTransition_AProviderReportedByARepeatBlockStillFacesTheProbe
```

- The other three pin behaviour this change does **not** alter — backup routing, the dead-session-cap
  release, and a reported provider that passes its gate — so a later change cannot move them quietly.

## Known limitations

- **The mid-epoch delta is untested.** `:3880` and `:4038` are where this actually changes routing,
  and nothing covers them.
- **The fail-closed placeholder writes `Reported: true` into a persisted, operator-facing field.** It
  flows through `withCarryOver()` into `blockedProviderRecords` and onto `/debug/provider-routing`,
  where `Reported` is documented (`block_reason.go:138`) as "whether the provider was ACTUALLY added
  to the reported-providers register". For the placeholder that is not true, and `withCarryOver`
  preserves it across every subsequent epoch. A local flag would get the branch right without
  getting the dashboard wrong.
- **`TestCheckAndUnblock_BackupRoutedToComprehensiveProbe` is now stale**
  (`consumer_session_manager_test.go:2443`): its comment documents the `!IsReported` guard this PR
  deletes, and its `:2479` assertion reads a register the function under test no longer consults.
- **`IsReported` has no production callers after this change** — only its definition remains.
- **Not verified on a running router.** The bug was found black-box; the fix has unit evidence only.

## Worth checking beyond this PR

Nothing in the process caught the lost merge. CI was green, the PR was reviewed, the status said
merged. It was found only because a behavioural test looked wrong.

**If stacked PRs are used regularly, other changes may have been lost the same way** — the signature
is a PR whose merge time is within seconds of its base's merge to `main`. Worth its own ticket
rather than a note here.
````

</details>

Smaller things, none blocking:

- `TestCheckAndUnblock_BackupRoutedToComprehensiveProbe` (`consumer_session_manager_test.go:2443`) — the comment describes the `!IsReported` guard this PR removes, and the `:2479` assertion passes with or without the fix. Worth updating in this PR since it is the same change that orphaned it.
- The comment at `consumer_session_manager.go:3066` (*"there has never been anything to remove"*) is true, but only because the single production caller is the defer at `:458`. Nothing in the method enforces that, and there are already two direct callers in tests.
- Branch is 8 commits behind `main`. The only overlap on the shared file is `be51990` moving an `errors` import, so the rebase should be uneventful.

Verified and **not** a problem, listed so it is clear these were checked rather than skipped:

- `awaitEpochReleasePass` cannot be satisfied by a bailed pass — the epoch-mismatch return at `:2996` sits above the cleanup defer's registration, so a pass that bails leaves `previousEpochBlockedProviders` non-empty and the helper times out. The "stays blocked" assertions are not vacuous.
- The "equivalent by construction" argument holds: `ReportProvider` is called only at `:2337` and `:2356`, both inside `blockProvider`, and both `RemoveReport` sites (`:2773`, `:3115`) accompany a release. Nothing can change a provider's report state while its block stands.
- Dropping `RemoveReport` from the immediate-unblock branch is safe — `Reset()` at `:526` always precedes the pass. It does mean a report landing in the 0–500ms window between them now survives where it used to be wiped, which reads like an improvement (it is fresh evidence from the new epoch) but is untested and unmentioned.
- Permanent-lockout risk is bounded by the pool-empty escape at `:1051`.

avitenzer (approved):

PR #349, opened 2026-08-31
feat(observability): record WHY an endpoint was taken out of rotation

Tomelia1999:
## Jira ticket

Jira ticket: MAG-2599

---

## The problem

Every provider block reads `all-endpoints-disabled`. **That is not a cause — it is a count.** It says
"every URL I have is off" and nothing about why any of them went off, so an operator reading it
learns only what the block itself already told them.

The causes are one level down, and they lead an operator to different places:

| Cause | Where to look |
|---|---|
| **unreachable** — timeout, connection refused, DNS, TLS, reset | the network path, the address, the credentials |
| **node-error** — the node answered, and the answer was its own failure | the node process; it is up and reachable but cannot serve |
| **http-server-error** — it answered 5xx | same, but we know least — this branch never read the body |

Before this, all three produced the identical line:

```
WRN disabled unhealthy endpoint endpoint=http://... refusals=50
```

**The information was already at both call sites and was being thrown away.** The relay path holds
the classified error (`:4067`) or the HTTP status (`:4091`) at the exact moment it decides to
disable. This records it.

## What changes

`MarkUnhealthy` now takes a reason. It is stored on the endpoint next to `disabledAt`, and surfaced
as `disable_reason` on the WARN line and `DisableReason` in `/debug/endpoint-state`.

```
WRN disabled unhealthy endpoint endpoint=http://... disable_reason=node-error refusals=50
```

**Two properties are load-bearing and pinned by tests:**

- **Edge-triggered.** Captured on the same `Enabled→false` transition that stamps `disabledAt`. A
  later failure of a *different* kind against an already-disabled endpoint must not rewrite the
  record of why it went down.
- **Cleared on re-enable.** An enabled endpoint carrying a stale reason would read, in
  `/debug/endpoint-state`, as simultaneously serving and disabled-because-of-X.

`http-server-error` is kept distinct from `node-error` deliberately. Both mean "it answered badly",
but that branch decides on the status without reading the body, so it carries no registry
classification at all — it is the case where we know least, and merging them would hide that.

An empty reason is treated as a bug (`unspecified`) rather than silently meaning "no reason needed",
matching how `BlockReasonUnspecified` is used one level up.

## Why this is separate from #340

**It was found while testing that PR, but it does not depend on it.**

Across every test scenario the provider `block_reason` had **one** value in 68 observations, and
`Reported` was **true** in all 46 — both invariant. The cause is structural: §2 deleted
`too-many-dead-sessions` and `never-served-successfully`, which were the two reasons that made the
field discriminate, and both surviving production callers of `blockProvider` pass
`reportProvider=true`.

MAG-2599's own justification for recording a reason was:

> *"a provider blocked for `too-many-dead-sessions` is never reported, so the reconnect loop can
> never release it, while one blocked for `all-endpoints-disabled` is"*

§2 deleted the half of that sentence that made it a distinction. **The endpoint level is where the
remaining distinctions actually are.**

The change itself is independent: it touches the two `MarkUnhealthy` call sites that already exist on
`main`, and needs nothing from #340. **When #340 lands it adds a third call site** — a node error
inside an HTTP 200 — which should pass `EndpointDisableNodeError`. One line.

## Tests

| Test | Pins |
|---|---|
| `IsRecorded` | each reason reaches both the endpoint and the health snapshot |
| `EmptyWhileEnabled` | no reason while serving, and none below the threshold |
| `ClearedOnReEnable` | cleared with `DisabledAt` on a successful relay |
| `NotRewrittenWhileDisabled` | a later failure of another kind does not overwrite it |
| `ProbeReEnableClearsThenRecordsAfresh` | a trial re-enable clears it; the next episode records its own |
| `EmptyBecomesUnspecified` | a missing reason is visible, not absent |
| `ListCoversEveryDeclaredConstant` | a per-reason gauge cannot acquire a series that never returns to 0 |

## Verification

`go build`, `go vet`, `gofmt` clean. Full `go test ./...` green.

## Follow-up, not in this PR

A per-reason metric. The `AllEndpointDisableReasons()` list exists so the gauge can publish a zero
for unused reasons and stay self-correcting — mirroring `AllBlockReasons()` — but no gauge is wired
yet.

### DIFF
--- a/protocol/lavasession/consumer_types.go
+++ b/protocol/lavasession/consumer_types.go
@@ -186,6 +186,11 @@ type Endpoint struct {
 	// after this instant, so a poll that succeeded BEFORE the disable can never re-enable (F1).
 	// Cleared (zeroed) on every re-enable.
 	disabledAt time.Time
+	// disableReason is WHY the endpoint was taken out, captured on the same Enabled→false edge that
+	// stamps disabledAt and cleared alongside it on re-enable. The provider-level BlockReason cannot
+	// answer this: it reads `all-endpoints-disabled`, which is a count of endpoints rather than a
+	// cause. See EndpointDisableReason.
+	disableReason EndpointDisableReason
 	// lastRecoveryPoll is the LastSuccessfulPoll value already counted toward the hysteresis streak,
 	// so a probe cadence faster than the poll cadence cannot count one successful poll twice — only a
 	// strictly newer successful poll advances the streak (F1: "distinct post-disable polls").
@@ -275,8 +280,12 @@ func (e *Endpoint) IsEnabled() bool {
 // never observes a half-applied MarkUnhealthy / RecordProbeVerdict transition. DisabledAt and
 // LastRecoveryPoll are zero while the endpoint is enabled / has never disabled.
 type EndpointHealthSnapshot struct {
-	Enabled                  bool
-	DisabledAt               time.Time
+	Enabled    bool
+	DisabledAt time.Time
+	// DisableReason is WHY the endpoint was taken out, captured on the same transition as
+	// DisabledAt and empty while enabled. The provider-level BlockReason cannot answer this — it
+	// reads `all-endpoints-disabled`, a count rather than a cause.
+	DisableReason            EndpointDisableReason
 	ConsecutiveHealthyProbes uint64 // distinct post-disable successful polls counted toward F1 re-enable
 	LastRecoveryPoll         time.Time
 	// RelayProbeMethod is the read-only method recorded as this disable episode's failing relay
@@ -304,6 +313,7 @@ func (e *Endpoint) HealthSnapshot() EndpointHealthSnapshot {
 	return EndpointHealthSnapshot{
 		Enabled:                  e.Enabled,
 		DisabledAt:               e.disabledAt,
+		DisableReason:            e.disableReason,
 		ConsecutiveHealthyProbes: e.consecutiveHealthyProbes,
 		LastRecoveryPoll:         e.lastRecoveryPoll,
 		RelayProbeMethod:         e.relayProbeMethod,
@@ -371,6 +381,10 @@ func (e *Endpoint) clearRecoveryStreakLocked() {
 func (e *Endpoint) clearRecoveryTrackingLocked() {
 	e.clearRecoveryStreakLocked()
 	e.disabledAt = time.Time{}
+	// The reason describes the episode that just ended, so it is cleared with the timestamp it was
+	// captured alongside. An enabled endpoint carrying a stale reason would read, in
+	// /debug/endpoint-state, as an endpoint that is somehow both serving and disabled-because-of-X.
+	e.disableReason = ""
 	e.dropRelayProbeEvidenceLocked()
 }
 
@@ -439,9 +453,21 @@ func endpointApproachingDisable(wasEnabled, transitioned bool, refusals uint64) 
 	return wasEnabled && !transitioned && refusals+endpointDisableWarningWindow >= MaxConsecutiveConnectionAttempts
 }
 
-// MarkUnhealthy increments connection refusals and disables endpoint if threshold exceeded.
-func (e *Endpoint) MarkUnhealthy() {
-	e.markUnhealthyAt(time.Now())
+// MarkUnhealthy increments connection refusals and disables the endpoint if the threshold is
+// exceeded, recording WHY on the disable itself.
+//
+// reason is required. Every call site already holds it — the relay path has the classified error or
+// the HTTP status in hand at the moment it decides to call this — and passing "" is treated as a bug
+// (EndpointDisableUnspecified) rather than silently meaning "no reason".
+func (e *Endpoint) MarkUnhealthy(reason EndpointDisableReason) {
+	e.markUnhealthyAt(time.Now(), reason)
+}
+
+// DisableReason reports why the endpoint was last taken out of rotation. Empty while enabled.
+func (e *Endpoint) DisableReason() EndpointDisableReason {
+	e.mu.RLock()
+	defer e.mu.RUnlock()
+	return e.disableReason
 }
 
 // markUnhealthyAt is MarkUnhealthy with an injectable clock (tests drive the disable instant so they
@@ -449,7 +475,10 @@ func (e *Endpoint) MarkUnhealthy() {
 // Enabled→false transition (edge-triggered): a repeated MarkUnhealthy on an already-disabled endpoint
 // must NOT push disabledAt forward, or it would silently invalidate post-disable poll evidence the
 // prober has already accumulated (F1).
-func (e *Endpoint) markUnhealthyAt(at time.Time) {
+func (e *Endpoint) markUnhealthyAt(at time.Time, reason EndpointDisableReason) {
+	if reason == "" {
+		reason = EndpointDisableUnspecified
+	}
 	e.mu.Lock()
 	e.ConnectionRefusals++
 	wasEnabled := e.Enabled
@@ -458,6 +487,10 @@ func (e *Endpoint) markUnhealthyAt(at time.Time) {
 	if disabled && wasEnabled {
 		e.Enabled = false
 		e.disabledAt = at
+		// Edge-triggered with disabledAt, and for the same reason: the reason belongs to the failure
+		// that actually took the endpoint out. A later failure of a different kind against an
+		// already-disabled endpoint must not rewrite the record of why it went down.
+		e.disableReason = reason
 		e.clearRecoveryStreakLocked()
 		// If this disable follows a probe re-enable that no successful relay ever validated, the probe's
 		// grant was wasted (the endpoint answers cheap polls but fails real relays). Escalate the next
@@ -475,6 +508,7 @@ func (e *Endpoint) markUnhealthyAt(at time.Time) {
 	if transitioned {
 		utils.LavaFormatWarning("disabled unhealthy endpoint", nil,
 			utils.LogAttr("endpoint", addr),
+			utils.LogAttr("disable_reason", reason),
 			utils.LogAttr("refusals", refusals),
 			utils.LogAttr("is_direct_rpc", isDirect),
 		)
--- a/protocol/lavasession/consumer_types.go
+++ b/protocol/lavasession/consumer_types.go
@@ -463,11 +463,29 @@ func (e *Endpoint) MarkUnhealthy(reason EndpointDisableReason) {
 	e.markUnhealthyAt(time.Now(), reason)
 }
 
-// DisableReason reports why the endpoint was last taken out of rotation. Empty while enabled.
-func (e *Endpoint) DisableReason() EndpointDisableReason {
-	e.mu.RLock()
-	defer e.mu.RUnlock()
-	return e.disableReason
+// disableLocked performs the Enabled→false transition and records the evidence that belongs to it:
+// the instant, the reason, and a cleared recovery streak. Caller must hold e.mu for writing and must
+// have already checked that the endpoint IS currently enabled — this is the edge, not a level.
+//
+// It exists so "every disable transition records disabledAt AND a reason" is an invariant of the
+// type rather than a property that each disabling caller has to remember. There are two such callers
+// (markUnhealthyAt on the relay path, and the dial-failure path in fetchEndpointConnectionFromConsumerSessionWithProvider),
+// and before this was extracted the second one stamped the timestamp but not the reason — producing
+// a snapshot that read {Enabled: false, DisabledAt: set, DisableReason: ""}, which the field docs
+// define as an enabled endpoint.
+func (e *Endpoint) disableLocked(at time.Time, reason EndpointDisableReason) {
+	if reason == "" {
+		// Unreachable on current callers — both pass a named constant. Kept so a future caller cannot
+		// introduce a silently empty reason, which would read as "enabled" in /debug/endpoint-state.
+		reason = EndpointDisableUnspecified
+	}
+	e.Enabled = false
+	e.disabledAt = at
+	// Edge-triggered with disabledAt, and for the same reason: the reason belongs to the failure that
+	// actually took the endpoint out. A later failure of a different kind against an already-disabled
+	// endpoint must not rewrite the record of why it went down.
+	e.disableReason = reason
+	e.clearRecoveryStreakLocked()
 }
 
 // markUnhealthyAt is MarkUnhealthy with an injectable clock (tests drive the disable instant so they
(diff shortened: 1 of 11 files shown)


avitenzer:
## Review: `DisableReason` misreports the cause on three realistic paths

Build, vet, and the `lavasession` / `rpcsmartrouter` tests all pass. The issues below are in the new label itself: the reason is derived by three different mechanisms (registry category at the JSON-RPC error site, raw HTTP status at the REST site, nothing at the dial-failure closure), so the same fault gets different labels and some faults get the wrong one.

### Must fix

**1. DNS / TLS / EOF failures are recorded as `node-error`** — `protocol/rpcsmartrouter/error_mapper.go` (`endpointDisableReasonFor`)

`DetectConnectionError` has no branch for `*net.DNSError`, x509 errors, or `io.EOF`, so `dial tcp: lookup rpc.example: no such host` classifies as `LavaErrorUnknown` (CategoryExternal, Retryable). That flows through `shouldMarkUnhealthy=true` → `MarkUnhealthy(node-error)`. The log line and `/debug/endpoint-state` then point the operator at the node process when no node ever answered. The PR's own definition of `unreachable` names "DNS failure, TLS mismatch". Only timeouts, `ECONNREFUSED`/`ECONNRESET`/`ENETUNREACH`/`EHOSTUNREACH`, and the string-fallback rows currently land on `unreachable` (verified with a throwaway probe).

Fix: carve out `Code == LavaErrorUnknown.Code` in `endpointDisableReasonFor` (→ `unreachable`), or add DNSError / x509 / EOF branches to `DetectConnectionError`.

**2. The same upstream 503 is `node-error` on JSON-RPC but `http-server-error` on REST** — `protocol/rpcsmartrouter/rpcsmartrouter_server.go` (site 2, ~L4238)

`sendJSONRPCRelay` wraps `statusCode >= 500` into an `HTTPStatusError` (`direct_rpc_relay.go` L451-466) → site 1 → registry → `LavaErrorNodeServiceUnavailable` → `node-error`. `sendRESTRelay` returns `(result, nil)` with `StatusCode=503` → site 2 → `http-server-error`. Same fault, two buckets on any dashboard grouped by reason. The comment "the registry has told us nothing" is misleading: REST's `ApplyNodeErrorClassification` already classified 503 to `NODE_SERVICE_UNAVAILABLE` but discards the `*LavaError`.

Fix: at site 2 call `common.ClassifyError(nil, chainFamily, common.TransportREST, statusCode, "")` and pass the result through `endpointDisableReasonFor`, then delete `EndpointDisableServerError`.

**3. The provider-relay dial-failure path never sets a reason** — `protocol/lavasession/consumer_types.go` (`connectEndpoint` closure, ~L1403-1407)

The closure sets `Enabled=false`, `disabledAt=time.Now()`, `clearRecoveryStreakLocked()` but not `disableReason`. `HealthSnapshot()` then returns `{Enabled:false, DisabledAt:<set>, DisableReason:""}`, which the field docs define as "enabled". The adjacent "disabling provider endpoint for the duration of current epoch" warning also lacks the `disable_reason` attr its sibling at L488 carries.

Fix: extract `disableLocked(at, reason)` used by both `markUnhealthyAt` and this closure, passing `EndpointDisableUnreachable` here (a dial failure is that constant's definition). This and #2 together collapse the three derivation paths into one.

### Should fix

- **`EndpointDisableUnspecified` is only ever produced by tests.** The nil branch in `endpointDisableReasonFor` is unreachable (`classifyEndpointHealth(nil)` returns false, and the fallback at server ~L4172 always assigns `classified`), and the `if reason == ""` guard in `markUnhealthyAt` is unreachable (both call sites pass named constants). The nil branch also contradicts the function's own doc ("only ever consulted once classifyEndpointHealth has already decided"). Either delete the dead branch or drop the "passing \"\" is a bug" framing in the docs so they agree.
- **`TestEndpointDisableReasons_ListCoversEveryDeclaredConstant` hand-copies the constants into `declared`**, so a constant added and omitted from both lists passes. The sibling `TestBlockReasons_ListCoversEveryDeclaredConstant` (`block_reason_test.go` L603-620) regexp-scans the source file precisely because comparing two hand lists "is circular and would pass with one missing", and the doc at `endpoint_disable_reason.go` L65-67 claims this test guards that. Mirror the sibling: `os.ReadFile` + `EndpointDisableReason\s*=\s*"([^"]+)"`. The trailing `seen` loop is also redundant with `ElementsMatch`.
- **`AllEndpointDisableReasons()` has zero production callers**, and its doc justifies it in the present tense by "a per-reason gauge" that does not exist. The `BlockReason` sibling exists because `blockedCountsByReasonLocked` feeds `smartrouter_csm_blocked_providers_by_reason`. Either land `smartrouter_csm_disabled_endpoints_by_reason{spec,apiInterface,reason}` from `HealthSnapshot().DisableReason` in the existing publish tick, or drop the list + test and make the doc forward-looking.
- **`disabledEndpoint(t, reason)` duplicates `disableAt(t, e, at)`** in `endpoint_probe_reenable_test.go` L20 (same loop to `MaxConsecutiveConnectionAttempts`, same precondition), and the messages have already drifted. Add a `reason` param to `disableAt` and delegate.
- **`Endpoint.DisableReason()` getter has no non-test callers.** The type documents `HealthSnapshot` as the only synchronized read path and has no per-field getters for `DisabledAt` / `ConsecutiveHealthyProbes` / `RelayProbeMethod` (the `IsEnabled()` exception is justified by a documented race fix and a hot-path caller). Delete it; tests can read `e.HealthSnapshot().DisableReason`, which `TestEndpointDisableReason_IsRecorded` already does.

### Pre-existing, surfaced by this PR (not blockers, worth a doc note)

- gRPC status errors from `Invoke` reach `relayInnerDirect` as `(result, nil)` with `StatusCode` 0-16 via `handleGRPCError`, so neither site fires and a gRPC direct endpoint can never carry a `DisableReason`. The "every call site already holds it" framing implies coverage gRPC doesn't have.
- No test asserts the new `"DisableReason"` key on a `/debug/endpoint-state` row (`rpcsmartrouter.go` ~L1462); `TestDebugEndpointState_WiringSafe` only checks the empty-array case. Matches the existing convention for that endpoint.

Tomelia1999:
Thanks — this was a genuinely useful review. I verified all three "must fix" items independently
rather than take them on trust, and **all three reproduce**. Two of them I was able to measure. Below
is what changed, what is deferred and why, and two places where I'd adjust the scope of a finding.

---

## Landed in this branch

**1. `endpointDisableReasonFor` now has tests.** This was the real gap — the existing tests pass a
reason in and assert it comes back out, so they would have agreed with any mapping at all, including
an inverted one. Added `protocol/rpcsmartrouter/endpoint_disable_reason_mapping_test.go`:

- a direct mapping test (Internal → `unreachable`, External → `node-error`, nil → `unspecified`);
- a characterisation test that runs faults through the **real** relay-path chain
  (`ClassifyError(DetectConnectionError(err), …)`) rather than hand-built `LavaError`s, so it measures
  what a real DNS or TLS failure actually produces;
- a test pinning the REST/JSON-RPC 5xx divergence you found, so it is a decision on the record rather
  than a surprise on a dashboard.

The DNS / TLS / EOF rows are asserted **as they behave today** and marked `KNOWN-WRONG` in the source,
with the ticket named. Asserting the desired value would just fail on main and tell nobody anything;
pinned this way, the wrongness is visible in the source and the test turns red the moment the
classifier is fixed — which is the prompt to update it.

**2. The list-coverage test now scans the source.** You were right that it could not fail: comparing
two hand-written lists is circular, and `endpoint_disable_reason.go` claimed it guarded the sync.
Now mirrors `TestBlockReasons_ListCoversEveryDeclaredConstant` — `os.ReadFile` +
`EndpointDisableReason\s*=\s*"([^"]+)"`. Verified it fails correctly: adding a constant to the source
but not to `allEndpointDisableReasons` now produces

```
EndpointDisableReason "canary" is declared but missing from AllEndpointDisableReasons()
```

The redundant `seen` loop is folded into the same pass.

**3. A merge note for #340**, in `endpoint_disable_reason.go`. Merging the two branches is a compile
error, not a silent mismatch, plus four textual conflicts in the two test files (#340 moved the loop
counter to `uint64`, this branch added the reason argument — both are needed). The third call site
#340 adds is the node-error-inside-HTTP-200 path, which is §2's whole purpose, so it must take
`EndpointDisableNodeError`. Resolving the compile error with `EndpointDisableUnspecified` would build
fine while shipping §2's main disable path with no reason at all. Written down so whoever merges
second does not have to reconstruct it under time pressure.

---

## Deferred, with tickets

### DNS / TLS / EOF reported as `node-error` — your item 1

Confirmed, and measured end to end through the real chain:

| Fault | classified | category | reason |
|---|---|---|---|
| DNS: no such host | UNKNOWN_ERROR | external | **node-error** |
| TLS: unknown authority / expired | UNKNOWN_ERROR | external | **node-error** |
| `io.EOF` / `ErrUnexpectedEOF` | UNKNOWN_ERROR | external | **node-error** |
| tls handshake failure | UNKNOWN_ERROR | external | **node-error** |
| broken pipe | UNKNOWN_ERROR | external | **node-error** |
| connection refused *(control)* | PROTOCOL_CONNECTION_REFUSED | internal | unreachable ✅ |

**Deferring this to its own ticket, deliberately.** It is pre-existing — `git log -S "DNSError"` and
`-S "x509"` over `error_classifier.go` return nothing, so those branches have never existed, and this
PR does not touch `protocol/common` at all. `classifyEndpointHealth` is unchanged here, so a
DNS-failing endpoint was disabled before this PR and is disabled after, identically. There is no
behavioural regression; what is new is only the label on a line that was already being emitted.

The fix belongs in the registry, and `Category` also drives retry eligibility, QoS availability
scoring and backoff. Moving these faults to Internal changes more than the label, so it wants its own
blast-radius review rather than riding along on an observability PR. Your suggested carve-out in
`endpointDisableReasonFor` would fix the label locally, but it would leave the same errors misfiled
for scoring and retry — the narrower fix in the more visible place.

### gRPC coverage — your "pre-existing, worth a doc note"

Agreed it is pre-existing, but I think it is bigger than a doc note, so it is getting a ticket too.

A gRPC status error carrying a response body returns `(result, nil)` with `StatusCode` = the gRPC
code (0–16) — the comment in `direct_rpc_relay.go` says so outright ("this branch returns err == nil,
so downstream sees a completed relay"). Site 1 needs `err != nil`; site 2 needs `>= 500 || == 429`.
Neither can fire.

So it is not only that a gRPC endpoint cannot carry a `DisableReason` — **`bench-after` (§2) is inert
for gRPC node errors entirely.** `UNAVAILABLE`, `INTERNAL` and `RESOURCE_EXHAUSTED` never touch
endpoint health, so a gRPC upstream failing every request is never benched. That is the same class of
gap §2 exists to close for JSON-RPC, still open on this transport — and there is precedent, since the
identical `>= 500 || == 429` gate is already called out in `rpcsmartrouter_server.go` as having
missed gRPC node errors for QoS scoring (MAG-2549). That was fixed for scoring; endpoint health was
not. Worth designing alongside #340's 2xx-with-node-error arm so both transports end up with one
"is the endpoint at fault" decision.

---

## Two scope adjustments

**Your item 3 (dial-failure closure) is real but narrower than "must fix."** Confirmed: the closure
sets `Enabled = false`, `disabledAt` and `clearRecoveryStreakLocked()` but not `disableReason`, and
the WARN alongside it lacks the attr its sibling carries. But the comment directly above it is
correct that this path disables **provider-relay** endpoints only — direct-RPC returns early and is
disabled solely via `MarkUnhealthy`. So it is inert for direct-RPC deployments today. Still worth the
`disableLocked(at, reason)` extraction you suggest, as a type invariant; not on the critical path.

I also checked the locking there, since a snapshot reader was the obvious risk: `endpoint.mu.Lock()`
is held with a deferred unlock from above, so those writes do not race `HealthSnapshot`. `-race` over
both packages is clean.

**Your item 2 (5xx divergence) — agreed on the facts, and I confirmed both halves.** JSON-RPC wraps
`>= 500` into an `HTTPStatusError` so it reaches site 1 and becomes `node-error`; REST returns
`(result, nil)` and reaches site 2 as `http-server-error`. I also measured the JSON-RPC half on a
live router: a direct-RPC upstream returning HTTP 500 disables with `disable_reason=node-error`, so
`EndpointDisableServerError` never fires on that transport at all. And you are right that the
justifying comment is inaccurate — `sendRESTRelay` does call `ApplyNodeErrorClassification` and site 2
discards it.

Where I landed differently: both labels still point the operator at the node, so this is a
consistency wart rather than a misdirection — unlike item 1, which sends you to the wrong system.
It is pinned by a test now so it cannot drift silently. **If you feel strongly that
`EndpointDisableServerError` should not ship at all**, say so and I will collapse it into the registry
path before merge — the argument for doing it now rather than later is vocabulary lock-in, since
renaming a reason afterwards breaks dashboards.

---

## Your "should fix" list

All four accepted, none done in this round — flagging rather than quietly dropping:

- the unreachable `unspecified` branch and the unreachable `if reason == ""` guard (confirmed:
  `ClassifyError` never returns nil, so the nil branch is dead in production);
- `AllEndpointDisableReasons()` with no production callers and a doc justifying itself by a gauge that
  does not exist;
- `Endpoint.DisableReason()` with no non-test callers;
- `disabledEndpoint` duplicating `disableAt`.

These are cleanups with no behavioural effect, so I would rather land them in one pass than
interleave them with the test fixes. Happy to do that on this branch before merge if you would like
it in the same PR — say the word.

Jira number still needs filling in.

Tomelia1999:
## Follow-up tickets are open

The two items deferred from the review above now have tickets. Neither is a regression, and neither
changes behaviour in this PR — both are pre-existing gaps that this PR made visible.

### [MAG-3563](https://magmadevs.atlassian.net/jira/software/projects/MAG/boards/100?jql=assignee%20%3D%20712020%3A623ee097-2ba8-4158-a7bf-71ba7e71f911&selectedIssue=MAG-3563) — `DetectConnectionError` misclassifies DNS, TLS and EOF as node errors

@avitenzer's item 1. `DetectConnectionError` has no branch for `*net.DNSError`, x509 errors or
`io.EOF`, and the string-fallback table has no matching row, so these classify as
`LavaErrorUnknown` — `CategoryExternal` — and land on `node-error` despite no node having answered.

Measured end to end through the real relay-path chain:

| Fault | classified | category | reason |
|---|---|---|---|
| DNS: no such host | UNKNOWN_ERROR | external | **node-error** |
| TLS: unknown authority / expired | UNKNOWN_ERROR | external | **node-error** |
| `io.EOF` / `ErrUnexpectedEOF` | UNKNOWN_ERROR | external | **node-error** |
| tls handshake failure | UNKNOWN_ERROR | external | **node-error** |
| broken pipe | UNKNOWN_ERROR | external | **node-error** |
| connection refused *(control)* | PROTOCOL_CONNECTION_REFUSED | internal | unreachable ✅ |

Fixed in its own ticket because `Category` also drives retry eligibility, QoS availability scoring
and backoff — moving these faults to Internal changes more than the label, so it wants its own
blast-radius review rather than riding along here.

**This PR already contains the guard for it.** The `KNOWN-WRONG` rows in
`protocol/rpcsmartrouter/endpoint_disable_reason_mapping_test.go` pin today's behaviour, so that test
turns red the moment MAG-3563 lands — which is the prompt to update the expectations and drop the
markers. The acceptance criteria on the ticket say so explicitly.

### [MAG-3564](https://magmadevs.atlassian.net/jira/software/projects/MAG/boards/100?jql=assignee%20%3D%20712020%3A623ee097-2ba8-4158-a7bf-71ba7e71f911&selectedIssue=MAG-3564) — gRPC node errors never disable an endpoint, so `bench-after` is inert on gRPC

@avitenzer raised this as "pre-existing, worth a doc note"; on investigation it is larger than the
label, hence a ticket rather than a note.

A gRPC status error carrying a response body returns `(result, nil)` with `StatusCode` set to the
gRPC code (0–16) — the comment in `direct_rpc_relay.go` says so outright: *"this branch returns
err == nil, so downstream sees a completed relay."* Site 1 needs `err != nil`; site 2 needs
`>= 500 || == 429`. Neither can fire.

So `UNAVAILABLE`, `INTERNAL` and `RESOURCE_EXHAUSTED` never touch endpoint health at all: a gRPC
upstream failing every request is never benched. That is the same class of gap FAILOVER-TASKS §2
exists to close for JSON-RPC, still open on this transport — and there is precedent, since the
identical `>= 500 || == 429` gate is already called out in `rpcsmartrouter_server.go` as having
missed gRPC node errors for QoS scoring (MAG-2549). That was fixed for scoring; endpoint health was
not.

Worth designing alongside #340's 2xx-with-node-error arm so both transports end up sharing one
"is the endpoint at fault" decision rather than three.


[MAG-3563]: https://magmadevs.atlassian.net/browse/MAG-3563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAG-3564]: https://magmadevs.atlassian.net/browse/MAG-3564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

avitenzer (approved):

PR #350, opened 2026-08-31
feat(chainlib): admit a provider per collection at boot (MAG-3326)

AnnaR-prog:
# Description

Stacked on [#345](https://github.com/Magma-Devs/smart-router/pull/345) — it uses that PR's `Endpoint`/addon plumbing. Review and merge #345 first; this diff is only the per-collection admission on top.

Admission was all-or-nothing. `ChainFetcher.Validate` returns on the first `Fail`-severity verification, so a provider serving the base collection perfectly and failing **one** add-on's verification was excluded from everything. Attach an archive url to an otherwise healthy provider, have it fail the archive check, lose the provider.

That is a live availability risk on every chain and has nothing to do with the add-on split that surfaced it — it is the reason this is worth scheduling rather than filing and forgetting. It is also step 3 of MAG-3296's own suggested fix: *"exclude a provider from the collections it fails, not from the whole (chain-id, api_interface) pair."*

## What changed

`ValidateCollections` runs the same verifications and **attributes** each failure. A `Fail` belonging to one add-on refuses that add-on and the provider keeps the rest. `ProviderAdmission.Apply` returns a **copy** of the node urls with refused add-ons removed, which the boot path hands to endpoint construction — so session selection, which already filters on addon, does the rest. No new eligibility axis.

The copy is the load-bearing detail: `applyReverification` re-reads the parsed configuration every epoch, and stripping it in place would make one transient failure permanent until restart. That is exactly the failure this codebase already guards against at provider granularity with `reverifyDemoteThreshold` (MAG-2445 — a 40-second blip costing a provider a full epoch).

## Two deliberate limits

**A failure with no add-on to attribute it to is still fatal** — the base collection's own verifications, a head probe that never answered, `GetVerifications` itself. A node that cannot serve the spec's primary surface has nothing narrower to fall back to.

**A base-collection failure does NOT derive `standalone-addons`.** MAG-3326 floats making that flag inferred rather than declared, and it is the one inference worth refusing: only an operator can tell *"this node serves only EVM by design"* from *"its Substrate side is down right now"*, and guessing the first from the second quietly admits a degraded node on a narrower surface. The explicit flag has no such ambiguity, so the assertion stays explicit.

**Boot only.** `applyReverification` keeps the all-or-nothing `Validate`, because the epoch path needs the same temporal hysteresis per add-on that it already has per provider before it can demote a collection safely. Marked `TODO(MAG-3326)` at the call site.

## Most critical to review

- `protocol/chainlib/chain_fetcher.go` — `ProviderAdmission`, and `Validate`/`ValidateCollections` sharing one loop via a `perCollection` flag so the two paths cannot drift.
- `protocol/rpcsmartrouter/rpcsmartrouter.go` — where the admissions map is written (once, before the closure is first called) and read in `convertProvidersToSessions`.

## Tests

The discriminating pair runs the **same** verification set through both entry points: `ValidateCollections` returns no error and refuses only the add-on, while `Validate` on identical input still returns one. That contrast is the assertion — it fails if the two paths ever converge.

Also covered: `Apply` leaves the input configuration untouched, keys by `(url, internalPath)` so the root path and `/v2` are refused independently, is identity when nothing was refused, and refuses nothing when verifications are skipped entirely.

## Not covered

Not exercised against a live upstream, unlike #344/#345 — the discriminating case needs an upstream that passes the base collection and fails one add-on, which none of the Acala or Shiden endpoints does. The unit tests drive both paths through the same mocked verification set instead.

## Jira ticket

Jira ticket: MAG-3326

---

## Author Checklist

I have...

### DIFF
--- a/protocol/chainlib/chain_fetcher.go
+++ b/protocol/chainlib/chain_fetcher.go
@@ -145,7 +145,10 @@ func (cf *ChainFetcher) Validate(ctx context.Context) error {
 		var latestBlock int64
 		if needsLatestBlock(url, verifications) {
 			for attempts := 0; attempts < 3; attempts++ {
-				latestBlock, err = cf.FetchLatestBlockNum(ctx)
+				// Probe the head with THIS url's collections. Using the base
+				// collection's directive for every node is what excluded a
+				// provider serving only a disjoint add-on surface (MAG-3296).
+				latestBlock, err = cf.FetchLatestBlockNumForCollection(ctx, url.Addons, url.InternalPath)
 				if stopValidateRetries(err) {
 					break
 				}
@@ -230,7 +233,9 @@ func (cf *ChainFetcher) ValidateCollect(ctx context.Context) []NodeURLValidation
 		var latestBlock int64
 		if needsLatestBlock(url, verifications) {
 			for attempts := 0; attempts < 3; attempts++ {
-				latestBlock, err = cf.FetchLatestBlockNum(ctx)
+				// Same collection-aware probe as Validate, so `smartrouter health`
+				// reports the head a node can actually serve (MAG-3296).
+				latestBlock, err = cf.FetchLatestBlockNumForCollection(ctx, url.Addons, url.InternalPath)
 				if stopValidateRetries(err) {
 					break
 				}
@@ -505,8 +510,25 @@ func (cf *ChainFetcher) CustomMessage(ctx context.Context, path string, data []b
 	return reply.RelayReply.Data, nil
 }
 
+// FetchLatestBlockNum probes the chain head using the base collection's
+// GET_BLOCKNUM directive. Callers that know which collections the node actually
+// serves should use FetchLatestBlockNumForCollection instead — see MAG-3296.
 func (cf *ChainFetcher) FetchLatestBlockNum(ctx context.Context) (int64, error) {
-	parsing, apiCollection, ok := cf.chainParser.GetParsingByTag(spectypes.FUNCTION_TAG_GET_BLOCKNUM)
+	return cf.FetchLatestBlockNumForCollection(ctx, nil, "")
+}
+
+// FetchLatestBlockNumForCollection probes the chain head with the GET_BLOCKNUM
+// directive of the collection this node URL actually serves.
+//
+// MAG-3296: this used to be the base collection's directive for every node,
+// which excluded any provider whose add-on is a disjoint api surface rather than
+// a superset — an EVM-only Acala node answers the `evm` collection's
+// eth_blockNumber and cannot answer the base collection's Substrate
+// chain_getHeader at all. On the admission path a failed head probe returns
+// before a single verification runs, so such a provider was dropped without ever
+// being asked the questions it could answer.
+func (cf *ChainFetcher) FetchLatestBlockNumForCollection(ctx context.Context, addons []string, internalPath string) (int64, error) {
+	parsing, apiCollection, ok := cf.chainParser.GetParsingByTagForCollection(spectypes.FUNCTION_TAG_GET_BLOCKNUM, addons, internalPath)
 	tagName := spectypes.FUNCTION_TAG_GET_BLOCKNUM.String()
 	if !ok {
 		return spectypes.NOT_APPLICABLE, utils.LavaFormatError(tagName+" tag function not found", nil, []utils.Attribute{{Key: "chainID", Value: cf.endpoint.ChainID}, {Key: "APIInterface", Value: cf.endpoint.ApiInterface}}...)
--- a/protocol/chainlib/chain_fetcher.go
+++ b/protocol/chainlib/chain_fetcher.go
@@ -101,6 +101,42 @@ func skipVerification(url common.NodeUrl, name string) bool {
 	return SkipAllVerifications || url.ShouldSkipVerification(name)
 }
 
+// verificationsForNodeUrl drops the base collection's verifications for a url that
+// has opted out of serving it (NodeUrl.StandaloneAddons).
+//
+// GetVerifications always adds the empty addon to whatever a url declares, so an
+// add-on node runs the base collection's verifications too. That is right when the
+// add-on EXTENDS the base surface and wrong when it REPLACES it: Acala's base
+// chain-id verification fires a Substrate chain_getBlockHash, which an EVM-only
+// node answers with -32601, and an omitted severity in a spec means Fail — so one
+// inherited verification excluded the provider (MAG-3296).
+//
+// This runs BEFORE needsLatestBlock for the same reason the skip filter does: a
+// verification that is not going to run must not drag a head probe out to the
+// upstream on its way past.
+//
+// Filtering here rather than inside GetVerifications keeps the parser's signature
+// free of node-url concerns — and dropping Addon=="" containers afterwards is
+// exactly equivalent to never having appended the empty addon.
+func verificationsForNodeUrl(url common.NodeUrl, verifications []VerificationContainer) []VerificationContainer {
+	if url.ServesBaseCollection() {
+		return verifications
+	}
+	kept := make([]VerificationContainer, 0, len(verifications))
+	for _, verification := range verifications {
+		if verification.Addon == "" {
+			utils.LavaFormatDebug("skipping base-collection verification for a standalone-addons url",
+				utils.LogAttr("verification", verification.Name),
+				utils.LogAttr("url", url.UrlStr()),
+				utils.LogAttr("addons", url.Addons),
+			)
+			continue
+		}
+		kept = append(kept, verification)
+	}
+	return kept
+}
+
 // needsLatestBlock reports whether any verification that will ACTUALLY RUN for this
 // node-url depends on the chain head, and therefore whether Validate has to spend a
 // FetchLatestBlockNum relay against the upstream before verifying.
@@ -138,6 +174,7 @@ func (cf *ChainFetcher) Validate(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
+		verifications = verificationsForNodeUrl(url, verifications)
 		if len(verifications) == 0 {
 			utils.LavaFormatWarning("no verifications for url", nil, utils.LogAttr("url", url.String()))
 		}
@@ -229,6 +266,7 @@ func (cf *ChainFetcher) ValidateCollect(ctx context.Context) []NodeURLValidation
 			results = append(results, nodeResult)
 			continue
 		}
+		verifications = verificationsForNodeUrl(url, verifications)
 
 		var latestBlock int64
 		if needsLatestBlock(url, verifications) {
--- a/protocol/chainlib/chain_fetcher.go
+++ b/protocol/chainlib/chain_fetcher.go
@@ -166,13 +166,98 @@ func stopValidateRetries(err error) bool {
 	return err == nil || errors.Is(err, common.StatusCodeError429)
 }
 
+// ProviderAdmission records which add-on collections a provider's node urls
+// failed to answer for, so a failure attributable to ONE add-on costs that
+// add-on rather than the whole provider (MAG-3326).
+//
+// Empty means everything a url declared was admitted.
+type ProviderAdmission struct {
+	failedAddons map[string]map[string]struct{}
+}
+
+func admissionKey(url common.NodeUrl) string {
+	return url.Url + "\x00" + url.InternalPath
+}
+
+func (a *ProviderAdmission) fail(url common.NodeUrl, addon string) {
+	if a.failedAddons == nil {
+		a.failedAddons = map[string]map[string]struct{}{}
+	}
+	key := admissionKey(url)
+	if a.failedAddons[key] == nil {
+		a.failedAddons[key] = map[string]struct{}{}
+	}
+	a.failedAddons[key][addon] = struct{}{}
+}
+
+// Any reports whether any add-on was refused.
+func (a ProviderAdmission) Any() bool { return len(a.failedAddons) > 0 }
+
+// Apply returns a COPY of nodeUrls with each url's refused add-ons removed.
+//
+// The input is never mutated: it is the parsed configuration, and the epoch
+// re-verification path re-reads it every cycle. Stripping it in place would make
+// one transient failure permanent — the failure this codebase already guards
+// against at provider granularity with reverifyDemoteThreshold.
+func (a ProviderAdmission) Apply(nodeUrls []common.NodeUrl) []common.NodeUrl {
+	if !a.Any() {
+		return nodeUrls
+	}
+	out := make([]common.NodeUrl, 0, len(nodeUrls))
+	for _, url := range nodeUrls {
+		refused := a.failedAddons[admissionKey(url)]
+		if len(refused) == 0 || len(url.Addons) == 0 {
+			out = append(out, url)
+			continue
+		}
+		kept := make([]string, 0, len(url.Addons))
+		for _, addon := range url.Addons {
+			if _, bad := refused[addon]; bad {
+				continue
+			}
+			kept = append(kept, addon)
+		}
+		copied := url
+		copied.Addons = kept
+		out = append(out, copied)
+	}
+	return out
+}
+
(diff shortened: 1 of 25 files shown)


protocol/chainlib/chain_fetcher.go
```diff
@@ -174,14 +299,26 @@ func (cf *ChainFetcher) Validate(ctx context.Context) error {
 				cf.verificationsStatus.Store(cf.getVerificationsKey(verification, cf.endpoint.ApiInterface, cf.endpoint.ChainID), false)
 				utils.LavaFormatWarning("failed verification on provider startup", err, utils.LogAttr("verification", verification.Name))
 				if verification.Severity == spectypes.ParseValue_Fail {
-					return utils.LavaFormatError("invalid Verification on provider startup", err, utils.Attribute{Key: "Addons", Value: addons}, utils.Attribute{Key: "verification", Value: verification.Name})
+					// A failure that belongs to one add-on costs that add-on, not
+					// the provider. Only a failure with no add-on to attribute it
+					// to — the base collection's own — is still fatal (MAG-3326).
+					if perCollection && verification.Addon != "" {
```
avitenzer, line 305: **This guard never fires for the case the PR leads with.** `archive` is an extension, not an add-on.

`VerificationKey` takes its two axes from different places (`base_chain_parser.go:805-808`):

```go
verificationKey := VerificationKey{
    Extension: parseValue.Extension,              // "archive"
    Addon:     apiCollection.CollectionData.AddOn, // ""
}
```

I scanned every spec under `specs/` — 50 (verification, value) rows, positive control passed — and every archive verification lands as `add_on='' extension='archive'`:

```
ethereum.json    ETH1       pruning   add_on=''      extension='archive'
lava.json        LAVA       pruning   add_on=''      extension='archive'
cosmoshub.json   COSMOSHUB  pruning   add_on=''      extension='archive'
aptos.json       APT1       pruning   add_on=''      extension='archive'
ethereum.json    ETH1       pruning   add_on='trace' extension='archive'   ← the only non-empty add_on
```

So for *"attach an archive url to an otherwise healthy provider, have it fail the archive check, lose the provider"* — the PR's opening sentence and the ticket's first bullet — `verification.Addon == ""`, this branch is skipped, and the `return admission, ...` below still costs the whole provider.

The feature does work, but only for genuinely add-on-scoped collections: ETH1 `trace`/`debug`/`bundler`, ARBITRUM `arbtrace`, BTC `blockdaemon`. That is a much smaller population than the body claims, and it excludes every `pruning` verification on every chain.

Two ways out, and they are different sizes: attribute on the extension axis too (which needs `Apply` and `Endpoint.Extensions` to gain a matching axis, since stripping `url.Addons` is not where an extension lives), or keep the add-on-only scope and restate the motivating case in the body. The second is honest and cheap; the first is what the ticket actually asks for.

protocol/chainlib/validate_collections_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+// TestValidateCollections_AddonFailureKeepsTheProvider is the MAG-3326 case that
+// affects deployments having nothing to do with add-on splits: attach an archive
+// url to an otherwise healthy provider, have it fail the archive verification,
+// and today you lose the whole provider.
+func TestValidateCollections_AddonFailureKeepsTheProvider(t *testing.T) {
+	url := common.NodeUrl{Url: "https://a.example.com", Addons: []string{"archive"}}
+	verifications := []VerificationContainer{failingVerification("pruning", "archive")}
```
avitenzer, line 61: This is why the suite is green on a case that does not work: the fixture puts `"archive"` in the add-on field.

`failingVerification` builds `VerificationKey{Addon: addon}` (`:22`) and every new test passes `"archive"` (`:61`, `:88`, `:101`, and throughout `per_collection_admission_test.go`). No shipped spec produces that shape — and the same package already has a test saying so:

```go
// base_chain_parser_addon_extension_test.go:15-26
//   "Mirrors the ethereum spec: archive lives in the extension set;
//    addons are debug/trace/bundler."
require.Empty(t, addons, "archive must not be classified as an addon")
```

The test named for the headline scenario is therefore testing a spec shape that cannot occur, and it would keep passing even if the guard were deleted for real archive verifications.

Building the fixture from `SeparateAddonsExtensions` output — or from an actual collection in `specs/ethereum.json` — would have caught this, and is the change I would make regardless of how the blocker above is resolved.

protocol/chainlib/chain_fetcher.go
```diff
@@ -130,29 +166,118 @@ func stopValidateRetries(err error) bool {
...
+		}
+		kept := make([]string, 0, len(url.Addons))
+		for _, addon := range url.Addons {
+			if _, bad := refused[addon]; bad {
+				continue
+			}
+			kept = append(kept, addon)
+		}
+		copied := url
+		copied.Addons = kept
```
avitenzer, line 221: **When this empties a `standalone-addons` url's list, the url starts serving base-collection traffic.** I ran it:

```
before Apply: Addons=[evm] ServesBaseCollection=false
after  Apply: Addons=[]    ServesBaseCollection=true   StandaloneAddons=true
```

`ServesBaseCollection()` is `!StandaloneAddons || len(Addons) == 0` (`common/endpoints.go:171`), so an empty list reads as *"serves base"* regardless of the flag. The boot path then completes it: `rpcsmartrouter.go:2454-2465` sees no declared add-ons, downgrades `standaloneAddons` to `false`, and logs a warning whose hint blames the operator's config (*"extension names belong in addons…"*) — a misleading diagnostic here, since the config was correct and admission emptied the list.

Concretely, for an Acala-shaped url (`addons: ["evm"], standalone-addons: true`) whose `evm` verification fails: `verificationsForNodeUrl` has already dropped every base-collection verification for that url, so every surviving verification has `Addon != ""` and this branch catches all of them. An EVM-only node that just failed its only check is admitted as a **base-collection endpoint** and answers Substrate requests with `-32601` — the routing #345's gate 5 exists to prevent, quoted verbatim in the comment at `consumer_types.go:329-332`.

Before this PR that provider was excluded. This is strictly less safe for the one population the base PR was written for.

`Apply` emptying a list is a lossy way to record a refusal: `Addons: []` already means *"declares no add-ons"*, and now also means *"every add-on was refused"*. Refusing the url outright when `kept` is empty and the url is standalone would keep the two apart — or, closer to the ticket's wording (*"a provider can be eligible for one collection and not another"*), record eligibility rather than rewrite the url list.

avitenzer (commented):
## Review — `NOT READY`, and the reason is substantive rather than procedural

Built and tested at `fd29b56` in a clean worktree. The per-collection idea is right and MAG-3326 is a real availability problem — but as it stands the change does not fix the case it leads with, and on one population it makes routing worse than the code it replaces.

### Blocking

1. **The headline case is not fixed: `archive` is an extension, not an add-on.** Detail inline on `chain_fetcher.go`. I scanned every shipped spec (50 verification/value rows, positive control passed): every archive verification is `add_on='' extension='archive'`. The only non-empty `add_on` in the tree is ETH1's `trace`.
2. **The test that covers it builds a spec shape that cannot exist** — inline on `validate_collections_test.go`, and contradicted by an existing test in the same package.
3. **A `standalone-addons` url whose only add-on is refused starts serving base traffic** — inline on `chain_fetcher.go:221`. Confirmed by running it.
4. **`go vet ./protocol/...` fails** — inline on `standalone_addons_routing_test.go:112`. Inherited from the base branch, but present here, and `lint` does not run on this PR (see below).
5. **MAG-3325 is a ticket obligation with no reviewer-visible answer.** MAG-3326 says *"Also folded in by this: MAG-3325 — the startup init relay is crafted once for the whole (chain-id, api_interface) pair from the base collection and sent to every provider."* `craftRelay` still calls the unscoped `GetParsingByTag` (`rpcsmartrouter_server.go:415`), and `grep -rn MAG-3325` returns nothing in the tree or the body. Either do it or say why it is deferred — right now a reviewer merges believing it landed.
6. **`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`** — conflicts in `spec_reverifier.go` against its own base. `6c407ea` on the base branch and the `9aae099` merged here are the same commit rebased, so the stack has drifted. Merge the base in again.

### The stacking costs real CI coverage

This PR runs **10 checks; #345 runs 21**. Missing: `lint`, `govulncheck`, `Validate Jira ticket`, `jira-ticket/validated` — because the base branch is not `main`. That is why the `go vet` failure above is green here. Worth weighing against MAG-3190/#337, which this team is still cleaning up in #348: a stacked PR whose base merged seconds earlier, whose fix never reached `main`, and which GitHub reported as MERGED throughout. The body is admirably explicit about the stack (*"Review and merge #345 first"*), so this is not a hidden dependency — but the checklist below it says `[x] targeted the main branch`, which is the one box that would have caught #337.

### Not blocking, but a reviewer should know

- **The epoch path reverses the boot admission ~2 ticks later** — inline on `spec_reverifier.go`.
- **A recovered add-on is never restored** — inline on `rpcsmartrouter.go:2610`.
- **The wiring has no test.** `validateProviderTier`'s new third return is discarded with `_` at all four call sites in `boot_resilience_test.go` (`:74`, `:94`, `:110`, `:128`), and the `admitAll` adapter hardcodes an empty `ProviderAdmission`, so `if admitted[i].Any()` is never taken by any test. Nothing builds a `lavasession.Endpoint` from an admitted url and asks it `CheckSupportForServices("")` — which is exactly where blocker 3 lives, and why the suite is green.
- `TODO(MAG-3326)` points at the ticket this PR closes, so closing it buries the deferred epoch work.
- `validateProviderCollections` is a near-verbatim copy of `validateProvider` and drops the comment explaining why `CloneChainParserForValidation` is needed, while keeping the call. Two bodies that must stay in step, with the reason documented in only one. Style, ranked last.

### Checked and cleared — listed so it is clear these were looked at

- **Pointer-keyed admissions map surviving an epoch**: fine. `configuredStatic: relevantStaticProviderList` (`rpcsmartrouter.go:2657`) is the same slice validated at boot, so lookups hold. That is what makes the *permanence* finding true rather than a *loss* finding.
- **The closure reading a nil map**: fine. The write at `:2610` precedes the first call at `:2689`; a too-early call would silently degrade to old behaviour, but there isn't one.
- **`ChainFetcher.Validate`'s callers**: one production caller (`spec_reverifier.go:574`), and behaviour is byte-identical via `validate(ctx, false)`.
- **Diff hygiene**: 25 files as GitHub reports, no skips, no `//nolint`, no credentials, no removed tests (population control 2030).

### Evidence (all run by me at `fd29b56`)

- `go build ./...` clean; `go vet ./protocol/...` **1 failure**
- `go test` on chainlib / rpcsmartrouter / lavasession / common / endpointstate — **all ok**
- Spec scan for archive classification — 50 rows, positive control passed
- The standalone+`Apply` interaction — executed, output in the inline comment

The body's *"Not covered"* section is honest about the absence of live verification, and the two deliberate limits are argued well — refusing to derive `standalone-addons` from a base failure is the right call and the reasoning is exactly right. The problem is narrower than the design: it is which field the refusal is keyed on.

protocol/lavasession/standalone_addons_routing_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	require.False(t, extensionOnly.CheckSupportForServices("", []string{"archive"}),
+		"archive traffic carries addon \"\", so the opt-out refuses it")
+	require.False(t, extensionOnly.CheckSupportForServices("", nil),
+		"and plain traffic too — the endpoint would serve nothing")
+
+	// Which is why the flag is downgraded before the endpoint is built. Same
+	// endpoint without it serves both, which is the behaviour an operator who
+	// wrote `addons: [archive]` expects.
+	downgraded := *extensionOnly
```
avitenzer, line 112: `go vet ./protocol/...` fails here:

```
protocol/lavasession/standalone_addons_routing_test.go:112:16: assignment copies lock value to downgraded:
  github.com/magma-Devs/smart-router/protocol/lavasession.Endpoint contains sync.RWMutex
```

`*extensionOnly` copies an `Endpoint`, which holds a `sync.RWMutex`. Harmless in this test in practice, but the repo just took the golangci baseline to zero and `govet` is in that set.

This came in with the base branch's `9aae099`/`6c407ea`, so it is not this PR's own work — but it is on this head, and **`lint` does not run on this PR** (10 checks here vs 21 on #345, because the base is not `main`). So nothing catches it on either PR's current check set. Constructing a fresh `Endpoint` with the fields you want, rather than dereferencing, fixes it.

protocol/rpcsmartrouter/spec_reverifier.go
```diff
@@ -534,17 +559,57 @@ func validateProvider(
...
+// provider survives with the rest.
+//
+// Boot only. applyReverification keeps the all-or-nothing Validate deliberately:
+// the epoch path already carries temporal hysteresis at provider granularity
+// (reverifyDemoteThreshold, MAG-2445, added because a 40-second blip cost a
+// provider a full epoch), and per-collection demotion there needs the same
+// hysteresis per add-on before it can be safe. Without it one transient failure
+// would strip an add-on until restart.
+//
+// TODO(MAG-3326): give the epoch path per-add-on hysteresis, then move it here too.
```
avitenzer, line 588: *"Boot only"* understates what happens next: the epoch path does not merely skip per-collection admission, it **undoes it**.

`applyReverification` re-probes every active provider with the all-or-nothing `validateProvider` (`:200`), built from `newEndpoint(provider.NodeUrls)` (`:539`, `:548`) — the untouched configuration, still declaring the add-on that failed at boot. So the same verification fails again on every tick, `demoteFailStreak` increments (`:309`), and at `reverifyDemoteThreshold = 2` (`:45`) the provider is demoted **wholesale** — roughly two epochs after boot.

A reviewer reading "boot only" reasonably infers *"the improvement applies at boot and the epoch path is no worse than today"*. What actually happens is that the stated benefit — *"attach an archive url to a healthy provider and lose the provider"* — is deferred by two ticks rather than removed. That belongs in the body under a limitations heading, whichever way the code goes.

(For what it is worth, the reasoning for *not* extending it yet is sound: per-add-on demotion without per-add-on hysteresis would strip an add-on on a transient blip, which is the MAG-2445 shape.)

protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -2556,11 +2592,23 @@ func (rpsr *RPCSmartRouter) CreateSmartRouterEndpoint(
...
+	// refused on its own and the provider keeps the rest, so adding an archive url
+	// to an otherwise healthy provider can no longer cost the provider. Read by
+	// convertProvidersToSessions below, which builds the endpoints.
+	//
+	// Written once here, before the closure is ever called, and read-only from then
+	// on — the closure also runs from the epoch path, which does not repopulate it.
+	for provider, admission := range backupAdmissions {
+		staticAdmissions[provider] = admission
+	}
+	admissionsByProvider = staticAdmissions
```
avitenzer, line 2610: This map is written once and never recomputed — so **a refused add-on is never restored**, which is the permanence the copy-not-mutate design is justified against.

The commit message argues the copy is load-bearing because stripping the config in place *"would make one transient failure permanent until restart"*. But the closure that reads this map (`:2437`) is handed to the epoch path (`:2656`) and to `retryFailedProviders` (`:2756`), and `applyReverification` calls it on every promotion (`spec_reverifier.go:354`). Pointer identity holds — `configuredStatic` is the same slice validated at boot — so a boot-era refusal key still matches an hour later.

If the add-on recovers, nothing clears the entry. The provider is re-promoted with its boot-time refusal re-applied, permanently, until the process restarts. The configuration is indeed untouched; the *decision* is the thing that became permanent.

A clear-on-successful-revalidation, or recomputing admissions when `applyReverification` promotes, would close it.

avitenzer:
Superseded by #358.

#345 has merged, so the MAG-3296 base this was stacked on is now on `main`. The per-collection admission commit (`f9db844`) has been rebuilt on `main` and targets `main` directly in #358 — byte-for-byte the same change, only hunk context shifted. Verified there: `go build ./...` clean and `./protocol/{chainlib,rpcsmartrouter,lavasession,endpointstate}/...` green from a clean worktree of the pushed commit.

Closing this one; review moves to #358.

PR #353, opened 2026-08-31
feat(lavasession): let stateful relays broadcast to the backup tier (MAG-3338)

Tomelia1999:
#Closes MAG-3338

Jira ticket: MAG-3338

## Why

A stateful relay — transaction submission — does not pick one provider. It **broadcasts** to every provider it selects at once and takes the first success. That shape is why a configured backup tier was invisible to it, in two independent ways: selection read the primary pool only, and a stateful relay never retries, so the "primaries exhausted → backup" route that every read uses can never fire for it.

Measured on a live router before this change, two primaries failing and a healthy backup configured:

| | primary-a | primary-b | backup-1 | client got |
|---|---|---|---|---|
| stateful write | 3 | 3 | **0** | the node error |

## What

`--stateful-to-backup` widens the stateful broadcast to include the backup tier. **Default `false`** — nothing moves for an existing deployment. The value is stored per session manager, and one manager serves one chain + api-interface, so the scope a per-chain override would need already exists.

**Both tiers are broadcast to at once, not one after the other.** A stateful relay has no second attempt to fall back into, so a backup absent from the first broadcast is never reached. Sending the same signed transaction to more nodes is what transaction broadcast already does and the hash is identical, so this raises no double-execution question — allowing a stateful *retry* would have, which is why that alternative was rejected.

**Each tier is ranked and capped on its own, primaries first.** Ranking them together would compare a warm primary against a cold backup, and since the cap decides who survives, a backup that had been serving a while would start pushing primaries out of the broadcast. Order does not decide who is relayed to — every selected provider is — only who survives the cap.

A backup joins on the same terms as a primary: not blocked, serves the addon and extensions asked for, and **not held off after a 429**. That last check is new for the backup tier; the existing fallback path does not do it.

### What turning it on costs

None of this is obvious from the flag name, so it is stated here and in the flag help:

- the backup is engaged on **every** stateful relay, including the ones the primaries serve fine — a per-request cost on a metered tier, not a per-outage one;
- a fast backup can **win the race**; the winner cancels the rest, so the spend can buy an answer that was already arriving. Cancelled relays carry no QoS penalty (MAG-2648), so this distorts spend, not scoring;
- the backup vendor sees every transaction.

### Two pre-existing defects fixed on the way

Both are older than this feature and both sit in the code it touches:

- the session builder resolved addresses through `csm.pairing` alone, so the first backup address to reach it would have hit `LavaFormatFatal` — a crash, not a fallback;
- the stateful selection sorted the **shared** cached address slice **in place** under a read lock, so two concurrent stateful relays sorted and ranged the same array. It now sorts a copy, with an address tie-break so an all-equal compute-unit list (every provider on a fresh router) still selects reproducibly.

## Test plan

- [x] `make test` at this commit — 33 packages ok. See "Known gaps" for one flake.
- [x] `make lint` — clean
- [x] `go test -race` on the new selection — clean
- [x] The shared-slice guard was **verified against a faithful pre-fix reproduction** (in-place sort of the caller's slice, fresh return slice — what the original did): it now fails **3/3 without `-race`** and reports 3 data races with it, and passes both ways against the code as shipped. An earlier revision of this description claimed the guard was load-bearing when it was not — see Review round 1.
- [x] 13 new unit tests: flag off/on, tier ordering, per-tier cap, blocked backup, already-tried, held-off backup, missing addon, no backups configured, determinism, both-tier lookup, the race, `GetSessions` end to end, and reads unaffected
- [x] Live router, 7 scenarios, fake upstreams on fresh ports:

| Scenario | primaries | backups | result |
|---|---|---|---|
| stateful, flag **off** | 3+3 | **0** | node error — identical to baseline |
| stateless read, flag off | 3+3 | 3 | still reaches backup via the fallback path |
| stateful, flag **on**, primaries failing | 3+3 | **3** | backup serves, client gets a success |
| stateful, flag on, primaries **healthy** | 3+3 | **3** | backup engaged anyway; won 2 of 3 races |
| stateless read, flag on | 3 | **0** | reads untouched by the flag |
| 5 primaries + 2 backups, flag on | 15 | **6** | 7 sends per relay, nothing trimmed |
| backup returns 429, flag on | — | 1 then 0 | one strike, 35s hold-off, excluded from later broadcasts |

Zero panics and zero fatals across seven router starts — specifically no hit on the `LavaFormatFatal` above.

### To reproduce the live runs

Fake upstreams answer the router's housekeeping honestly and fail only the customer relay, so the endpoint stays healthy and only the behaviour under test fails. Two traps that produced wrong answers first, both now guarded: every upstream must be healthy **before** the router boots (otherwise providers fail startup verification and never enter the pairing), and the fake chain heads must agree (otherwise the consistency pre-filter skips a provider without consuming the error budget).

## Review round 1 (@avitenzer) — both findings confirmed by reproduction, both fixed

**A provider configured in both tiers was broadcast to twice** (`0cc56e3`). `UpdateAllProviders` builds the two pools from separate inputs with no dedup and the file documents an address in both as legitimate, so the backup tier could re-add a provider the primary tier had already chosen. The duplicate was not the damage: it put the caller's target above what a map keyed by address can hold, so the "enough providers" check never fired, control fell through to a second full selection pass under the read lock, and that pass logged `Pairing list empty` while the relay was being served by every provider it had selected — on every stateful relay in that configuration. Reproduced: `selected` had 4 entries, `sessions` had 3.

The exclusion went where the other four already are, in `eligibleBackupsForStateful` beside already-tried / blocked / wrong-addon / held-off, rather than into a new mechanism at the call site.

One correction in the PR's favour: this is **not reachable through the product's config**. `ValidateUniqueProviderNames` groups both lists by chain + api-interface + name at boot, so a shared name fails before a listener binds. It is fixed because the file's own contract permits the overlap and other code in it handles it deliberately.

**The shared-slice guard could not fail against its own bug** (`f3cdb3b`). `setupStatefulCSM` left every provider at zero compute units, so the descending comparator never returned true, `sort.Slice` performed no swaps, and there was no reordering and no write to detect. My original verification was invalid: the "pre-fix" code I measured against also aliased the *return* value, so the race it reported came from `append`, not from the sort. Fixed with distinct compute units and an **order-sensitive** assertion — a membership comparison passes against an in-place permutation, and asserting order is what makes the guard reachable from `make test`, which runs no `-race`. The production fix was never in doubt; the guard was.

Two items from that review were deliberately **not** changed:

- the wrong-tier compute-units lookup needs no fix — with the duplicate excluded, a dual-pool address never reaches the backup ranking, so there is nothing left to rank by the wrong pool's counter;
- `eligibleBackupsForStateful` still does not check `currentlyBlockedProviderAddresses`, so a provider blocked out of the primary tier could serve as a backup. It is absent from `validAddresses` and therefore from the new exclusion. That is a policy decision rather than a defect and wants its own change. My recommendation: skip it — a node the router already judged bad gives no real redundancy and you would be paying the backup vendor for it.

## Known gaps — please read before approving

- **The Helm chart cannot set this flag.** `charts/smart-router/templates/routers_deployment.yaml` uses a fixed, enumerated arg list with no `extraArgs` escape hatch, so merging this delivers a flag no Kubernetes deployment can turn on. **Post-merge work, owner @Tomelia1999:** a companion chart PR adding a values key — ideally per-router, like `providerOptimizerAvailabilityWeight` already is, to match the per-chain scoping here.
- **Cap rule deviates from the original proposal.** The plan said "10 primaries, then all eligible backups"; this implements **10 per tier**. An unbounded tier in a fan-out is a latent problem, and no real deployment has more than 10 backups, so it is behaviourally identical today.
- **`TestBlockReason_SurvivesTheEpochTransition` failed once** under `make test` on this branch and passed on rerun. Not classified: branch `make test` 1 fail / 1 pass, fork-point `make test` 0 fail / 1 pass, the package alone 5/5 on both sides. Arguing against it being this PR's: `block_reason_test.go` sorts first so it runs *before* any test added here, and this diff touches none of `blockProvider` / `UpdateAllProviders` / `blockedRecord` / `CreateConsumerSessionManager`. Flagged rather than dismissed.
- **All primaries blocked + flag on** falls to the existing single-backup fallback, not the widened broadcast — the empty-pool check precedes the stateful branch. Not a regression; the flag only widens while at least one primary is healthy.
- **Log volume:** one INFO line per stateful relay while the tier is engaged. On a high-transaction chain that is per-transaction output.
- **Pre-existing fragility in the selection loops, not touched here.** The "do I have enough providers?" check sets its target from a slice length and compares it to a map size, uses `==` rather than `>=`, and returns a short group silently — you can ask for seven providers, get three, and nothing logs it. The duplicate above is one way to trip the first of those; there are others. Worth its own ticket rather than mixing a refactor of shared selection code into this change.
- **Ticket obligations not independently verified.** I could not read MAG-3338 (no Jira access from this environment), so the diff has not been checked against what the ticket actually asks for. A reviewer who can open it should do that.

https://claude.ai/code/session_01S95m47jBa9iHmksf7f1t4m

### DIFF
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -94,6 +94,15 @@ type ConsumerSessionManager struct {
 	// Production uses the process-wide holdoff.Shared; tests inject a clock-pinned one.
 	rateLimitHoldoff *holdoff.Registry
 
+	// statefulToBackup widens a stateful relay's broadcast to the backup tier
+	// (--stateful-to-backup). It lives here, rather than as a package variable, because one
+	// manager serves exactly one chain + api-interface: a customer who wants transaction
+	// broadcast mirrored to an expensive backup on one chain rarely wants it on all of them.
+	//
+	// The zero value is the safe one. Left unset, a stateful relay reaches primaries only,
+	// which is the behaviour of every release before this field existed.
+	statefulToBackup bool
+
 	// contains a sorted list of blocked addresses, sorted by their cu used this epoch for higher chance of response
 	currentlyBlockedProviderAddresses []string
 
@@ -1796,28 +1805,142 @@ func (csm *ConsumerSessionManager) GetSessions(ctx context.Context, wantedProvid
 	}
 }
 
-// csm must be rlocked here
-func (csm *ConsumerSessionManager) getTopTenProvidersForStatefulCalls(validAddresses []string, ignoredProvidersList map[string]struct{}) []string {
-	// sort by cu used, easiest to sort by that factor as it probably means highest QOS and easily read by atomic
-	customSort := func(i, j int) bool {
-		return csm.pairing[validAddresses[i]].atomicReadUsedComputeUnits() > csm.pairing[validAddresses[j]].atomicReadUsedComputeUnits()
+// statefulFanoutPerTier caps how many providers a stateful relay broadcasts to within one tier.
+// It bounds the fan-out, not the number of tiers: primaries and backups are capped separately, so
+// enabling the backup tier widens the broadcast instead of evicting primaries from it.
+const statefulFanoutPerTier = 10
+
+// getProvidersForStatefulCalls returns the providers a stateful relay broadcasts to.
+//
+// The primary tier is always included. The backup tier is appended only when this chain enables it
+// (--stateful-to-backup, off by default), and it is appended rather than merged: each tier is
+// ranked on its own and primaries come first. Ranking the two together would mean comparing compute
+// units served between a warm primary and a cold backup, a number that says nothing about either —
+// and since the cap is what decides who survives, a merged ranking would let a backup that has been
+// serving for a while start pushing primaries out of the broadcast.
+//
+// Order does not decide who is relayed to: every address returned here gets a session and its own
+// goroutine. It decides only who survives the cap.
+//
+// csm must be rlocked here.
+func (csm *ConsumerSessionManager) getProvidersForStatefulCalls(ctx context.Context, validAddresses []string, ignoredProvidersList map[string]struct{}, addon string, extensions []string) []string {
+	providers := csm.rankStatefulTier(validAddresses, ignoredProvidersList)
+	if !csm.statefulToBackup {
+		return providers
 	}
-	// Sort the slice using the custom sorting rule
-	sort.Slice(validAddresses, customSort)
-	addresses := []string{}
-	wantedLength := 10
-	for _, sortedAddress := range validAddresses {
-		// skip ignored providers
-		if _, foundInIgnoredProviderList := ignoredProvidersList[sortedAddress]; foundInIgnoredProviderList {
+
+	backups := csm.rankStatefulTier(csm.eligibleBackupsForStateful(ctx, ignoredProvidersList, addon, extensions), ignoredProvidersList)
+	if len(backups) == 0 {
+		return providers
+	}
+
+	// Only logged when the backup tier actually joins a broadcast, so a router with the flag off —
+	// or with no eligible backup — pays nothing and says nothing. When it does fire it is worth an
+	// INFO: a metered tier has just been engaged on a request the primaries may well serve.
+	utils.LavaFormatInfo("stateful relay broadcasting to the backup tier as well",
+		utils.LogAttr("primaries", providers),
+		utils.LogAttr("backups", backups),
+		utils.LogAttr("addon", addon),
+		utils.LogAttr("extensions", extensions),
+		utils.LogAttr("GUID", ctx),
+	)
+	return append(providers, backups...)
+}
+
+// rankStatefulTier orders one tier's addresses by compute units served, highest first, drops the
+// providers this request has already tried, and caps what is left at statefulFanoutPerTier.
+//
+// It never reorders its argument. validAddresses can be the cached addon-address slice shared with
+// every other reader of that router key, and sorting it in place — as this code used to — races
+// them under the read lock this runs beneath.
+//
+// Ties break on the address so the broadcast is reproducible. Compute units are equal for every
+// provider on a freshly started router, which makes the tie the common case rather than the corner
+// one.
+//
+// csm must be rlocked here.
+func (csm *ConsumerSessionManager) rankStatefulTier(addresses []string, ignoredProvidersList map[string]struct{}) []string {
+	ranked := make([]string, 0, len(addresses))
+	for _, address := range addresses {
+		if _, alreadyTried := ignoredProvidersList[address]; alreadyTried {
 			continue
 		}
-		// fill the slice until we have 10 providers who are not ignored
-		addresses = append(addresses, sortedAddress)
-		if len(addresses) >= wantedLength {
-			break
-		}
+		ranked = append(ranked, address)
 	}
-	return addresses
+
+	sort.Slice(ranked, func(i, j int) bool {
+		cuI, cuJ := csm.usedComputeUnits(ranked[i]), csm.usedComputeUnits(ranked[j])
+		if cuI != cuJ {
+			return cuI > cuJ
+		}
+		return ranked[i] < ranked[j]
+	})
+
+	if len(ranked) > statefulFanoutPerTier {
+		ranked = ranked[:statefulFanoutPerTier]
+	}
+	return ranked
+}
+
+// eligibleBackupsForStateful returns the backups a stateful broadcast may include: not already
+// tried by this request, not blocked, able to serve the addon and extensions asked for, and not
+// currently held off after a 429.
+//
+// The hold-off is a plain exclusion here, unlike filterRateLimitedProviders, which keeps the
+// soonest-to-expire candidate rather than leave a request with nowhere to go. That rescue cannot be
+// needed on this path: the backup tier is additive, and a stateful broadcast only gets here with at
+// least one primary already chosen.
+//
+// csm must be rlocked here.
+func (csm *ConsumerSessionManager) eligibleBackupsForStateful(ctx context.Context, ignoredProvidersList map[string]struct{}, addon string, extensions []string) []string {
+	eligible := make([]string, 0, len(csm.backupProviders))
+	for address, provider := range csm.backupProviders {
+		if provider == nil {
+			continue
+		}
+		if _, alreadyTried := ignoredProvidersList[address]; alreadyTried {
+			continue
+		}
+		if _, blocked := csm.blockedBackupProviders[address]; blocked {
+			continue
+		}
+		if !provider.IsSupportingAddon(addon) || !provider.IsSupportingExtensions(extensions, ctx) {
+			continue
+		}
+		if csm.rateLimitHoldoff != nil {
+			if _, heldOff := csm.rateLimitHoldoff.ProviderReadyAt(address); heldOff {
+				continue
+			}
+		}
+		eligible = append(eligible, address)
+	}
+	return eligible
+}
+
+// usedComputeUnits reads a provider's compute units served this epoch, whichever tier it belongs
+// to. Unknown addresses report 0 and therefore rank last, which is the right answer for the only
+// way one can occur: a provider dropped from the pairing between selection and this read.
+//
+// csm must be rlocked here.
+func (csm *ConsumerSessionManager) usedComputeUnits(address string) uint64 {
+	if provider := csm.providerByAddress(address); provider != nil {
+		return provider.atomicReadUsedComputeUnits()
+	}
+	return 0
+}
+
+// providerByAddress resolves an address to its session pool across both tiers.
+//
+// Everything that walks a selected-provider list has to go through this rather than reach into
+// csm.pairing directly, because a stateful broadcast can now carry backup addresses alongside
+// primary ones and csm.pairing does not hold backups.
+//
+// csm must be rlocked here.
+func (csm *ConsumerSessionManager) providerByAddress(address string) *ConsumerSessionsWithProvider {
+	if provider, ok := csm.pairing[address]; ok {
+		return provider
+	}
+	return csm.backupProviders[address]
 }
 
 // convertSelectionStatsToMetrics converts SelectionStats to metrics format with all provider scores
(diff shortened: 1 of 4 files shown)


protocol/lavasession/consumer_session_manager.go
```diff
@@ -1451,28 +1460,142 @@ func (csm *ConsumerSessionManager) GetSessions(ctx context.Context, wantedProvid
...
 	}
-	// Sort the slice using the custom sorting rule
-	sort.Slice(validAddresses, customSort)
-	addresses := []string{}
-	wantedLength := 10
-	for _, sortedAddress := range validAddresses {
-		// skip ignored providers
-		if _, foundInIgnoredProviderList := ignoredProvidersList[sortedAddress]; foundInIgnoredProviderList {
+
+	backups := csm.rankStatefulTier(csm.eligibleBackupsForStateful(ctx, ignoredProvidersList, addon, extensions), ignoredProvidersList)
```
avitenzer: **A provider configured in both tiers is emitted twice here.**

`eligibleBackupsForStateful` filters against `ignoredProvidersList`, `blockedBackupProviders`, addon/extensions and hold-off — but never against the primary set it is about to be appended to. This repo documents that overlap as legitimate (`consumer_session_manager.go:3079-3084`: *"built from separate inputs by UpdateAllProviders with no dedup, so a provider address can legitimately appear in both"*).

Reproduced with `dual-provider` in both maps:

```
selected = [dual-provider primary-b backup-1 dual-provider]   (len 4)
counts   = map[backup-1:1 dual-provider:2 primary-b:1]
INF stateful relay broadcasting to the backup tier as well
    primaries=dual-provider,primary-b  backups=backup-1,dual-provider
```

Downstream, all observed in that run:

- `wantedProviderNumber` is set to `len(providerAddresses)` = 4 (`:2161`), but `sessionWithProviderMap` is a map and dedups to 3, so the `len(sessionWithProviderMap) == wantedProviderNumber` early return at `:2202` **never fires**;
- control falls through to the refetch at `:2208` — a second full `getValidProviderAddresses` pass under the read lock — which logs `DBG Pairing list empty` and returns via the `err != nil && len(map) != 0` escape at `:2211`;
- the INFO line lists the same address under both `primaries` and `backups`.

It self-corrects, so this is not a crash — but it costs a redundant selection pass and a spurious "Pairing list empty" line on **every** stateful relay in that configuration.

Suggested shape — pass the already-selected set down so the backup tier can exclude it:

```go
providers := csm.rankStatefulTier(validAddresses, ignoredProvidersList)
if !csm.statefulToBackup {
    return providers
}
// A provider can be configured in both tiers (UpdateAllProviders does not dedup).
// Appending it twice inflates wantedProviderNumber past what the session map can
// hold, so the caller's "enough providers" check never fires.
selected := make(map[string]struct{}, len(providers))
for _, address := range providers {
    selected[address] = struct{}{}
}
backups := csm.rankStatefulTier(
    csm.eligibleBackupsForStateful(ctx, ignoredProvidersList, addon, extensions, selected),
    ignoredProvidersList)
```

Same root cause, worth a look while you are in here: `usedComputeUnits` resolves via `providerByAddress`, which prefers `csm.pairing` — so when ranking the **backup** tier, a dual-pool address is ranked by its *primary* entry's compute units.

Also: `eligibleBackupsForStateful` checks `blockedBackupProviders` but not `currentlyBlockedProviderAddresses`, so a dual-pool provider blocked out of the primary tier for repeated failures is still broadcast to as a backup — every stateful relay, while primaries are healthy, to a node already judged bad this epoch and for no added redundancy. The existing fallback path shares that blind spot, but only once primaries are exhausted, so this widens it. Worth an explicit decision in the comment either way.

No test covers the dual-pool case; one alongside the fix would close the gap.

protocol/lavasession/stateful_to_backup_test.go
```diff
@@ -0,0 +1,274 @@
...
+	require.NotNil(t, csm.providerByAddress("primary-a"))
+	require.NotNil(t, csm.providerByAddress("backup-1"), "a backup address must resolve, not return nil and crash the router")
+	require.Nil(t, csm.providerByAddress("nobody"))
+}
+
+// The selection used to sort its argument in place, and its argument is the addon-address slice
+// cached and shared with every other reader of that router key — under a READ lock. Concurrent
+// stateful relays therefore sorted and ranged the same backing array at once. Run with -race.
+func TestStatefulSelection_DoesNotMutateTheSharedAddressSlice(t *testing.T) {
+	csm := setupStatefulCSM(t, []string{"primary-a", "primary-b", "primary-c"}, []string{"backup-1"})
```
avitenzer, line 222: **This test cannot fail against the bug it guards.**

The fix itself is correct and worth having — I confirmed `filterRateLimitedProviders` returns the *same backing array* when `heldCount == 0` or the registry is nil (`:1665`, `:1690`), so the pre-fix code genuinely did sort the shared cached slice in place under an `RLock`. Real race, real fix.

But `setupStatefulCSM` leaves every provider at **zero** compute units. The descending comparator therefore always returns `false`, `sort.Slice` performs no swaps, and there is no write for the race detector to see. Restoring the pre-fix in-place sort and measuring:

| test | `DATA RACE` count vs. pre-fix code |
|---|---|
| this test as written | **0** |
| identical body, distinct CU values (10/20/30/40) | **5** |

The distinct-CU probe passes clean against the fixed code, so the fix holds — it is the guard that is inert.

This contradicts the description's *"fails with `WARNING: DATA RACE` against the restored pre-fix code, so the regression test is load-bearing rather than decorative"*. Could you re-run that claim, or correct it if it does not reproduce on your side too?

Separately, the assertion is order-insensitive: `sort.Strings(before)` and `sortedAfter` mean only membership is compared, which an in-place permutation preserves. So without `-race` it is a no-op guard in both directions.

One block in the setup fixes it:

```go
// Distinct CU values: with an all-equal key the descending comparator never
// returns true, sort.Slice performs no swaps, and the in-place mutation this
// test exists to catch produces no write for -race to flag.
require.NoError(t, csm.pairing["primary-a"].addUsedComputeUnits(10, 0))
require.NoError(t, csm.pairing["primary-b"].addUsedComputeUnits(20, 0))
require.NoError(t, csm.pairing["primary-c"].addUsedComputeUnits(30, 0))
```

Worth weighting: `make test` is `go test ./... -count=1` with no `-race`, and this repo's PR gate runs no Go tests at all — so unless this is also made order-sensitive, nothing automated will ever exercise it.

avitenzer (commented):
Two findings from review, both confirmed by reproduction rather than inspection. Design and reasoning are sound — per-tier ranking, default-off, and the cost disclosure in the description are all right. These are contained fixes.

Details inline on the two anchors below.

protocol/lavasession/consumer_session_manager.go
Tomelia1999: Confirmed and fixed in 0cc56e3. I reproduced your exact output before changing anything — `selected` 4 entries with `dual-provider` twice, `sessions` 3, the early return never firing, the refetch, and the `Pairing list empty` line. Every downstream consequence you listed is real.

**One correction, in the PR's favour rather than against it:** this is not reachable through the product's config. `ValidateUniqueProviderNames` (rpcsmartrouter.go:3098) groups *both* lists by chain + api-interface + name at boot, so a shared name fails before a listener binds, and all four `UpdateAllProviders` sites derive from that validated config. So it is a latent contract violation, not a live defect. Fixed anyway, for exactly the reason you cite: the file documents the overlap as legitimate and `GenerateReconnectCallback` already handles both sides deliberately, so this new code was the odd one out.

**I took your shape but with a linear scan instead of a set.** `alreadySelected` is capped at `statefulFanoutPerTier`, so building a set costs an allocation and a loop to save at most ten comparisons on a path already doing map lookups per candidate — and `slices.Contains` is the idiom this file already uses in three places. Same exclusion, placed where you suggested: in `eligibleBackupsForStateful` beside already-tried, blocked, wrong-addon and held-off, rather than a new mechanism at the call site.

**On `usedComputeUnits` preferring `csm.pairing`:** correct, and it needs no change once the duplicate is gone. A dual-pool address never reaches the backup ranking now, so there is nothing left to rank by the wrong pool's counter. I left it rather than add a second code path for a state that can no longer occur — say the word if you would rather have it explicit.

**On `currentlyBlockedProviderAddresses`:** you were right that it wants an explicit decision, and right that the comment is where it belongs. I had put it in the commit message and PR description, which is the wrong place — the question arises while reading the filter. Now recorded beside it in 861719b: deliberate, only reachable for a dual-pool address (one blocked as a primary is absent from `validAddresses`, so `alreadySelected` cannot exclude it), unreachable today because boot rejects that config, and the policy question deferred to the tier consolidation rather than widened into this change. My own view is that it should be skipped — a node the router already judged bad gives no redundancy and you would be paying the backup vendor for it — but that is a behaviour change and wanted its own decision.

**Test added** (`TestStatefulToBackup_ProviderInBothTiersAppearsOnce`). It asserts the invariant that actually broke — that the selection count is reachable by the session map — rather than only that the list has no repeats, so it also catches a future variant that produces a different unreachable count. Verified load-bearing by neutralising only the new skip and leaving the signature identical: fails with `elements differ`, passes with the skip restored.

protocol/lavasession/stateful_to_backup_test.go
Tomelia1999: You are right, and I have re-run it as you asked. **It does not reproduce, and I can tell you why my measurement was wrong.**

My "restored pre-fix" code also aliased the *return* value (`ranked := addresses`). The original returns a fresh slice — `addresses := []string{}` — and only the sort is in place. So the data race I measured came from `append(providers, backups...)` writing into the shared backing array, not from the sort at all. I verified the wrong code and then reported the guard as load-bearing on the strength of it. Entirely my error, and worse for having written both halves myself.

Against a faithful reproduction — in-place sort of the caller's slice, fresh return slice — your numbers hold exactly:

| variant | vs faithful pre-fix |
|---|---|
| test as written (all-zero CU) | **0 races**, 3/3 runs |
| + distinct CU (10/20/30) | **2, 5, 9 races** |

Fixed in f3cdb3b, taking both of your points:

- distinct compute units, so the descending comparator actually reorders;
- **order-sensitive assertion** — `require.Equal(before, after)` with no sorting. You were right that sorting both sides made it a no-op guard in both directions.

The second point turned out to matter more than the first. Making it order-sensitive means it no longer needs `-race` at all: against the faithful pre-fix code it now **fails 3/3 without `-race`**, and reports 3 data races with it. Against the code as shipped it passes both ways. So it is reachable from `make test`, which closes the gap you flagged — a `-race`-only guard would never have run anywhere automatically, given the PR gate runs no Go tests.

The PR description has been corrected; the old claim is gone and replaced with what I actually measured, with a note saying an earlier revision overstated it.

And thank you for the `filterRateLimitedProviders` check at :1665 and :1690 — confirming the same backing array is returned when `heldCount == 0` is the step that establishes the production bug was real, which my own verification had not actually shown.

avitenzer (approved):

PR #355, opened 2026-08-31
fix(chainlib): upgrade websockets on any path and keep them alive

nimrod-teich:
#Closes MAG-3344

Jira ticket: MAG-3344

Six commits, on `main`. #386 has merged, so this is no longer stacked. The last three answer the first review round.

## Why

Two defects on the consumer WebSocket path, both found while exercising subscriptions through a deployed gateway.

**The URL a caller guesses is the one that fails.** Dialing the bare endpoint — `wss://host/` — is answered **405**; only `/ws` and `/websocket` upgrade. That is the router's own routing table, not the proxy in front of it: the jsonrpc listener registers a POST catch-all beside `GET /ws` and `GET /websocket`, so a GET on `/` matches no handler.

The `ws://` / `wss://` scheme never reaches the server — TLS terminates at the edge and the router sees a plain HTTP GET — so the Upgrade header is the only signal there is. The rule this PR implements is therefore *any GET carrying the upgrade header is a WebSocket, whatever the path*, including the health path and a gateway that rewrites a path prefix to `/`.

**An idle subscription then dies at ~2 minutes.** The router sends nothing on a quiet connection, so a proxy closes the socket — with no close frame, so the client sees a transport error rather than an unsubscribe.

Fixing this in the router rather than in ingress config keeps it true for every deployment, not just the ones behind a particular gateway.

## What changed

**`faecb38` — upgrade on any path, and keep the connection alive**

- [`jsonRPC.go:572`](protocol/chainlib/jsonRPC.go#L572) — catch-all `GET /*` that hands upgrade requests to the same middleware + callback `/ws` uses. A GET without the header still gets 405.
- [`tendermintRPC.go:610`](protocol/chainlib/tendermintRPC.go#L610) — the same gate in front of the existing `handlerGet`, so Tendermint's URI-style GETs still reach the relay handler.
- [`consumer_websocket_manager.go:205`](protocol/chainlib/consumer_websocket_manager.go#L205) — `runWebsocketWriter` gains a keep-alive arm: it pings on an interval (default 30s; new `--websocket-keep-alive-interval` flag, `0` disables). The ping goes through the writer to preserve single-writer discipline, and a failed ping wakes the read loop the same way a failed frame write does.
- Same file: every frame the writer sends goes through `writeFrame`, which sets a write deadline (10s by default; see `991047e` below), and `ListenToMessages` waits on `writerDone` before returning. Without that wait a ping can fire after the handler returns and gofiber has recycled the connection — the race detector confirms it. The `websocketFrameWriter` interface gains `SetWriteDeadline` for it.

**`f36cda1` — release an idle websocket the client refuses to close**

The idle reaper sent its close frame through the writer and returned, leaving teardown to the client. A client that ignored that frame used to be reaped anyway by the proxy in front of the router — it closes a silent socket within ~2 minutes, which woke the read loop for us. The keep-alive above means the connection is never silent, so that backstop is gone and nothing would free the handler goroutine, its 128 KiB fasthttp read buffer or the per-IP limiter slot. That is the MAG-3722 leak by a different route.

- [`consumer_websocket_manager.go:161`](protocol/chainlib/consumer_websocket_manager.go#L161) — frames carry a `final` flag; [`:240`](protocol/chainlib/consumer_websocket_manager.go#L240) — the writer puts a final frame on the wire and then tears down with the same sequence the server-shutdown branch uses (immediate read deadline, read loop wakes, handler does the single `Close`). Every `conn` access stays on the writer goroutine, so the reaper needs no synchronisation of its own.
- `MaxIdleTimeInSeconds` becomes an atomic behind `Set`/`GetMaxIdleTimeInSeconds`, matching the keep-alive interval directly above it: written once from its startup flag, read by every live connection's reaper, and the regression test has to shorten it while other connections run. Its flag keeps its name and default but moves to the `viper.GetInt64` read the keep-alive flag uses — `websocket_tunables_cobra_viper_test.go` covers both flags from flags, YAML and defaults, because a broken binding would hand the code a `0` and silently disable either feature.

**`a0498fe` — keep `Allow` on the 405 the catch-all now answers**

Registering a GET route took fiber's own method-mismatch path out of play, and that path is what attached `Allow` to the 405. Measured against the fork point, a plain GET went from `405 Allow: POST` to `405 Allow:` *(absent)*. RFC 9110 §15.5.6 requires the header, so it is set explicitly. `HEAD` runs the same handler and is covered by the same line. A method with no route at all now reports `GET, HEAD, POST` rather than `POST` — that one takes fiber's normal mismatch path and is the route table telling the truth.

**`5159e59` — serve a websocket upgrade on the health path too** (review blocker 1)

`createAndSetupBaseAppListener` registers the health route before `Serve` registers the catch-all, and fiber matches in registration order, so an upgrade on the health path was answered with a health body; with `health-check-path: /` that was the bare-URL failure back again. The health handler now hands an upgrade on with `c.Next()`. Both listener tests run with the health route on `/` as well as `/lava/health`, dial a websocket there, and check that a plain GET is still the health check; they fail without the change.

Same commit: the Tendermint `handlerGet` branch the catch-all sits in front of now has coverage. A stub relay sender records what the relay path is asked for, so a plain `GET /status?height=7` is shown to reach it with path and query intact and to return its reply, while an upgrade on `/status` never does. The two stale comments (`test_smartrouter_sol.sh`, `consumer_types.go`) are corrected.

**`991047e` — write deadline as a flag, pings only after silence, startup warning** (review blocker 2 and two notes)

- `--websocket-write-timeout` (default 10s, `0` disables), atomic behind `Set`/`GetWebSocketWriteTimeout` like the other two tunables. The writer test asserts every frame goes out under a deadline of the configured length and that `0` sets none; the cobra/viper test covers it from flags, YAML and default.
- Every frame the writer sends resets the keep-alive ticker, so a busy connection is never pinged and a quiet one is pinged once per interval of silence, which is the silence the proxy measures. A writer test drives frames at a quarter of the interval and sees no ping, then goes quiet and sees them resume.
- The comment above the keep-alive arm now states what differs from `chainproxy/rpcclient/websocket.go` and why: no pong deadline, because the peer here is a customer and a late pong must not end its connection; a vanished peer is caught by the TCP retransmission timeout once pings stop being acknowledged, a peer that stopped reading by the write deadline.
- `--limit-websocket-connection-idle-time 0` with the keep-alive on holds a quiet connection for as long as its client keeps the socket: the reaper never runs and no proxy reaps a pinged connection. That is what disabling the idle limit asks for, but before the keep-alive the proxy did it anyway, so the combination now logs a warning at startup (`WebSocketKeepAliveOutlivesIdleReaper`, table-tested). Coupling the two, so that idle=0 also switched pings off, would have brought the two-minute proxy reap back for exactly the operators who asked for no limit.

**`9d30c25` — hand each upstream listener its own message channel** (the unclassified `DATA RACE`)

Reproduced: 1 in 25 runs of `TestSolanaRouterIDSurvivesUpstreamReconnect` under `-race`. A reconnect replaces `activeSub.messagesChan` under the manager lock while the listener read the field unlocked on every loop. In production the reconnect is only ever started from the listener's own error branch, after which that listener returns, so the ordering the test creates (reconnect driven directly while the first listener still selects) does not occur there. The channel is now a listener parameter, so each listener owns one channel for its life and the handoff is explicit. 25 further runs are clean.

## Behaviour notes

- The idle ceiling for a quiet subscription moves from ~2 minutes (proxy) to `--limit-websocket-connection-idle-time`, 20 minutes by default: `idleFor` is reset by inbound frames and outbound deliveries, not by pings. Operators with long-lived quiet subscriptions should raise it.
- A client that stops reading is disconnected once a frame write exceeds `--websocket-write-timeout`. There was no bound before this PR.

## How to verify

```
go test ./protocol/chainlib/... -run 'ChainListener_' -race -count=1
go test ./protocol/chainlib/ ./protocol/rpcsmartrouter/ -race -count=1
```

New tests: an upgrade succeeds on `/`, `/ws`, `/websocket`, a longer path and the health path (on `/` and on its default) on both listeners; a plain GET still answers 405 **and still carries `Allow: POST`**, and on the health path is still the health check; a plain Tendermint GET reaches the relay path with its path and query; a keep-alive ping arrives on an idle connection and not on a busy one; every frame goes out under the configured write deadline; a final frame ends the writer only after its frame reaches the connection; an idle connection whose client never reads or closes is released regardless; all three websocket tunables resolve from flags, YAML and their registered defaults; the keep-alive-without-idle-limit warning fires only for that combination.

Note that CI here runs builds, lint, govulncheck, GitGuardian and the Jira check — no `go test`. Local runs are the only test evidence.

Against a running router:

```
websocat wss://<router-host>/            # upgrades instead of 405
curl -sS -o /dev/null -w '%{http_code}\n' https://<router-host>/   # still 405
curl -sS -D- -o /dev/null https://<router-host>/ | grep -i '^allow:'   # Allow: POST
```

These three have not been run — no deployed router was available.

## Follow-up (not in this PR)

- The customer-facing endpoint docs still describe only the `/websocket` form; with this change the bare URL works, so they can state that both do.

### DIFF
--- a/protocol/chainlib/consumer_websocket_manager.go
+++ b/protocol/chainlib/consumer_websocket_manager.go
@@ -25,7 +25,33 @@ var (
 	MaxIdleTimeInSeconds = int64(20 * 60)   // 20 minutes of idle time will disconnect the websocket connection
 )
 
+// The server pings a live connection every keep-alive interval. Proxies in front
+// of the router (Cloudflare, the Envoy Gateway) close a silent WebSocket after
+// ~2 minutes and send no close frame, so an idle subscription reaches the client
+// as a transport error rather than an unsubscribe; pinging inside that window
+// keeps it open. The interval is a startup flag, held in an atomic so tests can
+// shorten it while other connections are running. Zero or less disables pings.
+var webSocketKeepAliveInterval = func() *atomic.Int64 {
+	interval := &atomic.Int64{}
+	interval.Store(int64(DefaultWebSocketKeepAliveInterval))
+	return interval
+}()
+
+func SetWebSocketKeepAliveInterval(interval time.Duration) {
+	webSocketKeepAliveInterval.Store(int64(interval))
+}
+
+func GetWebSocketKeepAliveInterval() time.Duration {
+	return time.Duration(webSocketKeepAliveInterval.Load())
+}
+
 const (
+	DefaultWebSocketKeepAliveInterval = 30 * time.Second
+	// websocketWriteTimeout bounds a single frame write so the connection
+	// goroutine is never pinned by a client that stopped reading: the write fails
+	// on the deadline, the writer exits, and the read loop tears the conn down.
+	websocketWriteTimeout = 10 * time.Second
+
 	WebSocketRateLimitHeader            = "x-lava-websocket-rate-limit"
 	WebSocketOpenConnectionsLimitHeader = "x-lava-websocket-open-connections-limit"
 
@@ -126,6 +152,9 @@ func (cwm *ConsumerWebsocketManager) ListenToMessages(ctx context.Context) {
 	websocketConn := cwm.websocketConn
 	logger := cwm.rpcConsumerLogs
 
+	// Closed by the writer goroutine on its way out.
+	writerDone := make(chan struct{})
+
 	webSocketCtx, cancelWebSocketCtx := context.WithCancel(context.Background())
 	guid := utils.GenerateUniqueIdentifier()
 	guidString := strconv.FormatUint(guid, 10)
@@ -133,6 +162,10 @@ func (cwm *ConsumerWebsocketManager) ListenToMessages(ctx context.Context) {
 	utils.LavaFormatDebug("consumer websocket manager started", utils.LogAttr("GUID", webSocketCtx))
 	defer func() {
 		cancelWebSocketCtx() // In case there's a problem make sure to cancel the connection
+		// gofiber recycles the connection the moment this handler returns, so wait
+		// for the writer goroutine to stop touching it — otherwise a keep-alive ping
+		// can land on a released conn. websocketWriteTimeout bounds the wait.
+		<-writerDone
 		utils.LavaFormatDebug("consumer websocket manager stopped", utils.LogAttr("GUID", webSocketCtx))
 	}()
 
@@ -149,13 +182,40 @@ func (cwm *ConsumerWebsocketManager) ListenToMessages(ctx context.Context) {
 		}
 	}
 
-	// Single writer goroutine: serves both regular messages and the server-shutdown
-	// close frame. Single-writer discipline guarantees we never call WriteMessage
-	// concurrently — the underlying gorilla/fasthttp websocket library is not safe
-	// for concurrent writes.
+	// writeFrame is the writer goroutine's only way to the connection. The
+	// deadline keeps a stalled client from blocking the goroutine forever, which
+	// would in turn stall the handler waiting on writerDone.
+	writeFrame := func(messageType int, data []byte) error {
+		_ = cwm.websocketConn.SetWriteDeadline(time.Now().Add(websocketWriteTimeout))
+		return cwm.websocketConn.WriteMessage(messageType, data)
+	}
+
+	// Single writer goroutine: serves regular messages, the keep-alive ping and the
+	// server-shutdown close frame. Single-writer discipline guarantees we never call
+	// WriteMessage concurrently — the underlying gorilla/fasthttp websocket library
+	// is not safe for concurrent writes.
 	go func() {
+		defer close(writerDone)
+		// A nil channel blocks forever, so a non-positive interval disables the
+		// keep-alive without a second select arm.
+		var keepAlive <-chan time.Time
+		if interval := GetWebSocketKeepAliveInterval(); interval > 0 {
+			keepAliveTicker := time.NewTicker(interval)
+			defer keepAliveTicker.Stop()
+			keepAlive = keepAliveTicker.C
+		}
 		for {
 			select {
+			case <-keepAlive:
+				// Ping from inside the writer goroutine to keep the single-writer
+				// discipline. A failed write means the connection is already gone:
+				// unblock the read loop with an immediate read deadline, the same way
+				// the shutdown branch below does, and leave the Close to the handler.
+				if err := writeFrame(websocket.PingMessage, nil); err != nil {
+					utils.LavaFormatTrace("error writing keep-alive ping to the websocket")
+					_ = cwm.websocketConn.SetReadDeadline(time.Now())
+					return
+				}
 			case <-ctx.Done():
 				// Server-wide shutdown — send a CloseGoingAway (1001) frame so the client can
 				// distinguish an intentional shutdown from a crash, then unblock the read loop by
@@ -166,7 +226,7 @@ func (cwm *ConsumerWebsocketManager) ListenToMessages(ctx context.Context) {
 				// goroutine is still inside Close() — a use-after-recycle data race. SetReadDeadline
 				// returns immediately (the read loop breaks on the resulting error) and lets the handler,
 				// the conn's sole owner, do the single Close on return.
-				_ = cwm.websocketConn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseGoingAway, "server shutting down"))
+				_ = writeFrame(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseGoingAway, "server shutting down"))
 				_ = cwm.websocketConn.SetReadDeadline(time.Now())
 				return
 			case <-webSocketCtx.Done():
@@ -176,7 +236,7 @@ func (cwm *ConsumerWebsocketManager) ListenToMessages(ctx context.Context) {
 				if !ok {
 					return
 				}
-				if err := cwm.websocketConn.WriteMessage(msg.messageType, msg.msg); err != nil {
+				if err := writeFrame(msg.messageType, msg.msg); err != nil {
 					utils.LavaFormatTrace("error writing msg to the websocket")
 					return
 				}
--- a/protocol/chainlib/jsonRPC.go
+++ b/protocol/chainlib/jsonRPC.go
@@ -565,6 +565,16 @@ func (apil *JsonRPCChainListener) Serve(ctx context.Context, cmdFlags common.Con
 		return err
 	}
 	app.Post("/*", handlerPost)
+	// A caller that guesses the bare endpoint URL (wss://host/) sends a GET
+	// carrying the upgrade headers — the ws/wss scheme never reaches the server,
+	// so that header is the only signal. Serve any such GET from the handler
+	// /ws and /websocket already use; a GET without it still answers 405.
+	app.Get("/*", func(fiberCtx *fiber.Ctx) error {
+		if !websocket.IsWebSocketUpgrade(fiberCtx) {
+			return fiber.ErrMethodNotAllowed
+		}
+		return fiberCtx.Next()
+	}, wsUpgradeMiddleware, websocketCallbackWithDappID)
 	// Go
 	addrChannel := make(chan string)
 	addrChannelSafe := common.NewSafeChannelSender(ctx, addrChannel)
--- a/protocol/chainlib/tendermintRPC.go
+++ b/protocol/chainlib/tendermintRPC.go
@@ -603,7 +603,16 @@ func (apil *TendermintRpcChainListener) Serve(ctx context.Context, cmdFlags comm
 	}
 
 	app.Post("/*", handlerPost)
-	app.Get("/*", handlerGet)
+	// A caller that guesses the bare endpoint URL (wss://host/) sends a GET
+	// carrying the upgrade headers — the ws/wss scheme never reaches the server,
+	// so that header is the only signal. Serve any such GET from the handler
+	// /ws and /websocket already use; every other GET is a URI-style query.
+	app.Get("/*", func(fiberCtx *fiber.Ctx) error {
+		if !websocket.IsWebSocketUpgrade(fiberCtx) {
+			return handlerGet(fiberCtx)
+		}
+		return fiberCtx.Next()
+	}, wsUpgradeMiddleware, websocketCallbackWithDappID)
 	//
 	// Go
 	addrChannel := make(chan string)
--- a/protocol/chainlib/jsonRPC_test.go
+++ b/protocol/chainlib/jsonRPC_test.go
@@ -799,3 +799,72 @@ func TestJsonRPCChainListener_GracefulShutdown_RejectsNewConnectionsAfterShutdow
 
 // Quiet the unused-import warning if sync isn't used.
 var _ = sync.WaitGroup{}
+
+// A WebSocket client that dials the bare endpoint URL (wss://host/) upgrades on
+// any path, because the ws/wss scheme never reaches the server — only the
+// upgrade header does. /ws and /websocket keep working, and a GET without the
+// header still answers 405.
+func TestJsonRPCChainListener_WebSocketUpgradesOnAnyPath(t *testing.T) {
+	serveCtx, cancelServe := context.WithCancel(context.Background())
+	defer cancelServe()
+	_, addr := startTestJsonRPCListener(t, serveCtx, false)
+
+	for _, path := range []string{"/", "/ws", "/websocket", "/lava@dapp"} {
+		client, resp, err := websocket.DefaultDialer.Dial("ws://"+addr+path, nil)
+		require.NoError(t, err, "WS dial on %q should upgrade", path)
+		require.Equal(t, http.StatusSwitchingProtocols, resp.StatusCode, "path %q", path)
+		_ = client.Close()
+	}
+}
+
+func TestJsonRPCChainListener_PlainGetIsStillMethodNotAllowed(t *testing.T) {
+	serveCtx, cancelServe := context.WithCancel(context.Background())
+	defer cancelServe()
+	_, addr := startTestJsonRPCListener(t, serveCtx, false)
+
+	httpClient := &http.Client{Timeout: 2 * time.Second}
+	resp, err := httpClient.Get("http://" + addr + "/")
+	require.NoError(t, err)
+	defer resp.Body.Close()
+	require.Equal(t, http.StatusMethodNotAllowed, resp.StatusCode)
+}
+
+// An idle subscription must not go silent: proxies in front of the router close
+// a quiet WebSocket after ~2 minutes without a close frame, so the server pings
+// on an interval to keep the connection alive.
+func TestJsonRPCChainListener_WebSocketSendsKeepAlivePings(t *testing.T) {
+	previousInterval := GetWebSocketKeepAliveInterval()
+	SetWebSocketKeepAliveInterval(50 * time.Millisecond)
+	defer SetWebSocketKeepAliveInterval(previousInterval)
+
+	serveCtx, cancelServe := context.WithCancel(context.Background())
+	defer cancelServe()
+	_, addr := startTestJsonRPCListener(t, serveCtx, false)
+
+	client, _, err := websocket.DefaultDialer.Dial("ws://"+addr+"/", nil)
+	require.NoError(t, err)
+	defer client.Close()
+
+	pinged := make(chan struct{}, 1)
+	client.SetPingHandler(func(string) error {
+		select {
+		case pinged <- struct{}{}:
+		default:
+		}
+		return nil
+	})
+	// Control frames are only processed while a read is in flight.
+	go func() {
+		for {
+			if _, _, readErr := client.ReadMessage(); readErr != nil {
+				return
+			}
+		}
+	}()
+
+	select {
+	case <-pinged:
+	case <-time.After(5 * time.Second):
+		t.Fatal("no keep-alive ping arrived within 5s")
+	}
+}
--- a/protocol/chainlib/tendermintRPC_test.go
+++ b/protocol/chainlib/tendermintRPC_test.go
@@ -500,3 +500,19 @@ func TestTendermintRpcChainListener_GracefulShutdown_RejectsNewConnectionsAfterS
 	}
 	require.Error(t, err, "GET after Shutdown should fail with connection error")
 }
+
+// Same as the jsonrpc listener: the bare endpoint URL upgrades. Tendermint's
+// URI-style GETs are unaffected — they carry no upgrade header and still reach
+// the relay handler.
+func TestTendermintRpcChainListener_WebSocketUpgradesOnAnyPath(t *testing.T) {
+	serveCtx, cancelServe := context.WithCancel(context.Background())
+	defer cancelServe()
+	_, addr := startTestTendermintListener(t, serveCtx)
+
+	for _, path := range []string{"/", "/ws", "/websocket", "/lava@dapp"} {
+		client, resp, err := websocket.DefaultDialer.Dial("ws://"+addr+path, nil)
+		require.NoError(t, err, "WS dial on %q should upgrade", path)
+		require.Equal(t, http.StatusSwitchingProtocols, resp.StatusCode, "path %q", path)
+		_ = client.Close()
+	}
+}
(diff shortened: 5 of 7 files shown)


avitenzer:
@nimrod-teich I reviewed this and pushed one commit onto the branch — **8220695** `fix(chainlib): release an idle websocket the client refuses to close`. Please review it as part of the PR; happy to drop it into a separate PR if you'd rather keep this one to the two original defects.

## What I found

The keep-alive is correct, but it quietly removes a backstop the idle reaper was relying on.

`consumer_websocket_manager.go` — the reaper sends its close frame through `sendWS` and returns. It never cancels `webSocketCtx` and never sets a read deadline, so teardown depends entirely on the client closing in response. Today a client that ignores that frame leaves a **silent** connection, and the proxy in front of the router reaps it within ~2 minutes — the exact behaviour this PR is fixing. That close woke `ReadMessage` and let the handler return.

Once we ping every 30s the connection is never silent, so the proxy won't reap it and nothing else will: the handler goroutine, its 128 KiB fasthttp read buffer and the per-IP limiter slot are held for the life of the process. That is the MAG-3722 leak reached by a different route.

## The fix

Frames gained a `final` flag. The writer puts a final frame on the wire and then tears down **exactly as your `ctx.Done()` shutdown branch already does** — immediate read deadline, read loop wakes, handler does the single `Close`. No new constant and no grace period: your existing `TestJsonRPCChainListener_GracefulShutdown_SendsWSCloseFrame_1001` already proves that sequence delivers the close code to the client rather than an RST. Every `conn` access stays on the writer goroutine, so the reaper needs no synchronisation of its own.

## The part worth arguing about

To regression-test this I have to shorten `MaxIdleTimeInSeconds`, and `-race` fails immediately: it is a plain global read by every live connection's reaper, so the write races connections left over from earlier tests. Benign in production (written once during flag parsing), not benign under test.

I converted it to an atomic behind `Set/GetMaxIdleTimeInSeconds`, mirroring what you did for `webSocketKeepAliveInterval` 30 lines above and for the same stated reason. That moves its flag from `Int64Var(&chainlib.MaxIdleTimeInSeconds, ...)` to the `viper.GetInt64` read the keep-alive flag already uses — **this is the bit I'd most like your call on**, since it changes operator-facing plumbing for a flag that isn't otherwise part of this PR. The alternative was draining `wsWG` in both test helpers, which papers over the shared global instead of fixing it.

Because that style silently yields `0` if the viper binding ever breaks — and `0` disables the keep-alive ping and the idle reaper outright, with no error anywhere — I added `websocket_tunables_cobra_viper_test.go`, modelled on `resp_cobra_viper_test.go`. It drives the real cobra command and asserts both flags resolve from flags, YAML and their registered defaults.

## Verification

- `go build ./...`, `go vet`, `golangci-lint run protocol/chainlib/...` → 0 issues
- `go test ./protocol/chainlib/ ./protocol/rpcsmartrouter/ -race -count=1` → ok, on a clean worktree checkout of 8220695
- Both new behavioural tests fail on e6b1d29 with the production change reverted and pass with it:
  - `a final frame must end the writer goroutine`
  - `the idle connection's handler never returned: the router is holding it for the life of the process`
- Also confirmed your `<-writerDone` wait is genuinely covered: removing it makes `TestJsonRPCChainListener_WebSocketSendsKeepAlivePings` fail with a `DATA RACE`, so the new ping test doubles as the regression for it.

## Three smaller things I did not change

1. **`health-check-path: /` silently shadows the bare-URL upgrade.** `createAndSetupBaseAppListener` registers `app.Get(healthCheckPath, ...)` before the new `/*` catch-all, so `GET /` would answer `200 Health status OK` and never upgrade. The default is `/lava/health` and nothing in this repo or smart-router-enterprise-internal overrides it — but `lavasession/consumer_types.go:816` documents the field as *`default is "/"`*. An operator following that comment hits the exact failure this PR fixes, silently. Cheapest fix is the stale comment; the robust one is letting the health handler fall through when `websocket.IsWebSocketUpgrade(c)`.

2. **The ceiling moves from ~2 min to ~20 min, not to "no limit".** `idleFor` is reset on inbound frames and on outbound subscription deliveries, so a subscription that is subscribed but producing no events resets neither and the router's own reaper still closes it at `MaxIdleTimeInSeconds`. Pings deliberately don't touch `idleFor` — correct, or the reaper would never fire — but it's worth a line in the PR body so operators know to raise `--limit-websocket-connection-idle-time` for long-lived quiet subscriptions.

3. **No pong deadline**, unlike `chainproxy/rpcclient/websocket.go:202-204`, which pairs its ping with `SetReadDeadline(now + wsPongTimeout)`. I think not dropping clients that fail to pong is the right call for a public-facing endpoint, but since the sibling implementation does the opposite it's worth a sentence saying so.

Routing-wise the rest checks out: `/ws`, `/websocket` and the health route are all registered before the catch-all so nothing is shadowed, the 405 preservation is real, and `HEAD` + upgrade headers (fiber's `Get` registers `HEAD` too) is safe — fasthttp's upgrader rejects non-GET, the `!= 101` check fires and the `wsWG.Add` stays balanced.

avitenzer (commented):
@nimrod-teich — branch housekeeping first, then two blockers and three smaller notes.

**I changed the branch.** #386 squash-merged, `mag-3722` was deleted and GitHub retargeted this to `main`, which left it conflicting: the branch still carried #386's eight pre-squash commits, content-identical to the squash commit but with different shas. I rebased the three real commits onto `main` (`--onto origin/main 626cd65`) and force-pushed. The rebase changed no content — `git diff old-base...old-head` and `git diff cc92100...new-head` are byte-identical. It's now 3 commits / 9 files / +398-13, down from 11 / 25 / +1658-260.

**Two of those three commits are mine**, so they need your eyes specifically — nobody else has looked at them:
- `f36cda1` — the idle reaper sent its close frame and returned, leaving teardown to the client. Before the keep-alive, a client that ignored it was reaped by the proxy within ~2 min, which woke the read loop for us. A pinged connection is never silent, so that backstop is gone. Frames now carry `final`; the writer tears down using the same sequence your `ctx.Done()` branch already uses.
- `a0498fe` — **your catch-all dropped the `Allow` header from the 405.** Measured against the fork point: `405 Allow: "POST"` → `405 Allow: ""`. Registering a GET route takes fiber's method-mismatch path out of play, and that path is what attached `Allow`. RFC 9110 §15.5.6 requires it, and the body says the answer is unchanged. Fixed, and the guarding test now asserts the header — it asserted status only, which is how this slipped through.

**Verification** (this repo's CI runs no Go tests, so local is the only evidence): on the rebased tree, `go build ./...`, `go vet ./...`, `golangci-lint` clean; `go test ./protocol/chainlib/ ./protocol/rpcsmartrouter/ -race -count=1` both ok. Both of my behavioural tests fail with the production change reverted. Your `<-writerDone` wait is genuinely covered — removing it makes the ping test fail with a DATA RACE.

**Not fixed, left for you** — two blockers inline below, plus three non-blocking notes.

Also outside the diff:
- The PR body still describes only your commit — 55% of the diff is undocumented for a reviewer.
- `scripts/pre_setups/test_smartrouter_sol.sh:35-36` now states something false: *"the root `/` only accepts HTTP POST and returns 405 on a ws upgrade"*.
- `lavasession/consumer_types.go:816` documents `health-check-path` as *`default is "/"`* when it's `/lava/health` (`common/endpoints.go:76`) — stale before this PR, but it's what makes the health-route blocker below reachable by someone following the comment.
- One `DATA RACE` in `TestSolanaRouterIDSurvivesUpstreamReconnect`, 1 of 5 full-package runs on the pre-rebase head; not reproduced in 4 more and I lost the stack. Clean at the fork point and on the rebased tree. Flagging it as unclassified rather than dismissing it.

protocol/chainlib/jsonRPC.go
```diff
@@ -565,6 +565,23 @@ func (apil *JsonRPCChainListener) Serve(ctx context.Context, cmdFlags common.Con
 		return err
 	}
 	app.Post("/*", handlerPost)
+	// A caller that guesses the bare endpoint URL (wss://host/) sends a GET
+	// carrying the upgrade headers — the ws/wss scheme never reaches the server,
+	// so that header is the only signal. Serve any such GET from the handler
+	// /ws and /websocket already use; a GET without it still answers 405.
+	app.Get("/*", func(fiberCtx *fiber.Ctx) error {
```
avitenzer, line 574: **Blocker — the health route shadows this, so the rule the PR body states isn't the rule that ships.**

`createAndSetupBaseAppListener` registers `app.Get(healthCheckPath, ...)` at `common.go:713`, which runs before `Serve` reaches this line. Fiber matches in registration order and that handler never calls `c.Next()`, so it wins for its own path regardless of the Upgrade header.

Measured on this branch:

```
path=/             UPGRADED  status=101
path=/ws           UPGRADED  status=101
path=/lava/health  no upgrade — request reaches the health handler
```

So even on the default path, *"any GET carrying the upgrade header is a WebSocket, whatever the path"* is already false. Worse, `HealthCheckPath` is operator-configurable (`lavasession/consumer_types.go:816`, yaml key `health-check-path`) — set it to `/` and the headline fix is defeated outright: `wss://host/` answers `200 Health status OK`, which is the exact bug MAG-3344 was filed for.

Neither new test can catch it: both harnesses hardcode `HealthCheckPath: "/lava/health"` (`jsonRPC_test.go:717`, `tendermintRPC_test.go:437`).

Suggested: let the health handler fall through when `websocket.IsWebSocketUpgrade(c)`, and add a `HealthCheckPath: "/"` case to the upgrade test. Same applies at `tendermintRPC.go:610`.

protocol/chainlib/consumer_websocket_manager.go
```diff
@@ -224,7 +315,7 @@ func (cwm *ConsumerWebsocketManager) ListenToMessages(ctx context.Context) {
 	idleFor.Store(time.Now().Unix())
 	requestsPerSecond := &atomic.Uint64{}
 	go func() {
-		if WebSocketRateLimit <= 0 && cwm.headerRateLimit <= 0 && MaxIdleTimeInSeconds <= 0 {
+		if WebSocketRateLimit <= 0 && cwm.headerRateLimit <= 0 && GetMaxIdleTimeInSeconds() <= 0 {
```
avitenzer, line 378: **Blocker — `--limit-websocket-connection-idle-time 0` plus the keep-alive holds a connection for the life of the process.**

`WebSocketRateLimit` defaults to `-1` (line 23) and `headerRateLimit` is 0 unless a header sets it, so with idle=0 all three are `<= 0` and this goroutine returns immediately. The reaper never runs, so the `final` close frame from the next commit never fires either — and the keep-alive ping means the connection is no longer silent, so nothing upstream reaps it.

Net: the handler goroutine, its 128 KiB fasthttp read buffer (`common.go:672`) and the per-IP limiter slot are held until the process exits. That is the MAG-3722 leak reached by a third route.

Latent today — nothing in this repo or `smart-router-enterprise-internal` sets 0 — but `websocket_tunables_cobra_viper_test.go` documents 0 as supported, so it's reachable by a supported flag value.

Suggested: when the reaper is disabled, don't start the keep-alive either — or warn at startup that the combination holds connections indefinitely.

protocol/chainlib/consumer_websocket_manager.go
```diff
@@ -22,10 +22,55 @@ import (
...
+	return time.Duration(webSocketKeepAliveInterval.Load())
+}
+
 const (
+	DefaultMaxIdleTimeInSeconds       = int64(20 * 60) // 20 minutes of idle time will disconnect the websocket connection
+	DefaultWebSocketKeepAliveInterval = 30 * time.Second
+	// websocketWriteTimeout bounds a single frame write so the connection
+	// goroutine is never pinned by a client that stopped reading: the write fails
+	// on the deadline, the writer exits, and the read loop tears the conn down.
+	websocketWriteTimeout = 10 * time.Second
```
avitenzer: Non-blocking, but this is the kind of constant that usually gets sent back here.

`websocketWriteTimeout` is a new hard ceiling on how slow a subscriber may be before the router tears its connection down — there was no write deadline at all before this PR. Three things about it:

- not derived from any bound; the repo's own comparable is `wsPingWriteTimeout = 5 * time.Second` (`chainproxy/rpcclient/websocket.go:36`)
- no flag, unlike the other two tunables this PR adds
- no test — both fakes return `nil` from `SetWriteDeadline`, so nothing exercises the slow-write teardown it introduces

A derivation from an existing bound, or a flag, would settle it.

protocol/chainlib/consumer_websocket_manager.go
```diff
@@ -109,30 +154,64 @@ func (cwm *ConsumerWebsocketManager) handleRateLimitReached(inpData []byte) ([]b
...
+	var keepAlive <-chan time.Time
+	if interval := GetWebSocketKeepAliveInterval(); interval > 0 {
+		keepAliveTicker := time.NewTicker(interval)
+		defer keepAliveTicker.Stop()
+		keepAlive = keepAliveTicker.C
+	}
+
 	for {
 		select {
+		case <-keepAlive:
```
avitenzer, line 265: Two differences from the sibling implementation at `chainproxy/rpcclient/websocket.go:187-207`, both of which look deliberate but aren't stated:

- **No pong deadline.** That one pairs its ping with `SetReadDeadline(now + wsPongTimeout)` so a half-open peer gets detected. Here a ping into a black-holed socket succeeds into the kernel buffer and nothing notices. I think *not* dropping clients that fail to pong is the right call for a public-facing endpoint — worth a sentence saying so, given the file next door does the opposite.
- **Pings unconditionally.** The sibling uses a `pingReset` channel so it only pings an idle connection; this fires every 30s even on a busy one.

The constants also diverge with no cross-reference: 30s/10s here, 60s/5s/30s there.

Non-blocking.

protocol/chainlib/tendermintRPC.go
```diff
@@ -603,7 +603,16 @@ func (apil *TendermintRpcChainListener) Serve(ctx context.Context, cmdFlags comm
 	}
 
 	app.Post("/*", handlerPost)
-	app.Get("/*", handlerGet)
+	// A caller that guesses the bare endpoint URL (wss://host/) sends a GET
+	// carrying the upgrade headers — the ws/wss scheme never reaches the server,
+	// so that header is the only signal. Serve any such GET from the handler
+	// /ws and /websocket already use; every other GET is a URI-style query.
+	app.Get("/*", func(fiberCtx *fiber.Ctx) error {
```
avitenzer, line 612: The one safety property the PR body claims for this listener — *"Tendermint's URI-style GETs are untouched"* — has no test.

`TestTendermintRpcChainListener_WebSocketUpgradesOnAnyPath` only dials websockets on four paths; nothing drives a plain GET through the `handlerGet` branch added here. It can't, as the harness stands: `startTestTendermintListener` builds the listener with a nil `relaySender` (`tendermintRPC_test.go:441`), and `handlerGet` dereferences `apil.relaySender.SendRelay` at `tendermintRPC.go:567`. fasthttp doesn't recover handler panics, so it would take the test binary down rather than fail cleanly.

So the branch this PR inserts in front of the tendermint relay path is the piece of it with zero coverage. Suggested: a stub `relaySender` in the harness plus one plain-GET assertion.

protocol/chainlib/jsonRPC.go
nimrod-teich: Confirmed and fixed in 5159e59: the health handler in `createAndSetupBaseAppListener` now hands an upgrade on with `c.Next()`, so the catch-all serves it whatever the health path is.

Both listener harnesses take the health path as a parameter now, and `TestJsonRPCChainListener_WebSocketUpgradesOnTheHealthPath` plus its Tendermint twin run with `/` and with `/lava/health`: a dial on the health path gets 101, a plain GET there still gets `200 Health status OK`. With the fall-through removed all four cases fail on the dial, so the shape you measured is pinned.

The `consumer_types.go` comment you traced this to now names `common.DEFAULT_HEALTH_PATH` and says an upgrade on the health path is still served as a websocket.

protocol/chainlib/consumer_websocket_manager.go
nimrod-teich: Confirmed. I took the warning rather than the coupling, and want to say why since you offered both.

With idle=0 the operator asked for no router-side limit. Before the keep-alive that flag never actually delivered it behind a proxy, which reaped the socket at ~2 minutes regardless. Coupling the two (idle=0 switches pings off) would bring that reap back for precisely the operators who asked for no limit, so the flag would keep lying. What is left when both are set is a connection the client is holding open and acknowledging pings on, bounded per IP by the connection limiter. A client that vanished is caught by the TCP retransmission timeout once its pings stop being acknowledged; one that stopped reading by the write deadline.

`WebSocketKeepAliveOutlivesIdleReaper` decides the warning (table-tested in `consumer_websocket_tunables_test.go`) and `rpcsmartrouter.go` logs it right after the flags are read, with both values. The PR body says the same under 991047e. If you would rather the combination refuse to start, that is a one-line change at the same site.

protocol/chainlib/consumer_websocket_manager.go
nimrod-teich: Agreed on all three. In 991047e it is `--websocket-write-timeout`, default 10s, `0` disables, held in an atomic behind `Set`/`GetWebSocketWriteTimeout` exactly like the two beside it and read once per writer.

`TestWebsocketWriter_BoundsEveryWrite` has the fake record every `SetWriteDeadline` and asserts each frame goes out under a deadline of the configured length, and that `0` sets none. `websocket_tunables_cobra_viper_test.go` covers the flag from flags, YAML and default.

I did not derive it from `wsPingWriteTimeout`: that one bounds a ping to a node the router chose, this one bounds a customer that stopped reading, and the two have no reason to move together. The slow-subscriber teardown end to end (socket buffers filling, then the deadline firing) is left to the library's deadline semantics rather than a buffer-filling integration test; if you want that one as well I can add it, it would be a test that needs a few hundred KB of frames and a sub-second timeout.

protocol/chainlib/consumer_websocket_manager.go
nimrod-teich: Both stated now, in the comment above the arm (991047e).

No pong deadline is deliberate: the sibling owns a connection to a node the router chose and drops it on a late pong; the peer here is a customer on whatever network they have, and a late pong must not end their subscription. A vanished peer is still caught, by the TCP retransmission timeout once the pings stop being acknowledged, and one that stopped reading by the write deadline.

The unconditional ping is gone: every frame the writer sends resets the ticker, so a busy connection is never pinged and a quiet one gets one ping per full interval of silence, which is the silence the proxy measures. Simpler than the sibling's `pingReset` channel because everything here is already on the writer goroutine. `TestWebsocketWriter_PingsOnlyAfterSilence` drives frames at a quarter of the interval and sees no ping, then goes quiet and sees them resume.

The constants stay different and the comment says so; they measure different things.

protocol/chainlib/tendermintRPC.go
nimrod-teich: Right, and it would have panicked rather than failed. 5159e59 adds `tendermintGetRelayStub`, the four-method slice of `RelaySender` the GET path uses (records the url it is asked for, answers a fixed reply), and `startTestTendermintListenerWithOptions` takes it.

`TestTendermintRpcChainListener_PlainGetReachesTheRelayPath`: `GET /status?height=7` returns 200 with the stub's reply and the stub saw exactly `status?height=7`; an upgrade on `/status` then gets 101 and the stub is still at one call. So the branch in front of `handlerGet` is covered in both directions.

nimrod-teich:
Thanks for the rebase and the two commits; the `Allow` catch in particular is one I would have shipped without. All five threads are answered inline. The three follow-up commits are 5159e59, 991047e and 9d30c25, and the body now describes all six.

On the items outside the diff:

- **PR body**: rewritten to cover every commit, with behaviour notes on the 20-minute idle ceiling for a quiet subscription and the new write bound.
- **`test_smartrouter_sol.sh` and `consumer_types.go:816`**: fixed in 5159e59.
- **The `DATA RACE`**: reproduced, 1 in 25 runs of `TestSolanaRouterIDSurvivesUpstreamReconnect` under `-race`. Write in `handleUpstreamDisconnect`, the reconnect replacing `activeSub.messagesChan` under the lock, against the listener's unlocked read of that field in its select. In production a reconnect is only ever started from the listener's own error branch, after which that listener returns, so the ordering the test creates (the reconnect driven directly while the first listener still selects) does not occur there, but the unlocked read was real. 9d30c25 makes the channel a listener parameter: each listener owns one channel for its life and the handoff on reconnect is explicit. 25 further runs are clean.
- **Verification on the new head**: `go test -race -count=1 ./protocol/chainlib/... ./protocol/rpcsmartrouter/ ./protocol/lavasession/` all ok; `golangci-lint run --new-from-rev origin/main ./protocol/...` 0 issues; the health-path tests fail with the fall-through removed and pass with it.

avitenzer (approved):

PR #357, opened 2026-09-01
ci: re-apply chart 6 PR-gate target and park dev-sim-prtests steps

nimrod-teich:
Jira ticket: MAG-3351

#Closes MAG-3351

Reverts #356 and takes the dev-sim-prtests box out of the PR gate instead.

## Why

The gate targets the chart-6 layout (`smart-router` namespace, `app.smart-router-id` selector), but the dev-sim-prtests box still runs the pre-chart-6 `lava-infra` layout. #356 papered over that by pointing the gate back at `lava-infra`, which reintroduces the old label the chart no longer emits.

## What

- Revert #356, so the workflow keeps the chart-6 namespace, selector and pull-secret source.
- Comment out the three box-dependent steps of the `build` job — `Check dev-sim-prtests access`, `Deploy and validate dev-sim-prtests Kubernetes rollout`, and `Deploy and validate latest Helm chart with PR image` — with a `TEMPORARY (MAG-3351)` header naming the restore condition.

- Comment out the `preflight-status` job as well. With nothing behind the status there is nothing to report, so `dev-sim-prtests / preflight` is not published at all rather than rubber-stamped green.

The non-blocking smart-router-automation smoke is untouched.

⚠️ The `Require PR Gate on main` repository ruleset still lists `dev-sim-prtests / preflight` as a required status check. A required check that is never published never arrives, so that ruleset has to be set to `disabled` in lockstep with this merge, and re-enabled when the gate comes back.

## Test plan

- `yaml.safe_load` parses the workflow; the `build` job's step list no longer contains any of the three commented steps, and `metadata` / `build` / `preflight-status` are all intact.
- `git diff --check` is clean.
- No uncommented `lava-infra` or `app.lavanet.io` reference remains in the workflow.

## Follow-up

Uncomment the block once the dev-sim-prtests box is migrated to chart 6 in the `smart-router` namespace.

### DIFF
--- a/.github/workflows/pr-gate-build-artifact.yml
+++ b/.github/workflows/pr-gate-build-artifact.yml
@@ -201,1090 +201,1095 @@ jobs:
           } >> "$GITHUB_OUTPUT"
           echo "Checked out smart-router-helm-chart main at ${chart_sha} (${chart_path}, chart ${chart_version})"
 
-      - name: Check dev-sim-prtests access
-        env:
-          SSH_HOST: ${{ secrets.DEV_SIM_PRTESTS_SSH_HOST }}
-          SSH_KEY: ${{ secrets.DEV_SIM_PRTESTS_SSH_KEY }}
-          PR_NUMBER: ${{ inputs.pr_number }}
-          HEAD_SHA: ${{ needs.metadata.outputs.head_sha }}
-        run: |
-          set -euo pipefail
-          preflight_result="failed"
-
-          write_summary() {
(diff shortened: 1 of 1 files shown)


avitenzer (approved):

idanhen1:
Superseded by #364. The `prtests` box now runs chart 6 in the `smart-router` namespace, so the real box-dependent gate stays enabled. Full end-to-end validation passed in run 33614410740.

PR #358, opened 2026-09-01
feat(chainlib): admit a provider per collection at boot (MAG-3326)

avitenzer:
# Description

Supersedes [#350](https://github.com/Magma-Devs/smart-router/pull/350), which was stacked on [#345](https://github.com/Magma-Devs/smart-router/pull/345). #345 has since merged, so this is the same commit rebuilt on `main` and targeting `main` directly. The delta is byte-for-byte identical to `f9db844` on the old branch — only hunk context shifted.

Admission was all-or-nothing. `ChainFetcher.Validate` returns on the first `Fail`-severity verification, so a provider serving the base collection perfectly and failing **one** add-on's verification was excluded from everything. Attach an archive url to an otherwise healthy provider, have it fail the archive check, lose the provider.

That is a live availability risk on every chain and has nothing to do with the add-on split that surfaced it — it is the reason this is worth scheduling rather than filing and forgetting. It is also step 3 of MAG-3296's own suggested fix: *"exclude a provider from the collections it fails, not from the whole (chain-id, api_interface) pair."*

## What changed

`ValidateCollections` runs the same verifications and **attributes** each failure. A `Fail` belonging to one add-on refuses that add-on and the provider keeps the rest. `ProviderAdmission.Apply` returns a **copy** of the node urls with refused add-ons removed, which the boot path hands to endpoint construction — so session selection, which already filters on addon, does the rest. No new eligibility axis.

The copy is the load-bearing detail: `applyReverification` re-reads the parsed configuration every epoch, and stripping it in place would make one transient failure permanent until restart. That is exactly the failure this codebase already guards against at provider granularity with `reverifyDemoteThreshold` (MAG-2445 — a 40-second blip costing a provider a full epoch).

## Two deliberate limits

**A failure with no add-on to attribute it to is still fatal** — the base collection's own verifications, a head probe that never answered, `GetVerifications` itself. A node that cannot serve the spec's primary surface has nothing narrower to fall back to.

**A base-collection failure does NOT derive `standalone-addons`.** MAG-3326 floats making that flag inferred rather than declared, and it is the one inference worth refusing: only an operator can tell *"this node serves only EVM by design"* from *"its Substrate side is down right now"*, and guessing the first from the second quietly admits a degraded node on a narrower surface. The explicit flag has no such ambiguity, so the assertion stays explicit.

**Boot only.** `applyReverification` keeps the all-or-nothing `Validate`, because the epoch path needs the same temporal hysteresis per add-on that it already has per provider before it can demote a collection safely. Marked `TODO(MAG-3326)` at the call site.

## Most critical to review

- `protocol/chainlib/chain_fetcher.go` — `ProviderAdmission`, and `Validate`/`ValidateCollections` sharing one loop via a `perCollection` flag so the two paths cannot drift.
- `protocol/rpcsmartrouter/rpcsmartrouter.go` — where the admissions map is written (once, before the closure is first called) and read in `convertProvidersToSessions`.

## Tests

The discriminating pair runs the **same** verification set through both entry points: `ValidateCollections` returns no error and refuses only the add-on, while `Validate` on identical input still returns one. That contrast is the assertion — it fails if the two paths ever converge.

Also covered: `Apply` leaves the input configuration untouched, keys by `(url, internalPath)` so the root path and `/v2` are refused independently, is identity when nothing was refused, and refuses nothing when verifications are skipped entirely.

Verified from a clean worktree of the pushed commit, not the working directory:

- `go build ./...` — clean.
- `go test ./protocol/chainlib/... ./protocol/rpcsmartrouter/... ./protocol/lavasession/... ./protocol/endpointstate/...` — all green.
- The seven new tests confirmed to execute (not skipped) under `-run 'ValidateCollections|PerCollection|Admission' -v`.

## Not covered

Not exercised against a live upstream, unlike #344/#345 — the discriminating case needs an upstream that passes the base collection and fails one add-on, which none of the Acala or Shiden endpoints does. The unit tests drive both paths through the same mocked verification set instead.

## Jira ticket

Jira ticket: MAG-3326

---

## Author Checklist

I have...


AnnaR-prog (commented):
## Review

Reviewing my own commit rebased by someone else, so I checked the rebase mechanically rather than eyeballing it, then re-read the change with fresh eyes. **The supersede claim is true.** Two findings in the underlying change — both mine originally, both the same theme: per-collection refusal interacts with lifetime in ways the commit does not account for. One I would want addressed before merge.

### Verifying the supersede claim

`da45c9b` vs the original `f9db844`, comparing only added/removed lines so hunk-context shift is excluded:

```
old delta lines: 390 | new: 390
IDENTICAL: added/removed lines match exactly
```

Worth checking specifically because `f9db844` predates `9aae099` (the #345 review fix that gave `FetchLatestBlockNumForCollection` its `allowBaseFallback` parameter), and `validate()` calls it. A byte-identical patch applied to a changed base is exactly where that argument could have been dropped. It was not — both call sites on the branch still pass `url.ServesBaseCollection()` (`chain_fetcher.go:273`, `:373`). Build, vet, gofmt and the full `./protocol/...` suite are green on `da45c9b`; CI is 20 pass / 1 skipped.

---

### 1. A `standalone-addons` url whose only add-on is refused flips to serving the base collection

`Apply` strips the refused add-on and leaves `StandaloneAddons` set. `ServesBaseCollection()` is `!StandaloneAddons || len(Addons) == 0`, so emptying `Addons` makes it return **true**. Demonstrated:

```
before:         Addons=[evm] Standalone=true  ServesBase=false
after refusal:  Addons=[]    Standalone=true  ServesBase=true
```

So an EVM-only Acala url whose `evm` chain-id verification fails is no longer excluded — it is admitted as a **Substrate** provider, and every base-collection request routed to it returns `-32601`. Before this commit that provider was excluded outright, because the failure was fatal under `Validate`.

The boot-path guard from #345 fires but does not save it: `Apply` has already emptied `Addons`, so `SeparateAddonsExtensions` finds none, the flag is downgraded to `false`, and the endpoint is built as an ordinary base-collection one. The warning it logs — *"url declares no add-on collection"* — is also misleading here, since the url did declare one and it was refused.

This is the mirror image of the extension-only finding on #345: that one made an endpoint serve *nothing*, this one makes it serve *the wrong thing*, which is worse because it produces failing relays rather than an idle endpoint.

The fix belongs in `Apply`, evaluated before `Addons` is overwritten — a url that served only its add-ons and kept none has nothing left:

```go
if len(kept) == 0 && !url.ServesBaseCollection() {
    // Served only its add-ons and none survived. Keeping it would hand a node
    // that opted out of the base collection back to base traffic.
    continue
}
```

Dropping the url is safe at the provider level: `rpcsmartrouter.go` already skips a provider whose endpoint list comes out empty.

### 2. An add-on refused at boot never recovers

`admissionsByProvider` is written once at `rpcsmartrouter.go:2610` and read inside `convertProvidersToSessions` at `:2437`. That closure is also handed to the epoch path (`:2656` → `spec_reverifier.go:353`), so every later promotion re-applies the **boot-time** refusal. A transient failure of one add-on's verification during boot costs that add-on for the process lifetime, even though `applyReverification` re-validates the provider every epoch and would otherwise restore it.

That is the same permanence the copy semantics were chosen to avoid. The commit is explicit that the epoch path stays all-or-nothing and TODOs it, which I still think is the right scope — but this consequence is not stated anywhere, and it is a behaviour change: before, a boot-time add-on failure excluded the provider and `retryFailedProviders` brought it back within 2s–3m; now the provider is kept and the add-on is gone until restart.

Not necessarily worth blocking on, but it should be in the commit message and the `TODO(MAG-3326)` rather than discovered later. If it is worth fixing now, the cheapest version is to clear a provider's entry from the map when the epoch path re-validates it cleanly.

---

### Scope

No objection to boot-only, and refusing to derive `standalone-addons` from a base-collection failure is the right call — inferring "serves only EVM by design" from "its Substrate side is down" would quietly admit a degraded node. Finding 1 is arguably the same principle applied one level down: a refused add-on should not be silently reinterpreted as base-collection capability either.

Happy to push both fixes to this branch if you would rather not carry them.

AnnaR-prog (commented):
## Follow-up — correcting my own review above

My first pass verified the rebase and found two lifetime issues. It missed the more important question: **does this change do what it says?** A deeper pass says no. Three findings, all verified against the PR head rather than reasoned about. I wrote this code, so these are my bugs, not the rebase's — the rebase is faithful.

### 1. The motivating case is not fixed — attribution ignores `Extension`

The PR exists to stop *"attach an archive url to an otherwise healthy provider, have it fail the archive check, lose the provider."* That still loses the provider.

`chain_fetcher.go:305` attributes on `verification.Addon` alone. But `VerificationKey` carries **both** `Addon` and `Extension`, and in every real spec the archive check arrives as an **extension on the base collection**. From `specs/ethereum.json`:

```
collection add_on='' | verification=pruning | extension='archive' | severity=None
```

So for `addons: ["archive"]`, `SeparateAddonsExtensions` classes `archive` as an extension, `GetVerifications` returns `VerificationKey{Addon: "", Extension: "archive"}`, and `verification.Addon == ""` makes the guard false — straight to the fatal `return`. Run against this branch with that exact container shape:

```
PR's motivating case -> err=true
                        admission.Any()=false   (nothing refused; provider dropped whole)
```

Same shape in `lava.json`, `cosmoshub.json`, `aptos.json`. The unit tests cannot catch it because `validate_collections_test.go:22` builds `VerificationKey{Addon: addon}` — `{Addon:"archive"}` is a container **no spec ever produces**. The helper encodes the same blind spot as the code.

Minimal fix: attribute on `Addon` **or** `Extension`. `url.Addons` contains `"archive"` verbatim, so `Apply` already matches it without further change.

### 2. The epoch path reverses the rescue within two ticks, permanently

For the add-on-keyed case that *does* work, it survives about 30 minutes.

`applyReverification` probes with the all-or-nothing `validateProvider` (`spec_reverifier.go:200`), whose fetcher endpoint is `newEndpoint(provider.NodeUrls)` (`:538`) — the **unstripped** config, add-on still declared. It fails again, `demoteFailStreak` increments, and at `reverifyDemoteThreshold = 2` the provider is demoted **in full**, add-ons and base alike. It does not come back: promotion needs `err == nil` from the same all-or-nothing `Validate`, and it is not in `failedStaticProviders`, so `retryFailedProviders` never touches it either.

"Boot only" reads as a scoping limit. It is not — the epoch path actively undoes the change, and does so into a worse state than before.

### 3. A `standalone-addons` url whose only add-on is refused flips to serving base

As in my earlier comment, and it compounds with (1) and (2): refusing the last add-on empties `Addons`, `ServesBaseCollection()` flips to true, and a node the operator declared cannot serve the base collection starts receiving base traffic.

---

### Verdict

Net effect on real specs: the headline case is unchanged (1); the case that does work is reverted two ticks later (2); and there is a new mis-routing regression (3). I would not merge this as it stands — not because the rebase is wrong, but because the original commit does not deliver MAG-3326.

Smaller things worth folding into the rework, all verified:

- **Over-attribution.** `admission.fail(url, verification.Addon)` discards `Extension`, so an `{Addon:"trace", Extension:"archive"}` failure refuses *all* of `trace` while leaving `archive` advertised — the opposite of what failed.
- **A head-probe failure is still fatal** even in per-collection mode (`chain_fetcher.go:280`), so an unreachable add-on url — the most common failure — still costs the whole provider.
- **`ValidateCollect` already does this sweep.** It runs every verification for every url without short-circuiting and returns `VerificationResult{Name, Addon, Extension, Ok, Error}` — including the `Extension` finding 1 needs. `ValidateCollections` reimplements a lossy subset of it, and the two exported names differ by one character with different semantics.
- **`validateProviderCollections` is a near-verbatim copy of `validateProvider`**, differing only in the final call — the same `perCollection bool` trick used one level down would avoid it.

Happy to take the rework: fix attribution to cover extensions, drop a url that kept no add-on, give the epoch path per-add-on hysteresis so it stops demoting whole providers, and rebuild the tests on container shapes the specs actually produce.

AnnaR-prog:
Superseded by #359, which reworks this after the review above. The three correctness findings — attribution ignoring `Extension` (so the motivating archive case was never fixed), the epoch path demoting the whole provider two ticks after boot, and a standalone-addons url flipping to serve the base collection — are all fixed there, along with the frozen admission map, the `expandInternalPaths` divergence, the duplicated `validateProvider`, and the test fixtures that manufactured a container shape no spec produces.

Suggest closing this one. The rebase itself was faithful — I verified the delta byte-for-byte identical and `allowBaseFallback` intact — the original commit just did not deliver MAG-3326.

AnnaR-prog:
Pushed the rework here as `942cfef`, on top of your `da45c9b` rather than rewriting it — so the PR now carries the original commit plus the response to the review.

What changed, all three verified against this branch:

**The motivating case was not fixed.** Attribution charged failures to `verification.Addon` alone, but every EVM and cosmos spec keys the archive check as an *extension on the base collection* (`specs/ethereum.json`: `add_on='' | pruning | extension='archive'`), so `Addon == ""` took the fatal path. `refusedServiceFor` now charges the extension first — which is also correct for `{Addon:"trace", Extension:"archive"}`, where refusing `trace` would drop working traffic and keep advertising the thing that actually failed.

**The epoch path undid the rescue.** `applyReverification` re-probed the unstripped config with all-or-nothing `Validate`, failed on the same service, and demoted the whole provider once `demoteFailStreak` hit the threshold — with no path back. It probes per-collection now, and records a fresh admission each tick, so a recovered service is re-admitted instead of staying refused until restart.

**A standalone-addons url that kept nothing flipped to serving base.** `AdmittedServices` returns `keep=false` and the boot path drops the url.

Also from your review: the admission is applied per-endpoint rather than before `expandInternalPaths` (so the subscription probe and the chain router still see the same raw config), `validateProvider` and the per-collection variant share one body, the key is a comparable struct rather than a NUL-joined string, and the tests are rebuilt on real container shapes — including a passing verification, the path every healthy provider takes, which nothing covered before. Plus a `go vet` copies-lock failure in a test I added on #345.

`go build ./...`, `go vet ./protocol/...` and the full `./protocol/...` suite are green on `942cfef`.

I opened #359 with the same change before you asked for it here — closing that one now, this PR is the place.

AnnaR-prog (commented):
## Review of the rework (`942cfef`)

**Caveat first: I wrote this commit, so this is a self-review and should not stand in for an independent one.** Posting it because a fresh adversarial pass found something real, not to clear the review requirement.

One finding, and it is in a claim I made rather than in behaviour that regressed.

### The recovery half of the epoch fix does not work for active providers

`942cfef`'s comment and commit message both say a refused service *"is re-admitted at the next tick instead of staying refused until the process restarts."* That is true only for a provider that goes through **promote**. For a provider that is already active — the common case — it is false.

The epoch tick does record a fresh admission: `probe` runs `validateProviderCollections` and calls `recordAdmission` before any reconciliation. But the session rebuild for an active provider reuses its existing endpoints:

```go
freshSession := lavasession.NewConsumerSessionWithProvider(
    oldSession.PublicLavaAddress,
    oldSession.Endpoints,   // rpcsmartrouter.go:3527 (and :3556 for backups)
    ...
)
```

Only the promote path goes back through the builder that reads the admission (`inputs.convertProvidersToSessions(toAdmit)`, `spec_reverifier.go:374`). So an active provider keeps the endpoint set it was built with at boot, and a service refused there stays refused for its lifetime unless it is demoted and later re-promoted.

**What the commit does fix stands:** the whole-provider demotion two ticks after boot is genuinely gone, because the probe no longer returns an error for a service-attributable failure. That was the severe one. The admission map is also genuinely refreshed — it is the *application* of the refreshed value that stops short.

Two ways forward, and I do not think this should block on its own:

1. Narrow the claim — say the refusal lasts until the provider is next rebuilt, and file recovery-for-active-providers as follow-up. Honest, and strictly better than the boot-frozen map it replaced.
2. Rebuild the session when a provider's admission changes across a tick. That means diffing old against new admission in `applyReverification` and routing changed providers through `convertProvidersToSessions` rather than the endpoint-reusing path. Contained, but it touches the epoch session refresh, which has its own invariants around connection lifetime — the demote path already has to close connections *after* `UpdateAllProviders` for exactly that reason.

I lean 1 for this PR and 2 as its own change, but it is a judgement call and I would rather it were made deliberately than inherited from a comment that overstates what the code does.

### Everything else I checked and did not find fault with

- `recordAdmission` is called from the `validateProviderTier` goroutines and from the epoch probe; the store takes a write lock on both paths, and `admissionFor` a read lock.
- Holdoff and rate-limit handling still wrap the probe, so a 429 remains inconclusive rather than a refusal.
- A standalone url can only fail on attributable verifications (`verificationsForNodeUrl` drops the base ones), so it can never take the fatal path — which is exactly why `keep=false` has to exist, and it does.
- A provider whose every url is dropped falls into the existing empty-endpoint skip, so it does not reach the pairing half-built.
- `AdmittedServices` returns the input slice when nothing was refused, and the test pins slice identity rather than equality.

CI is green (20 pass, 1 skipped), and `go build ./...`, `go vet ./protocol/...` and the full `./protocol/...` suite pass locally on `942cfef`.

### Still open from before

No live verification. The discriminating case needs an upstream that passes the base collection and fails exactly one service, which neither the Acala nor the Shiden endpoints provide. Everything here is unit-level.

AnnaR-prog:
Addressed in `84a6a16` — took option 2 rather than narrowing the claim.

`applyReverification` now notices when a provider's admitted set actually moved and rebuilds it, composing the two mechanisms already there instead of adding a third: the provider is left out of `healthyNames`, so the stale session is demoted and the caller closes its connections **after** the pairing swings over, and it is added to `toAdmit`, so a fresh session is built from config with the admission just recorded. That is the connection-lifetime ordering demotion already depends on, unchanged — which was the thing that made me hesitate about this option in the review.

Change detection is `ProviderAdmission.Equal`, compared per refusal rather than by count: the same service refused on two different urls gives equal counts and must still register. The test asserts that case and fails against a length-only implementation — verified by writing one.

A provider whose admission has not moved takes the default branch and stays in `healthyNames` exactly as before, so nothing changes for the healthy majority on every tick.

Full `./protocol/...` suite, `go build ./...` and `go vet ./protocol/...` green on `84a6a16`.

The remaining open item from that review is unchanged and still worth a look from someone else: none of this is verified against a live upstream, because the discriminating case needs a node that passes the base collection and fails exactly one service, which neither Acala nor Shiden provides.

AnnaR-prog:
## Live evidence, against a real upstream

Built from `84a6a16` and run against `specs/ethereum.json` with `ghcr.io/magma-devs/smart-router:main` as the control — the image the spec pipeline itself boots. Same spec dir, same config, same upstream, one variable.

**The scenario is the ticket's own, unmodified:** attach an add-on url to an otherwise healthy provider, have it fail that add-on's check.

Upstream `https://ethereum-rpc.publicnode.com` declared with `addons: ["trace"]`. It is a genuinely healthy ETH1 node — `eth_chainId` returns `0x1` — that genuinely does not serve trace: `trace_block` comes back `-32602 "Archive requests require a personal token"`. Nothing is stubbed or simulated; that is the public tier's real behaviour.

| | `:main` control | this branch |
|---|---|---|
| `invalid Verification on provider startup` | **14** (all `"verification":"trace"`) | **0** |
| `excluding from provider list` | **1** | **0** |
| `eth_chainId` through the router | `No pairings available` | `0x1` |
| `eth_blockNumber` through the router | `No pairings available` | `0x18af07d` |
| `trace_block` through the router | `No pairings available` | `No Providers For Addon {addon:trace}` |

The control row is the bug: a healthy Ethereum node serving nothing at all because one add-on check failed. The last row is the half that matters as much — trace is refused rather than quietly routed to a node that cannot serve it.

The attribution log line shows the rule firing on real data:

```
message: refusing one service for this url, keeping the provider
service: trace   addon: trace   extension: ""   verification: trace
```

### What could not be verified live, and why

The **extension**-keyed path — the one the rework fixed — has no live failure available. ETH1's archive check is `eth_getBlockByNumber("earliest")` expecting `0x0`, and every public endpoint retains genesis, so no real node fails it. I checked four before concluding that (`ethereum-rpc.publicnode.com`, `eth-rpc-acala.aca-api.network`, `rpc.shiden.astar.network`, `bsc-rpc.publicnode.com`) — all return `0x0`.

So I closed that gap the strongest way available instead, in `87103d0`: a test that loads `specs/ethereum.json` through `getServiceApis` — the same code the router runs — and asserts the archive check comes back keyed `{Addon:"", Extension:"archive"}` with an omitted severity, i.e. `Fail`. That is precisely the shape the original commit mis-attributed, and precisely what the old fixtures could not represent, since they built `{Addon:"archive"}` by hand. It covers the contrasting `trace` shape from the same file too, so both branches of `refusedServiceFor` are tied to real spec data rather than to a fixture that agrees with the code.

Live evidence for the feature end to end; real-spec evidence for the attribution rule. The one thing neither can reach is a node that fails an archive check, because that node does not appear to exist on the public internet.

CI is green on `84a6a16` (20 pass, 1 skipped) and the full `./protocol/...` suite passes locally on `87103d0`.

AnnaR-prog (commented):
## Review of `84a6a16` and `87103d0`

Fresh pass over the two commits added since my last review. Same caveat as before: I wrote them, so this is not the independent review this PR still needs.

Two findings, both about the rebuild mechanism, neither a correctness bug. The attribution and admission work I checked hardest is clean.

### 1. Rebuild-on-change has no hysteresis, while demotion does

`84a6a16` rebuilds a provider whenever its admitted set moves between ticks. Demotion, sitting a few lines away, deliberately does not work that way: `reverifyDemoteThreshold = 2` exists because *"a 40-second blip that happened to overlap an epoch tick demoted the provider outright and cost it a full epoch"* (MAG-2445). The rebuild path reintroduces that shape at service granularity.

A flapping add-on — one that intermittently times out rather than being cleanly up or down — now produces a full session rebuild on every tick it changes state. That is not free: the rebuilt session comes from `convertProvidersToSessions`, so it gets **new `Endpoint` objects with new `DirectRPCConnection`s**, where the ordinary epoch refresh reuses `oldSession.Endpoints` and therefore keeps their accumulated health state. So a flapping service costs the provider its endpoint health and its connections every 15 minutes.

The asymmetry is worth fixing, and I do not think it needs symmetric hysteresis. The two directions are not equally urgent:

- **newly refused** — rebuild immediately. The endpoint is currently advertising a service it cannot serve, and every relay to it fails until the rebuild lands.
- **recovered** — rebuild lazily. Nothing is broken while a working service stays refused; it is availability, not correctness.

Hysteresis on the re-admit direction only would bound the churn to one rebuild per outage rather than one per flap, and would leave the urgent direction sharp. That is a small change to the `case moved:` branch plus a per-service streak, and it is the shape MAG-2445 already argues for.

### 2. A rebuild whose connections fail drops the provider for one tick

If `convertProvidersToSessions` cannot create a connection for any url, it skips the provider (`rpcsmartrouter.go`, *"skipping static provider: all URL connections failed"*). On the rebuild path that provider has already been left out of `healthyNames`, so it is demoted — and then not re-added. A provider that validated successfully leaves the pairing because a transient dial failure happened to coincide with an admission change.

It self-heals: next tick it is `!wasActive`, so it goes through `toAdmit` and comes back. And it is not silent — the skip logs a warning. But it is a new way for a healthy provider to drop out, and it did not exist before this commit. Worth either a comment naming it as accepted, or keeping the old session when the rebuild yields nothing.

### What I checked and found clean

- `GroupLabel` and `StaticProvider` survive a rebuild — `convertProvidersToSessions` sets both from config, so the *"cross-validation group must survive epoch refresh"* invariant holds on this path too. This was the first thing I expected to be broken.
- `admissionMoved` is declared inside `applyReverification`, which runs once per tier, so the name-keyed map cannot collide across tiers the way `demoteFailStreak` would (hence its `tier|name` key).
- The moved set is written under a mutex during the concurrent probe phase and read only after `wg.Wait()`.
- `recordAdmission` returns false on the first sighting, so boot does not register as a change and nothing rebuilds on the first tick for that reason.
- A provider that failed at boot is never recorded, so its later recovery goes through the `!wasActive` branch rather than the rebuild branch.

### On `87103d0`

It couples CI to `specs/ethereum.json`, so a spec bump could fail it for reasons unrelated to the code. I think that is the right trade — the coupling is the point, since the bug it guards was a fixture disagreeing with reality — but the failure message should make that obvious to whoever hits it. `require.Empty(t, eth1.Imports, ...)` in particular would fail confusingly if ETH1 ever gained an import.

AnnaR-prog (approved):

PR #359, opened 2026-09-01
feat(chainlib): admit a provider per collection at boot and per epoch (MAG-3326)

AnnaR-prog:
# Description

Reworks [#358](https://github.com/Magma-Devs/smart-router/pull/358) (MAG-3326) after review found the original commit did not deliver the ticket. Off `main` directly. **#358 should be closed in favour of this.**

The review is on #358; this is what it produced. Three correctness problems, all with tests built on the container shapes `specs/` actually produces — the previous suite could not catch any of them because its fixtures manufactured a shape no spec builds.

## 1. The motivating case was not fixed

The ticket exists to stop *"attach an archive url to an otherwise healthy provider, have it fail the archive check, lose the provider."* It still lost the provider.

Attribution charged a failure to `verification.Addon` alone. But a `VerificationContainer` carries **both** an `Addon` and an `Extension`, and every EVM and cosmos spec keys the archive check as an **extension on the base collection**. From `specs/ethereum.json`:

```
collection add_on='' | verification=pruning | extension='archive' | severity=None
```

so `Addon == ""` and the fatal path fired. `refusedServiceFor` now charges the **extension first**, falling back to the add-on. That is also the right answer for ethereum's `{Addon:"trace", Extension:"archive"}`: what failed is the node's archive-ness, so refusing `trace` would drop working traffic *and* leave the failing extension advertised.

## 2. The epoch path undid the boot rescue

For the case that did work, it survived about two ticks. `applyReverification` re-probed the **unstripped** config with the all-or-nothing `Validate` on every tick, failed on the same refused service, and demoted the whole provider once `demoteFailStreak` reached `reverifyDemoteThreshold` — into a worse state than before the change, with no path back (promotion needs a clean all-or-nothing pass, and it is not in `failedStaticProviders` for the retry loop). It probes per-collection now.

## 3. A standalone-addons url that kept nothing flipped to serving base

Emptying `Addons` makes `ServesBaseCollection()` return true, so a node whose operator declared it cannot answer the base surface was handed base traffic — the MAG-3296 failure re-entered from the other side. `AdmittedServices` reports `keep=false` for such a url and the boot path drops it with a warning.

## Also addressed from the review

- **Admissions were frozen at boot.** The endpoint builder is also the retry and promote path, so a transient blip refused a service until process restart — the permanence the copy semantics were chosen to avoid, relocated into the map. They are recorded under a lock by boot **and** by every epoch re-validation, so a recovered service is re-admitted at the next tick.
- **`expandInternalPaths` was fed stripped urls.** Its subscription probe resolves a collection from the declared add-ons to decide ws-vs-http, and its doc requires its output to mirror the chain router's, which is built from the untouched config. The admission is applied per-endpoint instead, so both see the same raw urls.
- **`validateProvider` and the per-collection variant share one body** rather than being a hand-kept copy that must not drift.
- **The composite key is a comparable struct**, not a NUL-joined string.
- **`AdmittedServices` returns the input slice when nothing was refused** — the common case on every rebuild. The test asserts slice *identity*, since a deep compare would not notice the allocation coming back.
- **A passing verification is covered** — the path every healthy provider takes at boot, which nothing exercised before. Every previous fixture failed, so a build where the refusal branch was unreachable would have passed.

Also fixes a `go vet` failure (`copies lock value`) in a test added on #345.

## Most critical to review

- `chain_fetcher.go` — `refusedServiceFor` (extension-first is the whole correctness of the feature) and `AdmittedServices` (the `keep=false` case).
- `spec_reverifier.go` — the epoch probe and the shared `validateProviderImpl`.
- `rpcsmartrouter.go` — the guarded admission store and where it is refreshed.

## Not covered

Not exercised against a live upstream: the discriminating case needs a node that passes the base collection and fails one service, which none of the Acala or Shiden endpoints does. The unit tests drive both entry points through the same verification set instead — `ValidateCollections` returns no error where `Validate` on identical input returns one, so the two paths cannot silently converge.

`go build ./...`, `go vet ./protocol/...` and the full `./protocol/...` suite are green.

## Jira ticket

Jira ticket: MAG-3326

---

## Author Checklist

I have...

### DIFF
--- a/protocol/chainlib/chain_fetcher.go
+++ b/protocol/chainlib/chain_fetcher.go
@@ -166,13 +166,123 @@ func stopValidateRetries(err error) bool {
 	return err == nil || errors.Is(err, common.StatusCodeError429)
 }
 
+// admissionKey identifies one refused service on one node url. A comparable
+// struct rather than a joined string: the join needed a NUL sentinel, which is
+// an invariant every reader has to re-verify.
+type admissionKey struct {
+	url          string
+	internalPath string
+	service      string
+}
+
+// ProviderAdmission records the services a provider's node urls failed to answer
+// for, so a failure attributable to ONE service costs that service rather than
+// the whole provider (MAG-3326).
+//
+// "Service" is a name as it appears in NodeUrl.Addons, which mixes add-on and
+// extension names — the router's own config makes no distinction, and neither
+// does Endpoint.Addons/.Extensions, both of which are built from that one list.
+//
+// Empty means everything a url declared was admitted.
+type ProviderAdmission struct {
+	refused map[admissionKey]struct{}
+}
+
+// refusedServiceFor returns the service a failed verification should be charged
+// to, or "" when the failure belongs to no single service and is therefore fatal.
+//
+// EXTENSION FIRST, and this is the whole correctness of the feature. A
+// VerificationContainer carries both an Addon and an Extension, and the narrower
+// one is what was actually probed:
+//
+//   - {Addon:"", Extension:"archive"} — every EVM and cosmos spec keys the archive
+//     check this way, on the BASE collection. Charging it to Addon would find ""
+//     and take the fatal path, which is the motivating case of this whole ticket:
+//     "attach an archive url to a healthy provider, fail the archive check, lose
+//     the provider". Charging it to "archive" refuses archive and keeps the rest.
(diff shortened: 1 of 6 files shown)


AnnaR-prog:
Closing in favour of #358 — the same rework is now pushed there as `942cfef`, on top of avitenzer's `da45c9b`, which keeps the review and the response to it in one place.

PR #360, opened 2026-09-01
fix(rpcsmartrouter): relay subscription pushes named after their payload (MAG-3345)

AnnaR-prog:
Closes [MAG-3345](https://magmadevs.atlassian.net/browse/MAG-3345)

Jira ticket: MAG-3345

## The bug

No Substrate subscription had ever delivered a push frame through the router. The subscribe call itself always succeeded, which is what made the gap silent: nothing errored at boot or at subscribe time, frames simply never arrived. The ticket's suggested control case — `eth_subscribe` works, Substrate does not — localised it the same way it did for MAG-3297.

**Two independent defects sat on the delivery path**, both of them the same assumption: that a push frame can be recognised by its method name.

**1. `rpcclient`: the frame was never delivered.** `handleImmediate` only claimed a frame whose method ended in `_subscription` or `Notification`, or that carried a Tendermint query. Substrate names the frame after the *payload* — `chain_newHead`, `state_storage`, `author_extrinsicUpdate` — while using Ethereum's exact `{subscription, result}` params envelope, so no suffix rule could match it.

The frame was not merely dropped. It fell through to `handleCallMsg`, which has no id and no matching method to serve, so **the router answered the node with an `invalid request` error for every push it received.**

**2. `rewriteSubscriptionID`: the id was wrong.** It keyed on `msg.Method == "eth_subscription"`, so even once delivered, a Substrate frame reached the client carrying the **upstream** id — one the client was never issued and cannot match to its subscription.

Either defect alone leaves the subscription unusable. Both now match the params envelope instead of the method name; the frame's own method is echoed back rather than hard-coded, and params are rebuilt from what upstream sent so sibling fields survive.

## Verification

Each defect was proved to fail independently by reverting the fixes against the new tests:

| state | result |
| --- | --- |
| both fixes reverted | 5s timeout — *"no push frame reached the client"* |
| only the dispatcher fixed | fails at 0.02s — the frame **arrives**, carrying the upstream id |
| both applied | pass |

`protocol/chainlib/...`, `protocol/rpcsmartrouter` and `protocol/lavasession` are green under `-race`. `go vet ./...` and `gofmt` clean.

**Boundary:** verified against a mock upstream over a real WebSocket, not a live Substrate node, and the Phase 8 sweep was not re-run. This removes the mechanism that made all 24 probes fail; it is not a claim that 24/24 now pass.

## Tests

The harness could only speak EVM, which is why nothing here caught this — every existing test asserted against the one shape the router happened to handle. The mock upstream's notification method and subscription id are now fields, so it can serve a Substrate node, and the Substrate test uses a non-hex id (`Ck1rTHhOa1hxTGV3`) so nothing can pass by matching an `0x` prefix.

- `TestEndToEnd_SubstratePushFrameReachesClient` — runs over a real WebSocket, so the whole path is live. Asserts both defects separately: the frame arrives, keeps its own method, and carries the router id.
- `TestRewriteSubscriptionID_Substrate` — the unit half.
- `TestRewriteSubscriptionID_SolanaUntouched` — pins the boundary below.

## Two adjacent findings, deliberately not fixed here

**Solana pushes are dropped today, and still are.** `handleImmediate`'s `solanaNotificationMethodSuffix` branch is unreachable: its enclosing `isEthereumNotification()` predicate itself requires the `_subscription` suffix, which `accountNotification` does not have. Confirmed empirically. Both new guards require the subscription id to be a JSON string, which deliberately keeps Solana's integer ids out — delivering them needs an integer-id router mapping the id mapper cannot express. Worth its own ticket.

**The last hop is not a verbatim relay.** `consumer_websocket_manager.go` runs each push through `outputFormatter`, which injects the *subscribe request's* id into every notification — JSON-RPC notifications should not carry one. Pre-existing, identical for EVM, and it touches neither the method nor `params.subscription`, so it does not affect this fix. Flagging rather than changing it, since it would move EVM behaviour.

[MAG-3345]: https://magmadevs.atlassian.net/browse/MAG-3345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/protocol/chainlib/chainproxy/rpcclient/handler.go
+++ b/protocol/chainlib/chainproxy/rpcclient/handler.go
@@ -241,6 +241,11 @@ func (h *handler) handleImmediate(msg *JsonrpcMessage) bool {
 			h.handleSubscriptionResultEthereum(msg)
 			return true
 		} else if strings.HasSuffix(msg.Method, solanaNotificationMethodSuffix) {
+			// Unreachable: isEthereumNotification above already required the
+			// "_subscription" suffix, which no *Notification method has. Solana frames
+			// are served by the isSubscriptionNotification case below instead — left
+			// in place only so removing it is not confused with a behaviour change
+			// (MAG-3359).
 			h.handleSubscriptionResultSolana(msg)
 			return true
 		}
@@ -273,17 +278,16 @@ func (h *handler) handleImmediate(msg *JsonrpcMessage) bool {
 // fell through to handleCallMsg, which has no id and no matching method to serve, so the
 // router answered the node with an "invalid request" error for every push it received.
 func (h *handler) handleSubscriptionResultByParams(msg *JsonrpcMessage) {
-	var result ethereumSubscriptionResult
-	if err := json.Unmarshal(msg.Params, &result); err != nil {
+	id, ok := subscriptionIDFromParams(msg.Params)
+	if !ok {
 		utils.LavaFormatTrace("Dropping invalid subscription message",
-			utils.LogAttr("err", err),
 			utils.LogAttr("params", string(msg.Params)),
 		)
 		h.log.Debug("Dropping invalid subscription message")
 		return
 	}
-	if h.clientSubs[result.ID] != nil {
-		h.clientSubs[result.ID].deliver(msg)
+	if h.clientSubs[id] != nil {
+		h.clientSubs[id].deliver(msg)
 	}
 }
 
--- a/protocol/chainlib/chainproxy/rpcclient/handler.go
+++ b/protocol/chainlib/chainproxy/rpcclient/handler.go
@@ -17,6 +17,7 @@
 package rpcclient
 
 import (
+	"bytes"
 	"context"
 	"reflect"
 	"strconv"
@@ -299,7 +300,10 @@ func (h *handler) deliverSubscriptionPush(msg *JsonrpcMessage, subscriptionID st
 		sub.deliver(msg)
 		return true
 	}
-	if msg.hasValidID() {
+	// An explicit "id":null is not a request id — hasValidID only rejects objects and
+	// arrays, so `null` passes it. Letting that through to the call path answers the node
+	// with "invalid request" for a late push, which is the MAG-3345 symptom itself.
+	if msg.hasValidID() && !bytes.Equal(msg.ID, null) {
 		return false
 	}
 	// Late frames after an unsubscribe land here routinely, so this is a trace and not a
--- a/protocol/chainlib/chainproxy/rpcclient/handler.go
+++ b/protocol/chainlib/chainproxy/rpcclient/handler.go
@@ -255,8 +255,35 @@ func (h *handler) handleImmediate(msg *JsonrpcMessage) bool {
 		h.handleResponse(msg)
 		h.log.Trace("Handled RPC response", "reqid", idForLog{msg.ID}, "duration", time.Since(start))
 		return true
+	case msg.isSubscriptionNotification():
+		// Last, so the method-name cases above keep first claim on anything they
+		// already recognise. This one matches on the params envelope alone, which is
+		// how Substrate pushes arrive — see isSubscriptionNotification.
+		h.handleSubscriptionResultByParams(msg)
+		return true
 	default:
 		return false
+	}
+}
+
+// handleSubscriptionResultByParams delivers a subscription notification identified by its
+// params envelope rather than by its method name.
+//
+// Returning false for these from handleImmediate was worse than dropping them: the frame
+// fell through to handleCallMsg, which has no id and no matching method to serve, so the
+// router answered the node with an "invalid request" error for every push it received.
+func (h *handler) handleSubscriptionResultByParams(msg *JsonrpcMessage) {
+	var result ethereumSubscriptionResult
+	if err := json.Unmarshal(msg.Params, &result); err != nil {
+		utils.LavaFormatTrace("Dropping invalid subscription message",
+			utils.LogAttr("err", err),
+			utils.LogAttr("params", string(msg.Params)),
+		)
+		h.log.Debug("Dropping invalid subscription message")
+		return
+	}
+	if h.clientSubs[result.ID] != nil {
+		h.clientSubs[result.ID].deliver(msg)
 	}
 }
 
--- a/protocol/chainlib/chainproxy/rpcclient/handler.go
+++ b/protocol/chainlib/chainproxy/rpcclient/handler.go
@@ -243,13 +243,32 @@ func (h *handler) handleImmediate(msg *JsonrpcMessage) bool {
 		} else if strings.HasSuffix(msg.Method, solanaNotificationMethodSuffix) {
 			// Unreachable: isEthereumNotification above already required the
 			// "_subscription" suffix, which no *Notification method has. Solana frames
-			// are served by the isSubscriptionNotification case below instead — left
-			// in place only so removing it is not confused with a behaviour change
-			// (MAG-3359).
+			// are served by the shape-based case below instead — left in place only so
+			// removing it is not confused with a behaviour change (MAG-3359).
 			h.handleSubscriptionResultSolana(msg)
 			return true
 		}
 		return false
+	case msg.Method != "" && msg.Params != nil:
+		// Shape-based delivery: a method-bearing frame whose params name a subscription.
+		// The method-name cases above keep first claim on anything they already
+		// recognise; this catches the rest, which is how Substrate and Solana pushes
+		// arrive — see subscriptionIDFromParams.
+		//
+		// It sits BEFORE the StarkNet case deliberately. That predicate matches any
+		// id-less method-bearing frame carrying a result, so a push that also sets a
+		// top-level result would be swallowed there and answered with "invalid request"
+		// — the very MAG-3345 symptom this exists to remove. StarkNet's own frames put
+		// the id in result and carry no params, so they never enter this case.
+		//
+		// The id is parsed once and handed to the delivery path. The sibling handlers
+		// above re-parse inside the handler, which here would mean decoding the whole
+		// payload twice for every frame delivered.
+		id, named := subscriptionIDFromParams(msg.Params)
+		if !named {
+			return false
+		}
+		return h.deliverSubscriptionPush(msg, id)
 	case msg.isStarkNetPathfinderNotification():
 		if strings.HasSuffix(msg.Method, ethereumNotificationMethodSuffix) {
 			h.handleSubscriptionResultStarkNetPathfinder(msg)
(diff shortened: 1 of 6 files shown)


protocol/chainlib/chainproxy/rpcclient/handler.go
```diff
@@ -255,11 +255,38 @@ func (h *handler) handleImmediate(msg *JsonrpcMessage) bool {
...
 	}
 }
 
+// handleSubscriptionResultByParams delivers a subscription notification identified by its
+// params envelope rather than by its method name.
+//
+// Returning false for these from handleImmediate was worse than dropping them: the frame
+// fell through to handleCallMsg, which has no id and no matching method to serve, so the
+// router answered the node with an "invalid request" error for every push it received.
+func (h *handler) handleSubscriptionResultByParams(msg *JsonrpcMessage) {
```
avitenzer: **This is byte-for-byte identical to `handleSubscriptionResultEthereum` (line 308).** Same struct, same unmarshal, same `clientSubs` lookup, same `deliver`. The only difference is the log string.

The new case can route straight into the existing handler and this whole function disappears:

```go
case msg.isSubscriptionNotification():
    // Last, so the method-name cases above keep first claim…
    h.handleSubscriptionResultEthereum(msg)
    return true
```

The only argument for a separate function is the log wording, and `handleSubscriptionResultEthereum` is already the generic *string-id* handler — the name is the only Ethereum-specific thing about it. This is the one change I'd actually ask for.

---

Rider while you're here: **`params` gets unmarshalled twice per push frame.** `isSubscriptionNotification` parses it to decide, then this function parses the identical bytes again. The EVM path pays one (its predicate is a free suffix check); this path pays two, and the predicate now runs a full unmarshal on every message falling past the earlier cases.

At `chain_newHead` rates that's noise; at `state_storage` on a busy parachain it's a hot loop. I don't think it's worth restructuring the switch to thread the parsed value through — but if you take the collapse above you're consciously accepting the double parse, so it's worth a clause in the comment rather than leaving it to be rediscovered.

protocol/chainlib/chainproxy/rpcclient/json.go
```diff
@@ -97,6 +97,34 @@ func (msg *JsonrpcMessage) isEthereumNotification() bool {
...
+// envelope as Ethereum. No suffix rule can cover that, so nothing delivered those
+// frames and every Substrate subscription was silently dead (MAG-3345).
+//
+// The string check keeps this off Solana, whose ids are integers. Solana pushes are
+// unhandled today for a separate reason — handleImmediate's solana branch sits inside
+// the isEthereumNotification case, and that predicate itself requires the
+// "_subscription" suffix, which "accountNotification" does not have, so the branch is
+// unreachable. Delivering them needs an integer-id router mapping the id mapper cannot
+// express, which is its own change; MAG-3345 is the Substrate gap.
+func (msg *JsonrpcMessage) isSubscriptionNotification() bool {
```
avitenzer: **Coverage gap: this predicate has no test in its own package.** Its only coverage is the E2E WebSocket test over in `rpcsmartrouter` — there's no predicate test anywhere in `protocol/chainlib/chainproxy/rpcclient/*_test.go`.

That matters more than usual here because the **string-only guard is what keeps Solana out**, and right now that guard is asserted only indirectly, one layer up, in `TestRewriteSubscriptionID_SolanaUntouched`.

A table test in `subscription_test.go` over the boundary shapes would pin the contract at the layer that owns it — and fail in microseconds instead of via a 5s WebSocket timeout in a different package:

- array-shaped `params` (`["abc", {...}]`) → false
- integer id (`{"subscription":23784}`) → false
- no `subscription` key → false
- empty-string id → false
- `params: null` / `Method: ""` → false
- the Substrate envelope → true

protocol/rpcsmartrouter/direct_ws_subscription_manager.go
```diff
@@ -1810,39 +1810,65 @@ func createSubscriptionReplyFromRouterID(routerID string, requestID json.RawMess
...
+// routerID, or nil when params is not a subscription envelope carrying a string id.
+//
+// Sibling fields are preserved: the envelope is rebuilt from what upstream actually sent
+// rather than from a fixed {subscription, result} pair, so anything a chain adds alongside
+// them survives the rewrite.
+//
+// Returning nil for a non-string id keeps Solana out: its ids are integers, and a router id
+// is not one, so writing a string in would hand the client an id of a type its own
+// unsubscribe cannot use.
+func rewriteParamsSubscriptionID(params json.RawMessage, routerID string) json.RawMessage {
```
avitenzer: **`nil` is overloaded**, and it quietly changes EVM behaviour on one path.

It means both *"not a subscription envelope"* (expected → pass-through) and *"parse/marshal failed"* (unexpected → also pass-through). The two `json.Marshal` failures are unreachable in practice, but the `json.Unmarshal(params, &fields)` return is not:

- **Before:** an `eth_subscription` frame with unparseable params returned `(nil, err)`, the caller logged a warning and **skipped that client**.
- **Now:** it silently forwards the frame carrying the **upstream** id.

I checked the blast radius — `Unsubscribe` validates subscription ownership by `clientKey` before removing anything (line 863), so a leaked upstream id can't tear down a shared subscription. This is robustness, not security. But it's the one place EVM behaviour changed and the PR body doesn't mention it.

Either a `LavaFormatTrace` on the params-present-but-unparseable path, or return `(json.RawMessage, bool)` to separate the two meanings.

protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
```diff
@@ -1436,6 +1445,62 @@ func TestRewriteSubscriptionID_EVM(t *testing.T) {
...
 }
 
+// TestRewriteSubscriptionID_Substrate is the unit half of MAG-3345. Substrate uses the
+// same params envelope as EVM but names the frame after the payload, so matching on
+// method == "eth_subscription" passed these through untouched — the client received the
+// upstream id instead of the router id it was issued at subscribe time.
+func TestRewriteSubscriptionID_Substrate(t *testing.T) {
+	substrateMsg := &rpcclient.JsonrpcMessage{
+		Method: "chain_newHead",
+		Params: json.RawMessage(`{"subscription":"Ck1rTHhOa1hxTGV3","result":{"number":"0x1234"}}`),
```
avitenzer: **"Sibling fields are preserved" is the entire justification for the map-based rewrite, and nothing tests it.**

`rewriteParamsSubscriptionID`'s docstring claims it explicitly, and it's the sole reason for moving off the old two-field struct. I confirmed the behaviour holds:

```
in:  {"subscription":"abc","result":{"x":1},"extra":"keepme"}
out: {"extra":"keepme","result":{"x":1},"subscription":"rs_1"}
```

But none of the three rewrite tests carries a sibling field, so if someone later "simplifies" back to the struct, **every test still passes** and the behaviour regresses silently.

One extra key here closes it:

```go
Params: json.RawMessage(`{"subscription":"Ck1rTHhOa1hxTGV3","result":{"number":"0x1234"},"extra":"keepme"}`),
```

plus an assertion that `extra` survived the rewrite.

avitenzer (commented):
## Review — approve with minor comments

Correct diagnosis, well-scoped fix, and the tests are genuinely load-bearing. Nothing here blocks merge; the inline comments are ranked below.

### What I verified independently

From the committed sha `4a51956` (not a working dir), since this repo's PR gate doesn't run `go test`:

- `go build ./...`, `go vet` on both changed packages, `gofmt` — clean
- `./protocol/rpcsmartrouter/...` — **ok**, 97s under `-race`
- `./protocol/chainlib/...` (7 packages incl. `rpcclient`) — **ok** under `-race`

I mutation-tested the verification table in a throwaway worktree. Both halves reproduce, so the E2E test really does pin the two defects separately:

| mutation | result |
| --- | --- |
| `isSubscriptionNotification` forced `false` | **FAIL** at 5.00s — *"no push frame reached the client"* |
| rewrite re-gated on `method == "eth_subscription"` | **FAIL** at 0.03s — frame arrives carrying `Ck1rTHhOa1hxTGV3` instead of `rs_f91191_00001` |

I also checked the `outputFormatter` finding you flagged but didn't fix. `cacheformat.FormatterForRelayRequestAndResponseJsonRPC` does `sjson.SetBytes(inpData, IDFieldName, extractedID)` — top-level `id` only, never `method` or `params.subscription`. Your description is accurate and the fix reaches the client intact.

### Design call worth naming

Placing the shape-based case **last** in `handleImmediate` is what makes this safe to widen: the new predicate can only claim frames that previously hit `default: return false`. That's why chainlib and rpcsmartrouter stayed green without touching a single existing assertion. The comment saying so is load-bearing — a future refactor sorting the cases "logically" would silently change Solana and StarkNet routing.

### Ranked

- **handler.go** — delete the duplicate handler (one-line win)
- **test file + json.go** — two cheap test additions that protect the fix from a future refactor
- **direct_ws_subscription_manager.go** — judgement call, reasonable to close with a comment

Posting as a comment rather than an approval so the merge call stays yours.

protocol/chainlib/chainproxy/rpcclient/handler.go
AnnaR-prog: Right that it is a copy — but I went the other way, and I think the duplication was the symptom rather than the problem.

`handleSubscriptionResultByParams` is gone in the follow-up (#361), replaced by `deliverSubscriptionPush(msg, id)`, which is genuinely not the Ethereum handler:

- it takes the **pre-parsed** id, so the frame is decoded once — your rider, taken rather than commented
- it returns whether it *claimed* the frame, which the Ethereum handler has no notion of
- an id-bearing frame that matches no subscription now falls through to the call path, so a peer waiting on a reply still gets one. As written here it was claimed and swallowed
- an unmatched id-less push is claimed but traced, instead of vanishing

So routing into `handleSubscriptionResultEthereum` would have been correct today and a dead end tomorrow — that function cannot express "I did not claim this".

On the rider: you judged it not worth restructuring the switch. It turned out to be, once the id had to be threaded anyway. The case now guards on `Method != "" && Params != nil` before parsing, so frames that never had an envelope do not pay for one.

protocol/chainlib/chainproxy/rpcclient/json.go
AnnaR-prog: Agreed, and done — `subscription_notification_test.go` in that package, two table tests covering your list: array-shaped params, no `subscription` key, empty-string id, `null`, absent params, and the Substrate envelope.

One row is deliberately inverted from what you asked for: **integer id is now `true`.** The string-only guard you are leaning on to keep Solana out is removed on purpose in #361 (MAG-3359) — Solana subscriptions were dead for a different reason, and the fix is to follow the chain's id shape rather than to exclude it. `TestRewriteSubscriptionID_SolanaUntouched` is superseded there for the same reason.

Your underlying point stands and is what the new file is for: the contract now has tests at the layer that owns it, failing in microseconds rather than through a 5s WebSocket timeout in another package.

protocol/rpcsmartrouter/direct_ws_subscription_manager.go
AnnaR-prog: This was the sharpest of the four, and you were right that it is the one place EVM behaviour changed silently.

Took your option B in #361: the signature is now `(json.RawMessage, error)`. An envelope that is present but names its subscription in a shape no router id can stand in for returns an error, so the caller logs and skips that client exactly as before. Just pushed your trace on the remaining path — params present but not decodable as an object.

One correction to the reachability call: I believe that path is **not** reachable via delivery today. For a frame to arrive at the rewrite it must first have been delivered, and the dispatcher decoded those same params to find the subscription — so an `eth_subscription` frame with array-shaped params fails there and never reaches this function. The trace is there to make it visible if that ever stops being true, not because it fires now.

protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
AnnaR-prog: Best catch of the four — the docstring was asserting something nothing checked.

Added `"extra":"keepme"` to the Substrate rewrite test plus an assertion that it survives. And verified it is load-bearing rather than assuming: reverting the rebuild to a fixed `{subscription, result}` map makes exactly that assertion fail and nothing else in the suite notice. That is precisely the silent regression you described.

AnnaR-prog:
Thanks — this was a careful review, and the mutation-testing of the verification table is appreciated. All four comments were correct. Replies are on the individual threads; this is the summary, plus one thing the review could not have seen.

## What changed, per comment

| Your comment | Outcome | Where |
| --- | --- | --- |
| Duplicate handler — collapse into `handleSubscriptionResultEthereum` | Resolved differently — see below | #361 |
| Predicate has no test in its own package | Added, covering your boundary list | #361 |
| `nil` is overloaded; EVM behaviour changed silently | Took your option B, plus your trace | #361 + `3e62c66` |
| "Sibling fields are preserved" is untested | Added, and proved load-bearing | `3e62c66` |

**Duplicate handler.** You were right that it was a byte-for-byte copy, but I went the other way rather than collapsing. `handleSubscriptionResultByParams` is gone, replaced by `deliverSubscriptionPush(msg, id)`, which the Ethereum handler cannot stand in for: it takes the pre-parsed id (your rider — you judged it not worth restructuring the switch, and it turned out to be, once the id had to be threaded anyway), it returns whether it *claimed* the frame, and an id-bearing frame that matches no subscription now falls through to the call path so a waiting peer still gets a reply. As written in this PR such a frame was claimed and swallowed.

**Predicate tests.** One row is deliberately inverted from your list: **integer id is now `true`**. The string-only guard you are leaning on to keep Solana out is removed on purpose in #361 — Solana subscriptions were dead for their own reason, and the fix follows the chain's id shape instead of excluding it.

**Sibling fields.** Verified rather than assumed: reverting the rebuild to a fixed `{subscription, result}` map makes exactly the new assertion fail and nothing else in the suite notice — the silent regression you described.

## The thing this review could not see: **this PR should not merge alone**

You reviewed #360 in isolation, which is the right way to review it — but four defects in this diff are fixed only in #361, and merging this on its own lands them on `main`. Each is probed, not argued:

- **`null` and empty-string subscription ids are rewritten** while `isSubscriptionNotification` rejects the same frames. Two guards in one PR disagreeing about what an envelope is. `{"subscription":null,...}` comes back as `{"subscription":"rs_abc_00001",...}`.
- **The new case sits after `isStarkNetPathfinderNotification`**, whose predicate is shape-based (`ID == nil && Method != "" && Result != nil`) and matches any push that also carries a top-level `result` — including `result: null`, since `json.RawMessage("null")` is non-nil. Such a frame is preempted, `handleImmediate` returns false, and it gets answered `invalid request`: the exact MAG-3345 symptom. Your note that placing the case last is what makes it safe to widen is right about the *method-name* cases, but this one is not a method-name case. #361 moves the case above it.
- **Id-bearing frames are claimed and swallowed** with no reply.
- **Unmatched pushes are dropped with no log**, in the subsystem whose failure mode the ticket calls "a silent capability gap".

Two more, of a different kind:

- `TestRewriteSubscriptionID_SolanaUntouched` asserts the client receives the **upstream** id — verbatim the outcome this PR's own body calls fatal. It is superseded in #361.
- `json.go` ships the claim that delivering Solana "needs an integer-id router mapping the id mapper cannot express." That is **wrong**, and MAG-3359 disproved it: the mapper changed by +33 lines with zero map-type changes. The same sentence is in this PR's body. Both are corrected in #361.

**Suggested merge order:** fold #361 into this PR, or merge them back-to-back so the intermediate state never sits on `main`. Folding is cleaner and makes the disproved comment never ship.

## Still open, on both

- No approvals on either PR.
- Mock-verified only — no live Substrate or Solana node, and Phase 8's 24 Aventus methods were never re-probed. Stated in both bodies; the "fixed" claim on MAG-3345 should wait for that.
- Merging deploys nothing here, and neither body names a deploy owner.
- `protocol/lavasession` flakes on `TestBlockRecord_CarriedRecordTakesTheScopeItIsReBlockedIn` — 3 of 14 runs on the follow-up branch, 0 of 4 at this head. No code path from either diff reaches it; I could not identify the mechanism and did not run it at the merge base, so it is unattributed rather than cleared.

avitenzer (approved):

PR #361, opened 2026-09-01
fix(rpcsmartrouter): follow the chain's own subscription id shape (MAG-3359)

AnnaR-prog:
Closes [MAG-3359](https://magmadevs.atlassian.net/browse/MAG-3359)

Jira ticket: MAG-3359

> **Stacked on #360** (MAG-3345). Review that first; this branch builds on its params-envelope matching. The diff below is the second commit only.

## The bug

Solana subscriptions never delivered a push frame. Same silent symptom as MAG-3345 — subscribe succeeds, then nothing ever arrives — but a different cause: **Solana is the only supported chain that numbers its subscriptions rather than naming them**, and every id-handling site assumed a string.

`specs/solana.json` declares 9 subscribe/unsubscribe pairs on the `WS-ONLY` collection at 1000 CU each, on both `SOLANA` and `SOLANAT`.

Six sites, one theme:

| # | Site | Behaviour with a numbered id |
| --- | --- | --- |
| 1 | `handleImmediate` | never recognised `accountNotification` as a push, so — as in MAG-3345 — the router answered the node with an `invalid request` for every frame |
| 2 | `extractSubscriptionID` | decoded `23784` as a string, got `""`, and registered a mapping to nothing |
| 3 | `createSubscriptionReply` | handed the client a *string* router id it cannot pass to `accountUnsubscribe` |
| 4 | `rewriteSubscriptionID` | required a string, so frames reached the client carrying the upstream id |
| 5 | `extractSubscriptionIDFromUnsubscribe` | type-asserted the client's param to string and rejected it |
| 6 | upstream teardown | sent the router's canonical decimal string where the node needs its own number |

## The approach, and why it is smaller than the ticket feared

MAG-3359 predicted this would "change the id-generation contract" of `SubscriptionIDMapper`. It doesn't. There is **one canonical internal form** — the decimal string — and the string/number distinction lives only at the JSON boundary. Every map stays string-keyed.

The canonicalisation is shared with the rpcclient dispatcher through the newly exported `CanonicalSubscriptionID`, deliberately: the router's upstream id and the dispatcher's `clientSubs` key are the *same identity*, and a divergence between two private copies would silently strand every push.

Most sites read the shape off the message in front of them. Only two can't — the id issued to a *joining* client, and the param sent upstream on teardown — and those consult a `numericIDs` flag recorded on the subscription.

### Why the router issues its own numeric id

The ticket floated a cheaper Option B: relay the node's id and give up per-client dedup. **It was rejected because it is broken.** `handleUpstreamDisconnect` re-subscribes and the node answers with a *different* id, so a client holding the upstream one is silently left with an id naming nothing — pushes tagged with an id it never saw, and an unsubscribe that cannot be matched. Stability across reconnects is the entire reason the indirection exists.

`TestSolanaRouterIDSurvivesUpstreamReconnect` is that reasoning made executable: it drives the real `handleUpstreamDisconnect` against a live mock and asserts the client's id is unchanged while the upstream id moves.

## Verification

Every one of the six defects was individually reverted and confirmed to fail — the tests reference new symbols, so this is six targeted reverts rather than one file-level one:

| reverted | resulting failure |
| --- | --- |
| 1 dispatcher | `no push frame reached the client within 5s` |
| 2 upstream id | `the node must have been sent its own integer id` |
| 3 reply shape | `result must be a JSON number, got "1000000001"` |
| 4 push rewrite | id mismatch — arrives quoted |
| 5 client param | `unsubscribe with the issued numeric id must not be rejected` |
| 6 upstream teardown | node never told its own integer id |

`protocol/chainlib/...` and `protocol/rpcsmartrouter` are green under `-race`; `go vet ./...` and `gofmt` clean.

**Boundary:** verified against a mock upstream over a real WebSocket, not a live Solana node. This also closes the one claim MAG-3359 flagged as unproven — that the dispatcher's downstream machinery already handles integers via `strconv` — since the e2e only passes if it does.

**Unrelated flake seen while verifying:** `protocol/lavasession` fails intermittently on `TestBlockRecord_CarriedRecordTakesTheScopeItIsReBlockedIn` — 3 of 14 runs on this branch, 0 of 4 on the parent, which is consistent with a single ~20% flake rate rather than a difference between the branches. There is no code path from this diff to it: the changed functions have no callers in that package, and the test touches no JSON, WebSocket or subscription code. The failures clustered under heavy parallel load and included a second, different test whose message was explicitly a lock-contention timeout. I did **not** identify the mechanism — the ordering is deterministic (`sort.Slice`, one record per address) and `blockProvider` is called here with no expiry and `allowSecondChance` false, so my first two hypotheses were both wrong. Worth its own ticket; not addressed here.

## Tests

The harness could only speak EVM, which is why nothing caught this. The mock upstream can now number its subscriptions and mints **distinct ids per subscription** — without which neither the reconnect nor the join path could be expressed at all. It also rejects a quoted id when it is numbering, which is what makes site 6 observable; a lenient mock accepts the router's decimal string and hides the bug.

MAG-3345's `TestRewriteSubscriptionID_SolanaUntouched` is **superseded**. It pinned the old boundary deliberately and correctly — a string router id is one a Solana client cannot use — and this is the ticket that moves it.

The dead `solanaNotificationMethodSuffix` branch in `handleImmediate` is left in place with a comment pointing at the generic case, so removing it stays a separate, reviewable change rather than noise in this diff.

[MAG-3359]: https://magmadevs.atlassian.net/browse/MAG-3359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/protocol/chainlib/chainproxy/rpcclient/handler.go
+++ b/protocol/chainlib/chainproxy/rpcclient/handler.go
@@ -241,6 +241,11 @@ func (h *handler) handleImmediate(msg *JsonrpcMessage) bool {
 			h.handleSubscriptionResultEthereum(msg)
 			return true
 		} else if strings.HasSuffix(msg.Method, solanaNotificationMethodSuffix) {
+			// Unreachable: isEthereumNotification above already required the
+			// "_subscription" suffix, which no *Notification method has. Solana frames
+			// are served by the isSubscriptionNotification case below instead — left
+			// in place only so removing it is not confused with a behaviour change
+			// (MAG-3359).
 			h.handleSubscriptionResultSolana(msg)
 			return true
 		}
@@ -273,17 +278,16 @@ func (h *handler) handleImmediate(msg *JsonrpcMessage) bool {
 // fell through to handleCallMsg, which has no id and no matching method to serve, so the
 // router answered the node with an "invalid request" error for every push it received.
 func (h *handler) handleSubscriptionResultByParams(msg *JsonrpcMessage) {
-	var result ethereumSubscriptionResult
-	if err := json.Unmarshal(msg.Params, &result); err != nil {
+	id, ok := subscriptionIDFromParams(msg.Params)
+	if !ok {
 		utils.LavaFormatTrace("Dropping invalid subscription message",
-			utils.LogAttr("err", err),
 			utils.LogAttr("params", string(msg.Params)),
 		)
 		h.log.Debug("Dropping invalid subscription message")
 		return
 	}
-	if h.clientSubs[result.ID] != nil {
-		h.clientSubs[result.ID].deliver(msg)
+	if h.clientSubs[id] != nil {
+		h.clientSubs[id].deliver(msg)
 	}
 }
 
--- a/protocol/chainlib/chainproxy/rpcclient/handler.go
+++ b/protocol/chainlib/chainproxy/rpcclient/handler.go
@@ -17,6 +17,7 @@
 package rpcclient
 
 import (
+	"bytes"
 	"context"
 	"reflect"
 	"strconv"
@@ -299,7 +300,10 @@ func (h *handler) deliverSubscriptionPush(msg *JsonrpcMessage, subscriptionID st
 		sub.deliver(msg)
 		return true
 	}
-	if msg.hasValidID() {
+	// An explicit "id":null is not a request id — hasValidID only rejects objects and
+	// arrays, so `null` passes it. Letting that through to the call path answers the node
+	// with "invalid request" for a late push, which is the MAG-3345 symptom itself.
+	if msg.hasValidID() && !bytes.Equal(msg.ID, null) {
 		return false
 	}
 	// Late frames after an unsubscribe land here routinely, so this is a trace and not a
--- a/protocol/chainlib/chainproxy/rpcclient/handler.go
+++ b/protocol/chainlib/chainproxy/rpcclient/handler.go
@@ -243,13 +243,32 @@ func (h *handler) handleImmediate(msg *JsonrpcMessage) bool {
 		} else if strings.HasSuffix(msg.Method, solanaNotificationMethodSuffix) {
 			// Unreachable: isEthereumNotification above already required the
 			// "_subscription" suffix, which no *Notification method has. Solana frames
-			// are served by the isSubscriptionNotification case below instead — left
-			// in place only so removing it is not confused with a behaviour change
-			// (MAG-3359).
+			// are served by the shape-based case below instead — left in place only so
+			// removing it is not confused with a behaviour change (MAG-3359).
 			h.handleSubscriptionResultSolana(msg)
 			return true
 		}
 		return false
+	case msg.Method != "" && msg.Params != nil:
+		// Shape-based delivery: a method-bearing frame whose params name a subscription.
+		// The method-name cases above keep first claim on anything they already
+		// recognise; this catches the rest, which is how Substrate and Solana pushes
+		// arrive — see subscriptionIDFromParams.
+		//
+		// It sits BEFORE the StarkNet case deliberately. That predicate matches any
+		// id-less method-bearing frame carrying a result, so a push that also sets a
+		// top-level result would be swallowed there and answered with "invalid request"
+		// — the very MAG-3345 symptom this exists to remove. StarkNet's own frames put
+		// the id in result and carry no params, so they never enter this case.
+		//
+		// The id is parsed once and handed to the delivery path. The sibling handlers
+		// above re-parse inside the handler, which here would mean decoding the whole
+		// payload twice for every frame delivered.
+		id, named := subscriptionIDFromParams(msg.Params)
+		if !named {
+			return false
+		}
+		return h.deliverSubscriptionPush(msg, id)
 	case msg.isStarkNetPathfinderNotification():
 		if strings.HasSuffix(msg.Method, ethereumNotificationMethodSuffix) {
 			h.handleSubscriptionResultStarkNetPathfinder(msg)
@@ -260,35 +279,38 @@ func (h *handler) handleImmediate(msg *JsonrpcMessage) bool {
 		h.handleResponse(msg)
 		h.log.Trace("Handled RPC response", "reqid", idForLog{msg.ID}, "duration", time.Since(start))
 		return true
-	case msg.isSubscriptionNotification():
-		// Last, so the method-name cases above keep first claim on anything they
-		// already recognise. This one matches on the params envelope alone, which is
-		// how Substrate pushes arrive — see isSubscriptionNotification.
-		h.handleSubscriptionResultByParams(msg)
-		return true
 	default:
 		return false
 	}
 }
 
-// handleSubscriptionResultByParams delivers a subscription notification identified by its
-// params envelope rather than by its method name.
+// deliverSubscriptionPush hands a push frame to the subscription it names, reporting
+// whether this handler claimed the frame.
 //
-// Returning false for these from handleImmediate was worse than dropping them: the frame
-// fell through to handleCallMsg, which has no id and no matching method to serve, so the
-// router answered the node with an "invalid request" error for every push it received.
-func (h *handler) handleSubscriptionResultByParams(msg *JsonrpcMessage) {
-	id, ok := subscriptionIDFromParams(msg.Params)
-	if !ok {
-		utils.LavaFormatTrace("Dropping invalid subscription message",
-			utils.LogAttr("params", string(msg.Params)),
-		)
-		h.log.Debug("Dropping invalid subscription message")
-		return
+// Claiming matters as much as delivering. Before MAG-3345 an unrecognised push fell
+// through to handleCallMsg, which had no id and no matching method to serve, so the router
+// answered the node with an "invalid request" for every frame it received. But a frame that
+// carries a request id is a request the peer expects answered, and swallowing it silently
+// would leave the peer waiting on a reply that never comes — so an id-bearing frame that
+// matches no subscription is handed back to the call path, while an id-less one (a genuine
+// JSON-RPC notification, which nobody is waiting on) is claimed and dropped with a trace.
+func (h *handler) deliverSubscriptionPush(msg *JsonrpcMessage, subscriptionID string) bool {
+	if sub := h.clientSubs[subscriptionID]; sub != nil {
+		sub.deliver(msg)
+		return true
 	}
-	if h.clientSubs[id] != nil {
-		h.clientSubs[id].deliver(msg)
+	if msg.hasValidID() {
+		return false
 	}
+	// Late frames after an unsubscribe land here routinely, so this is a trace and not a
+	// warning — but it is logged, because a subscription registered under a key the push
+	// does not name is otherwise indistinguishable from a healthy idle one, which is
+	// exactly the failure mode MAG-3345 was.
+	utils.LavaFormatTrace("Dropping subscription push with no matching subscription",
+		utils.LogAttr("method", msg.Method),
+		utils.LogAttr("subscriptionID", subscriptionID),
+	)
+	return true
 }
 
 func (h *handler) handleSubscriptionResultStarkNetPathfinder(msg *JsonrpcMessage) {
@@ -382,11 +404,12 @@ func (h *handler) handleResponse(msg *JsonrpcMessage) {
 		go op.sub.run()
 		h.clientSubs[op.sub.subid] = op.sub
 	} else {
-		// This is because StarkNet Pathfinder is returning an integer instead of a string in the result
-		var integerSubId int
+		// This is because StarkNet Pathfinder is returning an integer instead of a string in the result.
+		// int64 rather than int so this key cannot drift from CanonicalSubscriptionID's on a 32-bit build.
+		var integerSubId int64
 		if json.Unmarshal(msg.Result, &integerSubId) == nil {
 			op.err = nil
-			op.sub.subid = strconv.Itoa(integerSubId)
+			op.sub.subid = strconv.FormatInt(integerSubId, 10)
 			go op.sub.run()
 			h.clientSubs[op.sub.subid] = op.sub
 		}
--- a/protocol/rpcsmartrouter/direct_ws_subscription_manager.go
+++ b/protocol/rpcsmartrouter/direct_ws_subscription_manager.go
@@ -1,6 +1,7 @@
 package rpcsmartrouter
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -43,10 +44,15 @@ type directActiveSubscription struct {
 	subscribeMethod    string // Method name (e.g., "eth_subscribe", "subscribe")
 
 	// numericIDs records that this chain numbers its subscriptions rather than naming
-	// them, read from the subscribe response the node actually sent. It is a property of
-	// the chain, so it survives a reconnect even though upstreamID does not. Only the two
-	// places that cannot read the shape off the message in front of them consult it: the
-	// id issued to a joining client, and the param sent upstream on unsubscribe.
+	// them, read once from the subscribe response the node actually sent. It is a property
+	// of the chain, so it survives a reconnect even though upstreamID does not.
+	//
+	// It is the ONLY source of this fact. Every site that renders an id for the wire — the
+	// subscribe reply, a joining client's reply, the per-frame rewrite, the param sent
+	// upstream on teardown — reads it rather than re-sniffing whatever JSON is nearest,
+	// because re-deriving invites the sites to disagree: an upstream that numbers the
+	// subscribe result but quotes the id in its pushes would otherwise hand the client a
+	// number at subscribe time and quoted ids in every frame after.
 	numericIDs bool
 
 	// Client tracking - multiple clients can share one upstream subscription
@@ -771,7 +777,7 @@ func (dwsm *DirectWSSubscriptionManager) StartSubscription(
 	// requestID is the client's original JSON-RPC id (extracted at the top of this function);
 	// passing it explicitly ensures the response echoes the caller's id verbatim, including
 	// non-numeric ids like string UUIDs (JSON-RPC 2.0 §4.2).
-	firstReplyData, err := createSubscriptionReply(routerID, requestID, firstMsg, dwsm.apiInterface)
+	firstReplyData, err := createSubscriptionReply(routerID, requestID, firstMsg, dwsm.apiInterface, numericIDs)
 	if err != nil {
 		upstreamSub.Unsubscribe()
 		dwsm.failPendingSubscription(hashedParams)
@@ -1137,7 +1143,6 @@ func (dwsm *DirectWSSubscriptionManager) UnsubscribeAll(
 	return nil
 }
 
-// getHashedParams extracts and hashes the subscription parameters from a protocol message
 // generateRouterID issues a router subscription id shaped like the ones this chain uses.
 //
 // Chains that number their subscriptions get a number, because the client hands the id
@@ -1152,6 +1157,7 @@ func (dwsm *DirectWSSubscriptionManager) generateRouterID(clientKey string, nume
 	return dwsm.idMapper.GenerateRouterID(clientKey)
 }
 
+// getHashedParams extracts and hashes the subscription parameters from a protocol message
 func (dwsm *DirectWSSubscriptionManager) getHashedParams(protocolMessage chainlib.ProtocolMessage) (hashedParams string, params []byte, err error) {
 	params, err = gojson.Marshal(protocolMessage.GetRPCMessage().GetParams())
 	if err != nil {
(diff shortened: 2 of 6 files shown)


protocol/chainlib/chainproxy/rpcclient/handler.go
```diff
@@ -255,36 +279,38 @@ func (h *handler) handleImmediate(msg *JsonrpcMessage) bool {
...
+// matches no subscription is handed back to the call path, while an id-less one (a genuine
+// JSON-RPC notification, which nobody is waiting on) is claimed and dropped with a trace.
+func (h *handler) deliverSubscriptionPush(msg *JsonrpcMessage, subscriptionID string) bool {
+	if sub := h.clientSubs[subscriptionID]; sub != nil {
+		sub.deliver(msg)
+		return true
 	}
-	if h.clientSubs[result.ID] != nil {
-		h.clientSubs[result.ID].deliver(msg)
+	if msg.hasValidID() {
```
avitenzer: **`"id": null` re-introduces the MAG-3345 symptom on unmatched pushes.**

The reasoning above is right — a frame carrying a request id is one the peer expects answered, so handing it back to the call path is correct. But `hasValidID()` is:

```go
return len(msg.ID) > 0 && msg.ID[0] != '{' && msg.ID[0] != '['
```

so `"id": null` reads as a **valid** id: `len("null") > 0`, and `'n'` is neither brace nor bracket.

So a late push arriving after an unsubscribe with `"id":null` falls through to `handleCallMsg`, which has no id and no matching method to serve, and the router answers the node with `invalid request` — precisely the symptom MAG-3345 removed. At the #360 base, `handleSubscriptionResultByParams` always claimed these.

Narrow in practice (needs a node that stamps ids on pushes *and* a non-`_subscription` method), but the fix is one clause, and the package already has the sentinel — `var null = json.RawMessage("null")` at json.go:47, used by `errorMessage`:

```go
if msg.hasValidID() && !bytes.Equal(msg.ID, null) {
    return false
}
```

Worth a case in `TestSubscriptionIDFromParams`' sibling coverage too, since `null` is exactly the shape that already fooled one decode in this diff.

protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
```diff
@@ -2858,3 +2964,370 @@ func TestEndToEnd_SubstratePushFrameReachesClient(t *testing.T) {
...
+	case notif := <-repliesChan:
+		require.NotNil(t, notif)
+		var frame struct {
+			Params struct {
+				Subscription json.RawMessage `json:"subscription"`
+			} `json:"params"`
+		}
+		require.NoError(t, json.Unmarshal(notif.Data, &frame))
+		assert.Equal(t, strconv.FormatInt(routerSubID, 10), string(frame.Params.Subscription),
+			"post-reconnect frames must still carry the id the client holds, not upstream %q", upstreamAfter)
```
avitenzer: **This assertion can be satisfied by a *pre*-reconnect frame, so it doesn't guard the restored path.**

I traced it through. The test calls `handleUpstreamDisconnect` directly, so:

1. `ReconnectWithBackoff` short-circuits at the `hasHealthy` check (`upstream_ws_pool.go:385`) and returns `nil` — **no reconnect actually happens**.
2. `GetConnection` therefore hands back the *same* connection.
3. The old upstream subscription (`23784`) is never `Unsubscribe`d, so the mock's goroutine for it keeps pushing every 20ms on that same connection, and the old `listenForUpstreamMessages` keeps forwarding to `routeMessageToClients`.
4. Both old and new frames rewrite to the identical client router id.

After the 200ms drain, the frame read here is as likely to come from `23784` as from `23785`. The scope note above acknowledges both listeners fan out to the same clients — but the conclusion drawn ("the assertions below are unaffected") is about *correctness* of the assertion, whereas the message reads as a claim about *provenance*: "post-reconnect frames must still carry...".

The two assertions that carry this test — `upstreamAfter != upstreamBefore` and `routerIDAfter == routerSubID` — are solid and do prove the design point the PR argues for. Either give the restored subscription distinguishable payload so provenance is assertable, or soften this message so it doesn't read as a guard it isn't.

protocol/rpcsmartrouter/subscription_id_mapper.go
```diff
@@ -62,6 +83,18 @@ func (m *SubscriptionIDMapper) GenerateRouterID(clientKey string) string {
...
+// GenerateNumericRouterID creates a router subscription ID for chains that number their
+// subscriptions rather than naming them (Solana). The id is returned in the same canonical
+// decimal-string form as every other id here — every map in this type stays string-keyed —
+// and is rendered as a JSON number only where it reaches the wire.
+//
+// Unlike GenerateRouterID the counter is not per-client: the value is visible to the client
+// and is matched against upstream ids on unsubscribe, so it must be unique across every
+// client this mapper serves.
+func (m *SubscriptionIDMapper) GenerateNumericRouterID() string {
+	return strconv.FormatUint(numericRouterIDBase+m.numericCounter.Add(1), 10)
```
avitenzer: **A single global counter drops the per-client namespacing the string path deliberately has.**

`GenerateRouterID` is `rs_{clientHash6}_{counter5}` — the hash means one client learns nothing about another's activity from its own ids. This generator is one process-wide monotonic counter handed to the client verbatim, so any Solana client can subscribe twice and read off exactly how many subscriptions the whole manager issued in between, across every dApp and IP it serves.

To be clear about severity: I checked, and this is **not** an authorization hole. The ownership loop in `Unsubscribe` gates on `clientRouterIDs[clientKey]` before either id comparison, so a guessed id gains an attacker nothing. It's an information leak — a rate oracle — and it's a property the string path was careful to have.

The docstring's reason for the global scope is sound (ids must be unique manager-wide and can't be hash-namespaced). But uniqueness doesn't require *predictability*: seeding `numericCounter` from `crypto/rand` in `NewSubscriptionIDMapper`, within the 2^40 → 2^53 headroom, keeps every property you documented and removes the oracle.

Flagging for an explicit decision rather than expecting a change — it may well be an acceptable trade here, but it's worth being deliberate about, since the string path's namespacing looks intentional.

protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
```diff
@@ -2858,3 +2964,370 @@ func TestEndToEnd_SubstratePushFrameReachesClient(t *testing.T) {
 		t.Fatal("no push frame reached the client within 5s — the subscription is dead end-to-end")
 	}
 }
+
+// --- MAG-3359: chains that number their subscriptions -----------------------------------
+
+// TestUnsubscribeParamsExtraction_SolanaNumeric covers the client sending its id back as a
+// JSON number. A string-only type assertion rejected it outright, so the unsubscribe failed
+// before it ever reached the ownership check.
+func TestUnsubscribeParamsExtraction_SolanaNumeric(t *testing.T) {
```
avitenzer, line 2982: **This passes `int64`; production passes `float64`, so the rounding the code comment worries about is never exercised.**

`JsonrpcMessage.Params` is `interface{}` (`rpcInterfaceMessages/jsonRPCMessage.go:29`), so by the time `extractSubscriptionIDFromUnsubscribe` re-marshals a client's numeric id it is a `float64` — exactly the path the comment at `direct_ws_subscription_manager.go:1202-1207` calls out. Every numeric test here (and the id passed to `Unsubscribe` in the e2e) hands it `int64(...)` or a Go-side `int64` variable, which round-trips trivially.

I probed the real path directly — decode `{"params":[N]}` into `interface{}`, re-marshal with gojson, decode into `json.RawMessage`, run `CanonicalSubscriptionID` — at 2^40, 2^40+1, 2^40+99999 and 2^53−1:

```
id=1099511627776    goType=float64  remarshalled=[1099511627776]    canonical="1099511627776"    match=true
id=1099511627777    goType=float64  remarshalled=[1099511627777]    canonical="1099511627777"    match=true
id=1099511727775    goType=float64  remarshalled=[1099511727775]    canonical="1099511727775"    match=true
id=9007199254740991 goType=float64  remarshalled=[9007199254740991] canonical="9007199254740991" match=true
```

So **no live bug**, and the 2^40 base choice is vindicated — gojson emits `f` format well below the 1e21 exponent threshold, so nothing is lost. But the guarantee currently lives only in a prose comment. One case here with `params: []interface{}{float64(1 << 40)}` would pin it, and would fail loudly if anyone ever raises the base toward 2^53.

protocol/rpcsmartrouter/subscription_id_mapper.go
```diff
@@ -30,9 +31,29 @@ type SubscriptionIDMapper struct {
...
+// own, so a router id can never collide with some other subscription's upstream id during
+// the unsubscribe lookup — without that lookup having to know which is which.
+//
+// The base is deliberately far above anything a node plausibly issues rather than merely
+// above what Solana is observed to issue today, so the guarantee does not rest on node
+// behaviour. The ceiling is the other side of it: ids are JSON numbers, and a JavaScript
+// client parses them as float64, so base plus counter has to stay under 2^53 to survive the
+// round trip. 2^40 leaves roughly four thousand times more headroom than any one process
+// will ever need.
+const numericRouterIDBase = 1 << 40
```
avitenzer, line 57: **The ceiling is derived; the base isn't — and it buys less than the comment implies.**

2^53 is a real bound (float64 mantissa) and `TestGenerateNumericRouterID` asserts it. 2^40 isn't derived from anything, and the comment concedes as much ("deliberately far above anything a node plausibly issues"). Reviewers here have pushed back on exactly this shape before, so worth pre-empting.

More usefully — I traced what the base actually protects. The comment says "so a router id can never collide with some other subscription's upstream id during the unsubscribe lookup", and that's precisely right, but it's also the *only* thing it protects: `routerToUpstream` and `upstreamToRouters` are separate namespaces, so a router id equal to some upstream id causes no trouble in the mapper itself.

That leaves one consumer — the `|| sub.upstreamID == subIDFromRequest` arm of the ownership loop, which exists as a legacy allowance for clients echoing back an upstream id they observed. Under this PR a Solana client is never shown the upstream id, so gating that arm on `!numericIDs` would remove the need for the constant entirely.

Optional — the current defense is harmless and cheap. But if the magic number draws fire in review, that's the exit.

protocol/rpcsmartrouter/direct_ws_subscription_manager.go
```diff
@@ -1799,59 +1898,68 @@ func createSubscriptionReplyFromRouterID(routerID string, requestID json.RawMess
...
 	raw, ok := fields["subscription"]
 	if !ok {
-		return nil
+		return nil, nil
 	}
-	var upstreamID string
-	if err := json.Unmarshal(raw, &upstreamID); err != nil {
-		return nil
+	if _, named := rpcclient.CanonicalSubscriptionID(raw); !named {
+		return nil, fmt.Errorf("subscription id is neither a string nor a number: %s", raw)
```
avitenzer: Nit: the message will be wrong in two of the cases that reach it.

`CanonicalSubscriptionID` reports `!named` for `null` and for `""` as well as for objects/arrays/floats — and `null` *is* a JSON null while `""` *is* a string, so "neither a string nor a number" would misdescribe both.

Unreachable in practice today (the dispatcher won't deliver a frame whose id names nothing, so `routeMessageToClients` never sees one), which is exactly why it'll be read cold by whoever it eventually fires on. `"subscription id names no subscription: %s"` covers every case the predicate actually rejects.

The test asserting `Contains(err.Error(), "neither a string nor a number")` would need the same edit.

avitenzer (commented):
Reviewed the second commit only (base is #360), and verified the claims against the source rather than the PR body.

**The design holds up.** One canonical internal form with the string/number distinction confined to the JSON boundary is the right call, and it lines up with what `handleResponse`'s pre-existing StarkNet integer fallback already produced — exporting `CanonicalSubscriptionID` turns that coincidence into an enforced invariant. Recording `numericIDs` on the subscription instead of re-sniffing per frame is also right, and the comment explains the failure mode it prevents. Rejecting the "relay the node's id" option on reconnect-stability grounds is correct.

**Case ordering in `handleImmediate` checks out** — I went predicate by predicate rather than trusting the comment, since a matched `case` cannot fall through, so inserting one *removes* messages from every case below it:
- `isResponse()` requires `Method == "" && Params == nil` — mutually exclusive with the new predicate, so responses can't be shadowed.
- `isStarkNetPathfinderNotification()` frames put the envelope in `Result` and carry no `Params`, so they can't enter the new case either.
- The "unreachable Solana branch" claim is accurate: `ethereumNotificationMethodSuffix = "_subscription"` and `solanaNotificationMethodSuffix = "Notification"` (json.go:41-42), and no method ends with both.

### Verification I ran locally

| check | result |
| --- | --- |
| `go build ./...`, `go vet` (both packages) | clean |
| `go test ./protocol/chainlib/chainproxy/rpcclient/... -race -count=1` | ok, 2.2s |
| `go test ./protocol/rpcsmartrouter/... -race -count=1` | ok, 95s |
| float64 round-trip probe, 2^40 … 2^53−1 (see the comment on the unsubscribe test) | exact at every value |

CI is fully green, but on this repo that is builds / lint / GoReleaser / GitGuardian only — **no Go tests run in CI**, so the run above is the only test evidence for this PR.

### What I'd like changed

Six inline comments. Two I'd want addressed before approval:
- **`handler.go`** — `"id": null` on an unmatched push re-introduces the MAG-3345 `invalid request` symptom. One clause.
- **the reconnect test** — the final assertion can pass on a pre-reconnect frame, so it doesn't guard what its comment says it guards.

One worth an explicit decision rather than a silent accept: the numeric id generator drops the per-client namespacing the string path deliberately has. The remaining three are polish.

Nothing here blocks on correctness — the production behaviour is sound as written.

---

One process note: the revert table in the description is the strongest evidence in this PR, and it's the one part a reader can't reproduce from the diff. Since CI never runs tests here, that claim is load-bearing — worth putting the revert list in MAG-3359 where it outlives the PR page.

protocol/chainlib/chainproxy/rpcclient/handler.go
AnnaR-prog: Confirmed and fixed in `b469fc6`. Verified `hasValidID` directly before touching it — `null` → `true`, exactly as you said, so an unmatched push carrying `"id":null` was handed to the call path and answered `invalid request`.

Took your clause verbatim, using the existing sentinel. Also added the case you asked for, though at the layer that owns the decision rather than in `TestSubscriptionIDFromParams` (which only sees `params`, never `msg.ID`): `TestDeliverSubscriptionPush_ClaimsUnmatchedNotifications` tables all four id shapes — absent, `null`, numeric, string. Reverting the clause fails only the `null` row, so it is load-bearing rather than decorative.

protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
AnnaR-prog: You are right, and the trace is correct — I checked `ReconnectWithBackoff` and it does return early on the healthy-pool check, so no reconnect happens, the old subscription is never torn down, and the mock keeps pushing on `23784`. Both ids rewrite to the same client id, so the frame read after the drain had no provenance at all.

Took the first of your two options rather than softening the message: the mock's notifications now carry their own upstream id in the payload, and the test reads until one demonstrably from the restored subscription arrives, failing if a frame comes from neither. The per-frame id assertion now runs on every frame rather than one arbitrary one.

Kept the scope note, corrected to say what actually happens — that no reconnect occurs and the old listener stays live — since that is the part I had wrong.

protocol/rpcsmartrouter/subscription_id_mapper.go
AnnaR-prog: Good catch, and I agree it should be deliberate. Changed it, with one correction to the proposed remedy.

Seeding `numericCounter` from `crypto/rand` hides the **absolute** total but not the oracle you describe: the leak is the *delta* between two ids issued to the same client, and a random starting point leaves that delta exactly equal to the number of ids issued in between. So the stride is what has to vary, not the seed. Each id now advances the counter by a random 1..2^20 from `crypto/rand`.

That keeps every property documented: monotonic, so uniqueness stays structural rather than probabilistic; and the span from `numericRouterIDBase` to the 2^53 ceiling still holds ~1.7e10 ids at the average stride.

Agreed on severity — the ownership loop gates on `clientRouterIDs[clientKey]` before either id comparison, so this was never an authorization issue.

protocol/rpcsmartrouter/direct_ws_subscription_manager_test.go
AnnaR-prog: Right, and thanks for probing the real path — that saved me from assuming. Added a case at 2^40, 2^40+1, 2^40+99999 and 2^53−1 through `[]interface{}{float64(...)}`, which is the shape `GetRPCMessage` actually produces.

Related: the comment on that function used to claim `json.RawMessage` "keeps a large numeric id exact; a float64 would not". That was already corrected earlier in this branch — it is false for exactly the reason your probe shows, since the rounding happens in the caller before the decode. The comment now says the decode is for handling both shapes in one branch, and that the rounding is upstream. Your test case pins the part that is actually guaranteed.

protocol/rpcsmartrouter/subscription_id_mapper.go
AnnaR-prog: Agreed on both halves, and the trace is right: `routerToUpstream` and `upstreamToRouters` are separate namespaces, so the constant only protects the `|| sub.upstreamID == subIDFromRequest` arm.

Keeping it for now, deliberately. Gating that arm on `!numericIDs` is the more principled fix and I would take it in a follow-up, but it changes the ownership loop — the one piece of this file that decides whether a teardown is authorised — for no behavioural gain today, and this branch has already grown well past its original scope. Recorded here rather than silently accepted.

One thing the base does still buy beyond that arm: a log line reading `1099511627777` is identifiable as a router id rather than a node id, which is the debugging affordance the `rs_` prefix used to give for free.

protocol/rpcsmartrouter/direct_ws_subscription_manager.go
AnnaR-prog: Fixed in `b469fc6` — `"subscription id names no subscription: %s"`, and the test assertion updated with it.

Your reasoning is the reason it was worth changing: it is unreachable today, so whoever it eventually fires on will be reading it cold, and a message that misdescribes `null` and `""` — the two shapes most likely to reach it — is worse than no message.

AnnaR-prog:
Thanks — going predicate-by-predicate through the switch rather than trusting the comment, and probing the float64 round-trip at four boundary values, is exactly the review this needed. All six comments were correct. Individual replies are on the threads; summary below.

Everything is in `b469fc6`. CI green, and the full suites re-run locally under `-race`.

## The two you'd hold approval for — both fixed

**`"id":null` re-introduced the MAG-3345 symptom.** Verified `hasValidID` before touching it: it only rejects `{` and `[`, so a literal `null` reads as valid, exactly as you said. An unmatched late push carrying it was handed to the call path and answered `invalid request`. Took your clause with the existing `null` sentinel.

Added the coverage at the layer that owns the decision rather than in `TestSubscriptionIDFromParams` (which only ever sees `params`, never `msg.ID`): `TestDeliverSubscriptionPush_ClaimsUnmatchedNotifications` tables absent / `null` / numeric / string. Reverting the clause fails only the `null` row.

**The reconnect test's final assertion.** Your trace is right, and I confirmed it: called directly, `handleUpstreamDisconnect` returns early on the healthy-pool check, so no reconnect happens, the old subscription is never torn down, the mock keeps pushing on `23784`, and both ids rewrite to the same client id — the frame read after the drain had no provenance at all.

Took the first of your two options. The mock's notifications now carry their own upstream id, and the test reads until one demonstrably from the restored subscription arrives, failing if a frame comes from neither. The per-frame id assertion now runs on *every* frame instead of one arbitrary one. The scope note is corrected to say what actually happens.

## The explicit decision you asked for — changed, with one correction

**The id oracle is real, and `crypto/rand` seeding does not close it.** The leak you describe is the *delta* between two ids issued to the same client, and a random starting point leaves that delta exactly equal to the number issued in between. So the **stride** varies instead: each id advances the counter by a random 1..2^20 from `crypto/rand`.

Every property in the docstring survives — monotonic, so uniqueness stays structural rather than probabilistic, and the span to the 2^53 ceiling still holds ~1.7e10 ids. Agreed on severity: the ownership loop gates on `clientRouterIDs[clientKey]` before either id comparison, so this was never an authorization issue.

## Polish

- **float64 vs int64** — pinned at 2^40, 2^40+1, 2^40+99999 and 2^53−1 through `[]interface{}{float64(...)}`. Related: the comment on that function used to claim `json.RawMessage` "keeps a large numeric id exact" — false for exactly the reason your probe shows, since the rounding happens in the caller. Corrected earlier in this branch; your case pins what is actually guaranteed.
- **Error wording** — now `"subscription id names no subscription"`. The old text misdescribed `null` and `""`, the two shapes most likely to reach it, and your point that it will be read cold is why it was worth changing.

## Declined, deliberately

**`numericRouterIDBase`.** Your trace is right that its only consumer is the `|| sub.upstreamID == subIDFromRequest` arm, and gating that on `!numericIDs` would remove the constant. I'd take that in a follow-up — but it changes the one piece of that file deciding whether a teardown is authorised, for no behavioural gain today, and this branch has already grown well past its original scope. One thing the base still buys: a log line reading `1099511627777` is identifiable as a router id rather than a node id, which the `rs_` prefix used to give for free.

## Two notes back

**Merge order.** Four defects in #360 are fixed only here, so #360 should not merge alone — detail in my comment on that PR. Fold this into #360, or merge them back-to-back.

**Your process point is right and I'd rather not lose it:** since CI runs no Go tests on this repo, the revert table is the only test evidence and the one part a reader cannot reproduce from the diff. It belongs in MAG-3359 where it outlives the PR page. I'll add it there.

avitenzer (approved):

PR #362, opened 2026-09-02
feat(observability): say who served the relay and why it stopped (MAG-3331)

Tomelia1999:
#Closes MAG-3331
Jira ticket: MAG-3331

## The problem

Two customer captures (`agent_docs/bug-reports/missing-logs/`), two questions the logs could not
answer.

**Solana `getBlock`** — the primary held the request for its full 10s timeout, the hedge ticker fell
over to a backup, and the backup answered 14 MB half a second later:

```
+10002ms DEB direct RPC relay failed      endpoint=onerpc-primary  (10s deadline)
+10549ms DEB direct RPC relay succeeded   endpoint=tatum-backup  latency=547ms   ← the answer
+10699ms INF ProcessingResult RETURNED    has_reply=true  reply_size=14405401
```

Which provider produced those bytes appears **once, at DEBUG**. At INFO the request looks like it
returned 14 MB from nowhere.

**Polygon `eth_sendRawTransaction`** — a single provider timed out after 11s and the request died
with no second attempt and no `[BackupProviders]` line at all. The cause is `relaypolicy.Decide`
returning `Stop("Stateful")`: a possibly-executed write must not run twice. Correct, and invisible —
indistinguishable from the router giving up.

## What this adds

Both answers ride the line that **already fires once per request** on the return path.
`ProcessingResult RETURNED` reported `has_result`/`has_reply`/`reply_size` and nothing about the
outcome. It becomes `relay finished`:

```
INF relay finished  served_by=tatum-backup  status=200  stop_reason=Success   has_reply=true
INF relay finished  served_by=              status=0    stop_reason=Stateful  has_reply=false
```

- **`served_by`** — who produced the returned bytes. Filled **only when something was served**:
  `GetProvider()` still returns a provider on a failed result, and reporting that would make the
  field lie on exactly the requests an operator reads most carefully. The failure is already
  attributed by the per-endpoint line and by `error`.
- **`stop_reason`** — why there was no further attempt. Carried state machine → `Done` instruction →
  processor, mirroring the existing cross-validation fail-fast reason. No interface change.

This deliberately does **not** promote `policy.Decide`'s `Stop` to INFO — that would add a line to
every single-attempt request to say what the terminal line now says once.

Also: the circuit-breaker warning read *"All providers exhausted, stopping retries"* immediately
before the Solana request **succeeded** on an in-flight relay. It stops new attempts, not in-flight
ones, and now says so.

## Local verification

Router run at `--log-level info` against controllable mock upstreams, four shapes:

| case | scenario | result |
|---|---|---|
| A | healthy | `served_by=onerpc-primary  stop_reason=Success` |
| B | **Solana shape** — primary stalls, backup answers | `served_by=tatum-backup  stop_reason=Success` |
| C | **Polygon shape** — stateful write, timeout | `served_by=<empty>  stop_reason=Stateful` |
| D | every upstream 429 | `served_by=<empty>  stop_reason=AllProvidersExhausted` |

**Case D initially returned `stop_reason=` empty.** The `SendStop` arm stops the request without
consulting the policy, so it never set a reason — the circuit breaker and the first-message failure
both exit that way. They now name their own (`AllProvidersExhausted`, `FirstMessageFailed`). That
was found by running it, not by reading it.

```bash
make lint    # go vet ./...            — clean
make test    # go test ./... -count=1  — exit 0, 34 packages, 0 FAIL
```

## Reviewer notes

- **golangci-lint was not run locally** — the local binary is v1.64.8 (go1.24) against this repo's
  Go 1.26 target and refuses to load the config. `go vet` is clean; CI gates the rest.
- **Not included, deliberately:** per-attempt latency, `returned_block`, and a write-executed verdict
  for stateful methods. All defensible; none is needed for the two blind spots these captures show.
  The write-executed one is the most interesting follow-up — a context deadline *after dispatch*
  leaves "did my transaction execute?" genuinely unknown, unlike a 429.
- Independent of #351 (different files); the two can land in either order.

https://claude.ai/code/session_01ES82b8aH1E9jj8zRWerT8h

### DIFF
--- a/protocol/relaycore/interfaces.go
+++ b/protocol/relaycore/interfaces.go
@@ -49,6 +49,11 @@ type RelayStateSendInstructions struct {
 	Done           bool
 	RelayState     *RelayState
 	NumOfProviders int
+	// StopReason is why the state machine stopped, carried on the Done instruction so the
+	// request's final log line can say it. The policy already computes this for every stop
+	// ("Stateful", "MaxRetriesReached", …) and it was previously visible only at DEBUG, which
+	// left "why was there no second attempt?" unanswerable in production.
+	StopReason string
 }
 
 func (rssi *RelayStateSendInstructions) IsDone() bool {
--- a/protocol/relaycore/interfaces.go
+++ b/protocol/relaycore/interfaces.go
@@ -49,10 +49,9 @@ type RelayStateSendInstructions struct {
 	Done           bool
 	RelayState     *RelayState
 	NumOfProviders int
-	// StopReason is why the state machine stopped, carried on the Done instruction so the
-	// request's final log line can say it. The policy already computes this for every stop
-	// ("Stateful", "MaxRetriesReached", …) and it was previously visible only at DEBUG, which
-	// left "why was there no second attempt?" unanswerable in production.
+	// StopReason is why the state machine stopped ("Stateful", "MaxRetriesReached", …), carried
+	// on the Done instruction so the request's final log line can say why there was no further
+	// attempt. Empty on a non-Done instruction.
 	StopReason string
 }
 
--- a/protocol/relaycore/relay_processor.go
+++ b/protocol/relaycore/relay_processor.go
@@ -56,8 +56,8 @@ type RelayProcessor struct {
 	crossValidationFailFastReason string
 
 	// stopReason is why the state machine stopped this request ("Stateful", "MaxRetriesReached",
-	// "ProcessingTimeout", "Success", …). Carried here so the request's final log line can say why
-	// there was no further attempt — previously only visible at DEBUG on policy.Decide.
+	// "ProcessingTimeout", "Success", …), so the request's final log line can say why there was
+	// no further attempt. Guarded by rp.lock.
 	stopReason string
 }
 
@@ -301,8 +301,7 @@ func (rp *RelayProcessor) GetCrossValidationRelayDeadline() time.Time {
 }
 
 // SetStopReason records why the state machine stopped this request. Set once, on the Done
-// instruction, so the request's final line can report it — the same carrier idea as the
-// cross-validation fail-fast reason below.
+// instruction — the same carrier idea as the cross-validation fail-fast reason below.
 func (rp *RelayProcessor) SetStopReason(reason string) {
 	rp.lock.Lock()
 	defer rp.lock.Unlock()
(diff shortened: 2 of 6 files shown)


nimrod-teich:
Reviewed, and pushed the follow-ups to this branch as `0bd30cc` rather than leaving a list — happy to drop any of them.

**One thing needs a paired PR before this merges:** the rename breaks six panels of the war-room dashboard, silently. `fleet-war-room.json` pins nine Loki queries to the literal `"message":"ProcessingResult RETURNED"`, and eight also match `has_result` — a field this deletes. Every query ends in `or on() vector(0)`, so the incident dashboard would have gone on rendering a healthy 0% error rate over 0 requests rather than failing. Fixed in smart-router-infrastructure#370 (MAG-3385), which matches both message strings so either merge order is safe.

**The tests didn't cover the feature.** I removed the wiring at all three sources — the success send, the policy's stop branch, the circuit-breaker arm — and both `./protocol/relaycore` and the full 90s `./protocol/rpcsmartrouter` suite stayed green. `TestRelayStateSendInstructions_CarriesTheStopReason` builds a struct literal with the field set and asserts the field is set, so it would pass against a struct containing nothing else. `relay_stop_reason_test.go` now drives the real state machine through `GetRelayTaskChannel` and asserts the reason on the Done instruction for four terminal paths; the same mutation fails three of them. Worth noting the fourth path taught me something: with the retry budget at zero a node error stops for `ErrorToleranceExceeded`, not `MaxRetriesReached` as I'd assumed.

**The deadline paths were losing the reason this PR exists to report.** `checkAndHandleTimeout` hard-coded `"ProcessingTimeout"` over an already-recorded policy reason, so the Polygon case — a stateful write that stopped because it must not run twice, and then hit the deadline — reported only the deadline. `stopReasonOr` keeps the specific reason when there is one. Case C in your table lands on `Stateful` only when the provider's error arrives before the deadline; at 11s in the capture I'd expect the deadline to win, so it may be worth re-checking that row against the original.

Smaller: `"SendStop"` named the code arm rather than a reason (now `BatchSendFailed`); `stopReason` got its own mutex instead of borrowing `relayStateLock` — every access is on the state machine's own goroutine, and the comment saying `validateReturnCondition` reads it from another describes something that doesn't happen; comments trimmed to present tense per the team convention.

I also demoted the per-task loop scaffolding (`🎬 STARTING TASK CHANNEL LOOP`, `📨 RECEIVED TASK FROM CHANNEL`, `UPDATING BATCH`, `LOOPING BACK TO RECEIVE NEXT TASK`) to DEBUG. Your argument against promoting `policy.Decide` to INFO applies with more force to these — they fire once per *batch* to say what the terminal line now says once per request. Say the word if you'd rather keep them and I'll revert that hunk. `GOROUTINES LAUNCHED` at line ~2148 is the same shape but outside this diff, so I left it.

The `served_by`-only-when-served gate is the right call and the circuit-breaker message correction is accurate — in-flight relays genuinely can still win. Unrelated: the branch's `lint` is red on `chain_fetcher_addon_probe_test.go:98`, inherited from main, and CI reports it as passing because the gate is `continue-on-error`.

nimrod-teich (approved):

PR #372, opened 2026-09-04
feat(session): cross-pod sticky sessions through the cache backend

Tomelia1999:
# Description

A request carrying `lava-stickiness: <id>` now resolves to the same upstream on **every router
replica**. Before this, each pod kept its own in-process affinity table, so behind a load
balancer one id pinned to a different upstream per pod — silently, with a healthy-looking
sticky gauge on every one of them.

Gated on `--shared-state`, the same operator signal the chain-tip and tracker-gate sharing
already use. Without it nothing changes.

## Why

A customer's block ingestion service asks for the chain head, then asks for that block. If the
head comes from one node and the block fetch lands on a node a few blocks behind, they get an
empty result for a block that exists, and ingestion records a false gap. **Retries do not cover
it** — the lagging node answers HTTP 200 with an empty result, so nothing retries.

## Jira ticket

Jira ticket: MAG-3136

---

## The three decisions worth reviewing

**1. First-writer-wins, not last-write-wins.** Replicas judge upstream health locally, so under
last-write-wins two pods with different views would overwrite each other indefinitely and the
session would never settle. It also gives "do not evict a peer's claim" for free — a pod whose
claim is refused simply keeps its own routing locally.

**2. The write returns the effective claim, not an empty reply.** A pod that loses the race
learns the winner from its own write and adopts it inside the same request, rather than
disagreeing with its peer until the claim expires.

**3. A resolved claim is carried as `selectedProvider`.** Under fail-closed, a fleet claim and a
header-pinned provider have identical semantics — route here or fail — so reusing that path
gives the guarantee exactly and leaves the locked selection code untouched.

## Fail closed

When the registry cannot be reached the request **fails** rather than being served off an
unverified pin. Serving anyway is the split this removes, and for the reporting customer a
failed call beats an empty block that reads as a real chain gap.

Two costs, stated plainly:

- The cache backend becomes a hard dependency **for sticky traffic only**. Traffic without the
  header is unaffected by a registry outage (there is a test for this).
- A pin whose upstream is unhealthy *on this pod* becomes an error rather than being quietly
  served elsewhere.

Whether that is the right default is an open product question; fail-open is one decision point
away.

## Both cache backends, deliberately

Claims travel through the `KVStore` seam so **both** backends carry them. The endpoint-observation
gate keeps its state beside the relay caches and is therefore silently dead on the RESP backend —
acceptable there (worst case: each pod polls locally), but a correctness contract cannot degrade
that way.

- **In-process:** a mutex-guarded map, not ristretto. Ristretto's `Set` is asynchronous and may be
  dropped under admission pressure, which the monotonic int64 ops absorb with confirm-and-retry.
  Sound for a counter, wrong for a claim: a dropped write means two pods both believe they won.
- **RESP:** an atomic Lua script. Not `SET ... NX GET`, which needs Redis 6.2+ and is not
  exercisable under miniredis.

## Other properties

- **Claim lifetime is unchanged from today**, expressed in epochs. Epochs derive from wall clock,
  so replicas agree on staleness without comparing clocks.
- **A confirmed claim costs no round trip.** Only a pod that has never seen the session pays one.
  Safe because the registry keeps a claim longer than a pod trusts its local copy, and
  first-writer-wins stops a peer claiming a different upstream in that window.
- **The session id travels as a digest.** It is customer-supplied and may identify an end user.
- **No network call happens while `csm.lock` is held** — resolution runs before any lock.
- A failed read is an error, never a miss. Reporting "no claim" when the truth is "could not find
  out" invents a free claim and splits the session.

## Test evidence

End-to-end harness in `docker/sticky-test/`: nginx round-robin → 3 router replicas → 4
interchangeable upstreams → 1 cache pod.

| | one session id, 30 requests | 12 distinct session ids |
|---|---|---|
| feature ON | **1 upstream**, across all 3 replicas | spread over the pool |
| `STICKY_SHARED_STATE=false` | **2 upstreams** | **9 of 12 sessions split** |

The OFF row is the reported bug reproduced — it proves the harness detects the defect it tests.

`adopted` is non-zero on every pod (the cross-pod signal) and `error` is zero.

*(An earlier version of this description offered "`claimed` summed 13 across three pods, exactly
the 13 session ids" as proof of first-writer-wins. That does not hold: a peer claiming the same
upstream this pod would have picked is indistinguishable from winning and is also counted as
`claimed`, so the sum is not the invariant it looked like. Removed rather than reworded.)*

**The unit test nearly passed for the wrong reason.** Both test pods seed their optimizers
identically, so the two-pod agreement test could have passed without the feature. A control was
added: unshared pods disagreed on all 8 ids. It now lives permanently beside the test it guards,
so it fails loudly if independent pods ever converge.

The harness also handles three ways it could have lied: connection reuse pinning to one pod,
response caching hiding which upstream was selected, and a lagging head skewing selection.

## The reported bug, reproduced

`docker/sticky-test/run-falsegap.sh` runs the reported sequence — ask for the head, then ask for
that block — against a pool deliberately out of lockstep, on a config shaped after a real
production ConfigMap (three primaries with add-on urls, a backup tier, listener on :3000,
metrics declared in the config).

| | false gaps in 20 rounds |
|---|---|
| no affinity | **4-8** |
| with affinity | **0**, across all three replicas, work still spread over every upstream |

Phase A must reproduce the bug and aborts the run if it cannot; a clean phase B means nothing
otherwise. Phase D breaks the pinned upstream and reports what the retry does without asserting,
since a pin dropped on retry is a known limit rather than a regression.

## Two review rounds — five routing regressions fixed

Adversarial review found five ways this could route worse than main. Each is fixed with a
regression test that had no coverage before.

**Round two (`a7cd716`) — the one worth reading:**

- **The refill silently abandoned the claim.** Provider selection tops itself up in a loop. An
  upstream that is valid and healthy but cannot take *this* relay — compute units for the epoch
  spent, or at its session cap — was set aside, and the refill re-picked **with no pin**. A
  different upstream then served the request with no error, no metric and no log line. That is
  the exact split this feature exists to remove, arriving under ordinary load rather than during
  an incident, and it is the most likely cause of a ~3% split previously seen in the container
  harness with no identified reason.

  Verified against the reproduction that found it: **3/3 silently rerouted before, 3/3 fail
  closed after.**

  This also affects `lava-select-provider`, since by that point a resolved claim and a
  header-pinned provider are the same value. That header is documented as a hard pin, so
  rerouting it was a bug there too — but it is a behaviour change beyond this feature and is
  called out rather than buried.

**Round one (`6fd8c2b`):**

- **A resolved claim wedged the session.** It was never re-validated or released, and the local
  fast path replayed a dead decision for up to two epochs. The code it replaced self-healed.
- **Claims spanned service classes.** One session id used for a plain and an archive call
  permanently broke one of the two. Claims are now scoped per add-on/extension class.
- **`lava-select-provider` was silently overridden**, inverting a precedence that predates this
  feature, and only on deployments running `--shared-state`.
- **The TTL ceiling contradicted the reader window** at epoch lengths the flag's own help text
  suggests. Its guard test compared two constants and could never have caught it.

Two metric outcomes were also misfiled — an empty pairing recorded as `error` would have paged
for a provider outage.

## Known limits

- A pin is dropped on retry (MAG-2228, unchanged). A node error mid-session still sends the retry
  elsewhere.
- The per-pod pin table has no size bound. Pre-existing and unchanged, but load-bearing once
  customers build on the feature — depends on their session-id cardinality.
- The pin is still dropped on retry (MAG-2228, unchanged), so a genuine error mid-session can
  send the retry to a different upstream.
- No single-flight: concurrent relays for one cold session id each perform their own registry
  round trip. Measured at 16 relays producing 32 calls. Wasteful, not incorrect.
- `--shared-state` is already enabled wherever `maxReplicas > 1`, so the fail-closed dependency
  on the cache backend arrives with no new operator opt-in.
- Separately: `calculateBlockAvailability` is referenced only by its own unit test. The router
  never considers the requested block when choosing an upstream, which is the general fix for this
  class of bug and needs no session id at all. Worth its own ticket.

## Not yet done

- The branch is behind `main`. Merged locally to check: no conflicts, build and vet clean, full
  `protocol/lavasession` suite green. Needs updating before merge, nothing hidden in it.
- The ticket's obligations have not been checked against the diff by an independent reader —
  no Jira access from the machine this was built on.

## Review order

1. `protocol/lavasession/consumer_session_manager.go` — `resolveStickyPin` and its hook in
   `GetSessions` (the behaviour)
2. `ecosystem/cache/core/kvstore.go` — the claim contract
3. `ecosystem/cache/redisstore/store.go` — the Lua script
4. `docker/sticky-test/driver.py` — the control phases

Three commits, each building standalone. `go build ./...` and `go vet ./...` clean; all touched
packages green.

https://claude.ai/code/session_01XSnTE23nFbN2dT1YuBSF3q


protocol/lavasession/consumer_session_manager.go
```diff
@@ -1619,6 +1623,37 @@ func (csm *ConsumerSessionManager) GetSessions(ctx context.Context, wantedProvid
...
+	//
+	// selectedProvider WINS over a sticky claim when both are present. That order predates this
+	// feature — getValidProviderAddresses handles selectedProvider and returns before it ever
+	// reaches the sticky block — and naming a provider explicitly is the more specific ask of
+	// the two. Resolving stickiness here unconditionally would silently invert it, and only on
+	// deployments running --shared-state.
+	if stickiness != "" && selectedProvider == "" && csm.sharedSticky != nil {
+		resolved, localKey, stickyErr := csm.resolveStickyPin(ctx, stickiness, cuNeededForSession, requestedBlock, addon, common.GetExtensionNames(extensions), stateful)
+		if stickyErr != nil {
+			return nil, stickyErr
```
avitenzer, line 1669: **Sticky requests never reach the backup tier when the primary pool is empty/blocked.**

With `--shared-state`, a cold sticky id whose primary pool is empty returns `PairingListEmptyError` from `resolveStickyPin` here, before `getSessionWithProviderOrError` runs. That skips the backup-provider cascade, `releaseBlockedProvidersIfPoolEmpty`, and the blocked-provider walk that non-sticky and pod-local sticky traffic still get.

Scenario: all primaries blocked after a bad deploy, backups healthy. A non-sticky request hits `PairingListEmptyError` in `getValidConsumerSessionsWithProvider` and the cascade at ~1520-1553 serves it from backups. A new `lava-stickiness` id goes through `stickyCandidate` (~3870), which only consults `csm.validAddresses` (`backupProviders` is a separate map it never looks at), gets the same error, and `GetSessions` returns it right here. `rpcsmartrouter_server.go:3940-3955` either surfaces the error or re-enters this same early return on the extension-degradation retry. Every new sticky session hard-fails for the whole incident while identical non-sticky traffic is served.

No test in `sticky_shared_test.go` populates `backupProviders` or empties the primary pool. Suggest letting a `PairingListEmptyError` from the hook fall through to the normal path (blank stickiness, no claim) rather than returning.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -1619,6 +1623,37 @@ func (csm *ConsumerSessionManager) GetSessions(ctx context.Context, wantedProvid
...
+		}
+		selectedProvider = resolved
+		stickiness = ""
+		// A resolved claim is enforced as a hard pin below, so an upstream that is no longer
+		// selectable here fails this request. Drop the local claim on that outcome, or the fast
+		// path replays the same dead decision on every following request until it ages out.
+		defer func() {
+			if errors.Is(errRet, SelectedProviderUnavailableError) {
```
avitenzer: **A pin to an unusable provider wedges the session for up to two epochs.**

This goes beyond the documented per-request fail-closed cost. `invalidateStickyPin` drops only the local copy, and only on `SelectedProviderUnavailableError`. The fleet claim is never released, and first-writer-wins means no pod can replace it, so a session pinned to a dead upstream fails on every request until the claim TTL (`2*epoch`, 30 min at defaults) expires.

Per request: local miss -> Fetch -> found and epoch-valid -> adopt (`adopted++`) -> hard pin -> `SelectedProviderUnavailableError` (~2001-2018) -> invalidate (`invalidated++`) -> fail. `used_providers.go:326-345` only bumps `batchNumber` when sessions were returned, so `resolvePinDirectives` (`rpcsmartrouter_server.go:3913`) re-honours stickiness on all three `SendRetry` attempts: four identical failures and three or four fleet round trips per client request, then `FirstMessageFailed`.

Pre-PR, the pod-local block at ~2051-2062 deleted the pin and re-picked in the same request. That block is now unreachable because line 1646 blanks `stickiness`.

Also: when the provider instead fails `validateComputeUnits` (refill at ~2474), `!connected` (~1729), or `MaximumNumberOfSessionsExceeded` (~1750), the refill returns `SelectedProviderAlreadyFailedError`, which this check does not match, so every later request is a `local_hit` into the same dead end. The `adopted` metric, documented as the fleet health signal, climbs fastest exactly when sessions are wedged.

Suggest: on any pin failure, release the fleet claim (delete or overwrite) and fall through to the optimizer in the same request, matching the pre-PR pod-local behaviour.

ecosystem/cache/ristretto_store.go
```diff
@@ -194,6 +212,9 @@ func (r ristrettoStore) Purge(ctx context.Context) error {
 		c.Clear()
 		c.Wait()
 	}
+	if r.cs.stickyPins != nil {
+		r.cs.stickyPins.clear()
```
avitenzer, line 216: **Purge / cache-be restart wipes every fleet claim but only the resetting pod drops its local pins, so the fleet splits.**

`FlushCache` (driven by any pod's `/debug/reset-all`) and a cache-be pod restart (`stickyPinStore` is an in-memory map, `cache/sticky_sessions.go:34-45`) clear all claims. `rpcsmartrouter.go:1252-1283` clears the *resetting* pod's `csm.stickySessions` and then flushes the shared cache, but every other pod keeps its `Confirmed` local pins, and the fast path answers those without a round trip (`consumer_session_manager.go:3897` never re-reads the store while `epoch+1 >= current`).

Pods A, B, C; session S claimed to P; A and B hold confirmed local pins. Reset runs on A. Next request for S on C: Fetch miss -> candidate Q -> claims Q. B keeps routing S to P from its local pin. S is served by P on B and by Q on A and C until `DeleteOldSessions` ages B's pin out, up to 30 min at the default epoch. That is the exact split this feature exists to remove.

The tests at `cache/sticky_sessions_test.go:117` and `redisstore/sticky_test.go:101` assert that flush drops claims but don't cover peers' retained copies. Options: have peers re-validate confirmed pins against the store on a cadence, or carry a store generation/epoch marker that a flush bumps so local pins self-invalidate.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -3768,3 +3816,159 @@ func (csm *ConsumerSessionManager) startStateSizesPublisher() {
...
+		csm.recordStickyOutcome(stickyOutcomeClaimed)
+	} else {
+		// A peer claimed this session first. Adopting its winner here — rather than using our
+		// own pick and letting the two pods disagree until the claim expires — is why the write
+		// returns the effective claim instead of a bare success.
+		csm.recordStickyOutcome(stickyOutcomeLostRace)
+		utils.LavaFormatTrace("sticky session: adopting a peer's claim after losing the race",
+			utils.LogAttr("ourPick", candidate), utils.LogAttr("winner", winner), utils.LogAttr("GUID", ctx))
+	}
+	csm.stickySessions.Set(localKey, &StickySession{Provider: winner, Epoch: winnerEpoch, Confirmed: true})
```
avitenzer, line 4144: **A TTL-live but epoch-stale claim causes two store round trips plus an optimizer run on every request, for up to a full epoch.**

A claim written at `E*dur+d` with TTL `2*dur` (`stickyClaimTTL`) is epoch-stale from `(E+2)*dur` (line ~3858: `epoch+1 >= current`) but the store holds it until `(E+2)*dur+d`. In that window each request: local pin stale -> Fetch -> found but stale -> `stickyCandidate` under RLock -> `PublishIfAbsent` -> the Lua `if cur then return cur end` (`redisstore/store.go:591-596`) or `setIfAbsent`'s `now.Before(expiresAt)` (`cache/sticky_sessions.go:52`) returns the stale claim, since neither store looks at `Epoch` -> this line stores `Epoch=winnerEpoch` (still stale) -> routed to it and counted `lost_race`.

Reproduced against the PR's own `fakeSharedSticky`: 5 requests at epoch E+2 for a claim made at epoch E gave 5 fetches, 5 writes, 5 optimizer runs.

`ClampStickyTTL`'s 1-minute floor widens the window for short epochs, and a corrupt `Epoch` of `MaxUint64` (`epoch+1` wraps to 0) makes it permanent for the claim's TTL. The `MaxStickyTTL` warning at `rpcsmartrouter.go:2432` only covers the opposite direction.

Fix: pass the epoch floor into set-if-absent so a stale entry is overwritten, or adopt a found-but-stale claim as-is instead of rejecting it on read.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -2422,8 +2457,21 @@ func (csm *ConsumerSessionManager) getValidConsumerSessionsWithProvider(ctx cont
...
+		// set just above, and a pin-less refill then served the request from a different
+		// upstream with no error and no metric. For a header-pinned provider that is the
+		// opposite of what the header asks for; for a resolved sticky claim it is the exact
+		// split cross-pod stickiness exists to remove, arriving under ordinary load rather than
+		// during an incident.
+		//
+		// Carrying it through routes the refill into the "selected provider already failed in
+		// this request" guard, so the request fails instead — which is what pinning means. The
+		// first fetch above has always passed it; only this one did not.
+		providerAddresses, err = csm.getValidProviderAddresses(ctx, 1, ignoredProviders.providers, cuNeededForSession, requestedBlock, addon, extensions, stateful, stickiness, selectedProvider)
```
avitenzer, line 2660: **Refill loop spins forever under `csm.lock.RLock` when the pod-local sticky pin fails `validateComputeUnits`.**

Pre-existing on `main`, but this PR rewrites the refill and the comment above asserts it's bounded, which is only true for `selectedProvider`.

With `sharedSticky == nil` (the default, or `--shared-state` without a cache): first fetch returns `[P]`; `pairing[P].validateComputeUnits` (`consumer_types.go:937`) fails (`UsedComputeUnits` accumulates all epoch against `MaxComputeUnits=999999999` from `rpcsmartrouter.go:2627`, roughly 1.1k rps of CU-1000 methods at a 15m epoch, or ~12 rps at `--epoch-duration=24h`); P is added to `ignoredProviders`; this refill re-passes `stickiness`; the sticky block at ~2051 checks only `slices.Contains(validAddresses, P)`, never the ignored set, and returns `[P]` again; `validateComputeUnits` fails again. The `for {}` at ~2424 has no ctx check or counter and holds RLock the whole time, so the next `UpdateAllProviders` writer queues behind it and every later `GetSessions` on that chain wedges with one core pegged.

No pre-existing pin is needed: ~2187 sets the pin from the CU-blind optimizer pick on the first request. The `selectedProvider` path has exactly this guard at ~2029. The sticky path needs the same: error out, or `Delete` the pin and fall through to the optimizer.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -1619,6 +1623,37 @@ func (csm *ConsumerSessionManager) GetSessions(ctx context.Context, wantedProvid
...
+	// feature — getValidProviderAddresses handles selectedProvider and returns before it ever
+	// reaches the sticky block — and naming a provider explicitly is the more specific ask of
+	// the two. Resolving stickiness here unconditionally would silently invert it, and only on
+	// deployments running --shared-state.
+	if stickiness != "" && selectedProvider == "" && csm.sharedSticky != nil {
+		resolved, localKey, stickyErr := csm.resolveStickyPin(ctx, stickiness, cuNeededForSession, requestedBlock, addon, common.GetExtensionNames(extensions), stateful)
+		if stickyErr != nil {
+			return nil, stickyErr
+		}
+		selectedProvider = resolved
```
avitenzer: **Every sticky relay now logs at Info as a header selection, and selection stats can name the losing candidate.**

Because the hook always converts the claim into `selectedProvider` here, `getValidProviderAddresses` takes the header branch and emits `LavaFormatInfo("Provider selected via header", ...)` at ~2042 with addon/extensions/validProviders/GUID attrs on every sticky request, warm `local_hit` included. Pre-PR, the sticky hit at ~2056 logged at Trace. It also mislabels a fleet claim as a `lava-select-provider` header.

On a cold id, `stickyCandidate` (~3873) runs the generic selection loop: `setSelectionStats` at ~2128 stores the candidate as `latestSelectionStats` and "Choosing providers" logs at Info (~2158). The pinned path returns at ~2048 without touching stats, so with `enableSelectionStats` on, the `SELECTION_STATS_HEADER_NAME` header (`rpcsmartrouter_server.go:4847-4855`) reports the candidate as `SelectedProvider` even when this pod lost the race and routed to the peer's winner.

Suggest picking the candidate via `ChooseUpstreamWithStats` directly (no stats/log side effects) and demoting or relabelling the ~2042 log when the pin came from a claim.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -1619,6 +1623,37 @@ func (csm *ConsumerSessionManager) GetSessions(ctx context.Context, wantedProvid
...
+	// have identical semantics — route here or fail — so reusing that path gives the guarantee
+	// exactly, and leaves the locked selection code untouched. With no store wired, stickiness
+	// passes through unchanged and keeps its pod-local meaning.
+	//
+	// selectedProvider WINS over a sticky claim when both are present. That order predates this
+	// feature — getValidProviderAddresses handles selectedProvider and returns before it ever
+	// reaches the sticky block — and naming a provider explicitly is the more specific ask of
+	// the two. Resolving stickiness here unconditionally would silently invert it, and only on
+	// deployments running --shared-state.
+	if stickiness != "" && selectedProvider == "" && csm.sharedSticky != nil {
```
avitenzer, line 1649: **The sticky hook runs regardless of `minGroups`, but group-diversity policies discard the result.**

Under a cross-validation policy with `MinGroups > 1`, `rpcsmartrouter_server.go:3913` resolves stickiness independently of the policy and 3929-3936 pass both into `GetSessions`. This hook then does Fetch + `stickyCandidate` + `PublishIfAbsent` (up to 600ms) and stores a local pin, after which ~2388-2408 logs a Warning (mislabelling the claim as a caller `selectedProvider` with `stickiness=""`) and calls `getValidProviderAddresses` with both blanked.

So two fleet round trips and an optimizer run are wasted, and a fleet claim is written and honoured for two epochs, adopted by non-CV requests on every pod, for a decision this pod never routed on.

`minGroups` is parsed at ~1619-1626, before the hook. Adding `&& minGroups <= 1` to the guard fixes it (or pass `stickiness` through untouched and let the existing warning fire).

docs/METRICS.md
```diff
@@ -402,7 +402,8 @@ went out, and which" — so read the alerting note under the table before buildi
...
 | `smartrouter_csm_blocked_backup_providers` | Gauge | `spec`, `apiInterface` | Size of the blocked-backup-providers store. Backups are tracked here only — they never appear in `smartrouter_csm_provider_blocked`. |
-| `smartrouter_csm_sticky_sessions` | Gauge | `spec`, `apiInterface` | Live sticky-session affinities. |
+| `smartrouter_csm_sticky_sessions` | Gauge | `spec`, `apiInterface` | Live sticky-session affinities held by THIS pod. With `--shared-state` this is a read-through cache of the fleet's claims, not the fleet total. |
+| `smartrouter_csm_sticky_claims_total` | Counter | `spec`, `apiInterface`, `outcome` | Cross-pod sticky-session claim resolutions. `outcome`: `local_hit` (answered from this pod, no round trip), `adopted` (took a claim a peer had already made), `claimed` (this pod made the claim), `lost_race` (claimed simultaneously with a peer and adopted its winner), `error` (claim could not be established, request failed rather than served off an unverified pin). **`adopted` staying at zero on a multi-replica fleet means the feature is wired but never firing.** |
```
avitenzer: **This row documents five outcomes but the code emits seven.**

`smartrouter_csm_sticky_claims_total` also emits `no_candidate` (`consumer_session_manager.go:3927`) and `invalidated` (`:3967`), and the Prometheus Help string at `smartrouter_metrics_manager.go:829-840` already describes both. Those two are exactly the series that fire in the failure modes: an all-primaries-blocked incident increments only `no_candidate`, and a claim pinned to a blocked provider increments only `invalidated`.

An alert built from this "closed set" that pages on `error > 0` would show sticky traffic healthy while every sticky request for those sessions fails. Please add both outcomes to the row.

protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -2416,6 +2417,28 @@ func (rpsr *RPCSmartRouter) CreateSmartRouterEndpoint(
 	// Create active subscription provider storage for each unique chain
 	activeSubscriptionProvidersStorage := lavasession.NewActiveSubscriptionProvidersStorage()
 	sessionManager := lavasession.NewConsumerSessionManager(rpcEndpoint, optimizer, smartRouterMetricsManager, smartRouterIdentifier, activeSubscriptionProvidersStorage)
+	// Cross-pod sticky sessions. Gated on --shared-state, the same operator signal the chain-tip
+	// and tracker-gate sharing use: "this chain runs on several replicas behind one cache". A
+	// single-replica router needs no registry, and its absence leaves stickiness pod-local.
+	// Claim lifetime is sized in epochs, which every pod derives from wall clock identically, so
+	// no clock agreement is needed between them.
+	if options.stateShare {
+		epochDuration := rpsr.epochTimer.GetEpochDuration()
```
avitenzer, line 2587: **Fleet-claim epochs are only comparable if every pod runs the same `--epoch-duration`, and nothing checks or warns about a mismatch.**

`protocol/common/epoch_timer.go:42-52` derives the epoch number as `floor((now - 2024-01-01) / epochDuration)`. The wiring here only warns when `epochDuration*2 > MaxStickyTTL`, and no epoch-duration value travels with a claim.

During a 15m -> 1h rollout (the value the flag's help text at ~3513 suggests), the 15m pods compute ~94k and the 1h pods ~23.5k. The 15m pods treat every 1h-pod claim as stale (Fetch -> candidate -> Publish returns existing winner -> Set with the rejected epoch, two round trips per request, `lost_race` inflated), while the 1h pods accept 15m-pod claims as future epochs (`stickyEpochValid` has no upper bound) and honour them until TTL. The design comment at `consumer_session_manager.go:3854-3857` ("every pod reaches the same verdict without comparing clocks") silently assumes uniform config.

At minimum: document the flag as fleet-uniform and log the epoch duration alongside the claim so a mismatch is diagnosable. Rejecting epochs `> current+1` closes the accept-anything half.

avitenzer (commented):
Review of the cross-pod sticky-session change. 9 inline comments below; one more finding has no nearby diff hunk so it lives here.

**`consumer_session_manager.go:643` — `DeleteOldSessions(previousEpoch)` on a same-epoch push deletes pins the reader rule still honours.**

`UpdateAllProviders` runs on every provider-set push, not only on epoch change: `rpcsmartrouter.go:4052` and `:4210` (readmit/merge) pass `currentEpoch`, and the comment at ~495-499 says same-epoch updates proceed. At 483, `previousEpoch` is the current epoch E; line 643 then deletes every pin with `Epoch < E`, i.e. claims made in E-1 that `stickyEpochValid` (~3858, `epoch+1 >= current`) still accepts through E. The next request for each such session misses the fast path, pays a Fetch (300ms budget, fail-closed on error), finds its own claim, and ~3915 records `adopted`. A single-replica router with periodic readmits shows `adopted > 0` (METRICS.md says adopted at zero means the feature isn't firing), and a fleet's `local_hit` ratio drops after every readmit. Pass `previousEpoch-1`, or make `DeleteOldSessions` use the same predicate as `stickyEpochValid`.

**Summary.** The three I'd block on are the backup-tier bypass (1643), the dead-provider pin that wedges a session for two epochs (1651), and the fleet split after a cache purge or cache-be restart (ristretto_store.go:216). Each defeats the feature's stated purpose during an ordinary incident and none has test coverage in the PR. The rest are efficiency, observability, and ops-hardening.

Not posted inline (confirmed cleanups, lower priority): `stickyPinStore` / `endpointObservationStore` duplication, a redundant Fetch before `PublishIfAbsent`, the `Confirmed` field is derivable, an unreachable `ttl < 0` guard, and the `Purge` doc comment is on the wrong function.

https://claude.ai/code/session_01A6LFGBAyHkeZKbBoUUXWwA

Tomelia1999:
Thanks — this was a genuinely useful review, and two of these were things I'd already reviewed twice without catching. Four fixed, three declined for now, the rest deferred. CI green at `ed82e0e`.

## Fixed

**Backup-tier bypass (`:1643`)** — `ed82e0e`. A sticky request that finds no primary to claim now proceeds **unclaimed** and reaches the backup cascade instead of failing. Fail-closed exists to protect a claim that *exists*; where no claim could be made in the first place there is nothing to be inconsistent with, so failing bought the caller nothing and cost it a tier that was still healthy. `no_candidate` still records the condition so it stays visible. Making backup providers claimable is separate work with its own ticket.

**Refill spin (`:2474`)** — `7d9b3ac`. You filed this as non-blocking; I'd rank it highest of the set. It fires on the default path — no `--shared-state` needed — and freezes a chain rather than degrading it. The sticky lookup now honours `ignoredProvidersList`, mirroring the guard the `selectedProvider` branch already carries for exactly this reason. The regression test carries its own deadline, because the failure mode is a hang and would otherwise surface as a ten-minute package timeout with no attribution. Verified in both directions: without the guard it trips the deadline, with it the call returns in about two seconds.

**METRICS.md (`:406`)** — `7d9b3ac`. All seven outcomes documented, with an explicit note that `no_candidate` and `invalidated` are the two that climb during an incident — so an alert built on `error` alone reads healthy while every sticky request for those sessions fails. Straightforwardly my error.

**Related, fixed just before this review** — `a7cd716`. The refill blanked `selectedProvider`, so a claimed upstream that was valid but merely busy was silently replaced by another with no error and no metric. Same root shape as `:2474`, found by an adversarial pass.

## Declined for now

**Releasing the fleet claim on pin failure (`:1651`)** — the wedge is real and understood, but releasing was considered and deliberately rejected: selectability is judged per pod, so one pod's local health view would evict a claim its peers are serving correctly. Keeping "fail, and invalidate the local copy only". The `AlreadyFailed` half you identified is a fair criticism of that choice and is on the follow-up list.

**Purge / cache-be restart split (`ristretto_store.go:216`)** — `/debug/reset-all` is a QA tool rather than a production path. The cache-be restart variant is real and is not addressed in this PR.

**Cross-validation (`:1640`)** — not changing now. Noted that the cost is not only wasted work: the claim that gets written is then adopted by other pods for ordinary requests.

## Deferred

Epoch-stale thrash (`:3948`), the same-epoch `DeleteOldSessions` purge, Info-level log volume (`:1645`), epoch-duration uniformity (`rpcsmartrouter.go:2426`), single-flight on cold ids, and the confirmed cleanups listed in your summary.

Two gaps worth stating plainly: the container harness is not wired into CI, so the headline evidence only re-runs by hand; and nothing has independently checked MAG-3136's obligations against the diff.

https://claude.ai/code/session_01XSnTE23nFbN2dT1YuBSF3q

protocol/lavasession/consumer_session_manager.go
avitenzer: Fixed in `09fe78b`, taking the option suggested here — a `PairingListEmptyError` out of the hook now blanks `stickiness` and falls through to normal selection instead of returning:

`consumer_session_manager.go:1652` — `case errors.Is(stickyErr, PairingListEmptyError):` sets `stickiness = ""`, so the request proceeds unclaimed and reaches the backup cascade, `releaseBlockedProvidersIfPoolEmpty` and the blocked-provider walk exactly as non-sticky traffic does. The session simply is not pinned until primaries return.

The metric split is in place too: this path records `no_candidate` rather than `error`, so an alert on `error` no longer pages for what is a pairing problem.

The coverage gap noted at the end of this thread is closed by `TestSharedSticky_NoPrimaryToClaimServesFromBackupUnpinned`, which populates `backupProviders` with an empty primary pool and asserts the request is served by `backupA`.

Re-verified at `09fe78b`: `go build ./...` and `go vet ./...` clean, `protocol/lavasession` green.

protocol/lavasession/consumer_session_manager.go
avitenzer: Fixed in `6e65834`, taking the "`Delete` the pin and fall through to the optimizer" option.

`consumer_session_manager.go:2228` now consults the ignored set before honouring the pin:

```go
_, alreadyTried := ignoredProvidersList[stickysession.Provider]
if !alreadyTried && slices.Contains(validAddresses, stickysession.Provider) { ... }
```

On a miss it deletes the pin and falls through to ordinary selection, so the refill can no longer be handed the same provider it just set aside. This mirrors the guard the `lava-select-provider` branch already carried — which, as noted here, was never mirrored into the sticky branch.

Covered by `TestStickyPodLocal_ExhaustedProviderDoesNotSpinTheRefillLoop`: it spends the pinned provider's `MaxComputeUnits` for the epoch (so the provider stays healthy and stays a valid address, matching the shape described above) and asserts `GetSessions` returns within an explicit 15s deadline. The deadline is deliberate — the failure mode is a hang, which would otherwise surface as a package timeout with no attribution.

Re-ran `protocol/lavasession` at `09fe78b`: green (64s).

docs/METRICS.md
avitenzer: Fixed — the `docs/METRICS.md` row now documents all seven outcomes: `local_hit`, `adopted`, `claimed`, `lost_race`, `error`, `no_candidate`, `invalidated`.

I checked each of the seven strings emitted by `recordStickyOutcome` against the row at `09fe78b`; all present, none extra.

The row also carries the alerting point this thread was really about, stated inline rather than left to the reader:

> note that `no_candidate` and `invalidated` are the two series that climb during an incident, so an alert built only on `error` will read healthy while every sticky request for those sessions is failing.

avitenzer:
Follow-up review of `09fe78bf87bc15fe72a6fca4f12a460da50f9075`: I would hold approval until the cache-backend restart split is addressed.

**Verified fixes:** the backup-tier fallback, pod-local sticky refill termination, preservation of the resolved pin during refill, and documentation of all seven metric outcomes are present. All 16 focused checked-in sticky-session tests passed.

**Remaining P1: cache-be restart can silently split the same session across upstreams.**

A cache-be restart initializes an empty claim registry ([server.go:126](https://github.com/Magma-Devs/smart-router/blob/09fe78bf87bc15fe72a6fca4f12a460da50f9075/ecosystem/cache/server.go#L126)), but router replicas retain their confirmed local pins. The [fast path at consumer_session_manager.go:4093](https://github.com/Magma-Devs/smart-router/blob/09fe78bf87bc15fe72a6fca4f12a460da50f9075/protocol/lavasession/consumer_session_manager.go#L4093) returns those pins without checking the registry. A replica that has not seen the session can then observe a miss and claim a different upstream.

I reproduced this on the current head in an isolated session-manager test using three replicas with the same healthy provider pool and different latency preferences. This simulates the registry state loss; it is not a Docker restart test.

| Scenario | Warm replica A | Warm replica B | Cold replica C |
| --- | --- | --- | --- |
| Registry intact | provider1 | provider1 | provider1 |
| Registry loses claims | provider1 | provider1 | provider2 |

All session acquisitions succeeded. The intact-registry control passed; the agreement assertion failed after registry loss. This recreates the failure MAG-3136 is meant to prevent: the upstream for one session depends on which router replica receives the request.

The QA-only argument applies to the reset endpoint, but it does not cover ordinary cache-be restarts. Please preserve fleet agreement through registry loss, or fail safely while agreement cannot be established, and add recovery coverage exercising both warm and cold router replicas.

On retaining a claim after a provider becomes unusable on one pod: the rationale is sound. One pod's local health judgment should not evict a claim its peers can still serve. I would not replace that with unconditional deletion. The resulting prolonged failures remain an availability tradeoff that needs an explicit product decision.

The branch is now up to date with main, conflict-free, and reported checks are green. The approval blocker above is a reproduced correctness issue.

Tomelia1999:
## Production-readiness testing

Ran a scenario suite against three router replicas behind a round-robin ingress, five upstreams deliberately out of lockstep, and one claim registry. Added as `docker/sticky-test/run-qa.sh` (`a2b480b`) so it is repeatable rather than a one-off.

Not only the happy path — the suite disturbs the infrastructure: it restarts a replica mid-run, and stops and restarts the registry.

| Scenario | Result |
|---|---|
| The reported sequence (head, then that block) across 3 replicas | **0 false gaps** in 12 rounds |
| Session id sent on both calls vs only the fetch | both = 0 gaps · fetch-only = 3 gaps |
| 24 concurrent requests on one session id | all 24 to a single upstream |
| `lava-select-provider` sent together with `lava-stickiness` | the explicitly named provider won |
| A replica restarts (rolling deploy, crash, eviction) | the restarted pod adopted the existing claim — no split |
| Registry unreachable | sticky requests failed closed (HTTP 500); traffic without the header unaffected |
| Registry restored | sticky sessions worked again |

**Negative control:** with `STICKY_SHARED_STATE=false`, **5 of the 7 fail**. That is what makes the green run mean something. The other two are recorded as non-discriminating rather than left looking stronger than they are — `lava-select-provider` does not depend on shared state at all, and the concurrency check passed once with the feature disabled.

Also re-ran the earlier harnesses on the branch rebased onto current main: `run-falsegap.sh` reproduces the reported bug without affinity (5 of 20 rounds) and shows 0 of 20 with it, across all three replicas.

## One gap found — not a blocker for this PR

Writing the suite surfaced something the earlier harnesses could not see, because they always warmed the system up first.

**A sticky request can be answered from the response cache before any upstream is chosen, so the pin plays no part.**

`eth_blockNumber` is the same request every time and therefore highly cacheable. If the head comes from the cache, it may be a value originally supplied by a node that is ahead, while the pinned fetch goes to the node the session is tied to, which may be behind — the original empty-result symptom, with the header correctly in use.

Measured on a freshly started stack:

- 24 sticky requests produced only **19** claim resolutions — five never reached provider selection
- 3 cold starts: **2 showed false gaps** (3 of 12 rounds, and 2 of 12)
- 4 runs once warm: **0 gaps in all four**

So it is a window, mainly the first traffic after a pod starts, not a constant failure. It is visible from our side rather than silent: the claim metric shows fewer resolutions than sticky requests, which is how it was found.

**Deliberately not addressed here.** How stickiness and caching should interact is a product decision with real trade-offs — bypass the cache for requests carrying the header, put the session id in the cache key, have callers send `lava-force-cache-refresh` on the head request (works today, no release needed), or accept the window. A follow-up ticket is going to product with the options and the evidence.

This does not change what this PR delivers: cross-pod stickiness works, and the cache interaction is pre-existing behaviour this PR neither introduces nor worsens.

## Also worth knowing

A QA guide for whoever picks this up is written and lives outside the repo — what the feature is, how to run all three harnesses **including the negative controls**, which scenarios actually discriminate, the known limits, and how to read `smartrouter_csm_sticky_claims_total`.

Two prerequisites are easy to miss and make the feature silently do nothing: a cache backend must be configured, and the router must run with `--shared-state`. With either missing, each pod falls back to its own private table and everything looks healthy.

https://claude.ai/code/session_01XSnTE23nFbN2dT1YuBSF3q

Tomelia1999:
Thanks — the finding is valid and I'm not disputing the reproduction. Separating the cache-be restart from the reset endpoint is the right distinction; that half was genuinely not covered by the QA-tool argument.

We're taking it as a follow-up ticket rather than holding this PR. The reasoning, so the decision is reviewable rather than just asserted:

**It is not a regression.** Before this change pods shared nothing, so a session split across replicas on essentially every request. After it, a session splits only in the window following a registry restart. The floor moves up either way, and nothing this PR adds makes the old behaviour worse.

**It is bounded and self-healing.** The divergence lasts until the warm local pins age out — at most two epochs, about 30 minutes at defaults — and the fleet then re-converges with no operator action.

**The fix is known rather than open-ended.** A generation marker on the registry, compared against the one a pin carries, so a restart invalidates stale local pins on the next request and everyone re-claims within seconds. That is the shape you suggested and we agree with it; it is scope, not uncertainty.

Two tickets are going to product together, both framed as decisions rather than bugs:

1. **Agreement through registry loss** — this one.
2. **Sticky requests answered from the response cache before any upstream is chosen**, so the pin plays no part. Found during production-readiness testing on this branch: 24 sticky requests produced only 19 claim resolutions, and two of three cold starts showed false gaps while four warm runs were clean. Same class of issue — a window rather than a constant failure.

Both go into the customer handover note as known windows, so the behaviour is disclosed rather than discovered in production.

On the retained claim after a provider becomes unusable on one pod: agreed, and noted that you would keep it too. The prolonged-failure tradeoff travels to product with the same two tickets.

https://claude.ai/code/session_01XSnTE23nFbN2dT1YuBSF3q

avitenzer (approved):

PR #374, opened 2026-09-06
fix(relay): make the per-attempt timeout trigger the next attempt, not kill this one (MAG-3362)

Tomelia1999:
**MAG-3362**

Jira ticket: MAG-3362

## The bug

The router had two time limits that should never have been one value:

- **the window** — how long a provider is expected to take on a method;
- **the request budget** — how long the whole request may take, across every attempt.

The window was used both to decide *when to try the next provider* and *when to give up on the
current attempt*. Because both jobs read the same number, the moment the next provider was tried was
the moment the current attempt was killed.

The consequence is not a slow request but an impossible one. **A method genuinely slower than the
window could never succeed on any provider**, however healthy, and however large the budget: three
attempts killed at the window, the retry limit exhausted by timeouts the router manufactured itself,
and the request abandoned with most of its budget unspent.

Two GK8 incidents were this shape — a Polygon write dying at 11s and a Stellar write at 7.0s, each
with ~170 of its 180 seconds unused. A write is the extreme case because it is never retried, so one
killed attempt was the entire request.

Evidence it was unfinished rather than intended:

- `GetRelayTimeout` ended with `// Set relay timout, increase it every time we fail a relay on
  timeout`. **No such escalation exists** — the value is computed once and never changed. The debug
  field carrying it was even named `newRelayTimeout`.
- The provider-side chain proxies never had this bug: `CapTimeoutForSend` has always handed an
  attempt the whole processing budget. Only the newer direct-RPC path capped it.

## The fix

The window now only **triggers** the next attempt. An attempt lives until the request ends — a
success, another provider winning, the policy stopping, or the budget expiring. **Dispatch cadence is
unchanged, so failover speed does not move.**

Writes need no special case: once nothing kills the attempt, a write runs to its full budget on its
own.

### Blame for a hang had to be rebuilt

This fix would otherwise have *removed* the blame that used to happen naturally. A hang previously
surfaced as `DeadlineExceeded`, which is not a cancellation, so it landed on the failure path by
itself. Now an attempt outlives its window and is cancelled when the request ends — which without
care routes every hang into the MAG-2648 no-penalty carve-out, recording nothing at all and letting a
permanently hung provider keep its score forever.

The verdict asks whether the **request ran out of budget**: an endpoint still inside its budget was
cut short and has proved nothing, while one still silent when the budget expired has. No threshold.

Two earlier attempts at this are worth knowing about, because both were wrong and both were caught by
review rather than by tests:

- **"Was it given its full window"** — wrong, and measurably so. A hedge is dispatched AT the window,
  so a race loser has *always* been silent longer than its window by the time it is cancelled. A
  healthy endpoint answering in 55s, cancelled at 35s when a faster one won, on a request that
  **succeeded**, was blamed. That is MAG-2648's structural penalty — every node but the fastest —
  and the pre-change code never did it, confirmed by running the same scenario on a merge-base build.
- **Reading the stop reason** — the first claimant keeps that field, so the timeout's note was
  discarded and a hang looked like a race loser. Measured on a read where every endpoint hangs: one
  blamed, one escaped, decided by which clock fired first.

The second is fixed at the claim rather than at the timeout. The circuit breaker was labelling the
pool `AllProvidersExhausted` while both providers were healthy and merely **busy with attempts from
this same request** — its own warning says `stopping new attempts — relays already in flight may
still answer`. That label only ever meant *no new attempts*, and filing it as the request's stop
reason was harmless while an attempt was killed at its window, because "cannot start more" and "the
request is over" were then the same instant. This PR decoupled them, so the claim became premature.
It now only names the stop reason when nothing is still in flight.

**The trade this accepts:** an endpoint that hangs while a different one answers is not blamed for
that request. That is the honest reading — we cancelled it early, so we do not know it hung — and a
genuinely hung endpoint is still caught on every request that has no other answer.

`OnSessionUnresponsive` rather than `OnSessionFailure`: same accounting today, but a hang and a
failure are different events and must be free to be punished differently later — the same reasoning
that keeps `OnSessionDiscarded` and `OnSessionCancelled` apart.

### A write with an unknown outcome no longer says "failed"

Both incidents ended with the transaction **broadcast** and the customer told it had failed. That is
the worst of the three available answers, because it invites a resubmit of a transaction that may
already be on chain. GK8's words: *"in some cases the tx succeeded even though they got an error and
now don't know it passed."*

A write that ends with no node having answered now returns, still HTTP 500 and in the existing REST
envelope:

```
transaction status unclear: timeout reached before any node responded; the transaction may
already have been submitted — verify on-chain before resubmitting
```

The rule is about **silence**, and about what "came back" actually means: it is a write, nothing
succeeded, and either some endpoint we asked did not answer at all, or one of them came back without
answering — a connection that died after the request was already on the wire settles nothing about
whether the node received it. That second clause uses the `MayHaveReachedNode` classification, which
also covers an error nobody could match, since an unmatched failure proves nothing either. Reading only the recorded errors is what got this
wrong first — a hung endpoint contributes none, so a sibling's fast HTTP 500 or connection refusal
decided the answer while the silent endpoint may have been holding the transaction.

Definite application rejections, such as REST HTTP 400, retain their response when every attempt
has a definite outcome. **Gateway HTTP 502/504 can hide a forwarded transaction.** The
REST dispatcher converts these responses to protocol errors before the results manager sees them,
just as the JSON-RPC path does. A write with no successful reply therefore returns HTTP 500 with
the unclear-write warning for either gateway status.

One consequence worth stating: a **dial** timeout against a firewalled host that drops packets is
indistinguishable from a response timeout at this layer, so it also reports "unclear". For a write
that is the safe direction — a false "failed" is the unrecoverable mistake — but it means not every
connect-phase failure reports a clean failure.

It is decided in `SendParsedRelay` rather than `buildFailureResult`, because the first change moved
the path: a hung write now ends with its goroutine still in flight and nothing recorded, so
`HasResults()` is false and the request returns through the no-results branch that
`buildFailureResult` never sees. Which of the two branches you land on was also a **race**, so judging
once above both is what stops the same incident producing two different customer-facing errors.

## Also in here

- Removed the comment promising escalation that was never built, and the `newRelayTimeout` name.
- The cross-validation straggler bound was `2 * relayTimeout`, which assumed an attempt could not
  outlive the window. Re-derived from the attempt budget; the old form would have reported a
  straggler still legitimately in flight as never received.
- A hang now reports `RelayOutcomeError` rather than `RelayOutcomeCancelled`, so the metrics agree
  with the availability sample.
- The per-node `timeout:` field stays **additive, deliberately**. It is set on nearly every endpoint
  in the shipped example configs, so turning it into a hard cap would kill attempts at 10s and make
  this whole change a no-op where it is used. **Not currently covered by a test** — worth adding.

## Verification

Ten scenarios replayed against the branch binary in one pass at `34b35a3`, real
`specs/ethereum.json`, two endpoints, `--min-relay-timeout 2s --default-processing-timeout 10s`.
Windows and budgets are taken from the router's own debug log, not from arithmetic. The `main`
column is a build from the merge base.

**The four that prove the fix**

| Scenario | Method | window / budget | `main` | this branch |
|---|---|---|---|---|
| Hedging: slow provider allowed to finish | `eth_gasPrice` | 1s / 10s | ❌ 500 at 2.0s | ✅ 200 at 4.0s |
| Heavy read slower than its window | `eth_estimateGas` (CU 100) | 10s / 60s | ❌ 500 at 20.0s | ✅ 200 at 14.0s |
| **Write slower than its window** | `eth_sendRawTransaction` | 28s / 60s | ❌ 500 at 28.0s | ✅ **200 at 40.0s** |
| Write that never gets an answer | `eth_sendRawTransaction` | 28s / 60s | 500 at 28.0s, *"failed … insufficient results"* | 500 at 60.0s, ***"transaction status unclear …"*** |

Row 3 is the headline: a write that was previously **impossible** now succeeds.

**Blame lands on the right endpoints, and only them**

| Scenario | Setup | Result |
|---|---|---|
| write, both endpoints hang | — | ✅ **both blamed** |
| read, both endpoints hang | — | ✅ **both blamed** — the circuit breaker logs `stillInFlight=2` and correctly stays quiet |
| write, one errors + one hangs | `up-a` HTTP 500 in <1ms, `up-b` hangs 60s | ✅ **both blamed**, client gets ***"status unclear"*** rather than the sibling's 500 |
| **write, healthy endpoint loses the race** | `up-a` 35s, `up-b` 55s | ✅ **`up-b` NOT blamed** — still inside its budget, merely slower. MAG-2648 preserved |
| **write, fast winner** | `up-b` 3s, `up-a` 30s | ✅ **`up-a` NOT blamed**, composite score byte-identical either side of the event |
| read, one errors + one hangs | — | ✅ both blamed, and the write message correctly does **not** appear on a read |
| write, no reachable provider | every endpoint on a closed port | ✅ `No pairings available` — same as a read. Nothing was dispatched, so nothing can be on chain |

**Failover cadence did not move** — the thing this change could most expensively have broken, now
covered by `TestFailoverCadence` as well as measured live. That test is mutation-checked: pointing
the ticker at the budget makes it fail, and its collection loop is bounded so the regression fails
rather than hanging CI. From the hedging run's upstream timeline:

```
port=9102  dispatched 4.30s   answered 9.30s
port=9101  dispatched 5.30s   answered 8.30s
dispatch gap = 1.00s   (window = 1s)
```

The gap is exactly the window, and the two attempts overlapped for three seconds with the later,
faster one winning — hedging genuinely hedging, which could not happen before this change.

**Tests.** The attempt-budget regression is at the real dispatch call site: it fails
if the caller passes the window instead of the budget, verified in both directions. An earlier round
of tests did **not** have that property — reverting the fix left them all green — which is why it
exists. All five touched packages pass at this commit: `rpcsmartrouter`, `relaycore`, `lavasession`,
`common`, `chainlib`.

`TestRESTListener_WriteOutcome` sends real HTTP requests through the REST listener, parser,
session selection, dispatcher, results manager, and client error envelope. It covers gateway 502/504
writes and a write that hangs until the request budget expires (all return HTTP 500 with the
unclear-write warning), a definite HTTP 400 rejection, a successful write, and gateway errors on reads.
The hung case holds the return from a real cancelled HTTP attempt until the client receives its
response, deterministically exercising the no-results branch without fabricating a result or stop
reason. Its attempt deadline is later than the request budget, so it also verifies that budget-driven
cancellation disables the URL at the refusal threshold, records `DisabledAt`, and sets endpoint health
to zero.

The session tests keep dispatched providers reserved until the production release method runs. They
check returned CU, released provider/session reservations, preserved dispatch counts, one unanswered
availability sample, and the optimizer's failure notification. Failure and unresponsive accounting are
each compared against an independently specified outcome.

**Mutation checks:** the following production regressions compile and fail the named tests. Each was
also checked across `rpcsmartrouter`, `lavasession`, and `relaycore` with `-count=1`.

| Production regression | Enforcing failure |
|---|---|
| Pass `isClientCancel` directly to `classifyEndpointHealth`, omitting the budget-expired exception | Hung REST write leaves the URL enabled |
| Remove both unclear-write returns from `SendParsedRelay` | Hung and gateway-write clients receive generic errors instead of the warning |
| Remove `AppendRelayFailure` from the availability-failure release helper | `TestOnSessionUnresponsiveRecordsBlame` receives no optimizer notification |

Removing either unclear-write return independently also fails its corresponding client test: the
hung case covers no recorded results, while gateway writes cover recorded protocol errors. Gateway
write cases also fail if the dispatcher stops converting gateway responses to protocol errors or if
`SendParsedRelay` stops using the write verdict.

All five packages above pass unmodified with `-count=1`. The REST listener and the session cancellation,
unresponsive, failure, and discard regression tests pass with the race detector.

**Reading a blame count.** A hung endpoint is blamed through either of two paths depending on which
clock fires first: its own attempt deadline (`DeadlineExceeded`, the pre-existing failure arm) or the
request-level cancel (`Canceled`, the new `OnSessionUnresponsive`). Counting only the new log line
under-reports. Check which endpoints reached *a* blame path.

## Behaviour changes reviewers should weigh

- **A request where *every* provider hangs now runs to the full budget** instead of giving up early.
  On a hanging or CU≥100 method that is 6× the processing timeout. This is what the budget is for, but
  it is a real change in worst-case latency and should be watched in percentiles after rollout.
- **More attempts are alive at once** — bounded by the hedge ceiling per request and the per-provider
  session cap, but sessions and connections are held longer.
- **Hedging now genuinely duplicates upstream work.** Note that cancelling on our side never stopped
  the node working — it had already received the request — so for most RPC nodes this was already
  being paid for and discarded. Gateways that *do* abort on client disconnect will now do more work,
  and only for requests slower than the window.

### Six defects found in review and fixed here (third and fourth commits)

- **A healthy endpoint was blamed for losing a race.** The verdict asked "was it given its full
  window", but a hedge is dispatched AT the window, so a race loser has *always* been silent longer
  than its window by the time it is cancelled — the test was true for every loser, always. Measured:
  a healthy endpoint answering in 55s, cancelled at 35s when a faster one won, on a request that
  **succeeded**, blamed at `waited=35s window=28s`. That is MAG-2648's structural penalty — every node
  but the fastest, on a broadcast — and the pre-change code never did it, confirmed by running the same
  scenario on a merge-base build. The verdict now turns on whether the **request ran out of budget**:
  still inside its budget means cut short and nothing proved. No threshold needed.
- **A write that never reached a provider was told it might have been submitted.** "No evidence
  recorded" also covers *nothing was ever dispatched*. Verified live: with no reachable upstream a read
  returned `No pairings available` while the write claimed the transaction may already be on chain —
  the same lie as the bug this fixes, pointing the other way. Now requires that the request actually
  used its budget.
- **The `2*relayTimeout` straggler bound existed in two places**; only one had been re-derived.

**Second review round — three more, all introduced by the two commits before them:**

- **The tests did not cover the fix.** Reverting the single behavioural line left all nine tests
  green: they called `SendDirectRelay` directly with a duration the test picked, proving that
  function honours its argument but never that the caller hands it the right one. Now covered at the
  real call site.
- **A write said "failed" while an endpoint was still silent.** The rule read only the responses that
  came back, and a hung endpoint contributes none — so a sibling's fast HTTP 500 decided the answer
  while the silent endpoint may have been holding the transaction. The rule now asks about
  **silence**: nothing succeeded and someone we asked did not answer ⇒ unclear.
- **A hung endpoint escaped blame on reads.** The verdict reads the stop reason and the first
  claimant keeps it. Fixed at the claim rather than at the timeout: the circuit breaker was labelling
  the pool `AllProvidersExhausted` while both providers were healthy and merely **busy with attempts
  from this same request** — its own warning says `stopping new attempts — relays already in flight
  may still answer`. That label was only ever about new attempts, and filing it as the request's stop
  reason was harmless while an attempt died at its window. This PR decoupled those, so the claim
  became premature. It now only names the stop reason when nothing is still in flight.

## Not in scope

- **Stellar testnet still has a single upstream.** None of this triggers when more than one provider
  is configured, so adding a second qualified primary remains the strongest fix for that chain.
- The REST error envelope shape is unchanged; reshaping it is a separate, breaking change.
- Writes are still never retried.
- Confirm `ReturnMaskedErrors` is off on the GK8 routers, or the new write message reaches nobody.

### DIFF
--- a/protocol/chainlib/common.go
+++ b/protocol/chainlib/common.go
@@ -572,16 +572,11 @@ func CompareRequestedBlockInBatch(currentLatestRequestedBlock, currentEarliestRe
 	return latestCallback(currentLatestRequestedBlock, parsedBlock), earliestCallback(currentEarliestRequestedBlock, parsedBlock)
 }
 
-// GetRelayTimeout returns the per-attempt WINDOW: how long an endpoint is expected to take on this
-// method. It is the interval after which the state machine dispatches another endpoint, and the
-// span an endpoint must be given before a non-answer counts against its availability.
+// GetRelayTimeout returns the per-attempt WINDOW: how long an endpoint is expected to take, and so
+// the interval after which the state machine dispatches another one.
 //
-// It is NOT how long an attempt may live — that is the processing budget, see GetTimeoutForProcessing
-// and sendRelayToDirectEndpoints. The two were one value until the window stopped killing the
-// attempt; a previous comment here promised the value would "increase every time we fail a relay on
-// timeout", which was never implemented and is deliberately not being implemented: growing the
-// window would slow failover, and the reason a slow method used to fail was the killing, not the
-// size of the window.
+// It is NOT how long an attempt may live — that is the processing budget, see
+// GetTimeoutForProcessing and sendRelayToDirectEndpoints.
 func GetRelayTimeout(chainMessage ChainMessageForSend, averageBlockTime time.Duration) time.Duration {
 	if chainMessage.TimeoutOverride() != 0 {
 		return chainMessage.TimeoutOverride()
--- a/protocol/chainlib/common.go
+++ b/protocol/chainlib/common.go
@@ -572,6 +572,16 @@ func CompareRequestedBlockInBatch(currentLatestRequestedBlock, currentEarliestRe
 	return latestCallback(currentLatestRequestedBlock, parsedBlock), earliestCallback(currentEarliestRequestedBlock, parsedBlock)
 }
 
+// GetRelayTimeout returns the per-attempt WINDOW: how long an endpoint is expected to take on this
+// method. It is the interval after which the state machine dispatches another endpoint, and the
+// span an endpoint must be given before a non-answer counts against its availability.
+//
+// It is NOT how long an attempt may live — that is the processing budget, see GetTimeoutForProcessing
+// and sendRelayToDirectEndpoints. The two were one value until the window stopped killing the
+// attempt; a previous comment here promised the value would "increase every time we fail a relay on
+// timeout", which was never implemented and is deliberately not being implemented: growing the
+// window would slow failover, and the reason a slow method used to fail was the killing, not the
+// size of the window.
 func GetRelayTimeout(chainMessage ChainMessageForSend, averageBlockTime time.Duration) time.Duration {
 	if chainMessage.TimeoutOverride() != 0 {
 		return chainMessage.TimeoutOverride()
@@ -585,7 +595,6 @@ func GetRelayTimeout(chainMessage ChainMessageForSend, averageBlockTime time.Dur
 	if chainMessage.GetApi().TimeoutMs > 0 {
 		relayTimeAddition = time.Millisecond * time.Duration(chainMessage.GetApi().TimeoutMs)
 	}
-	// Set relay timout, increase it every time we fail a relay on timeout
 	return extraRelayTimeout + relayTimeAddition
 }
 
--- a/protocol/common/error_codes.go
+++ b/protocol/common/error_codes.go
@@ -7,9 +7,8 @@ package common
 
 var (
 	// Connection errors (1001-1009)
-	// MayHaveReachedNode: assigned from net.Error.Timeout(), which covers a dial timeout AND a
-	// timeout waiting for the response. The two are indistinguishable here, so take the safe
-	// direction for a write — "unclear" is recoverable, a false "definitely failed" is not.
+	// MayHaveReachedNode: net.Error.Timeout() covers both a dial and a response timeout, and the
+	// two are indistinguishable here. For a write, "unclear" is the recoverable direction.
 	LavaErrorConnectionTimeout = registerError(&LavaError{
 		Code: 1001, Name: "PROTOCOL_CONNECTION_TIMEOUT", Category: CategoryInternal,
 		Description: "Network operation timed out connecting to provider", Retryable: true,
@@ -27,14 +26,13 @@ var (
 		Code: 1004, Name: "PROTOCOL_TLS_MISMATCH", Category: CategoryInternal,
 		Description: "HTTP/HTTPS protocol mismatch", Retryable: false,
 	})
-	// MayHaveReachedNode: a reset arrives on an established connection, so the request may already
-	// have been written and read by the upstream.
+	// MayHaveReachedNode: a reset arrives on an established connection.
 	LavaErrorConnectionReset = registerError(&LavaError{
 		Code: 1005, Name: "PROTOCOL_CONNECTION_RESET", Category: CategoryInternal,
 		Description: "Connection reset by peer", Retryable: true,
 		MayHaveReachedNode: true,
 	})
-	// MayHaveReachedNode: an EOF likewise means we got as far as an established connection.
+	// MayHaveReachedNode: an EOF likewise implies an established connection.
 	LavaErrorConnectionClosed = registerError(&LavaError{
 		Code: 1006, Name: "PROTOCOL_CONNECTION_CLOSED", Category: CategoryInternal,
 		Description: "Connection closed (EOF)", Retryable: true,
@@ -47,8 +45,7 @@ var (
 	LavaErrorContextDeadline = registerError(&LavaError{
 		Code: 1007, Name: "PROTOCOL_CONTEXT_DEADLINE", Category: CategoryInternal,
 		Description: "Caller's context.Context deadline expired before the relay completed (can fire at any layer — consumer request timeout, processing timeout, subrequest timeout)", Retryable: true,
-		// MayHaveReachedNode: the failure mode behind both GK8 write incidents. The upstream
-		// accepted the request and never answered; the transaction was broadcast regardless.
+		// MayHaveReachedNode: the upstream accepted the request and never answered.
 		MayHaveReachedNode: true,
 	})
 	LavaErrorContextCanceled = registerError(&LavaError{
--- a/protocol/common/error_codes.go
+++ b/protocol/common/error_codes.go
@@ -7,9 +7,13 @@ package common
 
 var (
 	// Connection errors (1001-1009)
+	// MayHaveReachedNode: assigned from net.Error.Timeout(), which covers a dial timeout AND a
+	// timeout waiting for the response. The two are indistinguishable here, so take the safe
+	// direction for a write — "unclear" is recoverable, a false "definitely failed" is not.
 	LavaErrorConnectionTimeout = registerError(&LavaError{
 		Code: 1001, Name: "PROTOCOL_CONNECTION_TIMEOUT", Category: CategoryInternal,
 		Description: "Network operation timed out connecting to provider", Retryable: true,
+		MayHaveReachedNode: true,
 	})
 	LavaErrorConnectionRefused = registerError(&LavaError{
 		Code: 1002, Name: "PROTOCOL_CONNECTION_REFUSED", Category: CategoryInternal,
@@ -23,13 +27,18 @@ var (
 		Code: 1004, Name: "PROTOCOL_TLS_MISMATCH", Category: CategoryInternal,
 		Description: "HTTP/HTTPS protocol mismatch", Retryable: false,
 	})
+	// MayHaveReachedNode: a reset arrives on an established connection, so the request may already
+	// have been written and read by the upstream.
 	LavaErrorConnectionReset = registerError(&LavaError{
 		Code: 1005, Name: "PROTOCOL_CONNECTION_RESET", Category: CategoryInternal,
 		Description: "Connection reset by peer", Retryable: true,
+		MayHaveReachedNode: true,
 	})
+	// MayHaveReachedNode: an EOF likewise means we got as far as an established connection.
 	LavaErrorConnectionClosed = registerError(&LavaError{
 		Code: 1006, Name: "PROTOCOL_CONNECTION_CLOSED", Category: CategoryInternal,
 		Description: "Connection closed (EOF)", Retryable: true,
+		MayHaveReachedNode: true,
 	})
 	LavaErrorNetworkUnreachable = registerError(&LavaError{
 		Code: 1009, Name: "PROTOCOL_NETWORK_UNREACHABLE", Category: CategoryInternal,
@@ -38,6 +47,9 @@ var (
 	LavaErrorContextDeadline = registerError(&LavaError{
 		Code: 1007, Name: "PROTOCOL_CONTEXT_DEADLINE", Category: CategoryInternal,
 		Description: "Caller's context.Context deadline expired before the relay completed (can fire at any layer — consumer request timeout, processing timeout, subrequest timeout)", Retryable: true,
+		// MayHaveReachedNode: the failure mode behind both GK8 write incidents. The upstream
+		// accepted the request and never answered; the transaction was broadcast regardless.
+		MayHaveReachedNode: true,
 	})
 	LavaErrorContextCanceled = registerError(&LavaError{
 		Code: 1008, Name: "PROTOCOL_CONTEXT_CANCELED", Category: CategoryInternal,
--- a/protocol/common/error_codes.go
+++ b/protocol/common/error_codes.go
@@ -48,9 +48,14 @@ var (
 		// MayHaveReachedNode: the upstream accepted the request and never answered.
 		MayHaveReachedNode: true,
 	})
+	// MayHaveReachedNode: a cancellation says when we stopped waiting, never how far the request
+	// got. The relay may already have been on the wire — cancelling the context does not recall it —
+	// so for a write the transaction may be held by the node. A race LOSER is excluded separately by
+	// the successes>0 check in the write verdict, which is why this flag costs nothing there.
 	LavaErrorContextCanceled = registerError(&LavaError{
 		Code: 1008, Name: "PROTOCOL_CONTEXT_CANCELED", Category: CategoryInternal,
 		Description: "Request context was canceled (client disconnect or relay race resolved)", Retryable: false,
+		MayHaveReachedNode: true,
 	})
 
 	// Provider availability (1010-1019)
@@ -251,13 +256,19 @@ var (
 		Code: 2013, Name: "NODE_RESOURCE_UNAVAILABLE", Category: CategoryExternal,
 		Description: "Resource exists but unavailable", Retryable: true,
 	})
+	// MayHaveReachedNode on both: a 502/504 is the PROXY reporting that its own upstream did not
+	// answer it in time, which is exactly the deadline case one hop further out. The proxy forwarded
+	// the request before it gave up, so for a write the node behind it may hold the transaction. The
+	// reasoning the deadline, reset and EOF entries above already carry applies here unchanged.
 	LavaErrorNodeGatewayTimeout = registerError(&LavaError{
 		Code: 2014, Name: "NODE_GATEWAY_TIMEOUT", Category: CategoryExternal,
 		Description: "Gateway timeout (HTTP 504 from provider)", Retryable: true,
+		MayHaveReachedNode: true,
 	})
 	LavaErrorNodeBadGateway = registerError(&LavaError{
 		Code: 2015, Name: "NODE_BAD_GATEWAY", Category: CategoryExternal,
 		Description: "Bad gateway (HTTP 502 from provider)", Retryable: true,
+		MayHaveReachedNode: true,
 	})
 	// NODE_UNAUTHORIZED: upstream rejected the smart-router's credentials
 	// (HTTP 401). Non-retryable because the same credentials are reused on
--- a/protocol/common/error_registry.go
+++ b/protocol/common/error_registry.go
@@ -497,6 +497,10 @@ type errorMapping struct {
 var LavaErrorUnknown = &LavaError{
 	Code: 0, Name: "UNKNOWN_ERROR", Category: CategoryExternal,
 	Description: "Unclassified error — no matcher matched", Retryable: true,
+	// An error nobody matched proves nothing about whether the request left this process, and for
+	// a write the unprovable direction is the safe one: "unclear" is recoverable, a false
+	// "definitely failed" invites a resubmit of a transaction that may already be on chain.
+	MayHaveReachedNode: true,
 }
 
 // ---------------------------------------------------------------------------
--- a/protocol/common/error_registry.go
+++ b/protocol/common/error_registry.go
@@ -97,6 +97,18 @@ type LavaError struct {
 	SubCategory ErrorSubCategory
 	Description string
 	Retryable   bool // retrying same relay with same params to a different provider has a chance of succeeding
+
+	// MayHaveReachedNode marks a failure that does NOT prove the request stayed inside this
+	// process. Connect-phase failures — refused, DNS, TLS, unreachable — do prove it: nothing was
+	// ever written to a socket. Anything that fails after the connection is established — a
+	// deadline, a reset, an EOF — leaves the upstream possibly holding a request we never read the
+	// answer to.
+	//
+	// It exists for writes, where the difference is the whole story: a transaction we could not
+	// prove undelivered may already be broadcast, and reporting that as a flat failure tells the
+	// customer something we do not know. Declared here, at registration, so adding a connection
+	// error forces the question to be answered rather than defaulted.
+	MayHaveReachedNode bool
 }
 
 func (le *LavaError) String() string {
--- a/protocol/common/error_registry.go
+++ b/protocol/common/error_registry.go
@@ -98,16 +98,10 @@ type LavaError struct {
 	Description string
 	Retryable   bool // retrying same relay with same params to a different provider has a chance of succeeding
 
-	// MayHaveReachedNode marks a failure that does NOT prove the request stayed inside this
-	// process. Connect-phase failures — refused, DNS, TLS, unreachable — do prove it: nothing was
-	// ever written to a socket. Anything that fails after the connection is established — a
-	// deadline, a reset, an EOF — leaves the upstream possibly holding a request we never read the
-	// answer to.
-	//
-	// It exists for writes, where the difference is the whole story: a transaction we could not
-	// prove undelivered may already be broadcast, and reporting that as a flat failure tells the
-	// customer something we do not know. Declared here, at registration, so adding a connection
-	// error forces the question to be answered rather than defaulted.
+	// MayHaveReachedNode reports that the failure does not prove the request stayed inside this
+	// process. Connect-phase failures (refused, DNS, TLS, unreachable) do prove it; a deadline,
+	// reset or EOF leaves the upstream possibly holding the request. Matters for writes, where a
+	// transaction we cannot prove undelivered may already be broadcast.
 	MayHaveReachedNode bool
 }
 
--- a/protocol/common/per_node_timeout_test.go
+++ b/protocol/common/per_node_timeout_test.go
@@ -1,0 +1,64 @@
+package common
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// A per-endpoint `timeout:` EXTENDS the attempt budget; it does not cap it.
+//
+// The distinction decides whether the attempt-budget change works at all. That field is set on
+// nearly every endpoint in the shipped example configs, so a well-meaning change turning `+` into a
+// cap would pin those attempts back to a few seconds and make the whole thing a no-op exactly where
+// it is deployed — silently, since every test that does not set the field would still pass.
+func TestNodeUrlTimeoutExtendsTheBudgetRatherThanCappingIt(t *testing.T) {
+	const budget = 30 * time.Second
+	const perNode = 5 * time.Second
+
+	deadlineIn := func(url *NodeUrl) time.Duration {
+		// A parent with far more room than either value, so the result reflects the helper's own
+		// arithmetic rather than CapContextTimeout clamping to the parent.
+		parent, cancelParent := context.WithTimeout(context.Background(), time.Hour)
+		defer cancelParent()
+		ctx, cancel := url.LowerContextTimeoutWithDuration(parent, budget)
+		defer cancel()
+		deadline, ok := ctx.Deadline()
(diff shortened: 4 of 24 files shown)


avitenzer:
## Review findings

Reviewed at `b3aa713` (PR head). The last commit only adds tests, so the production code is the same as at `bd62c4a`, where the verification was done. All five touched packages pass locally (`rpcsmartrouter`, `relaycore`, `lavasession`, `common`, `chainlib`), the three new test files pass, and `-race` runs on the new tests were clean. Findings are ranked; each has a concrete scenario and a file link. Note that `TestWriteOutcomeIsUnknown_Wrapper` covers a connect-phase failure and a mid-delivery cut-off beside a silent sibling, but not a node error beside a silent sibling, which is finding 1.

### Confirmed

**1. A node error from one endpoint bypasses the write-outcome rule while a sibling is still hung.** (high)
`SendParsedRelay` only consults `unknownWrite()` when `ProcessingResult` returns an error ([rpcsmartrouter_server.go:1236](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/rpcsmartrouter/rpcsmartrouter_server.go#L1236)), but `processNonCrossValidationResult` returns the first node error with `err == nil` whenever there are no successes ([relay_processor.go:1244](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/relaycore/relay_processor.go#L1244)). Scenario: a write fans out to A and B; A answers HTTP 200 with a JSON-RPC error body in 50ms, B hangs; `checkEndProcessing` needs both responses or a success, so the request waits the full stateful budget for B, then returns A's node error as the result (`analytics.Success = true`) while B may have broadcast. The live "up-a HTTP 500" scenario passed because a JSON-RPC 5xx is a protocol error (`classifyAndWrap`, `err != nil`); a JSON-RPC error inside a 200 is a node error and takes the nil path. The PR's own test row "one answered with an error, one still silent → true" describes exactly this case.

**2. A policy Stop with an attempt still in flight parks the request until the budget and then forfeits the hang blame.** (high)
The Stop arm ([unified_relay_state_machine.go:455](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/relaycore/unified_relay_state_machine.go#L455)) records the reason and only schedules `validateReturnCondition`, which returns silently while `CurrentlyUsed() > 0` ([unified_relay_state_machine.go:331](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/relaycore/unified_relay_state_machine.go#L331)); nothing cancels the in-flight attempt. At expiry `stopReasonOr` keeps the policy reason, so `requestRanOutOfRoad` is false and the hung endpoint is released through `OnSessionCancelled`. Scenario: stateless read, A hangs, the ticker dispatches B at the window, B returns a non-retryable node error → `Stop("NonRetryableNodeError")` ([policy.go:43](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/relaypolicy/policy.go#L43)); the client gets B's error at the 30s budget (60s / 180s for CU≥50 / hanging) instead of within a window or two, and A is not blamed. Same through `ErrorToleranceExceeded` and `MaxRetriesReached`. This is not the stated trade: A was not cut short, it consumed the entire budget. The description lists "the policy stopping" as something that ends attempts; nothing does.

**3. The endpoint-level health tracker no longer sees hangs.** (medium-high)
A budget cancel reaches `relayInnerDirect` as a client cancellation ([rpcsmartrouter_server.go:4331](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/rpcsmartrouter/rpcsmartrouter_server.go#L4331)) and `classifyEndpointHealth` returns `(false, false)` for that ([error_mapper.go:78](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/rpcsmartrouter/error_mapper.go#L78)): no `MarkUnhealthy`, no probe evidence, no `SetEndpointOverallHealth(false)`. Before the PR a hang arrived as `DeadlineExceeded` (CategoryInternal → unhealthy) and disabled the endpoint after `MaxConsecutiveConnectionAttempts` ([consumer_types.go:452](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/lavasession/consumer_types.go#L452), [common.go:40](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/lavasession/common.go#L40)). On the non-CV path the attempt's own deadline (dispatch + budget + `url.Timeout`) is always later than `processingCtx` (created earlier from the same budget), so the `DeadlineExceeded` path is unreachable there, and "blamed through either of two paths depending on which clock fires first" does not hold for non-CV. The provider-level block via `ConsecutiveErrors` (through `OnSessionUnresponsive`) still happens; the endpoint disable, the recovery probe and the overall-health metric do not.

**4. Gateway 502/504 responses carry `MayHaveReachedNode: false`.** (medium-high)
Checked empirically with a throwaway test: `HTTPStatusError{504}` → `classifyAndWrap` → `NODE_GATEWAY_TIMEOUT(false)`, and the classification `setErrorResponse` stores is the same; the REST path via `httpStatusRelayError(504)` also stores `NODE_GATEWAY_TIMEOUT(false)`. Registry entries at [error_codes.go:254](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/common/error_codes.go#L254). Scenario: a write where every endpoint sits behind a proxy that returns 504 after forwarding the signed tx to a stalled node → `answered == dispatched`, `cutOffMidDelivery == false` → "failed relay, insufficient results" → the client resubmits. The reasoning the PR applies to deadline / reset / EOF applies to 504 exactly: the gateway's upstream did not answer in time.

**5. The single-batch assumption behind `dispatched` is false.** (medium)
`Decide` retries a Stateful relay when every result so far was a rate limit ([policy.go:34](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/relaypolicy/policy.go#L34)). Scenario: A and B both answer 429 (protocol errors, `answered = 2`), the retry dispatches C in batch 2 (`SessionsLatestBatch` resets to 1), C hangs → `ProcessingTimeout` → `answered < dispatched` is false, and a 429 stores as `NODE_RATE_LIMITED(false)` → `unknownWriteOutcome` returns false → "failed" with a 503 (via `onlyRateLimited`) while C may hold the transaction. The comment at [rpcsmartrouter_server.go:1066](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/rpcsmartrouter/rpcsmartrouter_server.go#L1066) says the stateful gate keeps multi-batch requests out.

**6. Goroutine leak when more than three attempts are in flight at expiry.** (medium)
The deferred `SetResponse` is an unconditional blocking send ([relay_processor.go:374](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/relaycore/relay_processor.go#L374)) into a channel sized `fanOut(1) + RelayRetryLimit(2) = 3` for non-CV. After `ProcessRelaySend` returns nothing reads `rp.responses` on the non-CV path: `WaitForResults` exited on `processingCtx.Done()`, and `readExistingResponses` is only reachable through `NodeResults()` (no callers) and the CV watcher. Before the PR attempts died staggered at their window while the reader was re-armed; now every in-flight attempt pushes together after the request ended. `smartrouter_eth.yml` ships four endpoints: a stateful write fanned out to four hung endpoints, or a hedged read that reached four attempts, leaves one goroutine per request blocked forever, pinning its `RelayProcessor`.

**7. Ticker loop noise while an attempt is in flight and the pool is empty.** (low)
Each window the ticker arm ([unified_relay_state_machine.go:463](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/relaycore/unified_relay_state_machine.go#L463)) → `Decide(tickerFired)` → Retry (`AttemptNumber` does not advance on `PairingListEmptyError`) → `SendStop` → the circuit-breaker WARNING ([unified_relay_state_machine.go:397](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/relaycore/unified_relay_state_machine.go#L397)) plus `HedgeCount++`. A single hung endpoint, 1s window, 180s budget → roughly 180 warnings and `HedgeCount ≈ 180` for one request. Before the PR the request ended after about one window.

**8. The "unclear" reply can go out as HTTP 503.** (low)
`buildFailureResult` stamps 503 when every recorded error was a 429 ([relay_processor.go:1311](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/relaycore/relay_processor.go#L1311)); the second failure branch returns that result with `errUnknownWriteOutcome` ([rpcsmartrouter_server.go:1248](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/rpcsmartrouter/rpcsmartrouter_server.go#L1248)), and the listeners honour a non-zero status. A 429 sibling plus a hung endpoint returns a retry-inviting 503 under "verify on-chain before resubmitting". Stamp 500 explicitly.

### Plausible

**9. A client disconnect on websocket / gRPC is labelled `ProcessingTimeout`.**
Both timeout arms ([unified_relay_state_machine.go:302](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/relaycore/unified_relay_state_machine.go#L302), [unified_relay_state_machine.go:500](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/relaycore/unified_relay_state_machine.go#L500)) label any `processingCtx.Err()`, including a parent cancel. The websocket manager passes a per-connection context that is cancelled on close ([consumer_websocket_manager.go:346](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/chainlib/consumer_websocket_manager.go#L346)); the gRPC listener passes the stream context ([grpc.go:430](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/chainlib/grpc.go#L430)). The HTTP listeners build the request context from `context.Background()` and are immune. The relay goroutines are cancelled at the same instant as `sm.ctx`, so whether `GetStopReason()` at [rpcsmartrouter_server.go:2091](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/rpcsmartrouter/rpcsmartrouter_server.go#L2091) reads `""` or `"ProcessingTimeout"` depends on scheduling; when it reads the latter, a healthy in-flight endpoint gets `OnSessionUnresponsive` for the client's hang-up. Before the PR the label was inert. Fix: gate the label on `errors.Is(processingCtx.Err(), context.DeadlineExceeded)`.

**10. A drained cancel turns "unclear" into "failed".**
At expiry, if `WaitForResults` ([relay_processor.go:865](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/relaycore/relay_processor.go#L865)) is still in its select when a cancelled goroutine's deferred `SetResponse{Err: context.Canceled}` lands, the select picks randomly; `setErrorResponse` then files `PROTOCOL_CONTEXT_CANCELED` with `MayHaveReachedNode: false` (pinned by `TestMayHaveReachedNodeClassification`) → `answered == dispatched`, `cutOffMidDelivery == false` → "failed relay" for a transaction the node may hold. The window is small but nonzero under load. Setting the flag to true on `LavaErrorContextCanceled` closes it; race losers are already excluded by the `successes > 0` check.

**11. Crafted startup / health relays now block a full budget per hung endpoint.**
`sendCraftedRelays` runs on `context.Background()` with no deadline ([rpcsmartrouter_server.go:886](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/rpcsmartrouter/rpcsmartrouter_server.go#L886)), and `WaitForResults` at [rpcsmartrouter_server.go:840](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/rpcsmartrouter/rpcsmartrouter_server.go#L840) is bounded only by the attempt budget (30s default) rather than the window (about 1s). With `relays-health-enable` (default true) each monitor cycle runs `sendCraftedRelays(MaxRelayRetries = 6)`, so a hung endpoint can cost up to 6×30s per cycle; with the flag off, `Start` blocks at [rpcsmartrouter_server.go:331](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/rpcsmartrouter/rpcsmartrouter_server.go#L331). `initialized` has no production readers, so readiness is not directly affected.

### Nits

- The comment at [rpcsmartrouter_server.go:2110](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/rpcsmartrouter/rpcsmartrouter_server.go#L2110) says `RelayOutcomeError` keeps a hang in the latency series; `RecordDirectRelayEnd` only records latency on success, and `observeBatchSize` fires for a relay that never completed.
- `recordRelayProbeEvidence` still records the window as the replay budget ([rpcsmartrouter_server.go:4353](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/rpcsmartrouter/rpcsmartrouter_server.go#L4353), [recovery_probe.go:194](https://github.com/Magma-Devs/smart-router/blob/b3aa7130609ff2e42a9943683154d254d5bed3a6/protocol/rpcsmartrouter/recovery_probe.go#L194)), so a method that only succeeds past its window replays as still-failing.
- `TestFailoverCadence_NextEndpointDispatchedAfterOneWindow` leaks the state-machine goroutine and ticker (context never cancelled, 50000s mock budget, task channel buffer 1). Sibling tests share the pattern.
- The `MinimumTimePerRelayDelay` mutation in the call-site test raced clean across three `-race` runs; not an issue today.
- The longer session holds and the CV straggler footprint are acknowledged in the description and not re-litigated here.

https://claude.ai/code/session_01A6LFGBAyHkeZKbBoUUXWwA

Tomelia1999:
Thank you — this was a genuinely good review. Every claim I checked was accurate, and separating "confirmed" from "plausible" honestly made it much faster to act on. **All nine accepted findings are fixed**, each as its own commit; two are closed as won't-fix with reasoning below.

Rebased onto current main (`71fa41b`), CI green, `make test` 38 packages 0 failures.

## Fixed

| # | Commit | What changed |
|---|---|---|
| **1, 4, 8, 10** | `8f88cb7` | One change, four places — see below |
| **3** | `f6d44ad` | A budget-expiry hang reaches the per-URL health path again |
| **6** | `f5487fb` | `SetResponse` gives up on the request's context; buffer sized for a stateful broadcast |
| **7** | `533abc1` | `HedgeCount` counted on dispatch, warning once per request |
| **9** | `aede6e1` | Only a real deadline is labelled a timeout |
| **11** | `371fc69` | The router's own relays get their own short deadline |

**1, 4, 8, 10 landed together** because they are the same judgement in four places. The verdict is no longer gated on `err != nil` — `processNonCrossValidationResult` returns a node error with a nil error, which is exactly why your finding 1 case was skipped. Hoisting it is safe: `writeOutcomeIsUnknown` already returns false unless this is a stateful relay with no successes. 502/504 and `PROTOCOL_CONTEXT_CANCELED` now carry `MayHaveReachedNode` — a race loser is excluded by the `successes > 0` check instead, so flagging the cancellation costs nothing there. And `withUnclearWriteStatus` forces 500: fixing finding 1 introduced a *second* wrong status beside your 503, since a node error carries the node's own, typically 200.

**11** bounds each try rather than the loop — trying several endpoints is the point of the loop, and bounding the whole thing would squeeze later tries out on a slow boot. Named `internalRelayTryTimeout`, and the comment says explicitly that it is not the endpoint prober.

**6** gives up on the request's context rather than dropping on a full buffer: while the request is alive a full buffer is transient and the response is still wanted. That alone makes the leak impossible at any fan-out; the sizing fix only stops a legitimate response being crowded out. You sized the risk from a 4-endpoint config — worth noting it is 10 against 3, since a write fans out to the session manager's per-tier cap.

**3** is exactly one condition, per Tom: the request ran out of budget while this endpoint was still silent. Everything else stays exempt. Worth recording what was *not* lost — `OnSessionUnresponsive` shares its body with the ordinary failure path, so the provider-level QoS availability failure and `ConsecutiveErrors` were both still happening. Latency is deliberately untouched, which is right: a node that never answered has an availability problem, not a latency one. What was missing was only the per-URL layer.

## Not changed, deliberately

**`usedComputeUnits` preferring `csm.pairing` for a dual-pool address.** Correct, and it needs no fix once the duplicate is excluded — a dual-pool address never reaches the backup ranking, so there is nothing left to rank by the wrong pool's counter. (That one is on #353.)

**`currentlyBlockedProviderAddresses` not consulted.** You asked for an explicit decision in the comment either way, and you were right that the comment is where it belongs — I had put it in a commit message. It is now recorded beside the filter: deliberate, only reachable for a dual-pool address, unreachable today because boot rejects that config, and the policy question deferred rather than widened into this change.

## Closed without a fix

**Finding 2 — won't fix.** Reaching it needs two things together: an attempt still in flight (so slower than the window) *and* a non-retryable answer from the endpoint dispatched after it. At this deployment's `--min-relay-timeout 10s` the cost is bounded — the client waits the 30s read budget instead of about 10.2s. Two narrower fixes were considered and both ruled out: cancelling only on a non-retryable answer is unsafe (a non-retryable answer from one endpoint does not mean the others would fail the same way, and the router *does* serve a late success today), and "give it only its own attempt budget" is already what happens — both budgets are the same call with the same inputs, and the first attempt starts when the request does. **The mitigation is configuration, not code**: on the default 1s window the same case wastes nearly the whole budget.

**Finding 5 — won't fix**, as an edge case: it needs an all-429 round first. Re-checked your premise by running the policy, and you were right that writes retry there — `rateLimitRetrySafe` is the one carve-out, and only from a completed attempt, never a hedge.

## Three existing tests were pinning the bugs

- `TestResponseBufferSize` asserted *"every other selection launches one relay at a time"* and pinned Stateful at a fan-out of 1 — the assumption behind finding 6.
- `TestMayHaveReachedNodeClassification` pinned `PROTOCOL_CONTEXT_CANCELED` as proved-undelivered — finding 10.
- `MockProtocolMessage.CheckResponseError` returned `false` unconditionally, so that harness **could not express a node error at all** — every reply was filed as a success. That is why your finding 1 gap had no test: it was not expressible. Now opt-in per message, default false, so existing tests are unchanged.

All corrected with the reasoning in the comment rather than the value just flipped.

## Verification

Where it was possible, each fix was checked by restoring the pre-fix behaviour and confirming the test fails — the leak (`SetResponse blocked after the request ended`), the health condition, and the write verdict. My first attempt at the health test asserted `classifyEndpointHealth` directly, which passes with or without the change; the condition is now a named function with a truth table.

Two nits taken: the struct alignment (`85d6310`) and the stale `sendRelayToDirectEndpoints` call sites after main's MAG-3540 (`71fa41b`). The remaining nits are untouched.

https://claude.ai/code/session_01S95m47jBa9iHmksf7f1t4m

avitenzer (approved):

PR #375, opened 2026-09-06
docs(cross-validation): add a walkthrough and a demo lane per use case

avitenzer:
# Description

`docs/CROSS-VALIDATION.md` explains how to configure cross-validation but stops at the config block. There was nothing an operator — or a reviewer — could **run** to watch each use case of the Cross Validation Enhancements PRD actually happen, and no written record of which behaviours are proven against a live router rather than only in unit tests.

This adds a **Walkthrough** section in the shape of `docs/RESP-CACHE.md` — a lanes table, six numbered scenarios (one per shipped use case), and a troubleshooting table — plus the two lanes that drive it. Every console block in the doc is verbatim output from a real run.

### The lanes

| Lane | Covers |
| --- | --- |
| `scripts/pre_setups/init_smartrouter_cv_demo.sh` | Six simulator providers in three groups, six policies. Sub-demos `--uc1` `--uc2` `--uc4` `--uc5` `--uc6`, or `--all` |
| `scripts/pre_setups/init_smartrouter_cv_default.sh` | The same fleet with **no** `cross-validation:` block and **no** group labels — UC-7 |

UC-7 is a separate lane on purpose: the thing under test is the *absence* of configuration, which a router loaded with policies can never demonstrate whatever request you send it.

Both are ownership-safe in the same way the RESP cache lanes are — they reclaim only a process they recorded starting (pid plus a start-time fingerprint), never run `killall`, refuse to start on ports they do not own, and write generated configs into gitignored `debugging/`.

### What the lanes prove

| Use case | Demonstrated by |
| --- | --- |
| UC-1 per-method policy | Mandate with no headers; no-policy method untouched; caller opt-in still works; `forbid-caller-cv` ignores headers; floor and cap clamp in both directions |
| UC-2 group diversity | Cross-group quorum succeeds; a single-group count quorum is rejected as `diversity-unmet`; per-group quorum succeeds and fails as `group-quorum-unmet`; both startup capacity fail-fasts refuse to boot |
| UC-4 mismatch metric | Reply-time dissent and straggler dissent, with the `group` and `finality="finalized"` labels |
| UC-5 structured failure | `no-agreement` vs `insufficient-capacity`, the absent provider lists on a request-time fail-fast, and the contrast with a plain upstream error |
| UC-6 outlier exclusion | Outvoted under `2 of 6`; the same divergence is fatal under `6 of 6` |
| UC-7 no configuration | The policy-layout line is *absent* rather than `policies=0`; the header-driven path is unchanged |

UC-3 (value-threshold-triggered validation) was descoped, so there is no lane for it. The doc now records that as a caveat in behavioural terms, so a reader comparing it against the PRD does not assume it was missed.

### Why the simulator and not real endpoints

Half of these behaviours only exist when providers **disagree**, and real endpoints agree on finalized state — a shared-truth fleet can never manufacture a dissent. The `provider_simulator`'s per-method body override returns a valid-but-divergent result, which is exactly the one input the mismatch surface admits: a successful content outlier on a deterministic method. Its latency knob decides who wins the race to quorum, so the reply-time and straggler recording paths are selected deterministically rather than by luck.

### Two things found while writing it

1. **The three existing UC harnesses cannot start a router.** `test_uc1_smartrouter_lava.sh`, `test_uc2_smartrouter_lava.sh`, `test_uc4_smartrouter_sim.sh` and `init_smartrouter_sol.sh` still pass `--geolocation 1`, removed in v1.0.3, which now fails with `unknown flag`. Dropped.
2. **A runtime error string referenced a removed PRD use case.** `ValidateNoStatefulPolicies` said `(see UC-3)`; it now states the actual reason a write response cannot be cross-validated.

### Most critical to review

* `docs/CROSS-VALIDATION.md` — the new **Local testing lanes** and **Walkthrough** sections, and the new value-threshold caveat
* `scripts/pre_setups/init_smartrouter_cv_demo.sh` — in particular the fleet shapes that make each failure deterministic (distinct dissenting values for `diversity-unmet`, one dissenter per group for `group-quorum-unmet`)

## Jira ticket

Jira ticket: MAG-3450

---

## Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change — n/a, no API or client change
* [x] included the necessary unit and integration tests — n/a for new tests: the lanes are themselves executable checks (39 + 10, all passing), and the behaviours they cover already have unit tests in `rpcsmartrouter`, `relaycore` and `metrics`, which were re-run clean

### Verification

Note the PR gate does not run `go test`, so the local run below is the only test evidence.

```
init_smartrouter_cv_demo.sh --all      39 passed, 0 failed
init_smartrouter_cv_default.sh         10 passed, 0 failed
go build ./...                         clean
go vet ./protocol/rpcsmartrouter       clean
go test ./protocol/rpcsmartrouter ./protocol/relaycore ./protocol/metrics
  -run 'CrossValidation|Quorum|Group|Diversity|Straggler'   188 tests, all pass
shellcheck -S warning (both new lanes) clean
```

## Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

### DIFF
--- a/docs/CROSS-VALIDATION.md
+++ b/docs/CROSS-VALIDATION.md
@@ -141,14 +141,20 @@ recording path) for test suites that cannot scrape the metrics port — see
 ## Local testing lanes
 
 Two lanes, each bringing up its own stack, checking itself, and then printing how to drive
-it by hand. They use distinct ports and can run at the same time — which is the point of the
-second one, since comparing a configured router against an unconfigured one is the
-backwards-compatibility claim.
+it by hand. They run the **same six simulator providers**, so the only interesting difference
+between them is the configuration — which is the point of the second one, since comparing a
+configured router against an unconfigured one is the backwards-compatibility claim.
 
 | Lane | Covers |
 | --- | --- |
-| [`scripts/pre_setups/init_smartrouter_cv_demo.sh`](../scripts/pre_setups/init_smartrouter_cv_demo.sh) | Six simulator providers in three groups, six policies. Sub-demos `--uc1` `--uc2` `--uc4` `--uc5` `--uc6` (or `--all`) drive per-method policy, group diversity, mismatch metrics, the structured failure signal, and outlier exclusion. |
-| [`scripts/pre_setups/init_smartrouter_cv_default.sh`](../scripts/pre_setups/init_smartrouter_cv_default.sh) | The same fleet with **no** `cross-validation:` block and **no** group labels — the pre-feature shape, proving existing deployments and callers are untouched. |
+| [`scripts/pre_setups/init_smartrouter_cv_demo.sh`](../scripts/pre_setups/init_smartrouter_cv_demo.sh) | Six simulator providers in three groups, seven policies. Sub-demos `--uc1` `--uc2` `--uc4` `--uc5` `--uc6` (or `--all`) drive per-method policy, group diversity, mismatch metrics, the structured failure signal, and outlier exclusion. |
+| [`scripts/pre_setups/init_smartrouter_cv_default.sh`](../scripts/pre_setups/init_smartrouter_cv_default.sh) | The same six providers with **no** `cross-validation:` block and **no** group labels — the pre-feature shape, proving existing deployments and callers are untouched. |
+
+**Run them one at a time.** Their router ports are distinct, but the simulator is shared
+mutable state and both lanes drive the same provider ids, so a dissent injected by one is a
+fixture the other is relying on. Neither lane ever issues the simulator's unscoped
+`POST /reset/all` — each states the providers it owns explicitly — so they leave *other*
+harnesses alone, but they cannot help colliding with each other.
 
 Both are ownership-safe: they reclaim only the process they recorded starting (pid plus a
 start-time fingerprint), never run `killall`, and **refuse to start** if their ports are held
@@ -186,7 +192,7 @@ scripts/pre_setups/init_smartrouter_cv_demo.sh
 | | Where | Role |
 | --- | --- | --- |
 | provider_simulator | control `127.0.0.1:19000`, eth-sim on `18545-18547` + `18560-18562` | six upstreams whose answers you control |
-| Smart Router | `0.0.0.0:3396`, metrics `:7794`, debug `:6796` | six policies over three provider groups |
+| Smart Router | `0.0.0.0:3396`, metrics `:7794`, debug `:6796` | seven policies over three provider groups |
 
 The fleet is deliberately lopsided in a useful way — three groups of two:
 
@@ -201,15 +207,15 @@ It ends with a self-check, so a broken stack is caught before you start:
 ```
 [Smoke] policy load -> fan-out -> quorum
     router /lava/health                    200
-    policies loaded                        policies=6
+    policies loaded                        policies=7
     distinct groups                        distinctGroups=3
     mandated cross-validation              success
-    agreeing providers                     sim-2,sim-4
+    agreeing providers                     sim-1,sim-4
 
   SMOKE PASS — the stack is demo-ready.
 ```
 
-`policies=6` and `distinctGroups=3` come from the router's own startup line — the resolved
+`policies=7` and `distinctGroups=3` come from the router's own startup line — the resolved
 provider→group layout, logged once so an operator can confirm the diversity a config actually
 yields rather than the diversity it looks like it asks for.
 
@@ -359,9 +365,18 @@ scripts/pre_setups/init_smartrouter_cv_demo.sh --uc5
     failure-reason                         insufficient-capacity
   PASS  no provider lists — nothing was queried, and the headers say so
 
-[3] the contrast: a plain upstream error carries NO cross-validation headers
-  PASS  'quorum failure' is distinguishable from a generic upstream error
+[3] the contrast: a REAL upstream error on a cross-validated method
+    body      {"error":{"code":-32000,"data":{"error":"failed processing responses…
+    status    failed
+    failure-reason  insufficient-responses
+  PASS  the upstream error reaches the client as an error
+  PASS  not reported as a disagreement (reason: insufficient-responses)
 ```
+
+Step 3 is the one that is easy to get wrong. Every provider is made to return a JSON-RPC
+error, so nothing disagreed — nobody answered at all — and the router says exactly that:
+`insufficient-responses`, never `no-agreement`. A client that keys its retry logic on the
+reason therefore cannot mistake "your fleet is down" for "your fleet disagreed".
 
 The split in the reason enum is the client's decision procedure. A **quorum-time** reason
 (`no-agreement`, `insufficient-responses`, `diversity-unmet`, `group-quorum-unmet`) means
@@ -440,8 +455,9 @@ on any provider, and no thresholds:
 
 The policy-layout line being *absent* rather than reading `policies=0` is the honest signal:
 the resolver was never populated, so the router is on the path it ran before this feature
-existed, not on the new path in a disabled state. The two configs differ by exactly the block
-this page is about:
+existed, not on the new path in a disabled state. Because both lanes run the same six
+providers, the diff between their configs is the feature and nothing else — the
+`cross-validation:` block, the `group-label:` lines, and the listen port:
 
 ```bash
 diff debugging/smartrouter_cv_default.yml debugging/smartrouter_cv_demo.yml
@@ -470,5 +486,5 @@ policy floor instead.
 | A dissent shows up as *pending* instead of *disagreeing* | The quorum early-exited before the outlier answered. Give the honest providers a `latency_ms` so the outlier wins the race — that is exactly what the lane does to pick the reply-time path. |
 | `finality="unknown"` on the mismatch metric | The request did not carry a resolvable block number, or the chain tracker had not learned the head yet. Query a concrete finalized block, not `latest`. |
 | A cross-validated method answers without fanning out | Something served it from cache. These lanes configure no cache for that reason; if you add one, vary the request parameters. |
-| Router exits at startup with a cross-validation error | Working as intended for an unsatisfiable policy — compare `min-groups` and `max-participants` against the `distinctGroups` / `groupSizes` in the startup log. |
-| `/debug/cross-validation-events` returns 503 | The recorder is not installed: the router is missing `--debug-address`. A 503 is not an empty result — "nothing was recorded" and "nothing dissented" are opposite answers. |
+| Router exits at startup with a cross-validation error | Working as intended for an unsatisfiable policy. The error itself carries the numbers: `requiredGroups` / `configuredGroups` on the `min-groups` refusal, `groupSizes` on the per-group one. Don't look for the `distinctGroups` startup line — the validation runs *before* it, so a router that exits this way never logs it. |
+| `/debug/cross-validation-events` is refused / connection reset | The router was started without `--debug-address`, so there is no debug listener at all. A **503** is a different state — the listener is up but the recorder was never installed — and it is not an empty result: "nothing was recorded" and "nothing dissented" are opposite answers. |
--- a/docs/CROSS-VALIDATION.md
+++ b/docs/CROSS-VALIDATION.md
@@ -132,3 +132,343 @@ recording path) for test suites that cannot scrape the metrics port — see
 - **Cost & latency.** N relays per request. Scope policies to the methods that warrant it.
 - **Public endpoints are best-effort.** The example fleets use rate-limited community
   endpoints; for production, point at your own nodes or keyed gateways.
+- **No value-threshold escalation.** There is no knob that raises a method's policy for an
+  individual request based on the value it carries (parsing `eth_sendRawTransaction` for a
+  native amount, decoding ERC-20 calldata). That was descoped; a method's policy is the same
+  for every request to it. Callers that need a stricter quorum on a specific request can still
+  ask for one with the headers, up to the policy's cap.
+
+## Local testing lanes
+
+Two lanes, each bringing up its own stack, checking itself, and then printing how to drive
+it by hand. They use distinct ports and can run at the same time — which is the point of the
+second one, since comparing a configured router against an unconfigured one is the
+backwards-compatibility claim.
+
+| Lane | Covers |
+| --- | --- |
+| [`scripts/pre_setups/init_smartrouter_cv_demo.sh`](../scripts/pre_setups/init_smartrouter_cv_demo.sh) | Six simulator providers in three groups, six policies. Sub-demos `--uc1` `--uc2` `--uc4` `--uc5` `--uc6` (or `--all`) drive per-method policy, group diversity, mismatch metrics, the structured failure signal, and outlier exclusion. |
+| [`scripts/pre_setups/init_smartrouter_cv_default.sh`](../scripts/pre_setups/init_smartrouter_cv_default.sh) | The same fleet with **no** `cross-validation:` block and **no** group labels — the pre-feature shape, proving existing deployments and callers are untouched. |
+
+Both are ownership-safe: they reclaim only the process they recorded starting (pid plus a
+start-time fingerprint), never run `killall`, and **refuse to start** if their ports are held
+by anything else, printing a ready-made override. Generated configs land in `debugging/`,
+which is gitignored. Tear down with `--stop`; `--status` reports what is up.
+
+They drive the sibling **`provider_simulator`** rather than real endpoints, because half of
+these behaviours only exist when providers *disagree* and real endpoints agree on finalized
+state — a shared-truth fleet can never manufacture a dissent. The simulator's per-method body
+override returns a valid-but-divergent result, which is exactly the one input the mismatch
+surface admits: a successful content outlier on a deterministic method. Its latency knob
+decides who wins the race to quorum, so the reply-time and straggler paths are selected
+deterministically rather than by luck. Point the lanes at a checkout with `SIM_DIR=…` if it
+does not sit next to this one.
+
+The router-side unit tests need no infrastructure:
+
+```bash
+go test ./protocol/rpcsmartrouter -run 'CrossValidation|Quorum|Group' -v   # policy resolution, fail-fast, event recorder
+go test ./protocol/relaycore     -run 'Diversity|Straggler|Quorum'   -v   # quorum computation, group selection, the async watcher
+go test ./protocol/metrics       -run 'CrossValidation'              -v   # label shapes and cardinality
+```
+
+## Walkthrough
+
+Six scenarios. The first five run against one router that stays up between them, so you can
+interleave your own `curl`s; the sixth is the second lane.
+
+### Setup
+
+```bash
+scripts/pre_setups/init_smartrouter_cv_demo.sh
+```
+
+| | Where | Role |
+| --- | --- | --- |
+| provider_simulator | control `127.0.0.1:19000`, eth-sim on `18545-18547` + `18560-18562` | six upstreams whose answers you control |
+| Smart Router | `0.0.0.0:3396`, metrics `:7794`, debug `:6796` | six policies over three provider groups |
+
+The fleet is deliberately lopsided in a useful way — three groups of two:
+
+```yaml
+tier-1   sim-1  sim-2
+external sim-3  sim-4
+archive  sim-5  sim-6
+```
+
+It ends with a self-check, so a broken stack is caught before you start:
+
+```
+[Smoke] policy load -> fan-out -> quorum
+    router /lava/health                    200
+    policies loaded                        policies=6
+    distinct groups                        distinctGroups=3
+    mandated cross-validation              success
+    agreeing providers                     sim-2,sim-4
+
+  SMOKE PASS — the stack is demo-ready.
+```
+
+`policies=6` and `distinctGroups=3` come from the router's own startup line — the resolved
+provider→group layout, logged once so an operator can confirm the diversity a config actually
+yields rather than the diversity it looks like it asks for.
+
+Knobs, all optional: `SIM_DIR` for the simulator checkout, `ROUTER_PORT` / `METRICS_PORT` /
+`DEBUG_PORT` / `NEG_PORT` to move ports, `SKIP_SMOKE=1`.
+
+### 1. A method's policy outranks what the caller asked for (UC-1)
+
+```bash
+scripts/pre_setups/init_smartrouter_cv_demo.sh --uc1
+```
+
+```
+[1] 'eth_getBalance' has a policy with enabled: true — no caller headers sent
+    status                                 success
+    all-providers                          sim-1,sim-2,sim-3,sim-4,sim-5,sim-6
+    agreeing-providers                     sim-2,sim-3
+
+[2] 'eth_getBlockByNumber' has NO policy and the caller sent no headers
+  PASS  not cross-validated — a policy is scoped to its own method
+
+[4] 'eth_gasPrice' carries forbid-caller-cv — the same headers are IGNORED
+  PASS  cross-validation is off for this method whatever the caller asks
+
+[5] precedence on 'eth_getCode' — floor 2/cap 3 threshold, max-participants pinned to 3
+    a caller asking for MORE than the cap is clamped down to it:
+    requested                              max-participants 6, threshold 6
+    all-providers                          sim-2,sim-4,sim-6
+    a caller asking for LESS than the floor gets the floor:
+    requested                              max-participants 1, threshold 1
+    all-providers                          sim-1,sim-4,sim-6
+```
+
+Step 5 is the answer to "which wins, the header or the policy?": **neither, categorically** —
+each knob resolves as `clamp(caller, floor, cap)`. A caller who asks for six providers on a
+capped method gets three; a caller who asks for one gets three as well. The two directions are
+the same rule, and `forbid-caller-cv` (step 4) is the escape hatch for a method where the
+caller should have no say at all.
+
+The resolution is logged per request at debug level (`CrossValidation mode enabled
+(policy-resolved)`), so you can see which authority produced the numbers a given relay used.
+
+### 2. A quorum that isn't independent is not a quorum (UC-2)
+
+```bash
+scripts/pre_setups/init_smartrouter_cv_demo.sh --uc2
+```
+
+`eth_getTransactionCount` carries `min-groups: 2`. The interesting case is the failure: every
+provider outside `tier-1` is made to answer a *distinct* wrong value, so the only pair that
+agrees is `sim-1` + `sim-2` — a legitimate count quorum, drawn entirely from one vendor.
+
+```
+[2] a count quorum that all came from ONE group must be REJECTED
+    status                                 failed
+    failure-reason                         diversity-unmet
+```
+
+`eth_call` uses the stronger form (`per-group-quorum: true`): each of two groups must reach
+its *own* quorum of two and the winners must then agree. Breaking one provider in every group
+leaves no group able to reach its internal quorum:
+
+```
+[4] break ONE provider in EVERY group — no group can reach its own quorum
+    status                                 failed
+    failure-reason                         group-quorum-unmet
+```
+
+Two distinct reasons for two distinct policies, which is what lets a client tell "the fleet
+disagreed" from "the fleet agreed but not diversely enough".
+
+The lane then starts two throwaway routers with policies the fleet can never satisfy, and both
+refuse to boot:
+
+```
+[5] a policy the fleet can NEVER satisfy is refused at STARTUP, not per request
+  PASS  refused to start — min-groups: 4 over a 3-group fleet
+        ERR cross-validation min-groups policy cannot be satisfied: configured provider groups are fewer than required
+  PASS  refused to start — per-group-quorum with max-participants 3 < 2 * 2
+        ERR invalid cross-validation configuration error="… per-group-quorum needs max-participants >= min-groups * agreement-threshold"
+```
+
+The second one is caught by the config preflight, before a single provider is dialled. A
+misconfigured diversity policy is a deployment error, not a per-request failure — so it costs
+you a failed rollout rather than a silent quorum you were not actually getting.
+
+### 3. Divergence after quorum reaches the alerting surface (UC-4)
+
+```bash
+scripts/pre_setups/init_smartrouter_cv_demo.sh --uc4
+```
+
+`sim-6` (group `archive`) is made to answer first, and to answer wrong:
+
+```
+[1] reply-time dissent: sim-6 (archive) answers FIRST and answers wrong
+    status                                 success
+    disagreeing-providers                  sim-6
+    mismatch_total{group=archive}          0 -> 1
+
+smartrouter_cross_validation_mismatch_total{apiInterface="jsonrpc",finality="finalized",group="archive",method="eth_getBalance",spec="ETH1"} 1
+```
+
+The request succeeded — the honest majority answered — and the divergence is still visible.
+The labels are the point: `group` rather than provider address keeps cardinality bounded and
+stable enough to alert on, and `finality="finalized"` is the high-signal case, because
+divergence on settled state cannot be explained away as propagation lag. That label is why the
+lane queries block `0x1000000` rather than `latest`.
+
+The second scenario delays the same outlier past the quorum early-exit, so it is still in
+flight when the reply ships:
+
+```
+[2] straggler dissent: the SAME outlier, delayed past the quorum early-exit
+    status                                 success
+    pending-providers                      sim-1,sim-3,sim-5,sim-6
+    straggler_total{outcome=disagreed}     0 -> 1
+```
+
+It is reported as **pending**, never silently dropped and never accused in
+`disagreeing-providers` — which lists only dissent the router actually received. An async
+watcher compares its late answer against the reached consensus and increments the same
+mismatch counter, deduped per group. A dissenter that answers after the quorum closed still
+reaches alerting instead of vanishing.
+
+Per-request detail lives at `GET /debug/cross-validation-events` (the lane passes
+`--debug-address`), which names the provider, its group, and which path recorded it:
+
+```bash
+curl -s "http://127.0.0.1:6796/debug/cross-validation-events?outcome=disagreed" | jq
+```
+
+### 4. A quorum failure is not a generic upstream error (UC-5)
+
+```bash
+scripts/pre_setups/init_smartrouter_cv_demo.sh --uc5
+```
+
+```
+[1] QUORUM-TIME failure: all six answer differently, so nothing reaches 2
+    status                                 failed
+    failure-reason                         no-agreement
+    all-providers                          sim-1,sim-2,sim-3,sim-4,sim-5,sim-6
+
+[2] STRUCTURAL failure: a caller asks for more providers than exist
+    status                                 failed
+    failure-reason                         insufficient-capacity
+  PASS  no provider lists — nothing was queried, and the headers say so
+
+[3] the contrast: a plain upstream error carries NO cross-validation headers
+  PASS  'quorum failure' is distinguishable from a generic upstream error
+```
+
+The split in the reason enum is the client's decision procedure. A **quorum-time** reason
+(`no-agreement`, `insufficient-responses`, `diversity-unmet`, `group-quorum-unmet`) means
+responses came back and did not agree — a retry against a different set may help. A
+**request-time structural** reason (`insufficient-capacity`, `insufficient-groups`) means the
+candidate set could not even be assembled, so retrying this router will not help and the client
+should fall back. The structural case carries no provider lists precisely because no provider
+was queried; `failure-reason` is the whole signal.
+
+The same split is on the metrics side, so alerts can distinguish them without parsing logs:
+
+```
+smartrouter_cross_validation_failures_total{method="eth_getBalance",reason="no-agreement",…} 1
+smartrouter_cross_validation_failures_total{method="eth_getBlockByNumber",reason="insufficient-capacity",…} 1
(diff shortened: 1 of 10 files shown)


AnnaR-prog (commented):
### Review of the two new cross-validation lanes

15 findings, ranked most-severe first as inline comments below.

**What checks out** — worth stating, because it is most of the surface area: every hard-coded router constant the lanes assert on is correct. All 7 `lava-cross-validation-*` header names, all 6 `failure-reason` enum values, the 4 metric names and their label sets, the policy YAML schema including the `Bound` scalar shorthand, the 2 verbatim startup errors, and all 3 `go test -run` patterns (4/9/42 matching tests). The `--geolocation 1` removal is right (the flag no longer exists in Go) and the error-string reword in `cross_validation_policy.go` breaks no test — `cross_validation_policy_test.go:726` asserts only on the method name, and `UC-3` had no definition anywhere in the repo, so replacing it with a self-contained explanation is a genuine fix.

**What is broken is the harness around the assertions, not the assertions.**

The shape of it:

1. **One destructive defect** (#1): both lanes can kill a *different checkout's* router, because reclaim matches on a relative config path and runs before the port-ownership guard.
2. **Five checks that cannot fail** (#2-#6): a regex that matches the healthy startup line, a metric this lane can never emit, a negative grep that passes on a missing file, three cumulative counters that pass on any re-run, and a step that exercises neither half of its own claim. These would report all-green against a broken router.
3. **Three that produce false failures** (#7-#9): `timeout` is absent on macOS, which the lane targets; `--all`'s reuse branch skips `write_config` and `smoke` behind a gate that accepts a 503; and `sim_scenario`'s rejection is discarded at ~20 call sites, so a simulator-side failure is reported as a router bug.
4. **Three doc claims that are not true of the code** (#12-#14), and the 255-of-391-line duplication between the two lanes (#15), which is what forces #1 to be fixed twice.

Also cut below the line as lower-value: `metric()` re-scraping `/metrics` 9-33x per run, `timeout 90` x2 stalling ~3 minutes on the failure path, `reclaim_owned`'s unconditional `rm -f` before confirming death, no simulator-version guard, `python3` missing from the default lane's preflight, no reclaim path for an orphaned `NEG_PORT` router, the unexplained UC-3 numbering gap, the UC-7 sample block omitting steps [4]-[6], and the `failures_total` sample listing two reasons only `--uc2` produces.

<sub>Automated review via Claude Code.</sub>

scripts/pre_setups/init_smartrouter_cv_demo.sh
```diff
@@ -0,0 +1,1069 @@
...
+    echo ""
+
+    for tool in jq curl python3 lsof screen; do
+        command_exists "$tool" || { echo "ERROR: '$tool' is required."; exit 1; }
+    done
+    [[ -f "$SPEC_FILE" ]] || { echo "ERROR: spec not found: $SPEC_FILE"; exit 1; }
+
+    echo "[Setup] reclaiming this lane's previous run (if any)"
+    reclaim_owned "$PIDFILE" "router"
+    reclaim_by_identity "$ROUTER_PORT" "$CONFIG_REL" "router"
```
AnnaR-prog: **1. Reclaim kills another checkout's router, before the port guard**

Both lanes reclaim by a **relative** config path and a fixed screen name, and do it **before** the port-ownership guard, so running a lane from a second checkout silently kills the first checkout's router — the opposite of the safety contract the header promises.

`reclaim_by_identity`'s ownership proof is the needle `$CONFIG_REL` = `debugging/smartrouter_cv_demo.yml` (line 62), byte-identical in every checkout of this repo; `ROUTER_SCREEN="sr-cv-demo"` (line 66) is likewise global to the user.

Checkout A's demo lane runs on :3396. A developer runs the same lane from checkout B: B's `reclaim_owned` returns immediately (no pidfile of its own), then `reclaim_by_identity` (879) `lsof`s :3396, matches the relative needle in `ps -o command=`, and SIGTERMs **A's** router, while `screen -S sr-cv-demo -X quit` (880) kills A's session. Because reclaim runs at 877-880 and the port-blocked preflight only at 883, the port is free by the time the guard runs — no refusal, no warning.

Same defect at `init_smartrouter_cv_default.sh:97-99`. The header at lines 46-52 claims "it cannot reach another lane or another checkout" and "refuses to run if its ports are held by anything else".

**Fix:** use `$CONFIG_FILE` (absolute) as the needle and derive the screen name from the checkout path.

scripts/pre_setups/init_smartrouter_cv_demo.sh
```diff
@@ -0,0 +1,1069 @@
...
+    note "failure-reason" "$(hdr failure-reason)"
+    check "the request failed" "$(hdr status)" "failed"
+    check "with the per-group reason" "$(hdr failure-reason)" "group-quorum-unmet"
+
+    fleet_clean
+
+    echo ""
+    echo "[5] a policy the fleet can NEVER satisfy is refused at STARTUP, not per request"
+    assert_refuses_to_start "$NEG_DIVERSITY_REL" \
+        "min-groups|distinct.*group|insufficient-groups" \
```
AnnaR-prog: **2. min-groups regex matches the *healthy* startup line**

The min-groups fail-fast regex was loosened relative to the sibling lane's tight one, and now matches the HEALTHY startup line — so the check reports PASS for a router that started fine, and the health-probe guard that should have caught it can never fire.

The regex is `min-groups|distinct.*group|insufficient-groups`, applied with `grep -qiE` to the whole debug-level `$NEG_LOG`. A router whose min-groups guard regressed starts fine and logs `cross-validation per-method policies loaded ... distinctGroups=3 ...` (`rpcsmartrouter_server.go:158-162`, emitted whenever `HasPolicies()` is true), which `distinct.*group` matches case-insensitively; `relay_processor.go:925` logs `distinctGroups` on every served relay too. So the lane prints "PASS refused to start — min-groups: 4 over a 3-group fleet" for an instance that never refused.

The pre-existing sibling `scripts/pre_setups/test_uc2_smartrouter_lava.sh:349` already had the correct anchored pattern `min-groups policy cannot be satisfied|configured provider groups are fewer than required|insufficient-groups` — this lane weakened it.

The second guard cannot compensate: the `curl .../lava/health` at line 627 runs *after* the `timeout 90` subshell returns, i.e. after the process is dead either way. The comment at 609-611 describes a guard that does not exist.

scripts/pre_setups/init_smartrouter_cv_default.sh
```diff
@@ -0,0 +1,391 @@
...
+    note "agreeing-providers" "$(hdr agreeing-providers)"
+    check "it reports success" "$(hdr status)" "success"
+    check "having compared nothing — one provider queried" "$(count_csv "$(hdr all-providers)")" "1"
+    echo "    A client that needs real corroboration must read the provider lists,"
+    echo "    not 'status' alone. An operator policy is what forecloses this."
+
+    echo ""
+    echo "[6] the cross-validation counters only move for requests that asked for it"
+    curl -s -m 10 "http://127.0.0.1:${METRICS_PORT}/metrics" 2>/dev/null \
+        | grep '^smartrouter_cross_validation_requests_total' | sed 's/^/        /'
```
AnnaR-prog: **3. UC-7 step [6] asserts a metric this lane can never emit**

UC-7 step [6] asserts on a metric series that can never exist in this lane, so it passes unconditionally — and its printed conclusion contradicts what steps [3] and [5] just did.

`smartrouter_cross_validation_mismatch_total` is only created by a successful content **outlier** (`protocol/metrics/smartrouter_metrics_manager.go:1456-1457`: "NOT incremented for quorum failures, no-agreement, diversity-unmet, or node/protocol errors"). This lane never injects a dissent — `checks()` opens with `sim_reset` and all three providers return the identical `eth_getBalance` default `0x0` — so `grep -c '^smartrouter_cross_validation_mismatch_total'` is 0 whether or not policies loaded and whether or not cross-validation ran. It also passes when `METRICS_PORT` is unreachable: curl emits nothing and `grep -c` prints 0.

Meanwhile steps [3] and [5] **do** cross-validate (header-driven), so the printed "no mismatch series — nothing was cross-validated by policy" is false on its own lane. Ship a build that loads policies from a config with no `cross-validation:` block and step [6] still says PASS.

**Fix:** the discriminating series is `smartrouter_cross_validation_requests_total`, which line 221 prints but never asserts.

scripts/pre_setups/init_smartrouter_cv_default.sh
```diff
@@ -0,0 +1,391 @@
...
+checks() {
+    echo ""
+    echo "============================================"
+    echo "UC-7 — no cross-validation configuration"
+    echo "============================================"
+    sim_reset
+
+    echo ""
+    echo "[1] the startup log carries no policy layout at all"
+    if grep -q "cross-validation per-method policies loaded" "$ROUTER_LOG" 2>/dev/null; then
```
AnnaR-prog: **4. Negative grep swallows a missing log and reports PASS**

The lane's primary UC-7 assertion is a negative grep whose `2>/dev/null` swallows a missing or unreadable log file, so it reports PASS having read no evidence at all.

`grep -q PATTERN "$ROUTER_LOG" 2>/dev/null` exits non-zero both when the pattern is absent **and** when the file does not exist, and both land in the else branch printing "PASS absent, not 'policies=0' — the resolver was never populated".

Reachable via the documented `--check` mode (line 380) against a router started by hand or with a different `ROUTER_PORT`/`PROJECT_ROOT`, after `debugging/logs` was cleaned, or when the screen+tee pipeline died before writing.

**Fix:** add a positive anchor in the same file (assert a known startup line IS present) before concluding the policy-layout line is absent.

scripts/pre_setups/init_smartrouter_cv_demo.sh
```diff
@@ -0,0 +1,1069 @@
...
+    check "the quorum still formed around the honest answer" "$(hdr status)" "success"
+    check "the dissenter is named in the response headers" "$(hdr disagreeing-providers)" "sim-6"
+    if [[ "$after" -gt "$before" ]]; then
+        pass "the alerting counter moved for the outlier's group"
+    else
+        fail "mismatch_total{group=$GROUP_C} did not move (${before} -> ${after})"
+    fi
+    curl -s -m 10 "http://127.0.0.1:${METRICS_PORT}/metrics" \
+        | grep '^smartrouter_cross_validation_mismatch_total' | sed 's/^/        /'
+    if [[ "$(metric cross_validation_mismatch_total 'finality="finalized"')" -gt 0 ]]; then
```
AnnaR-prog: **5. Three checks read cumulative counters, so re-runs cannot fail**

Three checks assert on process-lifetime state instead of the current request, so they cannot fail on any re-run against the long-lived router this lane is designed around.

**(a)** uc4 step [1] line 664 tests `metric cross_validation_mismatch_total finality="finalized"` `-gt 0` cumulatively — two lines after the same function correctly used a before/after delta (649/651). On the second `--uc4` or `--all` the counter is already >= 1, so it passes even when THIS request resolved to `finality="unknown"`, the exact failure its own message names.

**(b)** uc5 step [4] lines 760-761 do the same for `failures_total` `reason="no-agreement"` and `reason="insufficient-capacity"`; a regression that stopped emitting `SetCrossValidationFailureMetric` entirely still prints its PASS line on run 2.

**(c)** uc1 step [6] line 528 greps the whole accumulated `$ROUTER_LOG` for `CrossValidation mode enabled (policy-resolved)`, which `smoke()` already guaranteed by issuing a mandated call before uc1 ran, and which `tee` only truncates on a router restart — so it passes even if uc1's own five requests bypassed the resolver.

**Fix:** all three need a per-run snapshot (a delta, or `tail -n +N` from a line count taken at the top of the sub-demo).

scripts/pre_setups/init_smartrouter_cv_demo.sh
```diff
@@ -0,0 +1,1069 @@
...
+    check "and a structural reason, not a quorum one" "$(hdr failure-reason)" "insufficient-capacity"
+    if [[ -z "$(hdr all-providers)" ]]; then
+        pass "no provider lists — nothing was queried, and the headers say so"
+    else
+        fail "provider lists present on a request-time fail-fast: '$(hdr all-providers)'"
+    fi
+
+    echo ""
+    echo "[3] the contrast: a plain upstream error carries NO cross-validation headers"
+    fleet_clean
```
AnnaR-prog, line 679: **6. uc5 step [3] cannot fail; neither half of its claim runs**

uc5 step [3] cannot fail, and never exercises the upstream error it claims to contrast against.

`call "eth_getBlockByHash" "[\"0xnotahash\",false]"` sends no `lava-cross-validation-*` request headers, and `eth_getBlockByHash` has no policy in the generated config, so `cv_present` is false by construction — structurally identical to uc1 step [2] at line 468, which already makes that point.

It also never produces an upstream error: the simulator's `METHOD_DEFAULTS` maps `eth_getBlockByHash` to a full block object, so the response is a plain JSON-RPC success. The check prints "PASS 'quorum failure' is distinguishable from a generic upstream error" without either half of that comparison having happened.

**Fix:** force a real upstream error (the simulator's per-method `error_stub` / `mode=error`) on a method that DOES carry a policy, or send CV headers.

scripts/pre_setups/init_smartrouter_cv_demo.sh
```diff
@@ -0,0 +1,1069 @@
...
+}
+
+# assert_refuses_to_start <config-rel> <error-regex> <what>
+# A time-boxed foreground router. A satisfiable config would instead run until the
+# timeout kills it and answer on its port — both of which fail this check, so a
+# port clash or a dead upstream cannot masquerade as a passing capacity test.
+assert_refuses_to_start() {
+    local cfg="$1" rx="$2" what="$3" rc
+    echo "    trying to start: $what"
+    ( cd "$PROJECT_ROOT" && timeout 90 smartrouter "$cfg" \
```
AnnaR-prog: **7. `timeout` does not exist on macOS; two assertions always FAIL**

`assert_refuses_to_start` depends on `timeout`, which does not exist on macOS — the platform the script explicitly targets — and the lane's own tool preflight is the one place that would have caught it and omits it.

Verified on a stock macOS box: neither `timeout` nor `gtimeout` is on `PATH`, in a login bash shell either, and no coreutils under `/opt/homebrew` or `/usr/local`. The preflight at line 872 is `for tool in jq curl python3 lsof screen` — no `timeout` — while line 254 documents that the script targets macOS bash 3.2.

`--uc2` or `--all` writes `timeout: command not found` into `$NEG_LOG` with rc=127, smartrouter is never launched, neither regex matches, and both calls at lines 600 and 603 report FAIL, so the run exits 1. The printed message is "expected a capacity error matching /.../ (rc=127)" — the lane reports the cross-validation startup fail-fast as BROKEN when the only thing missing is a coreutils binary.

Two guaranteed false regressions against the feature the lane exists to demonstrate. Either add `timeout` to the preflight with an install hint, or replace it with a background-PID + kill pattern.

scripts/pre_setups/init_smartrouter_cv_demo.sh
```diff
@@ -0,0 +1,1069 @@
...
+case "${1:-}" in
+    --stop)   teardown; exit 0 ;;
+    --status) status; exit 0 ;;
+    --uc1)    require_up; uc1 ;;
+    --uc2)    require_up; write_negative_configs; uc2 ;;
+    --uc4)    require_up; uc4 ;;
+    --uc5)    require_up; uc5 ;;
+    --uc6)    require_up; uc6 ;;
+    --all)
+        if router_up; then
```
AnnaR-prog: **8. `--all` reuse branch skips write_config and smoke, on a lax gate**

`--all`'s reuse branch skips both `write_config` and `smoke`, and the `router_up` gate it keys off returns true for ANY HTTP response — so all five sub-demos can run against a stale or unhealthy router with the gate that would have caught it disabled.

`router_up` (line 297) is `curl -s -o /dev/null -m 5 .../lava/health` with no `-w '%{http_code}'` and no `--fail`; curl exits 0 on a 503. `wait_ready` nine lines above (288) gets this right by comparing `%{http_code}` to `200` — the strict and loose forms sit side by side in the same file.

In `--all`, `if router_up; then ... write_negative_configs` (1044-1046) skips `write_config` AND `smoke`, so a router started by an older revision of the script, one using a hand-edited `debugging/smartrouter_cv_demo.yml`, or one answering 503 because its spec load failed, is accepted and driven through uc1..uc6. Every check then reports FAIL with `(got '', want 'success')` because `$HDR` carries no `lava-*` headers — pages of failures reading as "cross-validation is broken" instead of one accurate line.

`require_up` (299) and `init_smartrouter_cv_default.sh:130/381` share the laxness.

scripts/pre_setups/init_smartrouter_cv_demo.sh
```diff
@@ -0,0 +1,1069 @@
...
+        -H 'Content-Type: application/json' -d "$1")
+    if [[ "$(printf '%s' "$response" | tail -n1)" != "200" ]]; then
+        echo "  ! simulator rejected the scenario: $(printf '%s' "$response" | sed '$d')"
+        return 1
+    fi
+}
+
+# fleet_clean : every provider honest and fast. Always call this before injecting,
+# so no scenario inherits the previous one's latency or body override.
+fleet_clean() {
```
AnnaR-prog, line 140: **9. `sim_scenario`'s failure is discarded at ~20 call sites**

`sim_scenario`'s `return 1` on a rejected scenario is discarded at every one of its ~20 call sites, so a failed injection silently leaves the previous use case's overrides in place and the resulting FAIL is attributed to the router.

There is no `set -e`. `fleet_clean` (197), `dissent` (208) and `slow` (216) all propagate `sim_scenario`'s `return 1` (line 192), and every call site — 465, 543, 563-567, 596, 647-648, 674, 699, 716, 732, 748, 770, 789-792, 815, 827, 836, 941 — is a bare statement.

The simulator 400s on an unknown provider key or field and applies scenario updates per field, so a rejected `fleet_clean` leaves the PREVIOUS scenario fully intact, not partially. Concretely: uc4 step [1] sets `slow 1..5 400` and `dissent 6 eth_getBalance 0xdeadbeef`; if uc6's `fleet_clean` at line 789 is rejected, sim-1..5 are still at 400ms and sim-6 still returns `0xdeadbeef`, so uc6 FAILs with "the client received the outlier's value" — a simulator-side rejection reported as a router bug, behind one `!` line buried in ~100 lines of output.

The comment at 195-196 states this is exactly what `fleet_clean` exists to prevent.

scripts/pre_setups/init_smartrouter_cv_demo.sh
```diff
@@ -0,0 +1,1069 @@
...
+            tail -n 20 "$SIM_LOG" 2>/dev/null | sed 's/^/    /'
+            exit 1; }
+        echo "  simulator up (control $SIM_CONTROL, eth-sim on $SIM_PORTS)"
+    fi
+    fleet_clean
+    echo "  every provider honest and fast (clean slate for startup verification)"
+
+    # --- config ---------------------------------------------------------------
+    echo ""
+    echo "[Setup] generating ${CONFIG_REL} (+ two deliberately unsatisfiable ones)"
```
AnnaR-prog, line 847: **10. Unguarded config write can boot a stale config silently**

The config write is the one `bring_up` step with no error guard, so a failed redirection silently boots a stale config while the script reports the numbers it believes it wrote.

`bring_up` guards `make install` (904, `|| { ...; exit 1; }`) and `record_owned` (970, `|| exit 1`), but `write_config` (946) and `write_negative_configs` (947) are bare statements whose `{ ... } > "$CONFIG_FILE"` status is discarded, and with no `set -e` nothing aborts.

If `debugging/smartrouter_cv_demo.yml` is root-owned from an earlier `sudo` run, or the disk is full, bash prints one "Permission denied" line, the cosmetic `wc -c` echo at 948 reports the STALE file's size followed by the hardcoded "6 providers / 3 groups / 6 policies", and the router boots the old config. If that file happens to satisfy `smoke`'s two greps (`policies=6`, `distinctGroups=3`), the smoke passes and uc1-uc6 assert against thresholds and group labels that are not the ones the script wrote.

Same at `init_smartrouter_cv_default.sh:312`.

scripts/pre_setups/init_smartrouter_cv_default.sh
```diff
@@ -0,0 +1,391 @@
...
+            ready=1; break
+        fi
+        screen -list 2>/dev/null | grep -q "$ROUTER_SCREEN" || break
+        sleep 1
+    done
+    [[ "$ready" == "1" ]] || {
+        echo "ERROR: the router never became healthy. Tail of $ROUTER_LOG:"
+        tail -n 30 "$ROUTER_LOG" 2>/dev/null | sed 's/^/    /'
+        exit 1; }
+    record_owned "$PIDFILE" "$ROUTER_PORT" "router" || exit 1
```
AnnaR-prog, line 268: **11. No settle delay before an all-three-providers assertion**

The default lane goes straight from `/lava/health` 200 into an assertion that requires all three providers to be valid, with none of the settle delay the demo lane deliberately has.

`bring_up` ends at `record_owned` (337) and `checks()` runs immediately; the demo lane pays `sleep 5` at `init_smartrouter_cv_demo.sh:973` for the same transition. A 200 from `/lava/health` is not a provider-readiness signal — `ServeRPCRequests` starts the listener in a goroutine and provider validation continues after.

Step [3] (lines 191-198) then sends `lava-cross-validation-max-participants: 3` over exactly 3 providers and asserts `count_csv(all-providers) == 3` and `status == success`. If any one of sim-1..3 is not yet a valid address when `validateCrossValidationCapacity` runs (`rpcsmartrouter_server.go:657`: `MaxParticipants > candidateProviders`), the request returns `failed` / `insufficient-capacity` and both checks FAIL under the banner "UC-7 FAIL — existing deployments and existing callers need zero changes".

There is zero slack, because the caller asks for the entire fleet.

docs/CROSS-VALIDATION.md
```diff
@@ -132,3 +132,343 @@ recording path) for test suites that cannot scrape the metrics port — see
...
+- **No value-threshold escalation.** There is no knob that raises a method's policy for an
+  individual request based on the value it carries (parsing `eth_sendRawTransaction` for a
+  native amount, decoding ERC-20 calldata). That was descoped; a method's policy is the same
+  for every request to it. Callers that need a stricter quorum on a specific request can still
+  ask for one with the headers, up to the policy's cap.
+
+## Local testing lanes
+
+Two lanes, each bringing up its own stack, checking itself, and then printing how to drive
+it by hand. They use distinct ports and can run at the same time — which is the point of the
```
AnnaR-prog: **12. "Can run at the same time" is false: the simulator is shared**

The claim that the two lanes "can run at the same time" is false: both issue the unscoped `POST /reset/all` against the shared simulator and both drive eth-sim pids 1-3, so concurrent runs destroy each other's fixtures.

`sim_reset` (`init_smartrouter_cv_demo.sh:181`; `init_smartrouter_cv_default.sh:112`) posts to `/reset/all` with no `pool` argument. On the current simulator (`control_api.py` `_perform_reset`/`_scope`), `pool=None` means every provider in the simulator, not just this lane's. The demo lane calls it via `fleet_clean` at 21 sites; the default lane at lines 169 and 308.

The default lane's providers are ports 18545-18547 = eth-sim:1,2,3 = the demo lane's sim-1..sim-3. So uc5 step [1] (`for p in 1..6; dissent $p eth_getBalance 0x$p$p`) makes the default lane's step [3] fail its "header-driven path is untouched / 3 providers agreed" assertion, while the default lane's `sim_reset` simultaneously wipes uc5's injected dissent (uc2 [2] then sees all six agree -> `got 'success' want 'failed'`).

Only the router ports are distinct; the simulator is shared mutable state — and this doc presents concurrency as the reason the second lane exists. The simulator's reset accepts a `pool`, which neither lane passes.

docs/CROSS-VALIDATION.md
```diff
@@ -132,3 +132,343 @@ recording path) for test suites that cannot scrape the metrics port — see
...
+| A dissent shows up as *pending* instead of *disagreeing* | The quorum early-exited before the outlier answered. Give the honest providers a `latency_ms` so the outlier wins the race — that is exactly what the lane does to pick the reply-time path. |
+| `finality="unknown"` on the mismatch metric | The request did not carry a resolvable block number, or the chain tracker had not learned the head yet. Query a concrete finalized block, not `latest`. |
+| A cross-validated method answers without fanning out | Something served it from cache. These lanes configure no cache for that reason; if you add one, vary the request parameters. |
+| Router exits at startup with a cross-validation error | Working as intended for an unsatisfiable policy — compare `min-groups` and `max-participants` against the `distinctGroups` / `groupSizes` in the startup log. |
+| `/debug/cross-validation-events` returns 503 | The recorder is not installed: the router is missing `--debug-address`. A 503 is not an empty result — "nothing was recorded" and "nothing dissented" are opposite answers. |
```
AnnaR-prog: **13. Two Troubleshooting rows describe impossible router states**

Two of the six Troubleshooting rows describe states the router cannot produce, sending an operator to look for evidence that is never emitted.

**(a)** This row says `/debug/cross-validation-events` returns 503 when "the router is missing `--debug-address`". Without `--debug-address` there is no debug HTTP server at all (`rpcsmartrouter.go:386-400` gates the whole listener on it), so the operator gets connection-refused, never a 503. The code's own comment at `rpcsmartrouter.go:1855-1858` says the 503 is "Structurally impossible on a router that serves this endpoint at all" — the row documents the one state that cannot happen and omits the one that does.

**(b)** The row above tells the operator to "compare min-groups and max-participants against the `distinctGroups` / `groupSizes` in the startup log", but `validateCrossValidationStartup` runs at `rpcsmartrouter_server.go:153` and returns BEFORE the `LavaFormatInfo` at 158 that carries those fields — a router that exits on a cross-validation error never logs either. The min-groups error carries `requiredGroups`/`configuredGroups`; the per-group error carries `groupSizes`. Name those instead.

docs/CROSS-VALIDATION.md
```diff
@@ -132,3 +132,343 @@ recording path) for test suites that cannot scrape the metrics port — see
...
+
+Two lanes, each bringing up its own stack, checking itself, and then printing how to drive
+it by hand. They use distinct ports and can run at the same time — which is the point of the
+second one, since comparing a configured router against an unconfigured one is the
+backwards-compatibility claim.
+
+| Lane | Covers |
+| --- | --- |
+| [`scripts/pre_setups/init_smartrouter_cv_demo.sh`](../scripts/pre_setups/init_smartrouter_cv_demo.sh) | Six simulator providers in three groups, six policies. Sub-demos `--uc1` `--uc2` `--uc4` `--uc5` `--uc6` (or `--all`) drive per-method policy, group diversity, mismatch metrics, the structured failure signal, and outlier exclusion. |
+| [`scripts/pre_setups/init_smartrouter_cv_default.sh`](../scripts/pre_setups/init_smartrouter_cv_default.sh) | The same fleet with **no** `cross-validation:` block and **no** group labels — the pre-feature shape, proving existing deployments and callers are untouched. |
```
AnnaR-prog: **14. "The same fleet" is false; the doc's own `diff` refutes it**

This section calls the default lane "The same fleet" and says the two generated configs "differ by exactly the block this page is about", but they differ by three providers, six group-label lines and a port — so the `diff` the doc tells the reader to run undercuts the backwards-compatibility argument the section is making.

`init_smartrouter_cv_demo.sh:81` emits sim-1..sim-6 (18545-18547 + 18560-18562), each with a `group-label:`, listening on 3396. `init_smartrouter_cv_default.sh:46` emits only sim-1..sim-3 with no `group-label:`, listening on 3399.

A reader who runs this page's own `diff debugging/smartrouter_cv_default.yml debugging/smartrouter_cv_demo.yml` (line 447, echoed by the default lane's cheat sheet as "see exactly what the feature adds") gets a diff dominated by three extra provider blocks, six group-label lines and a changed listen port — not the `cross-validation:` block the sentence promises. Section 6's own sample output already contradicts "the same fleet": it shows `all-providers sim-1,sim-2,sim-3`.

**Fix:** either make the default lane emit the same six providers (dropping only the policy block and the labels), or drop the two claims.

scripts/pre_setups/init_smartrouter_cv_default.sh
```diff
@@ -0,0 +1,391 @@
+#!/bin/bash
```
AnnaR-prog, line 1: **15. 255 of 391 lines duplicate the demo lane — and fork finding #1**

This 391-line lane is a re-typed copy of the demo lane rather than a `--no-policies` flag on it. That duplication is the root cause of finding #1, which now has to be fixed in two divergent shapes, and the same ownership block already exists in five shapes across `pre_setups/`.

Only `write_config` (31 l), `checks()` (74 l) and `cheat_sheet` (28 l) are lane-specific. The rest — the ownership block (`proc_fingerprint`/`reclaim_owned`/`record_owned`/`reclaim_by_identity`/`teardown`), `pass`/`fail`/`check`/`note`, `call`/`addr`/`hdr`/`cv_present`/`count_csv`, `sim_up`/`sim_reset`, the 19-line port preflight, the 9-line `SIM_DIR` search, the screen+tee bring-up and the readiness poll — is duplicated: roughly 255 of 391 lines.

The header's argument ("a separate lane rather than a flag ... because the thing under test IS the absence of configuration") conflates a second router **instance** with a second **script file**: the demo lane already starts extra instances from alternative configs emitted by the same `emit_fleet` helper (`write_negative_configs`, lines 416-455).

The cost is already visible inside this one PR: `record_owned` is 16 lines in the demo (134) and 11 here (77); `check()` is 3 lines vs 1; the port-blocked remedy text differs. Repo-wide, `proc_fingerprint` has 7 definitions, `reclaim_by_identity` 6, `reclaim_owned` 6 in five divergent variants, and the port preflight appears verbatim in 7 files — so the cross-checkout kill in finding #1 must be found and fixed in every copy.

**Targets:** a sourced `scripts/pre_setups/_lane_lib.sh` for the helpers, and a `--no-policies` flag replacing this file.

scripts/pre_setups/init_smartrouter_cv_demo.sh
avitenzer: Fixed in 14986a3. `reclaim_by_identity` now takes the **absolute** `$CONFIG_FILE` as its needle and rejects a relative one outright (`die`), so a match is proof of this checkout. Screen names and pid files carry a per-checkout tag derived from `$PROJECT_ROOT` (`sr-cv-demo-<tag>`). Both lanes get this from the one implementation in `_cv_lane_lib.sh`, so it is fixed once. Also added a reclaim path for an orphaned `NEG_PORT` router.

scripts/pre_setups/init_smartrouter_cv_demo.sh
avitenzer: Fixed in 14986a3. Both patterns are now the router's own anchored error strings — `cross-validation min-groups policy cannot be satisfied` and `per-group-quorum needs max-participants` — so nothing in the healthy startup line or a served relay can match. `assert_refuses_to_start` also fails explicitly on rc=124 (still running when the bound expired), which is the guard the old comment described but did not implement.

scripts/pre_setups/init_smartrouter_cv_default.sh
avitenzer: Fixed in 14986a3. Step [6] no longer asserts on `mismatch_total`; it only reports it, with the reason it proves nothing here stated inline. The assertion moved to `requests_total{method}` as a **delta** across steps [2] and [3]: it must not move for the header-less call, and must move for the header-driven one. A build that loaded policies from this block-free config now fails step [2].

scripts/pre_setups/init_smartrouter_cv_default.sh
avitenzer: Fixed in 14986a3. The check now anchors on a line that must be present (`RPCSmartRouter started`) before concluding the policy-layout line is absent, and reports "cannot read this router's startup log — nothing was verified" as a FAIL when the log is missing or unreadable.

scripts/pre_setups/init_smartrouter_cv_demo.sh
avitenzer: Fixed in 14986a3, all three. (a) uc4's finality check is now a before/after delta on `mismatch_total{finality="finalized"}`, read from the same scrape as the group delta. (b) uc5's two `failures_total` reasons are deltas taken at sub-demo entry. (c) uc1 records `wc -l` of the router log on entry and greps only `tail -n +N` from it, so the smoke's earlier mandated call cannot satisfy it. Added `check_gt` for the delta assertions.

scripts/pre_setups/init_smartrouter_cv_demo.sh
avitenzer: Fixed in 14986a3. Step [3] now forces a real JSON-RPC error (`error_stub: revert`) on **every** provider of a method that **does** carry a policy (`eth_getTransactionByHash`, a seventh policy added for this). It asserts the client gets an error body, and that the failure-reason is not a disagreement reason. Live output: `insufficient-responses` — nobody answered, which is a different and honest statement from "they disagreed".

scripts/pre_setups/init_smartrouter_cv_demo.sh
avitenzer: Fixed in 14986a3. The dependency is gone rather than declared: `run_bounded` in `_cv_lane_lib.sh` backgrounds the process with `( cd "$workdir" && exec ... ) &`, polls `kill -0`, and returns 124 on expiry. `exec` makes `$!` the router's own pid so the kill reaches it. `env -C` was rejected for the same reason as `timeout` — absent on older macOS.

scripts/pre_setups/init_smartrouter_cv_demo.sh
avitenzer: Fixed in 14986a3. `--all` now always calls `bring_up`, so it can never drive a config it did not just write, and the smoke always runs. Every readiness probe is `router_healthy`, which compares `%{http_code}` to `200` — the lax `curl -o /dev/null` form is gone from both lanes, including `require_up` and the default lane's `--check` gate.

scripts/pre_setups/init_smartrouter_cv_demo.sh
avitenzer: Fixed in 14986a3. `sim_scenario` now calls `die` on a non-200, with a message saying why continuing is useless: any check after a dropped injection measures the previous scenario. That removes the need to guard ~20 bare call sites. `fleet_clean` also no longer calls `POST /reset/all` — it states this lane's six providers explicitly, so it stops wiping other harnesses' fixtures (see #12).

scripts/pre_setups/init_smartrouter_cv_demo.sh
avitenzer: Fixed in 14986a3. `write_config_atomic` runs the emitter into a temp file, checks its exit status, refuses an empty result, and `mv`s into place — dying with a specific message on any failure. Both lanes' config writes and both negative configs go through it, so a permissions or disk failure can no longer boot a stale file behind a cosmetic byte count.

scripts/pre_setups/init_smartrouter_cv_default.sh
avitenzer: Fixed in 14986a3. Added `wait_providers`, which polls a header-driven cross-validated call until `all-providers` reports the expected count — actual selectability, not a health code. Both lanes use it after `record_owned`. It warms on a method the checks do not assert on (`eth_getBlockByNumber`) so the per-method counters the checks read stay clean. The slack is better too: the default lane now runs six providers and asks for three (see #14).

docs/CROSS-VALIDATION.md
avitenzer: Fixed in 14986a3. The claim is gone; the section now says **run them one at a time**, and explains that the router ports are distinct but the simulator is shared mutable state and both lanes drive the same provider ids. Separately, neither lane issues `POST /reset/all` any more — each states the providers it owns via `/scenario` — so they no longer wipe *other* harnesses' fixtures, which was the wider half of this bug.

docs/CROSS-VALIDATION.md
avitenzer: Fixed in 14986a3. (a) The row now describes connection-refused as the symptom of a missing `--debug-address`, and keeps 503 as the distinct "listener up, recorder not installed" state. (b) The startup-error row no longer sends the reader to the `distinctGroups` line, which a router that exits this way never reaches; it names `requiredGroups`/`configuredGroups` and `groupSizes`, which the errors themselves carry.

docs/CROSS-VALIDATION.md
avitenzer: Fixed in 14986a3 by making the claim true rather than dropping it: the default lane now emits the **same six providers** on the same simulator ports. Verified — the diff is 6 `group-label:` lines, the listen port, and the `cross-validation:` block, and nothing else. The doc and the lane's cheat sheet both say precisely that now.

scripts/pre_setups/init_smartrouter_cv_default.sh
avitenzer: Done in 14986a3: `scripts/pre_setups/_cv_lane_lib.sh` now holds the ownership block, the port pre-flight, the assertion and request helpers, the metric helpers, the simulator control plane, `run_bounded` and `write_config_atomic`. The default lane drops from 391 to 282 lines and is now a config emitter plus `checks()`. I kept two entry points rather than a `--no-policies` flag — the two configs are genuinely different artefacts and the `diff` between them is part of the UC-7 demonstration — but the duplication argument is answered: there is one implementation of everything that can drift. The repo-wide dedup (7 copies of `proc_fingerprint`, the port pre-flight in 7 files) is real but belongs in its own change, not this PR.

AnnaR-prog (approved):

PR #376, opened 2026-09-09
feat(debug): add cache state endpoint

avitenzer:
Jira ticket: MAG-3478

## Summary

Adds read-only `GET /debug/cache-state`: which cache backend is actually serving, per-tier configuration and reachability, the addresses each tier uses, and the TTL policy the backend is applying.

## Response shape

`debugCacheStateResponse` is a package-level type with a `SchemaVersion`, matching the `/debug/runtime-config` convention so a consumer can unmarshal into a shared type.

Engine lives **on the tier**, not at the top level. The secondary is always a gRPC `*Cache` (`performance.InitCache` is its only constructor) while the primary becomes `*RespCache` whenever the `resp-cache` block is set — so a single top-level scalar both concealed a two-engine deployment and reported `none` for a secondary-only router. A top-level `engine` remains as a convenience, falling back to the secondary when no primary is configured.

`reachable` is a `*bool` that is always present: `true`, `false`, or `null` for "not determined yet". `null` is a real state — a RESP backend before its first probe returns, and a gRPC connection mid-dial. Collapsing it into `false` reports a healthy backend as unreachable to anything that polls promptly.

Because `reachable:false` means opposite things per engine — gRPC skips the tier entirely, RESP still issues every lookup and pays the full timeout — each tier carries `when_unreachable` stating which.

## Reachability is now read, not latched

`CacheReachable()` for the gRPC tier reads the `grpc.ClientConn` connectivity state instead of "a client pointer is installed". Two consequences:

- **The endpoint no longer mutates what it measures.** The old path went `CacheActive() → getClient()`, which spawns `go reconnectClient()` when the client is nil — so observing a down cache dialled it. `GetState()` has no side effects.
- **It stops being a latch.** The installed-pointer signal was cleared only by `resetOnConnectionError` on `codes.Unavailable` from a relay, so an idle router reported `reachable:true` for a backend killed an hour earlier.

A hanging backend still reads `reachable:true` — the transport genuinely is up — and that is documented rather than papered over.

## Bugs fixed

- **A router with no cache reported that it had one.** `SelectCacheBackend` returns `(*Cache)(nil)`; a typed nil in a non-nil interface still satisfies the reporter interface, so the default deployment answered `{"engine":"grpc","primary":{"configured":true}}`. Configuration is now discriminated on the address, not on the interface being non-nil.
- **`lifetimes` was always `0`/`0`.** The expiration flags are registered on the cache-server cobra command, never on the smartrouter one, so both `viper.GetDuration` calls returned zero on every deployment. Lifetimes now come from the backend that owns the policy, and are `null` for a `cache-be` tier whose TTLs live in another pod.
- **`RespCache.Close()` left `reachable` true** for a closed backend, observable during shutdown because the debug server and the cache close independently.
- **RESP addresses dropped `read-addresses` and `key-prefix`.** `Store.Ping` probes both clients, so a dead reader flipped `reachable` to `false` while the payload named only the healthy writer.
- **`NewWithClient`/`NewWithClients` produced a permanently empty address**, which is the seam every RESP test uses. Addresses are now derived from the client when no operator config is present.

## Interfaces

`var _ DebugCacheStateReporter` assertions added for both backends — the endpoint hangs on a runtime type assertion that would otherwise stop matching silently on a rename. `Cache.CacheAddress()` and `Cache.BackendEndpoint()` now read one field instead of two coincidentally-equal ones.

`debugMuxDeps` no longer stores the cache twice; the handler asserts on the existing `cache` field, so a test fixture cannot wire one endpoint and not the other.

## Security note

This is the first `/debug` route to expose the router's own backing-store addresses, and the debug listener has no authentication. No credential travels in these fields (`Password`/`PasswordFile` are separate config), but under `topology: sentinel` these are sentinel control-plane addresses. Documented with an explicit bind-to-loopback warning.

## Verification

Unit tests added covering every tier shape, the wire-key contract, the typed-nil and lifetimes regressions, and the read-only guarantee. Run from a clean worktree checkout of the head commit.

Note that CI on this repo does not run `go test` — the green checks below are builds, lint, govulncheck and the Jira gate — so the local run is the test evidence.

### DIFF
--- a/ecosystem/cache/redisstore/store.go
+++ b/ecosystem/cache/redisstore/store.go
@@ -62,8 +62,9 @@ type Store struct {
 	// configuration cannot answer this: under sentinel the serving node changes
 	// on every failover, and under cluster it depends on the key's slot.
 	// Observability only — nothing in the cache path reads these.
-	readEndpoint  *endpointTracker
-	writeEndpoint *endpointTracker
+	readEndpoint        *endpointTracker
+	writeEndpoint       *endpointTracker
+	configuredAddresses string
 
 	// stopWatcher terminates the credential poll loop (nil when the
 	// credentials are static).
@@ -152,6 +153,7 @@ func New(cfg Config) (*Store, error) {
 		}
 		return nil, err
 	}
+	store.configuredAddresses = strings.Join(cfg.Addresses, ",")
 	store.writeEndpoint = writeTracker
 	store.readEndpoint = readTracker
 	if cfg.PasswordFile != "" {
@@ -172,6 +174,14 @@ func New(cfg Config) (*Store, error) {
 		}
 	}
 	return store, nil
+}
+
+// ConfiguredAddresses returns the operator-provided primary RESP addresses.
+func (s *Store) ConfiguredAddresses() string {
+	if s == nil {
+		return ""
+	}
+	return s.configuredAddresses
 }
 
 // warnIfReadSplitIsDiscoveryScoped flags the one read-split shape that does
--- a/ecosystem/cache/redisstore/store.go
+++ b/ecosystem/cache/redisstore/store.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"sync/atomic"
@@ -62,9 +63,21 @@ type Store struct {
 	// configuration cannot answer this: under sentinel the serving node changes
 	// on every failover, and under cluster it depends on the key's slot.
 	// Observability only — nothing in the cache path reads these.
-	readEndpoint        *endpointTracker
-	writeEndpoint       *endpointTracker
-	configuredAddresses string
+	readEndpoint  *endpointTracker
+	writeEndpoint *endpointTracker
+
+	// configuredEndpoints is the STATIC configuration the trackers above
+	// deliberately do not answer — what an operator wrote down, so a debug reader
+	// can compare it against the node actually serving. Kept separate from that
+	// group precisely because their doc comment argues against static
+	// configuration for THEIR purpose.
+	//
+	// New() sets it from the operator's Config, which is the authoritative source.
+	// The NewWithClient/NewWithClients seams have no Config, so newStore derives
+	// what it can from the clients themselves: those two constructors are what
+	// every RESP test injects a miniredis client through, and leaving them empty
+	// is what would make this field untestable without a live Redis.
+	configuredEndpoints storeEndpoints
 
 	// stopWatcher terminates the credential poll loop (nil when the
 	// credentials are static).
@@ -153,7 +166,13 @@ func New(cfg Config) (*Store, error) {
 		}
 		return nil, err
 	}
-	store.configuredAddresses = strings.Join(cfg.Addresses, ",")
+	// The operator's own configuration is authoritative, so it replaces whatever
+	// newStore derived from the clients.
+	store.configuredEndpoints = storeEndpoints{
+		Addresses:     append([]string(nil), cfg.Addresses...),
+		ReadAddresses: append([]string(nil), cfg.ReadAddresses...),
+		KeyPrefix:     store.prefix,
+	}
 	store.writeEndpoint = writeTracker
 	store.readEndpoint = readTracker
 	if cfg.PasswordFile != "" {
@@ -176,12 +195,76 @@ func New(cfg Config) (*Store, error) {
 	return store, nil
 }
 
-// ConfiguredAddresses returns the operator-provided primary RESP addresses.
-func (s *Store) ConfiguredAddresses() string {
+// storeEndpoints is the operator-facing identity of a RESP deployment: every
+// address the store uses, plus the keyspace it uses them in.
+//
+// All three parts are load-bearing for a debug reader, and each was learnt by
+// asking what an operator would do wrong without it:
+//
+//   - ReadAddresses, because Ping probes BOTH clients and returns the first
+//     error. With reads split to another region, killing the reader reports the
+//     store unreachable while naming only the healthy writer — the operator curls
+//     the printed address, finds it fine, and blames the endpoint, while every
+//     cache LOOKUP is failing against an address nothing ever showed them.
+//   - KeyPrefix, because it decides which keyspace this router occupies. Two
+//     routers on one Valkey with different prefixes share zero entries and are
+//     otherwise indistinguishable in this output.
+type storeEndpoints struct {
+	Addresses     []string
+	ReadAddresses []string
+	KeyPrefix     string
+}
+
+// String renders the endpoints for the debug payload's address field. Read
+// addresses are named as such so the two roles are never confused for a list of
+// interchangeable nodes.
+func (e storeEndpoints) String() string {
+	if len(e.Addresses) == 0 && len(e.ReadAddresses) == 0 {
+		return ""
+	}
+	parts := make([]string, 0, 3)
+	if len(e.Addresses) > 0 {
+		parts = append(parts, strings.Join(e.Addresses, ","))
+	}
+	if len(e.ReadAddresses) > 0 {
+		parts = append(parts, "read="+strings.Join(e.ReadAddresses, ","))
+	}
+	if e.KeyPrefix != "" {
+		parts = append(parts, "prefix="+e.KeyPrefix)
+	}
+	return strings.Join(parts, " ")
+}
+
+// ConfiguredEndpoints renders every address this store is configured to use and
+// the key prefix it uses them under. Empty only when the store was built from
+// clients whose addresses could not be determined.
+func (s *Store) ConfiguredEndpoints() string {
 	if s == nil {
 		return ""
 	}
-	return s.configuredAddresses
+	return s.configuredEndpoints.String()
+}
+
+// clientAddresses recovers the configured addresses from an already-constructed
+// client, for the NewWithClient/NewWithClients seams that never see a Config.
+// Unknown implementations (including test doubles) yield nothing rather than a
+// guess.
+func clientAddresses(client redis.UniversalClient) []string {
+	switch c := client.(type) {
+	case *redis.Client:
+		return []string{c.Options().Addr}
+	case *redis.ClusterClient:
+		return append([]string(nil), c.Options().Addrs...)
+	case *redis.Ring:
+		addrs := make([]string, 0, len(c.Options().Addrs))
+		for _, addr := range c.Options().Addrs {
+			addrs = append(addrs, addr)
+		}
+		sort.Strings(addrs) // Ring addresses come from a map; keep the output stable
+		return addrs
+	default:
+		return nil
+	}
 }
 
 // warnIfReadSplitIsDiscoveryScoped flags the one read-split shape that does
@@ -224,12 +307,20 @@ func newStore(writeClient, readClient redis.UniversalClient, keyPrefix string) (
 	if !keyPrefixPattern.MatchString(keyPrefix) {
 		return nil, fmt.Errorf("invalid resp-cache key prefix %q: must match %s — SCAN MATCH patterns are globs, so glob characters could purge unrelated keys", keyPrefix, keyPrefixPattern.String())
 	}
+	// Derived rather than left empty so the exported NewWithClient/NewWithClients
+	// seams produce a usable address. New() overwrites this with the operator's
+	// Config, which knows the read addresses too.
+	endpoints := storeEndpoints{Addresses: clientAddresses(writeClient), KeyPrefix: keyPrefix}
+	if readClient != writeClient {
+		endpoints.ReadAddresses = clientAddresses(readClient)
+	}
 	return &Store{
-		write:         writeClient,
-		read:          readClient,
-		prefix:        keyPrefix,
-		readEndpoint:  &endpointTracker{},
-		writeEndpoint: &endpointTracker{},
+		write:               writeClient,
+		read:                readClient,
+		prefix:              keyPrefix,
+		readEndpoint:        &endpointTracker{},
+		writeEndpoint:       &endpointTracker{},
+		configuredEndpoints: endpoints,
 	}, nil
 }
 
--- a/protocol/performance/cache.go
+++ b/protocol/performance/cache.go
@@ -216,6 +216,15 @@ func (cache *Cache) CacheActive() bool {
 	return cache != nil && cache.clientStore.getClient() != nil
 }
 
+func (cache *Cache) CacheEngine() string { return "grpc" }
+func (cache *Cache) CacheAddress() string {
+	if cache == nil {
+		return ""
+	}
+	return cache.address
+}
+func (cache *Cache) CacheReachable() bool { return cache.CacheActive() }
+
 func (cache *Cache) SetEntry(ctx context.Context, cacheSet *pairingtypes.RelayCacheSet) error {
 	if cache == nil {
 		return NotInitializedError
--- a/protocol/performance/cache.go
+++ b/protocol/performance/cache.go
@@ -11,6 +11,7 @@ import (
 	"github.com/magma-Devs/smart-router/utils"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/emptypb"
 )
@@ -36,6 +37,27 @@ func newRelayerCacheClientStore(ctx context.Context, address string) (*relayerCa
 		address: address,
 	}
 	return clientStore, clientStore.connectClient()
+}
+
+// peekConnState reports the live connectivity of the installed connection WITHOUT
+// touching it. It exists because getClient() is not a read: it spawns
+// `go reconnectClient()` whenever the client is nil, so an observer that called it
+// to ask "is the cache up?" would dial the cache as a side effect of asking.
+// /debug/cache-state is documented read-only and must not do that.
+//
+// The second return distinguishes "no connection installed" from a state reading.
+// grpc.ClientConn.GetState() is a plain accessor — unlike Connect() and
+// WaitForStateChange() it neither triggers nor waits on a transition.
+func (r *relayerCacheClientStore) peekConnState() (connectivity.State, bool) {
+	if r == nil {
+		return connectivity.Shutdown, false
+	}
+	r.lock.RLock()
+	defer r.lock.RUnlock()
+	if r.client == nil || r.conn == nil {
+		return connectivity.Shutdown, false
+	}
+	return r.conn.GetState(), true
 }
 
 func (r *relayerCacheClientStore) getClient() pairingtypes.RelayerCacheClient {
@@ -174,8 +196,12 @@ func (r *relayerCacheClientStore) resetOnConnectionError(err error) {
 }
 
 type Cache struct {
+	// The configured address deliberately lives in exactly ONE place —
+	// clientStore.address, the string actually dialled — and is read back through
+	// BackendEndpoint. Cache used to carry a second copy that InitCache kept in sync
+	// by assigning the same argument to both; two accessors for one fact is how they
+	// eventually disagree.
 	clientStore *relayerCacheClientStore
-	address     string
 	serviceCtx  context.Context
 }
 
@@ -183,7 +209,6 @@ func InitCache(ctx context.Context, addr string) (*Cache, error) {
 	clientStore, err := newRelayerCacheClientStore(ctx, addr)
 	cache := &Cache{
 		clientStore: clientStore,
-		address:     addr,
 		serviceCtx:  ctx,
 	}
 	if err != nil {
@@ -216,14 +241,66 @@ func (cache *Cache) CacheActive() bool {
 	return cache != nil && cache.clientStore.getClient() != nil
 }
 
-func (cache *Cache) CacheEngine() string { return "grpc" }
-func (cache *Cache) CacheAddress() string {
-	if cache == nil {
-		return ""
+// DebugCacheState reports this tier for GET /debug/cache-state. Read-only: it
+// reads the connection's state rather than asking for a client, so observing a
+// down cache does not dial it.
+//
+// Reachability comes from the gRPC connectivity state, NOT from "a client pointer
+// is installed" (what CacheActive answers). The installed pointer is a latch: it is
+// cleared only by resetOnConnectionError, which returns early for every status code
+// except Unavailable and fires only from a relay's own GetEntry/SetEntry/Flush. So
+// on an idle router a cache-be killed an hour ago still reads as active, while the
+// connection has long since left Ready.
+//
+// The remaining gap is deliberate and documented rather than papered over: a cache-be
+// that ACCEPTS connections but hangs keeps its transport in Ready, so this reports
+// reachable while every operation times out. That is the honest reading of a
+// connectivity state — the backend is reachable and unresponsive, which are different
+// faults — and distinguishing them needs an active probe this read-only endpoint
+// must not perform.
+func (cache *Cache) DebugCacheState() DebugCacheState {
+	// A typed-nil *Cache is how an unconfigured cache travels inside the
+	// CacheBackend interface (see the wiring convention on CacheBackend), and a
+	// typed nil still satisfies this interface — so "configured" must be decided
+	// here, on the receiver, and never by the caller's `!= nil` on the interface.
+	//
+	// Both the configured test and the reported value read BackendEndpoint, so
+	// there is exactly one source for the address.
+	address := cache.BackendEndpoint()
+	if address == "" {
+		return DebugCacheState{Engine: CacheEngineGRPC}
 	}
-	return cache.address
+	state := DebugCacheState{
+		Configured: true,
+		Engine:     CacheEngineGRPC,
+		Address:    address,
+		// A gRPC tier that is not reachable is SKIPPED, not attempted: GetEntry
+		// returns NotConnectedError before any I/O, so an unreachable primary costs
+		// nothing per relay. The RESP tier does the opposite, which is why this is
+		// reported rather than assumed.
+		WhenUnreachable: CacheWhenUnreachableSkipped,
+	}
+	connState, connected := cache.clientStore.peekConnState()
+	if !connected {
+		state.Reachable = boolPtr(false)
+		state.Detail = "no connection installed"
+		return state
+	}
+	state.Detail = connState.String()
+	switch connState {
+	case connectivity.Ready, connectivity.Idle:
+		// Idle is a healthy connection with no active transport (gRPC drops one
+		// after an idle period); it dials on the next call. Reporting it as
+		// unreachable would make a quiet router look broken.
+		state.Reachable = boolPtr(true)
+	case connectivity.Connecting:
+		// Genuinely undetermined — the dial is in flight. Left nil rather than
+		// guessed either way.
+	default: // TransientFailure, Shutdown
+		state.Reachable = boolPtr(false)
+	}
+	return state
 }
-func (cache *Cache) CacheReachable() bool { return cache.CacheActive() }
 
 func (cache *Cache) SetEntry(ctx context.Context, cacheSet *pairingtypes.RelayCacheSet) error {
 	if cache == nil {
--- a/protocol/performance/cache_backend.go
+++ b/protocol/performance/cache_backend.go
@@ -2,6 +2,7 @@ package performance
 
 import (
 	"context"
+	"time"
 
 	pairingtypes "github.com/magma-Devs/smart-router/types/relay"
 )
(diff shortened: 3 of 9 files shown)


AnnaR-prog (commented):
### Review of `/debug/cache-state`

15 findings — 14 as inline comments below, plus the merge blocker at the end (its target file isn't in this diff).

**Two of them are reproduced, not inferred.** The handler was built in a test and the response read:

1. **A router with no cache reports that it has one.** `resp_config.go:113` hands back `(*Cache)(nil)` — a typed nil inside a non-nil interface still satisfies `DebugCacheStateReporter`, so the primary branch is taken and the endpoint answers `{"engine":"grpc","tiers":{"primary":{"configured":true,"reachable":false}}}` — byte-identical to a configured-but-down `cache-be` except for the missing `address` key. That is the **default** deployment, not an edge case: `docs/RESP-CACHE.md` says "Neither configured → the default cache, exactly as before", and 8 of 26 `pre_setups` lanes run that way.
2. **`lifetimes` is always `0`/`0`.** The expiration flags are registered on the cache-server cobra command only (`ecosystem/cache/command.go:70-71`), never on the smartrouter command, so both `viper.GetDuration` calls return zero on every deployment.

**The strongest novel finding is #3: `GET /debug/cache-state` is not read-only.** `CacheReachable() → CacheActive() → getClient()` reaches `if r.client == nil && !r.closed.Load() { go r.reconnectClient() }`, so observing a down cache *dials* it — a 3s blocking gRPC dial retried every 5s. Every sibling handler in this file documents the opposite ("Read-only; nil-router safe"). Four independent angles found it.

A single `buildDebugMux` table test catches both headline bugs in under a second — which is also finding #14: the endpoint, the interface and six new exported methods ship with no tests, and the PR's recorded verification is `go test -run ^$`, i.e. compile-only.

---

**15. Merge blocker — not a code defect.** The PR is currently RED on a required check: `jira-ticket/validated fail "Missing Jira ticket line"`, because the body uses a `## Jira` heading instead of the mandated single line.

VERIFIED LIVE via `gh pr checks 376`. The template states verbatim: "Add exactly one line using the format `Jira ticket: MAG-123`. The pull request cannot be merged unless the ticket exists in Jira." require-jira-ticket.yml:192 selects candidates with `test("^\\s*jira\\s+ticket\\s*:"; "i")` and fails at :199 when the count is 0; the body's `## Jira` + bare `MAG-3478` matches zero times. The ticket exists — this is purely a body-format violation blocking the merge.

---

Two more the sweep surfaced, below the cut:

- **`configuredAddresses` sits inside a doc-comment that argues against it** (`store.go:67`). The field joins the `readEndpoint`/`writeEndpoint` group whose comment says *"Static configuration cannot answer this: under sentinel the serving node changes on every failover"* — and the new field holds exactly that static configuration. godoc now attributes that paragraph to it. Give it its own comment, and state the `New()`-only invariant there (finding #8).
- **First `/debug` route to expose the router's own backing store.** The debug listener has no authentication, and under `topology: sentinel` these are sentinel control-plane addresses. No credential travels in `Addresses` (`Password`/`PasswordFile` are separate fields — confirmed in `Config.Validate`), so this is a threat-model call rather than a leak, but it wants an explicit decision and a documented "bind to loopback" warning.

<sub>Automated review via Claude Code.</sub>

protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -1045,6 +1056,33 @@ func buildDebugMux(deps debugMuxDeps) *http.ServeMux {
...
+		}
+		type tier struct {
+			Configured bool   `json:"configured"`
+			Reachable  *bool  `json:"reachable,omitempty"`
+			Address    string `json:"address,omitempty"`
+		}
+		state := deps.cacheState
+		engine := "none"
+		primary := tier{}
+		if reporter, ok := state.primary.(performance.DebugCacheStateReporter); ok {
```
AnnaR-prog: **1. The primary-tier type assertion succeeds for the typed-nil *Cache that SelectCacheBackend hands back when no cache is configured, so a router with no cache at all reports engine "grpc" and primary.configured=true.**

REPRODUCED. resp_config.go:113 sets `var cache CacheBackend = (*Cache)(nil)`; a typed-nil pointer in a non-nil interface still satisfies DebugCacheStateReporter, so line 1072 takes the branch. GET /debug/cache-state returns {"engine":"grpc","tiers":{"primary":{"configured":true,"reachable":false},"secondary":{"configured":false}}} — byte-identical to a configured-but-down cache-be except for the missing address key. cache_backend.go's own wiring convention says "Call sites holding a possibly zero-valued field must still nil-check the interface itself before probing". Discriminate on reporter.CacheAddress() != "", not CacheActive() (which would collapse configured into reachable).

protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -411,6 +412,8 @@ func (rpsr *RPCSmartRouter) Start(ctx context.Context, options *rpcSmartRouterSt
 			router:     rpsr,
 			cache:      options.cache,
 			qosClient:  smartRouterOptimizerQoSClient,
+			cacheState: cacheStateInfo{primary: options.cache, secondary: options.secondaryCache,
+				finalized: viper.GetDuration(cachepkg.ExpirationFlagName), nonFinalized: viper.GetDuration(cachepkg.ExpirationNonFinalizedFlagName)},
```
AnnaR-prog: **2. lifetimes is always 0/0: the expiration flags are registered only on the cache-server cobra command, never on the smartrouter command, so both viper.GetDuration calls return zero.**

REPRODUCED. `smartrouterd smartrouter --help` has no `expiration` flag; the flags live at ecosystem/cache/command.go:70-71 on cacheCmd only, and the router binds just its own (viper.BindPFlags(cmd.Flags()), line 3045). A non-zero value would still be a lie: resp_config.go:110 hardcodes core.DefaultPolicy() (1h/500ms) with the comment "router-side TTL tuning is a deliberate non-goal", and Policy.ForChain = max(avgBlockTime/8, NonFinalized) so the scalar is a floor, not the effective TTL. For cache-be the TTLs live in a different pod. Read them from the backend that owns the policy (RespCache holds engine.Policy) or report "unknown".

protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -1045,6 +1056,33 @@ func buildDebugMux(deps debugMuxDeps) *http.ServeMux {
...
+			Configured bool   `json:"configured"`
+			Reachable  *bool  `json:"reachable,omitempty"`
+			Address    string `json:"address,omitempty"`
+		}
+		state := deps.cacheState
+		engine := "none"
+		primary := tier{}
+		if reporter, ok := state.primary.(performance.DebugCacheStateReporter); ok {
+			engine = reporter.CacheEngine()
+			reachable := reporter.CacheReachable()
```
AnnaR-prog: **3. The advertised read-only GET mutates live connection state: CacheReachable() -> CacheActive() -> getClient() spawns `go r.reconnectClient()`.**

cache.go:48-50 reads `if r.client == nil && !r.closed.Load() { go r.reconnectClient() }`. With cache-be down, every GET — line 1074 for the primary and 1079 for the secondary — kicks a reconnect loop doing a 3s blocking grpc dial (WithBlock) retried every 5s, logging at Info. Every sibling read-only handler is documented as the opposite ("Read-only; nil-router safe", lines 1468/1569/1608). A monitoring scrape changes the state it measures: if the pod recovers between polls, the second reports reachable:true as a consequence of the first poll's dial. The secondary case reaches past the deliberately read-only performance.CacheReader seam whose doc says the guarantee is "structural, not conventional".

protocol/performance/cache.go
```diff
@@ -216,6 +216,15 @@ func (cache *Cache) CacheActive() bool {
...
 }
 
+func (cache *Cache) CacheEngine() string { return "grpc" }
+func (cache *Cache) CacheAddress() string {
+	if cache == nil {
+		return ""
+	}
+	return cache.address
+}
+func (cache *Cache) CacheReachable() bool { return cache.CacheActive() }
```
AnnaR-prog: **4. For the gRPC engine `reachable` is a latch, not a liveness signal — nothing ever re-probes it, and it is cleared only on codes.Unavailable.**

CacheReachable() == CacheActive() == "a *grpc.ClientConn is still installed". The only thing that clears it is resetOnConnectionError (cache.go:161-174), which returns early for every status code except Unavailable and fires only from a relay's GetEntry/SetEntry/Flush. So (a) kill cache-be on an idle router and the endpoint answers reachable:true indefinitely for a dead backend; (b) a cache-be that hangs returns DeadlineExceeded (bounded by --cache-timeout), which is not Unavailable, so the client stays installed and reachable stays true while 100% of cache ops fail. The RESP tier is the only one with a real probe.

protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -1045,6 +1056,33 @@ func buildDebugMux(deps debugMuxDeps) *http.ServeMux {
...
+			primary = tier{Configured: true, Reachable: &reachable, Address: reporter.CacheAddress()}
```
AnnaR-prog: **5. `reachable` carries opposite operational meanings for the two engines, with nothing in the payload marking the difference.**

gRPC: reachable:false means the tier is SKIPPED — GetEntry returns NotConnectedError at cache.go:201 before any I/O. RESP: RespCache.CacheActive() is `return cache != nil` unconditionally (resp_cache.go:179), so reachable:false means the router STILL issues every lookup and populator write against the dead backend, paying the full --cache-timeout per relay. Same JSON key, opposite realities. RESP reachable is also a snapshot up to ~13s stale (10s cadence + 3s ping timeout) with no freshness stamp, and the auth-vs-network classification the file preserves via classifyProbeError ("'unreachable' sends an operator to check networking when the real fault is a credential") is collapsed into one bit.

protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -1045,6 +1056,33 @@ func buildDebugMux(deps debugMuxDeps) *http.ServeMux {
...
+		}
+		secondary := tier{}
+		if reporter, ok := state.secondary.(performance.DebugCacheStateReporter); ok {
+			reachable := reporter.CacheReachable()
+			secondary = tier{Configured: true, Reachable: &reachable, Address: reporter.CacheAddress()}
+		}
+		resp := map[string]any{"engine": engine, "tiers": map[string]tier{"primary": primary, "secondary": secondary}, "lifetimes": map[string]float64{"finalized_seconds": state.finalized.Seconds(), "non_finalized_seconds": state.nonFinalized.Seconds()}}
```
AnnaR-prog: **6. `engine` is a single top-level scalar taken only from the primary; the tier struct has no engine field, and the secondary reporter's CacheEngine() is computed then silently discarded.**

The secondary is ALWAYS a gRPC *Cache — performance.InitCache at line 3312 is its only constructor — while the primary becomes *RespCache whenever the resp-cache block is set, and lines 3295-3298 explicitly anticipate that pairing. docs/SECONDARY-CACHE.md says "pairing *different* engines across the two tiers is not a supported configuration", yet SecondaryCacheConfig.Validate never checks for it. The endpoint reports one flat engine over a two-engine deployment that nothing validates — concealing the most valuable diagnosis it could offer. Since engine is also what tells a reader how to interpret reachable, dropping it for the secondary loses the disambiguator on the tier that most needs it. Move engine into tier.

ecosystem/cache/redisstore/store.go
```diff
@@ -152,6 +153,7 @@ func New(cfg Config) (*Store, error) {
 		}
 		return nil, err
 	}
+	store.configuredAddresses = strings.Join(cfg.Addresses, ",")
```
AnnaR-prog: **7. configuredAddresses joins cfg.Addresses only, dropping cfg.ReadAddresses — while Store.Ping probes BOTH clients, so a dead reader flips reachable to false with only the healthy writer printed.**

Multi-region deployment with `read-addresses: [reader.eu-west-1:6379]` (docs/RESP-CACHE.md: "Reads go to read-addresses, writes ... to addresses"). Store.Ping pings write and read and returns the first error, so killing the reader gives reachable:false with address=redis-primary:6379. The operator curls the printed address, finds it healthy, and concludes the endpoint is broken — while every cache LOOKUP fails against an address the payload never names. Same omission drops KeyPrefix, which decides which keyspace this router uses: two routers on one Valkey with different prefixes share zero entries and return byte-identical cache-state output.

ecosystem/cache/redisstore/store.go
```diff
@@ -174,6 +176,14 @@ func New(cfg Config) (*Store, error) {
 	return store, nil
 }
 
+// ConfiguredAddresses returns the operator-provided primary RESP addresses.
+func (s *Store) ConfiguredAddresses() string {
```
AnnaR-prog: **8. configuredAddresses is set only in New(); newStore() never sets it, so the exported NewWithClient/NewWithClients constructors produce a Store whose ConfiguredAddresses() is permanently empty.**

newStore (store.go:220-232) initialises readEndpoint/writeEndpoint but not configuredAddresses, and both wrapper constructors route straight to it. Combined with `json:"address,omitempty"` the key vanishes and a configured, healthy tier renders as {"configured":true,"reachable":true}. These two constructors are the only seam every RESP test uses to inject a miniredis client, so whoever writes the missing endpoint test gets an empty address and either asserts the wrong shape or needs a live Redis — which is how the address path ends up untestable.

<sub>Anchored here because the code it names (`ecosystem/cache/redisstore/store.go:211`) is outside this PR's diff.</sub>

protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -1045,6 +1056,33 @@ func buildDebugMux(deps debugMuxDeps) *http.ServeMux {
...
 	mux := http.NewServeMux()
+	mux.HandleFunc("/debug/cache-state", func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet {
+			http.Error(w, "GET only", http.StatusMethodNotAllowed)
+			return
+		}
+		type tier struct {
+			Configured bool   `json:"configured"`
+			Reachable  *bool  `json:"reachable,omitempty"`
+			Address    string `json:"address,omitempty"`
```
AnnaR-prog: **9. `Address string json:"address,omitempty"` makes an UNKNOWN address disappear rather than reporting it empty, so "cannot name the backend" is indistinguishable from "field does not apply".**

Two live paths produce configured:true with no address key: a *RespCache built through NewWithClient/NewWithClients, and the typed-nil *Cache of finding 1 whose CacheAddress() returns "" by its nil guard. Both render as {"configured":true,"reachable":false}, which reads like the unconfigured tier's {"configured":false} plus a stray flag. Because address is the ONLY field distinguishing finding 1's phantom gRPC tier from a real one, omitempty removes the one signal a consumer could key on. Drop omitempty, as Configured correctly does.

protocol/performance/resp_cache.go
```diff
@@ -176,6 +179,17 @@ func (cache *RespCache) CacheActive() bool {
...
 }
 
+func (cache *RespCache) CacheEngine() string { return "resp" }
+func (cache *RespCache) CacheAddress() string {
+	if cache == nil || cache.store == nil {
+		return ""
+	}
+	return cache.store.ConfiguredAddresses()
+}
+func (cache *RespCache) CacheReachable() bool {
```
AnnaR-prog: **10. RespCache.Close() stops the health loop and shuts the store's clients but never resets `reachable`, so the accessor keeps returning true for a closed, dead backend.**

Close() (resp_cache.go:240-250) closes healthStop and calls store.Close(); nothing writes reachable.Store(false). Shutdown ordering makes it observable: RPCSmartRouter.Stop() closes the cache backend at rpcsmartrouter.go:482 while the debug server is torn down independently by the ctx watcher at line 421, with no ordering between them. A GET in that window reports reachable:true for a cache whose connections are closed. The gRPC tier does the opposite (Close() sets client=nil so CacheActive() flips false), so the two engines disagree even about what a closed cache looks like.

protocol/performance/resp_cache.go
```diff
@@ -33,6 +34,7 @@ type RespCache struct {
 
 	healthStop chan struct{}
 	closeOnce  sync.Once
+	reachable  atomic.Bool
```
AnnaR-prog: **11. `reachable atomic.Bool` zero-values to false and is written only by the health-loop goroutine, so a healthy RESP backend reports reachable:false until the first probe returns.**

NewRespCache returns immediately; healthLoop's first probe() is a PING bounded by respCachePingTimeout (3s) plus scheduling. Any integration test or readiness script that starts the router with --debug-address and polls promptly sees reachable:false for a healthy Redis and flakes or reports a false outage. The wire format already has the three-state slot for this — `Reachable *bool`, nil meaning unknown, used for the unconfigured tier — but "not probed yet" is collapsed into "unreachable" instead of nil.

protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -513,7 +516,15 @@ type debugMuxDeps struct {
 	// lie on any deployment running with --cache-be (MAG-1764). Reached
 	// through cacheFlusher rather than the concrete *performance.Cache so
 	// tests can inject a fake without standing up a gRPC client.
-	cache cacheFlusher
+	cache      cacheFlusher
```
AnnaR-prog: **12. options.cache is now stored TWICE in debugMuxDeps — as `cache cacheFlusher` and again as `cacheState.primary` — and every existing test fixture populates only the first.**

Both are assigned options.cache at lines 413/415, so production cannot diverge, but nothing enforces it. debug_server_test.go:687/704/724/743/767 build debugMuxDeps{cache: &fakeCacheFlusher{active:true}} with cacheState zero: on that one mux, POST /debug/reset-all flushes a live cache while GET /debug/cache-state answers engine:"none", configured:false — two endpoints contradicting each other. It also undercuts the comment above line 519 (cacheFlusher exists "so tests can inject a fake without standing up a gRPC client"): a fake serving both handlers must now implement CacheBackend's 5 methods plus DebugCacheStateReporter's 3. CacheBackend is a strict superset of cacheFlusher, so the handler can assert on deps.cache and the primary field can go away.

protocol/performance/cache_backend.go
```diff
@@ -34,6 +34,15 @@ type BackendEndpointReporter interface {
 	BackendEndpoint() string
 }
 
+// DebugCacheStateReporter exposes the runtime facts needed by the debug cache
```
AnnaR-prog, line 107: **13. cache_backend.go adds DebugCacheStateReporter with no `var _` compile-time assertions, so the whole endpoint hangs on a runtime type assertion that can stop matching silently; and *Cache now has two accessors for one fact reading two DIFFERENT fields.**

The new interface is declared directly beside the existing `var _ BackendEndpointReporter = (*Cache)(nil)` / `(*RespCache)(nil)` block but adds no assertions of its own. Rename any of the three methods on either backend and it is not a build failure — the comma-ok assertion stops matching and the endpoint silently degrades to {"engine":"none","primary":{"configured":false}} on a router that is caching fine; neither go build nor go vet catches it. Separately, CacheAddress() returns cache.address while the pre-existing BackendEndpoint() returns cache.clientStore.address; they are only coincidentally equal because InitCache assigns the same addr to both (cache.go:181-190).

protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -1045,6 +1056,33 @@ func buildDebugMux(deps debugMuxDeps) *http.ServeMux {
 	const maxDebugOffsetSeconds = float64(24 * 3600) // 86400 s
 
 	mux := http.NewServeMux()
+	mux.HandleFunc("/debug/cache-state", func(w http.ResponseWriter, r *http.Request) {
```
AnnaR-prog, line 1187: **14. The new handler, the new interface and six new exported methods ship with zero tests, against CODING_GUIDELINES.md's explicit requirement — and the PR's own verification was compile-only.**

CODING_GUIDELINES.md:111 "Provide unit tests for every unit of your code if possible" and :115 "Do not decrease code test coverage. Explain in a PR if test coverage is decreased." The PR body records only `go test ... -run ^$`, which runs nothing. Every prior /debug endpoint landed with a dedicated test file. This is the root cause of findings 1 and 2 — a single buildDebugMux table test reproduces both. The response shape compounds it: unlike the sibling /debug/runtime-config (package-level routerConfigResponse + SchemaVersion:1 + a wire-key contract test), this handler declares `type tier` INSIDE the closure and emits map[string]any, so no consumer can unmarshal into a shared type; the encode also drops its error (`_ =`) where runtime-config marshals to a buffer first and returns 500 "rather than a half-written body".

protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -1045,6 +1056,33 @@ func buildDebugMux(deps debugMuxDeps) *http.ServeMux {
...
+		type tier struct {
+			Configured bool   `json:"configured"`
+			Reachable  *bool  `json:"reachable,omitempty"`
+			Address    string `json:"address,omitempty"`
+		}
+		state := deps.cacheState
+		engine := "none"
+		primary := tier{}
+		if reporter, ok := state.primary.(performance.DebugCacheStateReporter); ok {
+			engine = reporter.CacheEngine()
```
VicSheCodes: `engine` is read from the primary only, so a secondary-only router reports `engine: "none"` while its secondary is configured and serving reads.

That topology is documented as valid rather than exotic. `docs/deployment/cache/secondary.md` says the secondary "is valid with no primary configured at all: reads work, nothing backfills, and the router logs an advisory warning for that topology", and `secondary_config.go` has a dedicated branch that warns and starts for exactly that shape.

So on such a router this endpoint reports no engine while one is in use. A test reading it would conclude no cache is in play.

Two ways to close it, both small:

1. Fall back to the secondary's reporter when there is no primary. This matches the documented rule that both tiers run the same engine, so one field stays correct.
2. Move `engine` inside each tier.

I would take the first, because per-tier engines are explicitly not a supported configuration.

Context on why we care: a secondary-only router is one of the shapes the secondary-cache test topology may deploy (MAG-2659), and `engine` is the field we would use to state which backend a result was measured against.

Everything else here does exactly what MAG-3478 asked for. Reading `engine` from the reporter that is actually serving, rather than from the configuration, is the part that mattered most and it is right. Making `reachable` a pointer so it is omitted when a tier is unconfigured is also right, because it lets a test tell "not configured" from "unreachable".

protocol/rpcsmartrouter/rpcsmartrouter.go
avitenzer: Fixed. Reproduced exactly as described, and it is the default deployment.

`Configured` is now decided by the backend on its own receiver — `Cache.DebugCacheState()` returns `{Engine: grpc}` with `Configured: false` when the receiver is nil or has no address — so the caller never gets to infer it from the assertion succeeding. Discriminating on the address, as you suggested, rather than on `CacheActive()`.

Test: `TestCacheStateTypedNilCacheIsNotConfigured`, plus `TestDebugCacheStateTypedNilReportsUnconfigured` in the performance package, which asserts the trap itself (`require.True(t, ok, "a typed nil still satisfies the interface")`) so the reason the guard exists is on the record.

protocol/rpcsmartrouter/rpcsmartrouter.go
avitenzer: Fixed, and taken the way you framed it: read them from the backend that owns the policy, report nothing where none is owned.

`viper` is gone from this path entirely. `RespCache` reports `Finalized`/`NonFinalized`/`NodeErrors` off `engine.Policy`; the gRPC tier returns `nil`, which renders `"lifetimes": null`. A `cache-be` tier's TTLs are configured in and applied by another pod, so `null` is the honest answer where `0` asserted a value no deployment uses.

Also documented in the struct that these are the policy's base values and that `non_finalized_seconds` is a floor, since `ForChain` takes `max(avgBlockTime/8, NonFinalized)`.

Test: `TestCacheStateLifetimesComeFromTheBackend`, including `require.NotContains(body, "finalized_seconds":0)`.

protocol/rpcsmartrouter/rpcsmartrouter.go
avitenzer: Fixed. This was the best finding in the review — the endpoint was advertised read-only and was not.

Reachability no longer goes near `getClient()`. New `peekConnState()` takes the read lock, returns `conn.GetState()`, and spawns nothing; `GetState()` is a plain accessor, unlike `Connect()`/`WaitForStateChange()`. Both backends now answer through `DebugCacheState()`, whose interface doc states the no-mutation requirement as a contract rather than a habit.

Test: `TestDebugCacheStateDoesNotTriggerReconnect` asserts with `require.Never` over 200ms — a bare `False` would pass just by beating the goroutine — and then calls `getClient()` on the same store and requires the loop DOES start, so the guard has a positive control and cannot pass vacuously.

protocol/performance/cache.go
avitenzer: Fixed by the same change as #3, which is what made it worth doing that way rather than special-casing the handler.

`reachable` is now the connectivity state, not "a client pointer is installed": `Ready`/`Idle` → true, `TransientFailure`/`Shutdown` → false, `Connecting` → `null` (undetermined, not guessed). Your case (a) is closed — the connection leaves `Ready` on its own, with no relay needed to notice.

Case (b) is not, and I have documented it rather than claimed otherwise: a backend that accepts connections and then hangs keeps its transport in `Ready`, so this still reports reachable while operations time out. I think that is the honest reading of a connectivity state — reachable and unresponsive are different faults — and separating them needs an active probe, which is exactly what #3 says this endpoint must not do. The comment on `DebugCacheState` says so explicitly so the next reader does not think it was missed.

protocol/rpcsmartrouter/rpcsmartrouter.go
avitenzer: Fixed, with the disambiguator you identified plus the consequence stated outright.

`engine` moved onto the tier (see #6), so a reader always has it for the tier they are looking at. On top of that each tier carries `when_unreachable`: `"skipped"` for gRPC (bypassed before any I/O, costs nothing) and `"attempted"` for RESP (`CacheActive()` is unconditionally true, so every lookup still runs and pays the full timeout). Same flag, opposite bill, now said in the payload instead of inferred from the engine name.

Staleness and the auth/network collapse are both fixed too: `reachable_checked_at` stamps the RESP snapshot (empty for the gRPC tier, which is read live), and `reachable_detail` carries the classification. It renders through the existing `safeProbeDetail`, so an auth rejection reads "backend rejected the configured credentials" and no credential can reach the payload — the same rule the logs already follow.

Test: `TestCacheStateReportsWhatUnreachableCosts`, and the timestamp/detail cases in `TestCacheStateWireContract`.

protocol/rpcsmartrouter/rpcsmartrouter.go
avitenzer: Fixed — `engine` is now on the tier, as you asked.

I kept a top-level `engine` as well, because @VicSheCodes raised the complementary case: a secondary-only router (documented as valid, with a dedicated branch in `secondary_config.go`) reported `engine: "none"` while its secondary served every read. It now falls back to the secondary when no primary is configured, and `"none"` means neither tier is configured. Its doc comment points readers at the per-tier field for exactly the case it cannot express.

So both hold: the flat field stays useful for the common single-engine deployment, and the mixed deployment nothing validates is now readable rather than averaged away.

Test: `TestCacheStateEngineFallsBackToSecondary`, including `a mixed-engine deployment is visible per tier`.

ecosystem/cache/redisstore/store.go
avitenzer: Fixed. Both omissions, and the multi-region scenario you described is what the test is written against.

The field is now a `storeEndpoints{Addresses, ReadAddresses, KeyPrefix}`, rendered as `redis-primary:6379 read=reader.eu-west-1:6379 prefix=sr`. Read addresses are labelled rather than concatenated, so the two roles are never mistaken for a list of interchangeable nodes — which was the trap: `Ping` returns the first error from either client, so the dead one has to be nameable.

Key prefix included for the reason you gave; the doc comment records it, since two routers on one Valkey with different prefixes are otherwise indistinguishable in this output.

ecosystem/cache/redisstore/store.go
avitenzer: Fixed at the root rather than by documenting the gap.

`newStore` now derives addresses from the client itself (`clientAddresses` type-switches over `*redis.Client` / `*redis.ClusterClient` / `*redis.Ring`), so `NewWithClient`/`NewWithClients` produce a real address — miniredis included. `New()` still overwrites with the operator's `Config`, which is authoritative and additionally knows the read addresses.

That was the point worth taking: leaving it `New()`-only is what would have made the address path untestable without a live Redis, so the missing test would have been written around the bug.

Also gave the field its own doc comment stating both invariants, per your note below the cut — it had joined the `readEndpoint`/`writeEndpoint` group whose comment argues against static configuration, and godoc was attributing that paragraph to it.

Test: `TestRespCacheDebugState` asserts the miniredis address and `prefix=srtest` survive the `NewWithClient` seam.

protocol/rpcsmartrouter/rpcsmartrouter.go
avitenzer: Fixed — `omitempty` is gone from every field on the tier, not just `address`.

The struct comment now states the rule, since it is the endpoint's whole purpose: telling states apart is what this route is for, and `omitempty` erases the difference between "the router cannot name this" and "the field does not apply". A configured tier the router cannot name is a real, distinct condition.

`reachable` stays a `*bool` but is likewise always present, rendering `null` for undetermined (see #11). "Is this tier configured" is answered by `configured`, never by a field's absence.

Test: `TestCacheStateWireContract/an_unnameable_address_is_reported_empty,_not_omitted` asserts `"address":""` literally in the body.

protocol/performance/resp_cache.go
avitenzer: Fixed. `Close()` now publishes the closed state before tearing the clients down, so the last probe cannot stand forever after the loop that would have corrected it has stopped.

The comment records the shutdown-ordering window you found — `RPCSmartRouter.Stop()` closing the cache while the debug server is torn down by its own ctx watcher, unordered — because that is the only reason this is observable rather than theoretical, and it is not obvious from the file.

Also called out that the gRPC tier already does the opposite on close, so the two engines now agree about what a closed cache looks like.

Test: `TestRespCacheCloseClearsReachability`.

protocol/performance/resp_cache.go
avitenzer: Fixed, using the three-state slot you pointed at.

`atomic.Bool` is replaced by `atomic.Pointer[respCacheHealth]`, where a nil pointer IS the "not probed yet" state — it cannot be confused with a false verdict the way a zero-valued bool could. `DebugCacheState` maps that to `Reachable: nil` → `"reachable": null`, with `reachable_detail: "no health probe has completed yet"`.

The snapshot also carries the probe timestamp and its detail in one atomic value, so the verdict, its age and its reason can never be read torn apart from each other.

Test: `TestRespCacheReachabilityIsUnknownBeforeFirstProbe` builds the cache without starting the health loop, so the pre-probe state is asserted deterministically instead of by racing it.

protocol/rpcsmartrouter/rpcsmartrouter.go
avitenzer: Fixed as you suggested: the handler asserts on `deps.cache` and the duplicate field is gone. `debugMuxDeps` now holds only `secondaryCache` extra, which genuinely has no other home.

You are right that `CacheBackend` is a strict superset of `cacheFlusher` — nothing had to change about the flush path, and the fake injection the `cacheFlusher` comment protects still works, since a fake only implements the reporter if the test needs cache-state.

Test: `TestCacheStateAndResetAllSeeTheSameCache` builds one mux and asserts both endpoints agree a cache exists, so the two can no longer be wired apart.

protocol/performance/cache_backend.go
avitenzer: Fixed, both halves.

`var _ DebugCacheStateReporter = (*Cache)(nil)` / `(*RespCache)(nil)` added to the existing block, with a comment saying what they prevent: the endpoint hangs on a comma-ok assertion, so a rename is not a build failure — it silently degrades to "no cache" on a router that is caching fine, and neither `go build` nor `go vet` catches it.

I also cut the surface the assertion has to match. The three accessors are now a single `DebugCacheState()` returning a struct, so there is one name to keep rather than six.

On the two accessors for one fact: `CacheAddress()` is gone and `Cache.address` with it. The configured address now lives only in `clientStore.address` and is read through `BackendEndpoint()` — including by the "is it configured" test, so there is exactly one source. You were right that they were only coincidentally equal.

protocol/rpcsmartrouter/rpcsmartrouter.go
avitenzer: Fixed. Two new test files, and you were right that one table test reproduces both headline bugs in under a second — writing it first would have caught them.

`protocol/rpcsmartrouter/debug_cache_state_test.go`: typed-nil primary and secondary (#1), engine fallback and mixed-engine (#6), `when_unreachable` (#5), three-state reachability (#11), lifetimes from the backend and `null` rather than `0` (#2), the read-only guarantee (#3), the wire-key contract, non-GET, and both endpoints seeing one cache (#12).

`protocol/performance/debug_cache_state_test.go`: the typed-nil trap at the interface, no reconnect on read with a positive control, pre-probe unknown, `Close()` clearing reachability (#10), and the address surviving the `NewWithClient` seam (#8).

On the response shape: `debugCacheStateResponse` is now a package-level type with `json` tags and `SchemaVersion: 1`, matching `routerConfigResponse` so a consumer can unmarshal into a shared type; the encode marshals to a buffer and returns 500 rather than writing half a body, for the reason `runtime-config` gives.

I ran these from a clean `git worktree` checkout of the head commit. Worth noting for anyone reading the checks: CI here does not run `go test` at all, so the local run is the evidence either way.

protocol/rpcsmartrouter/rpcsmartrouter.go
avitenzer: Both cases are covered now, and thank you for the topology context — it changed the design.

I took your option 1 *and* @AnnaR-prog's #6, because they close different halves and neither alone is enough:

- Top-level `engine` now falls back to the secondary when no primary is configured, so a secondary-only router names the engine that is actually serving instead of `"none"`. Exactly your reasoning: per-tier engines are not a supported configuration, so one field staying correct is the right default.
- Each tier ALSO carries its own `engine`, because the unsupported pairing is not prevented by anything — `SecondaryCacheConfig.Validate` does not check it, and the secondary is always a gRPC client while the primary becomes RESP whenever the `resp-cache` block is set. The flat field cannot express that, and it is the state an operator would most want this endpoint to show.

So `engine` remains the single field to read on a normal deployment, and the mixed one is visible rather than averaged away.

Test: `TestCacheStateEngineFallsBackToSecondary` — `secondary-only router names the secondary's engine`, `primary wins when both are configured`, and `a mixed-engine deployment is visible per tier`.

On the MAG-2659 point: `engine` is safe to use as "which backend was this measured against" for a secondary-only router now. If a case ever needs it for a mixed deployment, read `tiers.<tier>.engine` — that is what it is there for.

Separately, `reachable` changed shape slightly from what you praised. It is still a pointer, but it is now ALWAYS present, rendering `null` when reachability has not been determined (a RESP backend before its first probe, a connection mid-dial) — Anna's #11 showed that collapsing that into `false` reports a healthy backend as down to anything polling at startup. Your property is unaffected: "not configured" versus "unreachable" is still distinguishable, by `configured`, and `null` is now distinct from `false` as well.

VicSheCodes (approved):

PR #383, opened 2026-09-13
test(cache): lock the cached entry's field list, so a new field cannot ship unseen

VicSheCodes:
## Jira ticket

Jira ticket: MAG-3621

## Why

The cached entry a router serves from a second cache tier has six fields. Nothing in this repository checks that it still has six.

`types/relay/cache_test.go` already holds four tests, and all four check that an old cache and a new cache can still read each other's entries. Each names only the one or two fields it cares about, so a seventh field can be added tomorrow and all four stay green.

That matters because the entry crosses a zone boundary. A reader in another zone decides what to trust and what to drop, field by field — it keeps the body, cuts the stored headers to two, drops the writer's signature, and replaces that zone's view of the newest block with its own. A field added without anybody noticing is a field with no defence and no test.

## What changed

Two tests, in the file that already exists. Each reads the declared field list and compares it against a written one.

## Why reflection and not `json.Marshal` alone

Worth reading, because the first version of this was wrong and the mutation caught it.

Marshalling a populated value asks `encoding/json` the same question the cache asks it at run time. That is the right question, and it cannot answer this one. A field added later as `` `json:"x,omitempty"` `` is absent from any literal written today, so it marshals to nothing — and a marshal-based check stays green while the field ships.

That is measured, not argued: the first version of this test passed against exactly that mutation. Reflection reads the **declaration**, so it sees every field whatever its tag says.

Marshalling is kept as the cross-check that the declaration and `encoding/json` agree about the names.

## How to verify

```
go test ./types/relay/ -run FieldListIsLocked -v
```

Proved by mutation, five ways, each against a copy of the struct with one change:

| mutation | result |
|---|---|
| a seventh field added | fails, naming it |
| a seventh field with `omitempty` | fails — this is the one that defeated the first version |
| a field with no JSON tag | fails with its own message |
| a field tagged `json:"-"` | fails with its own message |
| an existing field renamed on the wire | fails |

The third and fourth get their own message because either one changes what crosses the zone boundary, and that should be a deliberate decision rather than a silent one.

## When it fails

The fix is one line in the test — and a message to whoever tests this cache from outside, because their coverage list is keyed on these names. The failure message says so.


avitenzer:
Reviewed against the current head (`c9f2775`, post-rebase) — **the test is correct.** I re-ran the five mutations in your table and all five reproduce. I also added four aimed at wrong verdicts rather than missed ones:

| mutation | result | |
|---|---|---|
| field removed | fails | correct |
| two fields reordered | passes | correct — JSON field order isn't wire-semantic |
| unexported field added | passes | correct — matches `encoding/json`'s own rule |
| duplicate `json` tag | fails | correct |

No false pass and no false failure at the scope it asserts. The reflection-over-marshalling argument holds up — I confirmed the `omitempty` mutation defeats a marshal-only check exactly as you describe. Both expected lists still match the declarations after the rebase.

### One thing worth fixing, on the path this test was built for

**The marshal cross-check's only reachable failure carries a message that says the opposite of what happened.** Because the reflection assertion runs first under `require` (so, `FailNow`), every tag-level disagreement dies at line 144. The only way to reach line 147 is: someone adds an `omitempty` field, correctly updates `expected` per the first failure, and doesn't populate the literal. They then get *"what encoding/json emits and what the struct declares disagree"* — but the declaration and `encoding/json` agree perfectly; the literal is just missing the field. Reproduced end to end.

Suggested: `"...disagree — if you just added a field, populate it in the literal above; an omitempty field left at zero vanishes from the marshalled output."`

Related, since the body says the fix is "one line in the test": that's true for a plain field (verified — updating `expected` alone passes), but the `omitempty` case needs two, and that's the case this PR is built around.

### A question rather than a request

`Metadata` isn't locked, and it's reachable from both locked structs (`cache.go:26`, `relay.go:219`). I added a third field to it and both tests stayed green. Since `filterForeignReplyMetadata` appends whole `Metadata` values (`sanitize.go:91`), a new field would cross the trust boundary attached to every allowlisted header — against a sanitizer whose documented job is dropping identity-bearing data. Same gap for `BlockHashToHeight`.

Two more `declaredWireNames` calls would close it. But MAG-3621 only asked for `CacheRelayReply` and `RelayReply`, so this is equally fine as a follow-up if you'd rather keep scope tight.

### Two latent nits, low priority

- `declaredWireNames(t, &CacheRelayReply{})` panics (`reflect: NumField of non-struct type`) rather than failing. No call site hits it today.
- An embedded struct is reported as *"carries no JSON name … Give it an explicit tag."* Fails safe, but suggests the wrong remedy — embedding flattens fields into the parent object rather than nesting them.

### Separately, and not for this PR to fix

Nothing in `.github/workflows/` runs `go test` — confirmed with a positive control: `go build`/`go mod download` appear 4×, `go test`/`make test`/`gotestsum` 0×, and `lint.yml` only type-checks test files via `tests: true` without executing them. `make test` exists at `Makefile:130` and nothing invokes it.

So this alarm won't fire on a PR that adds a seventh field — all checks would go green. Given the ticket's title is "so a new field fails the router's build," that gap seems worth its own ticket rather than being bolted onto a two-test PR.

avitenzer (approved):

PR #384, opened 2026-09-16
chore(scripts): normalize shebangs and group the cache lanes under pre_setups/cache/

avitenzer:
# Description

Housekeeping pass over `scripts/`. No Go code, no router behaviour change — the diff is entirely shell scripts.

## Jira ticket

Jira ticket: MAG-3705

---

## What changed

**1. Shebang normalization — 22 scripts**

`#!/bin/bash` → `#!/usr/bin/env bash`. Every script under `scripts/` that carries a shebang (27 of them) is now on the same one, so there's no mixed convention left to drift. Resolving `bash` through `PATH` picks up Homebrew bash 5.x on macOS instead of the 3.2 Apple still ships at `/bin/bash`, and it works on systems where bash isn't at `/bin/bash` at all.

To be straight about the scope: **this is not fixing a live break.** These scripts are deliberately written 3.2-compatible — see the comment at `scripts/pre_setups/init_smartrouter_cv_demo.sh:179` explaining that a `case` is used precisely *because* `declare -A` is unavailable on 3.2. Every changed script parses clean under both 3.2 and 5.3. Call it convention plus headroom.

(`scripts/useful_commands.sh` has no shebang and is left alone — it's a sourced function library, not an executable.)

**2. Cache lanes grouped — 11 scripts moved**

The RESP / Redis / Valkey / secondary-cache lane scripts move from `scripts/pre_setups/` into `scripts/pre_setups/cache/`, along with `test_resp_lane_ownership.sh`, the regression harness that covers the RESP lane. `pre_setups/` had grown to ~20 files where half were cache-backend variants of one lane, which buried the per-chain lanes.

The move invalidated **38 copy-paste paths the moved scripts print in their own usage headers** (each advertised `scripts/pre_setups/init_smartrouter_eth_*.sh`, which no longer exists), plus one cross-reference from `init_smartrouter_eth_redis_demo.sh` to the lavap lane. All 38 are repointed, so no moved script documents a path that isn't there.

**3. Repo-root resolution made depth-independent**

The move added a directory level, and every moved script derived the repo root by counting `../` levels from its own location:

```bash
PROJECT_ROOT=$(cd "${__dir}"/../.. && pwd)   # → <repo>/scripts, not <repo>
```

No script sets `-e` and `<repo>/scripts` exists, so the `cd` succeeded and the wrong value propagated silently into configs, specs, pidfiles and `make -C`. Rather than bump the count to `../../..`, the root now comes from git:

```bash
PROJECT_ROOT=$(git -C "$__dir" rev-parse --show-toplevel 2>/dev/null || true)
[[ -f "$PROJECT_ROOT/go.mod" ]] || { echo "ERROR: cannot locate the smart-router repo root from $__dir"; exit 1; }
```

Counting levels is what broke, and it would break again on the next move; this can't. The `go.mod` assertion is there because `rev-parse` returns empty outside a checkout, which would otherwise make every derived path absolute-from-`/`. Two follow-ons from the same root cause: `resp_cache` / `secondary_cache` took `LOGS_DIR` from their own `../../` count and now hang off `PROJECT_ROOT` like the other eight already did; and `secondary_cache` sourced `../useful_commands.sh` (broken by the move) and `../vars/variables.sh` (already dead — `scripts/vars/` isn't tracked). It calls nothing from either, so both lines are gone.

**4. Six obsolete test scripts deleted — 2,453 lines**

`test_chaintracker_chainstate_sim.sh`, `test_uc1_smartrouter_lava.sh`, `test_uc2_smartrouter_lava.sh`, `test_uc4_smartrouter_sim.sh`, `uc1_curls.sh`, `scripts/test_mag2062_e2e.sh`. Nothing in the repo invokes any of them.

Two scripts were in this list and have been kept, because "nothing in the repo invokes it" is the wrong test for a manually-run harness — a human running a script by hand leaves no reference anywhere in the tree:

- `test_resp_lane_ownership.sh` — a live regression harness for a destructive-action safety property: the RESP lane must never signal a process merely for being named `smartrouter`, nor force-remove a same-named container it can't prove it created. Moved to `cache/` beside the lane it covers.
- `test_smartrouter_sol.sh` — `8d0a03a` (#355) updated it on `main` *after this branch was cut*, to describe the new upgrade-on-any-path websocket behaviour. It's the manual verification lane for that fix. This was also the sole cause of the branch's merge conflict.

## Known debt this PR leaves behind

Deliberately out of scope to keep the diff scripts-only — four **non-script** files still document the pre-move cache paths, so those copy-paste commands are broken until repointed:

| File | Stale refs |
|---|---|
| `docs/RESP-CACHE.md` | 15 |
| `docs/SECONDARY-CACHE.md` | 4 |
| `docs/LOCAL-COMPOSE.md` | 1 |
| `docker/docker-compose.secondary-cache.yml` | 1 (comment) |

Also stale but harmless: `scripts/pre_setups/init_smartrouter_eth_cv_events.sh:58` reserves ports away from `test_uc4_smartrouter_sim.sh`, which this PR deletes — the comment now names a script that doesn't exist. Left as-is since the port reservation itself is still the operative information.

## Verification

**CI on this repo gates builds, lint, govulncheck and the Jira ticket — it never executes anything under `scripts/`. Green checks say nothing about this diff.** Local evidence, run against a clean `git worktree` checkout of the pushed commit rather than my working directory:

- **`scripts/pre_setups/cache/test_resp_lane_ownership.sh` → 8 passed, 0 failed**, docker container cases included. This is the check that matters: it executes the RESP lane end-to-end through its refusal path, so it exercises the whole `PROJECT_ROOT` → config → spec → pidfile chain. On the previous push of this branch it was 5 passed / 3 failed.
- `bash -n` clean on 28/28 scripts in the committed tree
- shebang audit: 27 shebangs, all `#!/usr/bin/env bash`, zero remaining `#!/bin/bash`
- zero `${__dir}/../` depth-counting left anywhere under `pre_setups/cache/`
- zero references to the old cache paths anywhere under `scripts/`; no `cache/cache/` double-application
- rebased onto `main` — `git merge-tree` reports no conflicts

Renames are recorded as renames (R097–R099), so the moves review as small diffs rather than delete+add.

A text-level audit (`bash -n` plus grepping for stale path strings) was what the first push relied on, and it cannot detect a `$__dir` depth change — which is the one failure mode a directory move actually introduces. Executing a lane is the check that catches it.

---

## Author Checklist

* [x] confirmed `!` in the type prefix if API or client breaking change — n/a, no API change
* [x] included the necessary unit and integration tests — no CI harness covers `scripts/`; `test_resp_lane_ownership.sh` is the manual regression check and passes 8/8
* [x] updated the relevant documentation or specification — in-script usage headers repointed; the four non-script doc files are called out as known debt above

---

## Review findings (2026-09-20)

Reviewed head [`fbcc593`](https://github.com/Magma-Devs/smart-router/commit/fbcc5932bd816b36440f4ed565a0c4324b9181ed).

### [P2] Update the documented commands together with the cache-script move

The moved scripts now resolve the repository root correctly, but the checked-in runbooks still invoke the deleted locations. For example, [the Redis walkthrough](https://github.com/Magma-Devs/smart-router/blob/fbcc5932bd816b36440f4ed565a0c4324b9181ed/docs/RESP-CACHE.md#L366) and [the secondary-cache demo](https://github.com/Magma-Devs/smart-router/blob/fbcc5932bd816b36440f4ed565a0c4324b9181ed/docs/SECONDARY-CACHE.md#L233) both fail before starting any validation. The old paths also occur in documented `--stop`, `--status`, and `--failover` commands.

Reproduced from a clean checkout of this head:

```text
bash scripts/pre_setups/init_smartrouter_eth_redis_demo.sh
# exit 127: No such file or directory
bash scripts/pre_setups/init_smartrouter_eth_secondary_cache.sh
# exit 127: No such file or directory
```

The PR's known-debt section already identifies the remaining references, but leaving them behind makes the shipped cache setup and operations instructions unusable as soon as the move lands. Please repoint the references in `docs/RESP-CACHE.md`, `docs/SECONDARY-CACHE.md`, `docs/LOCAL-COMPOSE.md`, and the Compose comment, or retain forwarding scripts at the old locations. Check every documented executable path against the resulting tree.

Validation: all 23 changed shell files passed `bash -n`; the moved ownership harness passed **8/8**, including its Docker-container checks. The earlier root-resolution and ownership-harness findings are fixed in this head.


AnnaR-prog (changes requested):
## Request changes — the directory move breaks all 10 moved scripts at runtime

The shebang normalization is fine and the deletions are mostly justified. The `cache/` move is not: every moved script derives the repo root from its own directory with a hardcoded `../..`, and none of them were re-depthed. Verified empirically against a clean worktree of `fe3bea0`.

### 1. `PROJECT_ROOT` is off by one level in all 10 moved scripts — blocking

Every moved script carries:

```bash
__dir=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
PROJECT_ROOT=$(cd "${__dir}"/../.. && pwd)
```

`__dir` is now `scripts/pre_setups/cache`, so `PROJECT_ROOT` resolves to `<repo>/scripts` instead of `<repo>`. No script sets `set -e`, and `<repo>/scripts` exists, so the `cd` succeeds and the wrong value propagates silently.

Lines: `lavap_cache.sh:23`, `redis_cluster_remote.sh:81`, `redis_demo.sh:49`, `redis_multiregion.sh:52`, `redis_oss_remote.sh:75`, `redis_sentinel.sh:63`, `resp_cache.sh:40`, `secondary_cache.sh:24`, `valkey_cluster_remote.sh:79`, `valkey_remote.sh:72`.

Everything downstream of it is now wrong:

| Consumer | Resolves to | Exists |
|---|---|---|
| `make -C "$PROJECT_ROOT" install` | `<repo>/scripts` | no Makefile |
| `--use-static-spec $PROJECT_ROOT/specs/ethereum.json` | `<repo>/scripts/specs/` | missing |
| `$PROJECT_ROOT/config/smartrouter_examples/*.yml` | `<repo>/scripts/config/` | missing |
| generated configs, pidfiles, `.pw` files | `<repo>/scripts/debugging/` | wrong location |

Running the lane in the PR tree, on free ports:

```
$ ROUTER_PORT=39461 METRICS_PORT=39462 VALKEY_PORT=39463 \
    bash scripts/pre_setups/cache/init_smartrouter_eth_resp_cache.sh
ERROR: missing config/smartrouter_examples/smartrouter_eth_resp_cache.yml
```

Failure mode differs per script and it's worth knowing which: `redis_demo.sh:436` guards `make -C "$PROJECT_ROOT" install || { echo "ERROR: make install failed"; exit 1; }` and dies clean. `resp_cache.sh:159` is unguarded with no `set -e`, so without the config check above it would limp past a failed `make` into a router launch pointed at a nonexistent spec.

**Fix:** `../..` → `../../..` on those 10 lines. Verified — after that single change `resp_cache.sh` finds its config, completes `make install`, and proceeds to start valkey.

### 2. `LOGS_DIR` — same class, 2 scripts

`resp_cache.sh:36` and `secondary_cache.sh:19`:

```bash
LOGS_DIR=${__dir}/../../debugging/logs   # → <repo>/scripts/debugging/logs
```

`mkdir -p` then creates it, so the PR tree grows a stray `scripts/debugging/logs`. (`.gitignore:25` `debugging/` is unanchored and does cover it, so nothing gets committed — just scratch state in the wrong place.) Needs `../../../debugging/logs`. Note `secondary_cache.sh:21` does `rm $LOGS_DIR/*.log` against that wrong directory.

### 3. `secondary_cache.sh:16` sources a path that no longer exists

```bash
source "$__dir"/../useful_commands.sh   # → scripts/pre_setups/useful_commands.sh
```

The file is at `scripts/useful_commands.sh`; all nine unmoved siblings use the same `../` and resolve correctly from `pre_setups/`. Should be `../../useful_commands.sh`.

In practice this one is noisy rather than fatal — no `set -e`, and the script calls none of the library's functions — so it prints an error on every run and continues. Worth fixing regardless.

Line 17 `. "${__dir}"/../vars/variables.sh` also fails, but `scripts/vars/` isn't tracked on `main` either, so that one predates this PR. Good opportunity to delete both lines if neither is load-bearing.

### 4. Don't delete `test_resp_lane_ownership.sh`

The other seven deletions look right. This one isn't obsolete — it's a live regression harness for a destructive-action safety property: the RESP lane must never signal a process just because it's named `smartrouter`, and must never force-remove a same-named container it can't prove it created. It's safe by construction (throwaway ports 39361-3, refuses before the build stage) and its invocation is documented in its own header.

On `main` it passes 8/8, docker case included. Run verbatim against this PR's tree it reports **5 passed, 3 failed** — `LANE="$__dir/init_smartrouter_eth_resp_cache.sh"` no longer resolves, so `bash "$LANE"` exits 127 and the refusal-message assertions fail.

So the PR breaks a passing test and deletes it in the same commit. "Nothing in the repo invokes any of them" is the wrong criterion for a manually-run harness. Please keep it, move it to `cache/` with the others, and repoint `LANE` — it's the only automated check that touches this lane.

### On the verification section

`bash -n` and grepping for old path strings are text-level checks; neither can detect a `$__dir` depth change, which is the one failure mode a directory move actually introduces. The PR body is candid that CI never executes `scripts/`, which makes that gap the whole ballgame here.

Cheapest durable fix so the next move can't reintroduce this: derive the root depth-independently, e.g. `PROJECT_ROOT=$(git -C "$__dir" rev-parse --show-toplevel)`, and keep the ownership harness as the smoke check.

### Verified good

- Shebangs: 25 in the tree, all `#!/usr/bin/env bash`, zero `#!/bin/bash` left. Matches the claim.
- Renames recorded as renames (R097-R099), modes preserved at 100755.
- The other seven deleted scripts are genuinely unreferenced; the one dangling mention (`init_smartrouter_eth_cv_events.sh:58`) is disclosed in the PR body.
- `redis_demo.sh:332` `stop_all_lanes` still resolves — all four lanes it drives moved together.
- No unmoved script references a moved one, by full path or bare filename.
- Known doc debt is disclosed. Minor: `docs/RESP-CACHE.md` carries stale refs on 15 lines, not the 11 stated.

avitenzer:
Thanks — all four confirmed and fixed in `fbcc593`. The branch is rebased onto `main` and the conflict is gone.

**1 + 2. `PROJECT_ROOT` and `LOGS_DIR`** — took the depth-independent route you suggested rather than bumping `../..` to `../../..`, since counting levels is the thing that broke and would break again on the next move:

```bash
PROJECT_ROOT=$(git -C "$__dir" rev-parse --show-toplevel 2>/dev/null || true)
[[ -f "$PROJECT_ROOT/go.mod" ]] || { echo "ERROR: cannot locate the smart-router repo root from $__dir"; exit 1; }
```

The `go.mod` assertion is load-bearing: `rev-parse` returns empty outside a checkout, which would silently make every derived path absolute-from-`/` — the same failure class in a new disguise. Applied to all 10 lanes plus the harness. `LOGS_DIR` now hangs off `PROJECT_ROOT` in `resp_cache` and `secondary_cache`, matching what the other eight already did; the `rm "$LOGS_DIR"/*.log` you flagged is quoted while I was there.

**3. `secondary_cache.sh:16-17`** — deleted both lines rather than repointing. Confirmed against all 18 functions in `useful_commands.sh`: the script calls none of them, and `scripts/vars/` isn't tracked, so neither line was doing anything.

**4. `test_resp_lane_ownership.sh`** — restored, moved into `cache/` next to the lane it covers, `LANE` resolves as a sibling again. You're right that "nothing in the repo invokes it" is the wrong criterion for a manually-run harness. **It now reports 8 passed, 0 failed on the committed tree, docker cases included** — matching your `main` baseline.

That criterion also caught one you didn't flag: `test_smartrouter_sol.sh` was likewise uninvoked, but `8d0a03a` (#355) updated it on `main` after this branch was cut, to describe the new upgrade-on-any-path websocket behaviour — it's the manual lane for that fix, and it was the sole cause of the merge conflict. Kept, shebang normalized with the rest. Net deletions are now six scripts / 2,453 lines.

**On the verification gap** — agreed, and the harness is now the smoke check you asked for. `bash -n` plus grepping for stale path strings are both text-level and structurally blind to a `$__dir` depth change, which is the one thing a directory move introduces. Executing a lane is what catches it; I've swapped the PR body's verification section to lead with the harness result rather than the static audit.

**Doc debt** — you're right, `docs/RESP-CACHE.md` is 15, not 11. `docs/SECONDARY-CACHE.md` was also off (4, not 5). Table corrected. Still out of scope here to keep the diff scripts-only; happy to fold the 21 refs in if you'd rather not leave them dangling.

AnnaR-prog (approved):

avitenzer:

## Review findings (2026-09-20)

Reviewed head [`fbcc593`](https://github.com/Magma-Devs/smart-router/commit/fbcc5932bd816b36440f4ed565a0c4324b9181ed).

### [P2] Update the documented commands together with the cache-script move

The moved scripts now resolve the repository root correctly, but the checked-in runbooks still invoke the deleted locations. For example, [the Redis walkthrough](https://github.com/Magma-Devs/smart-router/blob/fbcc5932bd816b36440f4ed565a0c4324b9181ed/docs/RESP-CACHE.md#L366) and [the secondary-cache demo](https://github.com/Magma-Devs/smart-router/blob/fbcc5932bd816b36440f4ed565a0c4324b9181ed/docs/SECONDARY-CACHE.md#L233) both fail before starting any validation. The old paths also occur in documented `--stop`, `--status`, and `--failover` commands.

Reproduced from a clean checkout of this head:

```text
bash scripts/pre_setups/init_smartrouter_eth_redis_demo.sh
# exit 127: No such file or directory
bash scripts/pre_setups/init_smartrouter_eth_secondary_cache.sh
# exit 127: No such file or directory
```

The PR's known-debt section already identifies the remaining references, but leaving them behind makes the shipped cache setup and operations instructions unusable as soon as the move lands. Please repoint the references in `docs/RESP-CACHE.md`, `docs/SECONDARY-CACHE.md`, `docs/LOCAL-COMPOSE.md`, and the Compose comment, or retain forwarding scripts at the old locations. Check every documented executable path against the resulting tree.

Validation: all 23 changed shell files passed `bash -n`; the moved ownership harness passed **8/8**, including its Docker-container checks. The earlier root-resolution and ownership-harness findings are fixed in this head.

PR #385, opened 2026-09-16
fix(chainlib): forward the client's Content-Type on REST bodies (MAG-2745)

AnnaR-prog:
## Ticket

[MAG-2745](https://magmadevs.atlassian.net/browse/MAG-2745) — the router overwrote every customer's `Content-Type`; decision on the ticket is to stop and proxy the client's header. Follows [MAG-2744](https://magmadevs.atlassian.net/browse/MAG-2744) (0a0d287), which only let a *spec* directive reach the wire.

## Why the client's header never survived

Two layers, and the ticket names neither:

1. `BaseChainParser.HandleHeaders` drops any header the spec does not declare. The listener collected the customer's `Content-Type`; it was gone before transport.
2. `sendRESTRelay` hardcodes `ContentType: "application/json"` as the default, which is then what reached the node.

Every chain whose node takes a non-JSON body needed a per-spec workaround (`pass_send` on CARDANO/STACKS/ALGORAND/TEZOS, `pass_override` on XLM); every chain without one silently sent JSON.

## The change

Rule: **whoever authors the body owns `Content-Type`.** REST forwards the client's body byte-for-byte, so `HandleHeaders` now forwards an undeclared `content-type` on REST collections. Deliberately narrow:

- **POST/PUT/PATCH only.** A GET has no body to describe.
- **REST only.** JSON-RPC/tendermint re-marshal the body they send, so the content-type describes the router's bytes, not the client's. gRPC's is transport-owned (MAG-2643).
- **Undeclared names only.** A spec that declares `content-type` keeps its exact say: `pass_send` forwards as before, `pass_ignore` still drops it, and `pass_override` / `pass_nullify` are appended after the client's headers and win on the wire (`DoHTTPRequest` writes the `application/json` default first, then applies forwarded headers in order). Stellar `/transactions` behaves exactly as after MAG-2744. Clients that send no `Content-Type` still get the JSON default.
- Request direction only; empty values are treated as absent.

Untouched on purpose: `SendRequest` in `direct_rpc_connection.go` (the line the ticket points at) is only used by the internal probes, which author their own JSON.

## Cache key

`hashCacheRequest` marshals the whole `RelayPrivateData`, metadata included. Against the real gate (`tryCacheWriteResolved`: not stateful, block resolvable), **280 of the 496 REST POST/PUT/PATCH APIs in the lava-specs catalog are cache-eligible, across 22 specs** — EOS, TRON, TON and Canton read through POST. Forwarding the header as-is would have put a client that sends `Content-Type: application/json` and one that sends nothing into separate lanes for the same body on all of them.

So the second commit masks client body headers in the hash the way it already masks the salt and the request id (restored on return). Safe because the header only says how to read bytes that are already in the key, and a value the node cannot read is a non-2xx answer the cache never writes. A test pins the key equal with and without the header, and still moved by any other forwarded header.

## Scope note for reviewers

The forwarded header travels in `RelayPrivateData.Metadata`, so it also crosses to Lava providers on the provider path (where the provider's own `HandleHeaders` decides again).

## Tests

- `base_chain_parser_headers_test.go`: the carve-out and each boundary (methods, interfaces, direction, other headers, `pass_send` once, `pass_ignore` drops, `pass_override` is the single entry and wins, `pass_nullify` deletes).
- `cache_hash_content_type_test.go`: key stability with/without the header, sensitivity control on another header, metadata handed back untouched.
- `rest_content_type_passthrough_test.go`: direct relays through the real parser and transport asserting the wire `Content-Type` — client value forwarded, default when absent, spec override wins, JSON-RPC unaffected.
- The forwarded cases fail on `main`; `go test ./protocol/chainlib/... ./protocol/lavasession/... ./protocol/rpcsmartrouter/...` and `go vet` pass on the branch.

[MAG-2745]: https://magmadevs.atlassian.net/browse/MAG-2745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MAG-2744]: https://magmadevs.atlassian.net/browse/MAG-2744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AnnaR-prog:
## Self-review: `/code-review high` on `1a73748`

Scope: full diff `main...HEAD`, the REST listener/transport that consumes the forwarded metadata, the cache-key and cache-write paths. `RelayPrivateData.Metadata` is built from `HandleHeaders`' filtered output, so the cache-key strip is scoped as intended. All new tests pass. Each finding was re-verified after the review; the verdict says how far.

### 1. Medium — behaviour change for REST clients that send JSON under a "sloppy" `Content-Type` — **CONFIRMED**

`protocol/chainlib/base_chain_parser.go:157`

Before this branch every REST POST reached the node as `application/json` no matter what the client sent. Now the client's value goes through verbatim, and strict nodes reject it. Two common client shapes:

- `curl -d '{...}' http://router/v1/view` — curl defaults to `application/x-www-form-urlencoded`
- browser `fetch(url, {method: "POST", body: JSON.stringify(...)})` with no headers — sends `text/plain;charset=UTF-8` to stay CORS-simple

Both links checked. Router side: `/view` and `/transactions` live in the REST `POST` collection of `specs/aptos.json` with `"headers": []`, so `forwardsClientBodyHeader` applies and the header goes through — the same path this branch's `TestRESTRelay_ClientContentTypeReachesTheWire` exercises. Node side: live probe against `fullnode.mainnet.aptoslabs.com/v1/view` with the same JSON body — `application/json` → 200, `application/x-www-form-urlencoded` → **415**. Lenient nodes (gRPC-gateway LCDs, EOS) are unaffected.

This is the ticket's intent (pure proxy), so not a bug in the change — but it is a breaking change for clients that were leaning on the rewrite. The JSON-RPC test comment names exactly this hazard for the other interface. Options: call it out in the release notes; or narrow the carve-out so the two well-known sloppy defaults are not forwarded when the body parses as JSON.

### 2. Low — the cache-key merge rests on "a value the node cannot read is a non-2xx answer", which not every REST node honours — **PLAUSIBLE**

`protocol/chainlib/chain_fetcher.go:967`

`tryCacheWriteResolved` skips non-2xx and `IsNodeError`; `RestMessage.CheckResponseError` marks only 5xx / 429 / cosmos-tx errors as node errors. A node that misparses the body under the wrong `Content-Type` and answers **200 with an error envelope** (TRON-style `{"Error": ...}`) or a default/empty result (a form parser reading `{"a":1}` as a key with no value) produces a cache-eligible entry, which is then served to a client that sent the same bytes with the correct header. Only bites when `reqBlock != NOT_APPLICABLE`. The mechanism is real; I have not demonstrated a node that does this. The trade-off may be fine, but the comment states it as an invariant.

### 3. Low — PUT/PATCH are in the method set, but the REST listener never forwards their body — **CONFIRMED**

`protocol/chainlib/base_chain_parser.go:134`

`rest.go:417` routes only POST to `handlerPost`; PUT/PATCH fall through to `handlerUse` (`rest.go:419`), which calls `SendRelay(..., "", fiberCtx.Method(), ...)` with an empty body (`rest.go:378`). So on those methods the router would forward a `Content-Type` describing bytes it dropped. Harmless today — the bundled specs have zero PUT/PATCH REST collections, and such a request was already broken — but the comment's "forwards that body byte-for-byte" overstates the guarantee. Either restrict to POST or have `handlerUse` forward the body.

### Verified and not flagged

- `pass_override` / `pass_nullify` still win: append-last ordering is honoured by both `DoHTTPRequest` and the legacy `SendNodeMsg`.
- Metadata strip/restore in `hashCacheRequest` matches the existing pattern for the other masked fields; nil-vs-empty metadata marshals consistently on the production path.
- The shared cache server keys on the router-supplied hash, so no second hashing site needs the strip.

PR #388, opened 2026-09-17
feat(resp-cache): let the resp-cache block carry expirations (MAG-3631)

avitenzer:
## The gap

Both cache paths run **one engine over one `core.Policy`** — but only one of them could be told what to put in it.

| the cache is | who builds the Policy | from what | finalized TTL |
| --- | --- | --- | --- |
| the sidecar we ship | `ecosystem/cache/handlers.go:47` | its own flags, set by the chart | **90 min** |
| the customer's Redis | `protocol/performance/resp_config.go:110` | `core.DefaultPolicy()`, hardcoded | **60 min** |

Nothing anywhere passed a non-default `Policy` into `NewRespCache`. `DefaultPolicy`'s comment says it is the table a router-embedded backend uses *"unless configured otherwise"* — there was no otherwise.

So a deployment moving its cache to Redis silently dropped the chart's **own shipped defaults**. `expiration_multiplier: 1.5` and `expiration_non_finalized_multiplier: 1.25` (`values.yaml:578-579`) kept configuring a sidecar no longer in the request path, while the router ran the unscaled table — more traffic to the customer's own nodes, slower repeat answers, and the `1.5` still sitting in their values file looking as though it did something.

## The second multiplier matters more than it looks

`expiration-non-finalized` is a **floor**, not a duration: the effective TTL is `max(averageBlockTime/8, floor)`. It only decides below a **5s** block time.

That threshold is easy to measure past. On a 13s chain both paths give 1.625s and look identical. Most chains in `specs/` are well under 5s:

| avg block time | eighth | sidecar (floor 625ms) | RESP (floor 500ms) | |
| --- | --- | --- | --- | --- |
| 250ms – 2s | 31–250ms | 625ms | 500ms | **differs** |
| 6.5s | 812ms | 812ms | 812ms | same |
| 13s | 1.625s | 1.625s | 1.625s | same |

Both multipliers are carried here, not just the finalized one.

## The change

The `resp-cache:` block gains the expiration keys, under the **cache server's own flag names** (`expiration`, `expiration-multiplier`, `expiration-non-finalized`, `expiration-non-finalized-multiplier`, `expiration-finalized-node-errors`, `expiration-blocks-hashes-to-heights`) so a value moves across unchanged rather than needing a second vocabulary.

```yaml
resp-cache:
  addresses: ["cache:6379"]
  expiration-multiplier: 1.5                 # the chart's own default
  expiration-non-finalized-multiplier: 1.25  # the chart's own default
```

`ForRelayEntry`, `ForChain` and every TTL rule are untouched — this only decides which numbers they are handed.

**Every key is optional.** An omitted key keeps the cache server's flag default, so existing `resp-cache` deployments are byte-for-byte unchanged; `TestRespCacheExpirationsConnectionOnlyBlockKeepsDefaults` pins exactly that.

### One deliberate divergence from the cache server

A key that is **present and non-positive** aborts startup instead of being clamped. The sidecar accepts `--expiration-multiplier 0` and writes a zero TTL, which Ristretto reads as "never expires" — contained, because the store dies with the pod. Here it is a key with **no expiry at all**, in a store the customer owns and may share, surviving every restart and reclaimed only by eviction. Omitting a key is how you ask for the default.

## Verifying it

`/debug/cache-state` already reports the engine's policy, so the resolved table is readable without new surface:

```
curl -s localhost:<debug>/debug/cache-state | jq .tiers.primary.lifetimes
# finalized_seconds: 5400  → the multipliers were carried over
# finalized_seconds: 3600  → they were not
```

Startup also logs the resolved table (`expiration-finalized=1h30m0s expirations-configured=true`) rather than the raw keys — a multiplier alone doesn't tell an operator what it became.

## Tests

`protocol/performance/resp_expirations_test.go` and one addition to `backend_select_test.go`. The end-to-end one asserts on **miniredis's own TTL for the written key**, not on the resolved `Policy` — the Policy resolving correctly was never the claim; the claim is that it arrives at the engine that writes. That test reads 1h on `main` and 90m here.

Also covered: the fast-chain floor divergence above, multiplier-scales-explicit-base ordering, omitted-key-is-not-an-explicit-zero, non-positive rejection per key, and int64 saturation on an absurd multiplier (wrapping negative would turn "very long" into "no caching").

`go build ./...`, `go vet ./...`, and `go test ./protocol/... ./ecosystem/...` all pass. golangci-lint reports only a pre-existing `h2c.NewHandler` deprecation in `connection_cmd.go`.

## Scope

Router-side only, matching how the RESP backend itself is configured today. The published Helm chart has **no `resp-cache` support at all** (`grep -riE "resp-cache|redis|valkey" charts/smart-router/` is empty), so the RESP path is reachable only through the router's config file or flags. When chart support lands, these keys are what it should pass through — and `miscellaneous.cache.expiration_multiplier` should feed them, or the same gap reopens one layer up.

Docs: `docs/RESP-CACHE.md` gains an **Expirations** section. Its opening claim that TTLs are "identical to the default cache" was half true and is now corrected — the *table* is shared, the *values* are yours to set.

Jira ticket: MAG-3631

---

## Review findings (2026-09-20)

Reviewed head [`cecd3a1`](https://github.com/Magma-Devs/smart-router/commit/cecd3a11329a8846b3b3ce84af59d7fd244925c0).

### [P2] Reject a positive expiration that scales down to zero

[`scaleExpiration`](https://github.com/Magma-Devs/smart-router/blob/cecd3a11329a8846b3b3ce84af59d7fd244925c0/protocol/performance/resp_expirations.go#L165-L170) protects against overflow, but converting a positive product below one nanosecond to `time.Duration` truncates it to zero. The loader validates the two inputs individually, so this configuration is accepted:

```yaml
resp-cache:
  addresses: ["127.0.0.1:6379"]
  expiration: 1ns
  expiration-multiplier: 0.5
```

The resulting `Policy.Finalized` is `0s`. The Redis adapter uses zero for a SET with **no expiry**, so a setting intended to shorten retention creates a permanent key—the behavior the new non-positive validation explicitly promises to prevent.

Reproduced through `SelectCacheBackend` and the real cache engine using miniredis: a finalized relay entry had `TTL=0s` and still existed after advancing the server clock by 24 hours. The diagnostic used a temporary Go test overlay; repository source files were not changed.

Please validate the effective scaled duration, rejecting or rounding up a positive result that becomes zero, and add a regression that checks the stored entry's TTL. This is also the effective-TTL issue already recorded on [#405](https://github.com/Magma-Devs/smart-router/pull/405); this PR's separate scaling implementation needs the same protection.

Validation: the existing expiration-policy and backend-selection tests passed. The additional assertion that an accepted positive configuration must not create a permanent key failed on this head.


avitenzer:

## Review findings (2026-09-20)

Reviewed head [`cecd3a1`](https://github.com/Magma-Devs/smart-router/commit/cecd3a11329a8846b3b3ce84af59d7fd244925c0).

### [P2] Reject a positive expiration that scales down to zero

[`scaleExpiration`](https://github.com/Magma-Devs/smart-router/blob/cecd3a11329a8846b3b3ce84af59d7fd244925c0/protocol/performance/resp_expirations.go#L165-L170) protects against overflow, but converting a positive product below one nanosecond to `time.Duration` truncates it to zero. The loader validates the two inputs individually, so this configuration is accepted:

```yaml
resp-cache:
  addresses: ["127.0.0.1:6379"]
  expiration: 1ns
  expiration-multiplier: 0.5
```

The resulting `Policy.Finalized` is `0s`. The Redis adapter uses zero for a SET with **no expiry**, so a setting intended to shorten retention creates a permanent key—the behavior the new non-positive validation explicitly promises to prevent.

Reproduced through `SelectCacheBackend` and the real cache engine using miniredis: a finalized relay entry had `TTL=0s` and still existed after advancing the server clock by 24 hours. The diagnostic used a temporary Go test overlay; repository source files were not changed.

Please validate the effective scaled duration, rejecting or rounding up a positive result that becomes zero, and add a regression that checks the stored entry's TTL. This is also the effective-TTL issue already recorded on [#405](https://github.com/Magma-Devs/smart-router/pull/405); this PR's separate scaling implementation needs the same protection.

Validation: the existing expiration-policy and backend-selection tests passed. The additional assertion that an accepted positive configuration must not create a permanent key failed on this head.

AnnaR-prog:
## Review: `/code-review high` on `cecd3a1`

Scope: full diff `main...cecd3a1`, the sidecar's flag handling it mirrors (`ecosystem/cache/server.go`), `core.Policy`, the RESP store's TTL write, and the two open PRs that touch the same block (#397, #405). Each finding was re-verified after the review against a worktree at this head; the verdict says how far. The scaled-to-zero P2 already on this PR stands at this head and is not repeated on its own — it reappears under finding 3 as one of three instances of one missing check.

**Top-line, before the findings:** #405 is open, same author, same ticket, and implements the same feature under a different schema — nested `expiration.finalized-multiplier` there, flat `expiration-multiplier` here. They cannot both merge. Which schema is the decision for MAG-3631?

### 1. High — the expiration keys and #397's strict block decoding cannot coexist — **CONFIRMED**

`protocol/performance/resp_expirations.go:100`, `protocol/performance/resp_config.go:36`

The `resp-cache:` block is decoded twice, into `redisstore.Config` and into `RespCacheExpirations`, each ignoring the other's keys. #397 (open; #405 is stacked on it) turns on `ErrorUnused` for the `redisstore.Config` decode of that same block. Merged `origin/fix/MAG-3521-cache-keyspace-isolation` onto `cecd3a1` (one import-block conflict, nothing else) and ran this PR's own e2e test:

```
--- FAIL: TestSelectBackendRespCarriesConfiguredExpirations
invalid resp-cache block (unknown keys are rejected so a misspelled setting cannot fall back to a default unnoticed):
'redisstore.Config' has invalid keys: expiration-multiplier
```

The breakage is symmetric: if #397 lands first, this PR's rebase fails this test; if this PR lands first, #397's rebase starts refusing every config that uses these keys. CI catches it either way, but the fix is a design choice rather than a rebase: decode the block once into a struct that knows both halves, or carry the expiration fields on `redisstore.Config` as #405 does.

### 2. Medium — a misspelled expiration key silently keeps the default — **CONFIRMED**

`protocol/performance/resp_expirations.go:100`

`v.UnmarshalKey` without `ErrorUnused`: `expiration-multiplyer: 1.5` loads without error and the finalized entry is written with TTL `1h0m0s` (miniredis). That is the ticket's own failure shape — the chart's multiplier not applying and nothing saying so — reproduced one layer down, with the docs' "omit a key to get its default" now also covering "misspell a key". Strict decoding is the fix, which is finding 1 from the other side.

### 3. Medium — the resolved `Policy` is never validated, only the inputs' sign, and there are more ways past that guard than the P2 records — **CONFIRMED**

`protocol/performance/resp_expirations.go:44`, `protocol/performance/resp_expirations.go:124`

`LoadRespCacheExpirations` checks each key's sign; nothing checks what `Policy()` resolves them to. Three inputs that pass, and what the store then holds (all against miniredis):

| config | resolved `Finalized` | stored TTL |
| --- | --- | --- |
| `expiration: 1ns`, `expiration-multiplier: 0.5` | `0s` | none — permanent key (the P2 already on the PR) |
| `expiration-multiplier: 1e-12` | `3ns` | `1ms` — go-redis rounds a sub-millisecond `PX` up |
| `expiration: 3600` | `3.6µs` | `1ms` |

The last row is the one an operator will actually type: mapstructure maps a bare YAML integer straight onto the `int64`, so a unitless value is nanoseconds. The sidecar's `--expiration 3600` is refused by pflag (`missing unit`), so this is one place "same names, same numbers" is weaker than it reads, and the docs table lists `1h` / `500ms` without saying a unit is mandatory. The startup log does print `expiration-finalized=3.6µs`, so it is visible — to whoever reads that line. The connection keys (`dial-timeout`) have the same decode, but a 500ns dial fails loudly; a 1ms TTL just stops caching, `expirations-configured=true` in the log.

One check on the resolved table after scaling — every field at least 1ms, the precision a RESP `PX` can express — closes all three rows together, and is where the P2's "reject or round up" should live rather than in `scaleExpiration`.

### 4. Nit — the docs' "under 5s" is the chart-configured threshold, not the default's

`docs/RESP-CACHE.md:106`

The threshold where the floor stops mattering is `8 × floor` (`core/policy.go:55`): 5s at the chart's 625ms, 4s at the 500ms default the table two rows above lists. Right for the comparison the paragraph is making, but stated flat, next to a default it does not hold for. "under 8× the floor — 5s with the chart's 1.25" reads true at both.

### Verified and not flagged

- Key names match `ecosystem/cache/server.go:27-32` verbatim, and `Policy()` applies each multiplier after its base resolves, exactly as `server.go:79-80`.
- `ForChain` is `max(avg/8, NonFinalized)` (`core/policy.go:55`), so the docs formula itself is right.
- `scaleExpiration` saturates on the product, not the inputs; `+Inf` saturates, `NaN` is rejected by the `> 0` check.
- `/debug/cache-state` reports `engine.Policy` for the RESP tier (`resp_cache.go:252`), so the verification recipe in the PR body works.
- The `--resp-cache-*` flags carry no expirations; a flag-only deployment gets the defaults, which is what it gets on `main` — consistent with "the full surface lives in the YAML block", nothing lost.

avitenzer:
Closing in favour of #405, which implements the same feature for MAG-3631 with the expiration fields carried on the store config, so it coexists with the strict block decoding #397 introduces, and which already validates the effective lifetime after scaling (the scaled-to-zero finding). The schema question raised in the review is settled by keeping #405; the other findings here are re-checked there where its own decoding differs.

PR #396, opened 2026-09-18
fix(redisstore): refuse a tls block that was written without its switch

avitenzer:
# Description

**Bug (MAG-3683).** A `resp-cache.tls` block written with `ca-file` / `cert-file` / `key-file` but without `tls.enabled` was silently inert: the files are only opened when the switch is on, so the router started, logged `tls=false` among six other startup fields, opened a plaintext connection, and its first write carried the configured username and password in the clear. Three nonexistent certificate paths were no obstacle to booting.

**Fix.** `redisstore.Config.Validate()` now treats a `tls` block carrying any key other than the switch (`ca-file`, `cert-file`, `key-file`, `server-name`, `insecure-skip-verify`) as dangling configuration and refuses to start, naming `tls.enabled` and the two ways out (set it, or remove the block). This is the rule the same file already applies to `sentinel-*` credentials on a non-sentinel topology, and the ticket's first option.

**Design call, stated so it can be challenged.** An explicit `tls.enabled: false` beside certificate paths is refused too. At struct level the two are indistinguishable, and a block that says "off" next to three certificate paths is the same trap for the next reader. A templated config that renders the block with the switch off must wrap the block in a conditional. The demo lanes under `scripts/pre_setups/` already render `enabled: true` and are unaffected.

Files to review first: `ecosystem/cache/redisstore/config.go` (the rule) and `ecosystem/cache/redisstore/tls_test.go` (the wire-capture test).

## Tests

The ticket asks for both directions. `TestTLSBlockWithoutEnabledNeverSendsPlaintext` drives a listener that records what reaches it and never answers, the same instrument that captured the ticket's evidence, in four steps:

1. no `tls` block, with a password: the password **is** readable in the capture (instrument check; that configuration is deliberate and stays allowed);
2. the ticket's block without the switch: refused at construction naming `tls.enabled`, nothing reaches the wire;
3. the same block with the switch on: the listener sees a TLS handshake record and never the credential (the control without which step 2 passes on a store that sends nothing);
4. the switch on with a missing CA file: construction aborts naming the file (the ticket's second control).

Also: `TestConfigValidateMatrix` gains one case per `tls.*` key, `TestNewFailsFastOnBadInputs` covers construction, and `TestLoadRespCacheConfigRefusesTLSBlockWithoutSwitch` pushes the ticket's YAML through the real loader, with the switched-on control failing on the first file as it always should have.

The ticket suggests a binary-level test in the automation repo's `tests/infrastructure/binary_startup/` harness. The in-repo wire-capture test covers the same two directions at the store layer; a binary-level lane there is still worthwhile as an end-to-end check and is left to that repo.

## Docs

`docs/RESP-CACHE.md`: the `tls.enabled` reference row, and a troubleshooting row for the new startup error.

## Out of scope

The related finding that `insecure-skip-verify` is accepted with no log line is filed separately, per the ticket.

Jira ticket: MAG-3683

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable: a previously-silent misconfiguration now fails startup; no API changes)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3683)

## Reviewers Checklist

I have...


AnnaR-prog:
## Review: `/code-review high` on `dcd32c4`

Scope: the full diff `main...dcd32c4`, the two loaders that call `Validate` (`redisstore.New`, `LoadRespCacheConfig`), the changelog tooling the commit message feeds, and the two open PRs stacked on this one that touch the same block (#397, #402). Each finding was re-verified after the review against a worktree at this head; the verdict says how far.

**Stack note:** this branch is the base of two chains (→ #397 → #405, and → #400 → #401 → #402), all opened against `main`, so their "Files changed" tabs include this diff until it merges. Each was reviewed against its parent head, not `main`.

### 1. Medium — the explicit `enabled: false` refusal is a breaking change the release tooling will not see, and it is a placement choice, not a struct-level necessity — **CONFIRMED**

`ecosystem/cache/redisstore/config.go:175`, `protocol/performance/resp_config.go:32`, `docs/RESP-CACHE.md:697`

Probe through the real boot path (`SelectCacheBackend → LoadRespCacheConfig → redisstore.New`): `tls: {enabled: false, ca-file: /ca.pem}` starts on `main` and is refused at `dcd32c4`. The PR body says so and calls it a design call, so two consequences to decide on:

- The subject carries no `!` and the body no `BREAKING CHANGE:` footer, which are the only two signals `scripts/changelog-bump.sh` reads (`:143`, `:167`), so the next release's notes file this under Bug fixes with no ⚠ block. Repo precedent for a config-shape refusal is `chore!: drop deprecated static-providers/backup-providers config keys (#135)`. The body's third paragraph is already the migration note; it just needs the footer.
- "At struct level the two are indistinguishable" is true of the struct, but the only production caller already tells absent from present with `v.IsSet(RespCacheViperKey)`, and `v.IsSet("resp-cache.tls.enabled")` is `true` for a written `false` and `false` for the ticket's switchless block (probe, viper v1.21). A loader-level `IsSet("resp-cache.tls") && !IsSet("resp-cache.tls.enabled") && hasMaterial()` refuses exactly the shape MAG-3683's first option names while a deliberate off-toggle keeps booting. Trade-off to state if you take it: a Go caller of `redisstore.New` with zero-value `Enabled` plus material would dial plaintext unless the struct rule is kept too. Keeping the struct rule and adding the footer is the smaller change; I'd take either over the current state, which is the strict rule with no release signal.

Whichever way: the new troubleshooting row says the block "was written without its switch" and offers "Add `enabled: true`" or "remove the block" — wrong for the operator who turned TLS off by hand while keeping the paths (probe: `enabled: false`, `"false"`, `0`, `null` and `""` all hit this message), and the accepted off-shape `tls: {enabled: false}` alone is stated nowhere. Row 74 is fine; its bold rule disambiguates.

### 2. Low — the harm the message names is silent in every shape that stays allowed — **CONFIRMED**

`protocol/performance/resp_config.go:106`, `ecosystem/cache/redisstore/config.go:176`

`resp-cache: {addresses: [cache.internal:6379], username: cacheuser, password: s3cret}` with no `tls` block (subtest 1's shape, "deliberate and stays allowed") boots with `tls=false` as one of five Info attributes and the first write carries the password (probe captured `hello 3 auth cacheuser …`). The refusal only covers the syntactic shape. The repo's pattern for this is `protocol/common/grpc_auth.go:108` `TokenOverInsecureWarning`: warn, don't refuse, name the endpoint. One `LavaFormatWarning` in `SelectCacheBackend` when any of `username`/`password`/`password-file`/`sentinel-password(-file)` is set and `!TLS.Enabled` covers every shape, fires correctly on the two remote demo lanes and harmlessly on the loopback ones, and breaks nothing. Complementary to this PR's refusal, not a replacement.

Related: the message asserts "the configured credentials would cross the network readable" without consulting them — `tls: {server-name: cache.internal}` on a credential-less internal cache gets the same sentence (probe). The two sibling dangling-config errors (`config.go:160`, `resp_config.go:59`) use `<what> — dangling configuration (<remedy>)`; the docs row and all seven test assertions key on the prefix, so shortening to that shape keeps everything green, and the rationale already lives in the `TLSConfig` doc comment.

### 3. Low — `hasMaterial()` hand-enumerates the five fields and nothing locks the list — **CONFIRMED**

`ecosystem/cache/redisstore/config.go:107`

Overlay 1: add ``MinVersion string `mapstructure:"min-version"` `` to `TLSConfig`, honour it in `build()`, leave `hasMaterial` alone → `Config{Addresses, Password, TLS: {MinVersion: "1.3"}}` passes `Validate`, `build()` returns `nil, nil`, `New` dials plaintext with the password, full package green. Overlay 2: delete `c.KeyFile != ""` → `TestConfigValidateMatrix`, `TestNewFailsFastOnBadInputs`, the wire test and `TestLoadRespCacheConfig*` all stay green (`config_test.go:44` is the only KeyFile case and pairs it with CertFile). `c.Enabled = false; return c != TLSConfig{}` is equivalent on all 64 field combinations, vet-clean, catches the next key, and stops compiling if a non-comparable field is ever added.

### 4. Low — four test-shape points in `TestTLSBlockWithoutEnabledNeverSendsPlaintext` — **CONFIRMED**

`ecosystem/cache/redisstore/tls_test.go:294`, `:317`, `:332`, `:237`

- Subtest 1 is the only subtest whose pass depends on the 500 ms Ping context covering go-redis's background dial plus the HELLO write. Deterministic probe: dialer delayed 600 ms → `pingOnce` still errors as required, the late conn is parked idle with no HELLO written, and `require.Eventually` at `:322` fails after 2 s with "password not readable on the wire". 10/10 `-race` on loopback, so this is the loaded-CI-runner flake shape, on a subtest that proves what `TestAuthAgainstMiniredis` (`credentials_test.go:122-134`) already implies. Either drop it or re-issue the Ping inside the `Eventually` condition (a second Ping initialises the parked conn and the password then appears).
- Subtests 2 and 4: `require.Empty(received())` is dead — dominated by the preceding `ErrorContains`, and it still passes with `Validate` moved after both `buildClient` calls (go-redis dials lazily; probe with the rule removed entirely captured 0 bytes after construction). The header comment at `:288-289` says subtest 2 is the control that needs subtest 3; subtest 2 passes because no store exists. Both cases already live in `config_test.go:219-229` and `resp_config_test.go:183`; if they go, strengthen `config_test.go:220` from bare `require.Error` to `ErrorContains("/does/not/exist")` so "names the file" stays covered.
- The closure `newStore` at `:294` shadows the package constructor `newStore(writeClient, readClient, keyPrefix)` at `store.go:303` (compiles only because it is function-local); `newTestStore` is taken by `store_test.go:16`, so something like `newStoreWithCreds`, and let `pingWithTLS` delegate to it.
- `wireCaptureListener` re-implements `listenLocal(t)` (`config_test.go:57`) and hand-rolls a read loop that is `io.Copy` into a locked writer; the package now has three copies of listen-and-cleanup and three spellings of accept-and-drain. One trap for whoever folds it: the buffer must be a named field, not embedded — embedding `bytes.Buffer` promotes `ReadFrom`, `io.Copy` takes it and bypasses the lock (DATA RACE under `-race` in the overlay).

### Covered by the PRs stacked on this one — not flagged here

- Mis-indented, misspelled (`enable:`) or snake_case (`ca_file`) tls keys decode to `TLSConfig{}` and the router starts plaintext with the password (probe captured the AUTH for all three), while rows 74/697 promise a refusal. #397's `ErrorUnused` on this block closes it; worth one clause in those rows that unknown keys are refused rather than ignored, once #397 lands.
- `tls: {insecure-skip-verify: true}` alone: the remedy text steers into TLS that accepts any certificate and nothing announces it (probe: self-signed server receives the decrypted `hello 3 auth …`; startup log says only `tls=true`). #402's warning closes it.

### Verified and not flagged

- `tls: {enabled: true}` alone (system pool), `tls: {}` and `tls: {enabled: false}` alone are all still accepted (probe through the loader).
- `Validate` runs before `build()` and both `buildClient`s in `New`, and the loader calls it too, so nothing dials before the rule.
- Refuted: subtest 3 reading a split ClientHello (`got[1] != 0x03`) — it is one 1501-byte record in one Write; with a 200 ms DialTimeout overlay the listener sees merged whole records, 3/3 pass.
- The new tests pass at this head (`go test ./ecosystem/cache/redisstore/ ./protocol/performance/`).
- In-repo lanes (`scripts/pre_setups/*`, `docker/docker-compose.resp-cache.yml`) render `enabled: true` or no block; nothing shipped in-repo is refused.

avitenzer:
Fixed in 7051815. The root commit is amended to 44571b1 (message only) and five commits sit on top; the diff against `main` is what the "Files changed" tab shows once the two chains stacked on this one are re-pushed.

1. Kept the struct-level rule, since the loader-level alternative would leave a Go caller of `redisstore.New` dialling plaintext with a zero-value `Enabled` beside material, which is the exact shape the rule exists for. Added the release signal instead: the subject is now `fix(redisstore)!:` and the body carries the `BREAKING CHANGE:` footer with the migration note, so `scripts/changelog-bump.sh` files it under the ⚠ block. The troubleshooting row (7051815) now says the message fires whether the switch is missing or written as anything but `true`, that a switch turned off by hand with the paths left in place is refused the same way, and that a bare `tls: {enabled: false}` or no block is how to be off; the reference row states the accepted off-shape too.

2. The warning is there (719ddc2), emitted from `redisstore.New` beside `warnIfReadSplitIsDiscoveryScoped` rather than from `SelectCacheBackend`, so every caller of the constructor gets it; the loader is the only caller today, and `TestSelectBackendWarnsWhenCredentialsCrossPlaintext` drives it through `SelectCacheBackend` with the same credentials under TLS and no credentials at all as the two controls. It names the endpoints and which of the six credential keys are set, never a value (`TestConfiguredCredentialKeys`). The dangling-tls error is shortened to the sibling shape, `… — dangling configuration (set tls.enabled: true, or remove the other tls.* keys)`; the prefix the docs row and the seven assertions key on is unchanged. Capturing the warning needed the synchronized sink, so #400's `log_capture_test.go` now lands here at the root of the chain (e736a1a) with the nesting contract enforced, and #400 inherits it.

3. `hasMaterial` clears the switch and compares the rest to the zero value (de351a5); a key added to `TLSConfig` is covered without anyone listing it, and a non-comparable field stops the build.

4. All four (57221b1). Subtest 1 stays but re-issues its Ping inside the `Eventually` condition, so a budget that runs out between the dial and the HELLO write is retried on the next tick rather than failing the test. Subtests 2 and 4 are gone; the refusal is pinned in the matrix, in `TestNewFailsFastOnBadInputs` (which now asserts the missing file is named, for both the credential file and the CA) and through the loader. The closure is `newStoreWithCreds` and `pingWithTLS` delegates to it. `wireCaptureListener` sits on `listenLocal` and copies into a locked writer whose buffer is a named field, for the reason you gave; the handshake-timeout test at the top of the file uses `listenLocal` too. The test is renamed `TestTLSSwitchDecidesWhatCrossesTheWire`, since the refusal is no longer what it proves.

Verified: `go build ./...`, `go vet` and `golangci-lint run --config .golangci.yml` over `./ecosystem/cache/... ./protocol/performance/... ./protocol/rpcsmartrouter/...` (the two lint hits left in touched files, `nilnil` on `build()` and the assertion in `TestClusterOptionsMapping`, are on `main`); `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/... ./protocol/rpcsmartrouter/... ./types/relay/...` passes at 7051815; the wire test, the warning test and the credential-keys test pass 3× under `-race`.

Not done: the "unknown keys are refused" clause on the two docs rows, which waits for #397's strict decoding as you noted; and a decode hook, since the strict decoder is #397's.

AnnaR-prog (approved):

PR #397, opened 2026-09-18
fix(cache): give every router its own keyspace on a shared cache

avitenzer:
# Description

**Bug (MAG-3521).** Two routers that declare the same chain but read different nodes served each other's cached answers as ordinary cache hits. The cache key carried no router identity (chain, request hash, block), the chain tip that resolves `latest` was one value per chain, and on the shipped gRPC sidecar there was no setting to separate deployments. The RESP backend had `key-prefix`, but nothing a customer could deploy was able to set it (MAG-3687), and a misspelled key in the block was silently ignored and fell back to the shared default (MAG-3677). Of the three tickets MAG-3521 absorbed, MAG-3687 and MAG-3677 are covered here; MAG-3686 is re-filed as MAG-3755 (below).

**Fix.** The keyspace is a per-router setting on both backends, with one character set shared through `core`:

| Backend | Setting | Mechanism |
| --- | --- | --- |
| gRPC sidecar (`cache-be`) | `cache-be-key-prefix` / `--cache-be-key-prefix` | The client stamps a new `key_prefix` field on every relay and sticky-session message; the server's engine folds it into every key it derives (entry, chain tip, block-hash heights, shared-state tip, sticky claim). |
| RESP | `--resp-cache-key-prefix` (flag form of `resp-cache.key-prefix`) | Store-layer prefix, as before; now reachable from a flag. |

Plus strict decoding of the `resp-cache:` block: unknown keys, nested blocks included, are a startup error.

**Design decisions, stated so they can be challenged.**

- **Where the scope sits.** After the kind prefix (`rel:f:<prefix>:ETH1:…`), not at the head of the key like the RESP store's own prefix. The in-process store routes on the kind prefix, and the sidecar holds one store for every router, so the scope has to travel inside the key.
- **Empty is the identity.** A router without a prefix writes exactly the keys it wrote before, persisted RESP entries survive the upgrade, and a router that predates the field is unchanged against a new server. Tested.
- **The default stays shared.** Isolating by default would silently cost every existing multi-replica deployment its hit rate on upgrade, and the shared case (replicas reading the same nodes) is the documented one and is correct. Deriving a keyspace automatically from the node set was considered and not done: it is a product call, and a config change would invalidate a fleet's cache mid-rollout. Opt-in, documented, with the condition written down: routers sharing a keyspace must read the same nodes.
- **Endpoint observations stay unscoped.** They are facts about one upstream URL; sharing them across routers is correct, and a router on other nodes never matches their ids. Commented at the key function.
- **The secondary cache tier is unchanged.** It reads the unprefixed keyspace; a prefix for it is outside this ticket and a small addition on the same mechanism if wanted.

**Compatibility hazard, named on purpose.** A new router with a prefix against an *old* cache server gets no isolation and cannot detect it: the field is dropped on the wire silently. The docs, the flag help and the field comment all say the sidecar must be upgraded alongside.

**Flush asymmetry** (the ticket asked). RESP flush is prefix-scoped as before. The sidecar's `FlushCache` empties every keyspace, because ristretto cannot enumerate by prefix; documented.

**Re-filed, not dropped.** The intra-fleet chain-tip anti-lie guard, the part of MAG-3686 a prefix cannot fix per the absorbing comment on MAG-3521, is re-filed as [MAG-3755](https://magmadevs.atlassian.net/browse/MAG-3755) and linked to the ticket.

**Merge order.** This branch is rebased on top of #396 (MAG-3683) because both touch `docs/RESP-CACHE.md` and `protocol/performance/resp_config_test.go`; merge #396 first. Until then this PR's diff shows that commit as well.

Files to review first: `ecosystem/cache/core/keys.go` (`ScopedChainId`, `ValidateKeyPrefix`), `ecosystem/cache/core/engine.go`, `protocol/performance/cache.go`, `protocol/performance/resp_config.go`.

## Tests

- `core`: `TestKeyPrefixIsolatesEveryKey`. Every key of a write is scoped (entry, chain tip, heights, shared tip); another prefix and the unscoped keyspace see none of it, by concrete block or by `LATEST`; an unprefixed writer's keys are byte-for-byte unchanged. The fake store now records chain tips by key, because it used to ignore the key argument and a tip-scoping assertion against it would have been vacuous.
- `performance_test`: `TestKeyPrefixIsolatesRoutersOnOneCacheServer`. The ticket's scenario over a real gRPC hop: three routers on one server (two prefixes, one legacy); the second router does not serve the first's entry by block or by `LATEST`, with the positive control the ticket demands (each router serves its own write back, and none of those writes reach the first). Also sticky-claim isolation and the debug-state rendering.
- Selection: a glob-unsafe `cache-be-key-prefix` aborts startup; two routers selected with different prefixes share nothing.
- Loader: `--resp-cache-key-prefix` outranks YAML and works with the address flag alone; a `key_prefix` typo and a typo inside `tls:` are refused; every defined key still loads.
- Shipped command: both flags registered and landing through the real cobra/viper sequence.
- Wire: legacy payloads decode to the shared keyspace; all four messages carry the field.

The ticket suggests a live isolation test and a cluster configuration guard in the automation repo, and its PR 937 there already adds both as strict xfails. Merging this PR alone leaves that xfail red on purpose: turning it green is a configuration change outside this repo. Each router deployment on the cluster needs its own `cache-be-key-prefix` (in the automation repo's router configs or the Helm chart), and the cache sidecar must run a build that carries this change. Once the routers have distinct prefixes the live test should pass, and the cluster guard should group by keyspace rather than fail on every shared cache address.

## Docs

`docs/RESP-CACHE.md`: new section "Sharing a backend between routers" (the condition, both settings, the sidecar upgrade requirement, the flush asymmetry, what a prefix does not do), the `key-prefix` reference row, the strict-decoding note, and the flush section. `docs/LOCAL-COMPOSE.md` and the two example configs point at it. The lines the ticket cites in `Magma-Devs/docs` (`deployment/cache/index.md`, the "shared across multiple router replicas" sentences and the key description; `deployment/cache/redis.md`, the prefix advice) live in that repo and need the same condition written in; follow-up there.

`go.mod`: `github.com/go-viper/mapstructure/v2` moves from indirect to direct. It is the type viper's decoder option names, needed for strict decoding.

Jira ticket: MAG-3521

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable: the new wire field is additive, empty is the identity, and the default keyspace is unchanged; a misspelled `resp-cache` key now fails startup where it was silently ignored)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3521)

## Reviewers Checklist

I have...

[MAG-3755]: https://magmadevs.atlassian.net/browse/MAG-3755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AnnaR-prog:
## Review: `/code-review high` on `156542e`

Scope: this PR's commit (`dcd32c4...156542e`, without #396's diff the "Files changed" tab still shows): `core/keys.go` and every `ScopedChainId` call in the engine and the sticky handlers, the client-side stamping in `performance/cache.go`, the loader's strict decoding and the new flags, the wire types, both example configs, and the docs. Re-verified against a worktree at this head; the key-prefix, loader, cobra/viper and wire tests pass there, and both bundled example configs load under the strict decoder.

**Stack note:** stacked on #396 (base `main` on GitHub); #405 is stacked on this one. #388 (same author, MAG-3631, open) cannot merge after this without a rebase — its flat `expiration-*` keys are unknown to the strict decoder; #405 is the version that fits, as noted on #388.

The design decisions are sound and the mechanism is complete: one `ScopedChainId` per request covers the entry, the tip that resolves `LATEST`, the heights, the shared-state tip and the sticky claim; empty is the identity so persisted keys survive; the client copies rather than mutates, so the secondary tier sees the caller's request. Findings are about what the PR names and chooses not to do.

### 1. Medium — "a new router with a prefix against an old cache server gets no isolation and cannot detect it" is detectable, on the wire this PR already changes — **CONFIRMED**

`types/relay/cache.go:92`, `protocol/performance/cache.go:220`, `ecosystem/cache/handlers.go:91`

This is MAG-3677's failure class one level up: the operator set the one setting that must differ between deployments, and it falls back to the shared keyspace without a word. The PR documents it in three places; it can be closed instead, cheaply, in either of two ways on surfaces this PR touches:

- Echo `key_prefix` in `CacheRelayReply` / `StickySessionReply`. A server from this version returns what it scoped by; an older one leaves the field empty. A client that sent a non-empty prefix and reads an empty echo knows, on its first reply, that it is unisolated — warn once per connection (the reconnect loop already has a once-per-connection seam), and render `/debug/cache-state` as `prefix=<x> (unconfirmed)` until an echo arrives.
- Or `Health`'s `CacheUsage`, today an empty struct, carries a capability the client checks on connect.

Either way the failure stops being silent. Given that the operator explicitly asked for isolation, refusing to serve until confirmed is defensible too; the warning is the minimum.

### 2. Low — strict decoding turns any config with a stray key under `resp-cache:` from "starts" into "refuses to start", with no release signal — **CONFIRMED**

`protocol/performance/resp_config.go:46`

The right change (the ticket is exactly the misspelled-key case), and every in-repo lane and both example configs load under it. But a deployment whose chart renders one extra key — or one that misspelled a key and has been running on the default since — fails startup on upgrade, and the subject has no `!` / body no `BREAKING CHANGE:` footer, the only two signals `scripts/changelog-bump.sh` reads (`:143`, `:167`). Same point as on #396 and #401; the three land as one release and should carry the footer consistently. Worth checking the Helm chart's rendered `resp-cache:` block against the struct before this ships (the chart lives outside this repo).

### 3. Low — the server takes any `key_prefix` the wire hands it; the grammar the client enforces is not enforced where the keys are made — **CONFIRMED**

`ecosystem/cache/core/keys.go:39`, `ecosystem/cache/sticky_sessions.go:112`, `ecosystem/cache/handlers.go`

`ValidateKeyPrefix` forbids `:` so a scoped chain id "can never be read as a different chain", but only the router validates; the sidecar folds whatever arrives. A client that sends `a:ETH1` for chain `X` derives the same keys as prefix `a` on chain `ETH1:X` — contrived from this repo's client, not from any other. One `core.ValidateKeyPrefix(req.KeyPrefix)` per handler (`codes.InvalidArgument`) keeps the key grammar unambiguous at the one place that owns it, and costs a regexp match on a string that is usually empty. Related, for the docs section: the prefix is a cooperative namespace, not a tenant boundary — the sidecar has no authentication, so any client can name any keyspace; saying so stops someone reading "isolation" as a security property.

### Nits

- `--cache-be-key-prefix` set while a `resp-cache` block is present: RESP wins and the gRPC prefix scopes nothing, but the loader's warning covers only `cache-be` empty, and the "both configured" warning names the address, not the prefix.
- `FlushCache` on the sidecar empties every keyspace — documented; blast radius is unchanged from before (everyone shared one keyspace), so not flagged, but the docs sentence could say that explicitly so an operator does not read the prefix as a flush boundary.

### Verified and not flagged

- Empty prefix is byte-for-byte the pre-PR key on every kind (the core test), and a legacy payload decodes to the shared keyspace (wire test).
- The gRPC handler's metrics label keeps the raw chain id: `ScopedChainId` is applied to a local, never to the message.
- Endpoint observations stay unscoped for the reason given; the key is the endpoint id, which a router on other nodes never matches.
- Validation sits in `SelectCacheBackend` rather than the constructor because the constructor's error is the dial's; correct call, and the reason is written at both sites.
- Strict decoding reaches nested blocks: `tls: {enable: true}`, `tls: {enabled: true, ca_file: /x}` and a mis-indented `tls:` are all refused at this head (probe through the loader).

avitenzer:
Rebased onto #396's 7051815; no content change. Tip is 659aa75. The one conflict was the docs reference table, where this branch's `key-prefix` row and #396's reworded `tls.enabled` row sit on adjacent lines; both kept as written.

Verified at 659aa75: `go build ./...`, `go vet` and `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/... ./protocol/rpcsmartrouter/... ./types/relay/...` pass.

avitenzer:
Fixed in 5a8ee5b, on the branch rebased onto #396's new head 67d0e5c (main advanced to 970ab87 under it; #405 follows on cc4650a). The root commit is amended (message only) and three commits sit on top.

1. The failure is detectable now, on the wire this PR already changes, and I took the echo rather than the capability: it costs one string on replies the router reads anyway, and it judges the server actually answering rather than the one that answered `Health`. The server echoes the prefix it scoped by on every relay and sticky reply, hit or miss (4bd1099). A router that sent a prefix and reads an empty echo warns once per connection, using the seam you named: the reconnect loop resets the verdict and the once-flag when it installs a connection, so a server upgraded behind a reconnect is judged afresh. `/debug/cache-state` qualifies the prefix by what the current connection's server has said: `prefix=tenant-a (unconfirmed)` until the first reply, `prefix=tenant-a (ignored by the cache server)` once a reply came back without the echo, bare `prefix=tenant-a` once confirmed. Serving goes on either way; I did not take the refusal, since it turns a missing sidecar upgrade into an outage, and the warning plus the marker is what the operator needs to see it. The echo is informational, never trusted for data; it joins `CacheRelayReply`'s locked field list with a note that a cross-zone reader must not read an empty echo as a malformed reply. `TestKeyPrefixEchoNamesAServerThatIgnoresIt` stands a server in that drops the field: four replies, one warning, the marker on the debug state, serving continues; an unprefixed router draws nothing on the same server; the current server confirms without a word. The handler test covers hit, miss, sticky claim and sticky miss, and the shared keyspace echoing as empty.

2. The subject is `fix(cache)!:` and the body carries the `BREAKING CHANGE:` footer with the migration note (correct or remove the key; fix a chart that renders an extra key before upgrading), so `scripts/changelog-bump.sh` files it under the ⚠ block beside #396's and #401's; the PR body's checklist line says the same. On the chart: the enterprise chart repository renders no `resp-cache` key of any spelling in any file type today (grep over the whole tree), so nothing in-house is refused; a chart that starts rendering the block needs its keys checked against `redisstore.Config`, which the footer says.

3. Every handler that takes a prefix runs `core.ValidateKeyPrefix` on it and answers `InvalidArgument` before anything is read or written (5a171e7), so the grammar the client enforces holds where the keys are made; the empty prefix is not a prefix. `TestHandlersRefuseAnInvalidKeyPrefix` sends `a:ETH1` and `glob*` to all four and checks nothing landed under chain `ETH1:X` in keyspace `a`. The docs say what a prefix is not: a tenant boundary, since the sidecar has no authentication and the server refuses only a malformed prefix, never one that belongs to someone else.

Nits: a `cache-be-key-prefix` beside a `resp-cache` block now draws its own warning naming the gRPC prefix and the RESP keyspace in force, and the dangling shape keeps its own line (`TestSelectBackendWarnsWhenTheGRPCPrefixIsOutranked`); the flush section says the prefix is not a flush boundary on the sidecar.

Verified at 5a8ee5b from a detached checkout: `go build ./...`, `go vet` and `golangci-lint run --config .golangci.yml` over `./ecosystem/cache/... ./protocol/performance/... ./protocol/rpcsmartrouter/... ./types/relay/...` add nothing over `origin/main`'s findings; `go test -race -count=1` over the same packages passes; the keyspace, echo, debug-state and selection tests pass `go test -race -count=3` in all three packages. The rebase of the stack is mechanical: #400, #401 and #402 sit on 67d0e5c with trees identical to before outside `scripts/`, and #405 on this branch's tip with no conflict.

Not done: the "unknown keys are refused" clause on #396's two docs rows still waits for this PR to land, as noted there; and the `Health` capability, since the echo covers the same case per reply.

AnnaR-prog (approved):

avitenzer:
Rebased onto `main` after #396 merged there (3cf9bcf); the branch no longer carries #396's commits, so the conflict on `docs/RESP-CACHE.md` GitHub reported is gone. Tree identical to 5a8ee5b; tip is 8103bec. #405 follows on 90b90bd.

PR #398, opened 2026-09-18
fix(redisstore): trim both sides of the credential file, not only the right

avitenzer:
# Description

**Bug (MAG-3685).** A RESP password file beginning with a space or a blank line sent those characters as part of the credential. The store answered `WRONGPASS`, the auth path deliberately withholds the server's reply, and nothing pointed at the file's formatting. In the `username:password` form the leading whitespace landed on the username. The ticket's own correction stands: the file *was* trimmed, but on the right side only, which is what made the report easy to dismiss as already fixed.

**Fix.** `trimCredential` trims both sides. The trailing trade was already shipped, so the only credential that becomes inexpressible is one beginning with whitespace, which nobody writes on purpose. The `FileCredentials` doc comment described the intent rather than the code and now matches.

Files: `ecosystem/cache/redisstore/credentials.go` and its test.

## Tests

- `TestFileCredentialsTrimBothSides`: leading space, newline, tab and CRLF on each side through the parser, and the combined form with a leading space, proving the username comes out clean.
- `TestPasswordFileWithLeadingWhitespaceAuthenticates`: round trips against miniredis with `RequireAuth` and `RequireUserAuth`, so only the exact trimmed credential is accepted, plus a wrong-credential control so the server is the one deciding.

MAG-3726 in the automation repo is the binary-level check for the same defect (a password file that starts with a space, captured on the wire). Its cases 1 and 2 are strict expected failures naming MAG-3685, so once this merges that daily run turns red until those markers are removed; it accepts either remedy, and this PR takes the trim. Its controls (no space, trailing newline) keep passing because the right side is still trimmed.

Jira ticket: MAG-3685

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3685)

## Reviewers Checklist

I have...


AnnaR-prog:
## Review: `/code-review high` on `3776d5d`

Scope: the diff `main...3776d5d`, both callers of `trimCredential` (`FileCredentials.Credentials`, `Config.sentinelPassword`), `redisstore.New`'s fail-fast, the auth-failure log path in `protocol/performance/resp_cache.go`, and the operator docs. Each finding was re-verified against a worktree at this head; the verdict says how far.

**Stack note:** #399 and #407 are stacked on this branch (GitHub base `main`), so their "Files changed" include this diff until it merges.

### 1. Medium — the trailing trim is not "already the one that shipped": `TrimSpace` strips a wider right-hand set than the `TrimRight` it replaces — **CONFIRMED**

`ecosystem/cache/redisstore/credentials.go:80-85`

The old cutset was the four ASCII bytes `"\r\n \t"`; `strings.TrimSpace` is every Unicode `White_Space` rune, so `\v`, `\f`, U+0085, U+00A0 (NBSP), U+2028/9, U+3000 and the rest are now also stripped from the END of the credential. Probe at this head:

| file | `main` | this PR |
| --- | --- | --- |
| `"secret \n"` | `"secret "` | `"secret"` |
| `"secret\v\n"` | `"secret\v"` | `"secret"` |
| `"secret　"` | `"secret　"` | `"secret"` |

So the doc comment ("a credential ending in whitespace could never be expressed in this file") and the PR body ("the only credential that becomes inexpressible is one beginning with whitespace") both state a guarantee the code does not provide, on the side the PR says it did not touch. A deployment whose password ends in an NBSP (a paste artifact that a matching ACL was created with) authenticates on v1.5.2 and gets `WRONGPASS` after this upgrade — with the same withheld-reply diagnosis the PR is fixing, since `safeProbeDetail` (`resp_cache.go:172`) collapses it to "backend rejected the configured credentials". Rare, but the comment is what the next reader relies on. `strings.Trim(raw, "\r\n \t")` makes the claim literally true (byte-identical on the right, the same four bytes on the left); if `TrimSpace` is the intended contract, say "Unicode whitespace on both sides" and add one test row for the widened set so a revert to `TrimRight` cannot pass green — today's rows (`credentials_test.go:64-68`) pass identically against either implementation.

### 2. Medium — the failure class the PR closes is "invisible leading bytes", and three common members of it survive the trim — **CONFIRMED**

`ecosystem/cache/redisstore/credentials.go:55`, `:85`

All three verified at this head through `FileCredentials.Credentials()`:

- **UTF-8 BOM.** `unicode.IsSpace(0xFEFF)` is false, so a file saved by Windows Notepad ("UTF-8 with BOM") or PowerShell 5 `Out-File` reads back as `"﻿secret"`; in the combined form it lands on the username (`user="﻿cacheuser"`), where the `parsed-username` Info line renders it invisibly as `cacheuser`. Same for U+200B (Cf, not White_Space). `strings.TrimPrefix(raw, "﻿")` before the trim, or a once-only warning when the trimmed value still starts with a non-printable rune, closes it.
- **Whitespace next to the colon.** The trim is on the file's outer edges only, so the YAML-habit spelling `cacheuser: secret` gives `user="cacheuser", pass=" secret"` and `" user :pass"` gives `user="user "`. The only breadcrumb (`parsed-username=cacheuser`) then actively reassures the operator, because the stray byte rides on the never-logged half. The test comment at `credentials_test.go:101-102` ("one trim fixes both halves") holds only for file-edge whitespace. Either `TrimSpace` each half after `strings.Cut` (the same trade the PR already accepts for the whole file) with `"cacheuser: placeholder-credential\n"` as a test row — it fails at this head — or narrow the comment.
- **Empty after trim.** `" \n\t\n"` (and, new with this PR, an NBSP-only file) trims to `""`, `Credentials()` returns `(username, "", nil)`, `New()`'s fail-fast at `store.go:140` accepts it, and go-redis then sends `HELLO` with no `AUTH` at all: probe against miniredis with `RequireAuth` → `New` nil error, `Ping` → `NOAUTH Authentication required.`; against a server without `requirepass` it connects unauthenticated with `password-file` configured. Pre-existing for `\n`-only files, widened here. `password-file` was configured explicitly, so an empty result is never intended: reject it in the startup fail-fast naming the path (the refresh path should warn-and-keep-previous, since a Kubernetes mount can be transiently empty mid-rotation).

### 3. Low — `sentinel-password-file` changes meaning through the shared helper, unmentioned and untested — **CONFIRMED**

`ecosystem/cache/redisstore/config.go:196-200`

`sentinelPassword()` is the second caller of `trimCredential`, so the control-plane file also gets both-sides / Unicode trimming — desirable, but the PR body scopes the change to `credentials.go` and its test, the new `trimCredential` comment narrates the `username:password` form that this path never parses, and the only sentinel coverage (`TestFailoverOptionsMapping`, `config_test.go:156` / `:178`) uses a trailing newline. That value is captured once into `FailoverOptions.SentinelPassword` (restart-only rotation), so finding 1's trailing-NBSP case fails discovery before any data connection. `sentinelPassword` also hand-rolls `os.ReadFile` + `trimCredential`, which is `readCredentialFile` (`credentials.go:88-94`) minus the error wrap — folding it gives one reader for the empty-file / BOM checks above to land in, and one leading-whitespace row in `TestFailoverOptionsMapping` pins the sentinel plane.

### 4. Low — the trim contract is stated in Go comments only; the operator docs still document the `:` constraint and nothing else — **CONFIRMED**

`docs/RESP-CACHE.md:69`, `:71`

The `password-file` row and the rotation section spell out the colon limitation and say nothing about whitespace, so the class this PR makes inexpressible (leading), the pre-existing one (trailing), and the widened one (Unicode) live only in an unexported function's comment. MAG-3685 was itself an operator not knowing the trim rule. One sentence next to the colon constraint — "surrounding whitespace and newlines are trimmed; every other byte, including a BOM, is sent as part of the credential" — closes it, and the same sentence belongs on the sentinel row.

### Nits

- `store.go:137` / `:140`: `New` builds two `FileCredentials` (a third under sentinel, `config.go:387`), each with its own `warnedCombined`, so the "logged once" line fires twice at every startup with a combined-form file — visible in this PR's own `username:password file` subtest output. Pre-existing; `source := cfg.credentialsSource()` once fixes it.
- `credentials_test.go:63`: `TestFileCredentialsTrimBothSides` is `TestFileCredentialsParsing` (`:44`) with more rows, iterated from a `map` (random subtest order); one slice-backed table with `{name, content, username, wantUser, wantPass}` rows covers both plus the combined-form copy at `:79`.
- `credentials_test.go:125`: the control asserts bare `require.Error`, which a refused dial also satisfies; `require.ErrorContains(err, "WRONGPASS")` pins what the comment claims, and a `RequireUserAuth` control with a wrong username would pin the combined form's other half.

### Verified and not flagged

- The reported shape is fixed: leading space / blank line / tab / CRLF all reach the store trimmed, and both positive miniredis subtests fail against `main`'s `TrimRight` (build overlay), so the tests are not vacuous.
- `Refresh()` compares post-trim values, so a rotation whose only delta is now-trimmed whitespace no longer pushes — contrived enough to leave.
- The new tests pass at this head (`go test ./ecosystem/cache/redisstore/`).

avitenzer:
Fixed in 30a1a14. Two commits on top of 3776d5d: the synchronized log sink from #399 moves down to this branch (the tests here count the lines the credential reader writes, and the swap-style capture was the shape you flagged on #399; #399 inherits it and its own copy collapsed in the rebase), and the fix itself.

1. Taken the way you spelled it: `strings.Trim(raw, "\r\n \t")`. The cutset is the four ASCII bytes the right-hand trim always stripped, on both sides, so the right side is byte for byte what v1.5.x sent and the doc comment's claim is now true of the code. Not `TrimSpace`, for the reason you gave: a password ending in a non-breaking space authenticated before and would have answered `WRONGPASS` after the upgrade, behind the same withheld reply. Two rows pin the set, a trailing U+00A0 and a leading U+3000 that stay part of the credential; both fail against `TrimSpace`, so a widening cannot pass green.

2. All three members, and what follows from them. A leading UTF-8 byte order mark is dropped before the trim, in the one file reader (see 3), with rows for the password-only and combined forms. Each half of the combined form is trimmed on its own with the same cutset, so `cacheuser: placeholder-credential` is user `cacheuser` and the password without the space; your row is in the table and fails at 3776d5d. A file that is empty once trimmed is an error naming the path from `Credentials()`: `New` refuses it at startup on both credential files (`TestNewRefusesAnEmptyCredentialFile`), and `Refresh` treats it as a failed read that keeps the previous credentials (`TestRefreshKeepsPreviousCredentialsWhenTheFileIsEmpty`). On this branch that is the per-tick warning; #399's once-per-outage report lands on top after the rebase, and a test there (`TestEmptyCredentialFileIsOneOutage`) holds an empty file to one line, with connections opened meanwhile handed the last credentials read. And what the trim still cannot remove is said out loud: a credential that begins with a non-printable rune after all of that (U+200B, U+00A0) is sent as written and reported once, naming the file, the half and the code point, never the value (`TestFileCredentialsWarnOnceAboutAnInvisibleLeadingRune`: three reads, one line, the value absent from the capture).

3. `sentinelPassword` reads through the same `readCredentialFile`, so the BOM, the empty-file refusal and the invisible-rune report hold for the control-plane file too, wrapped with the setting's name. `TestFailoverOptionsMapping` now reads its file with leading whitespace, and `TestSentinelPasswordFileIsReadLikeTheDataNodeFile` pins the trim, the BOM and the refusal on that path. The PR body scopes the change to both files.

4. The `password-file` and `sentinel-password-file` rows say what is trimmed and what is kept, and the rotation section has a paragraph on it: ASCII whitespace and a leading BOM go, every other byte is sent, an invisible leading rune is reported, an empty file is refused at startup and a momentarily empty one mid-rotation is a failed read that keeps the previous credentials.

Nits. One credential set per store: `New` resolves the source, the provider and the sentinel password once (`clientCredentials`) and hands them to every client build, so the file is read once at startup and the combined-form line fires once; `TestCombinedFormIsReportedOncePerStartup` counts it through a real store and fails at 3776d5d with two. The two parsing tests are one slice-backed table with `{name, content, username, wantUser, wantPass}` rows, in order. The miniredis control asserts `WRONGPASS`, and a `RequireUserAuth` control with a wrong username pins the combined form's other half.

Verified: from a clean checkout of 30a1a14, `go build ./...`, `go vet` and `golangci-lint run --config .golangci.yml` over `./ecosystem/cache/... ./protocol/performance/...` (no findings beyond `main`'s), and `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/...` pass; the credential, watcher and rotation tests pass `go test -race -count=3`. With the new test file overlaid on 3776d5d, the space-after-the-colon and space-before-the-colon rows, both BOM rows, both cutset rows, every empty-file case, the startup refusal, the refresh case, the invisible-rune cases and the once-per-startup count all fail there and pass here; the rows that 3776d5d already fixed pass on both.

Not done: a change of the trim contract beyond the four bytes. A credential that begins with any other invisible rune is reported rather than trimmed, since the report is what the operator needs and a wider trim is the upgrade hazard from finding 1.

AnnaR-prog (approved):

avitenzer:
Rebased onto `main` after #396 merged there (3cf9bcf). The one conflict was `docs/RESP-CACHE.md`: #396 reworded the `username` / `password` row (the plaintext warning) and this branch reworded the neighbouring `password-file` row (the trim rule); both rows are kept as each side wrote them. Code merged clean. Tip is 0b07d9e; re-verified from a clean checkout (build, lint against `main`'s baseline, `go test -race` over the cache, performance, router and relay packages).

PR #399, opened 2026-09-18
fix(redisstore): report an unreadable credential file once per outage, not per tick

avitenzer:
# Description

**Bug (MAG-3690).** When the RESP password file becomes unreadable while the router runs, the refresh loop keeps the credentials it has, which is right, and warns about it on every tick with no end: 21 identical lines in 22 seconds at a one-second interval, about 8,600 a day at the shipped ten seconds. The health loop in the same feature wrote once over the same window, because it reports transitions.

**Fix.** `Refresh` reports the first failed read of an outage, stays silent for the rest of it, and reports the outage's end once. That answers the ticket's open question about what happens when the file comes back. A later outage is a new event. Nothing changes about what the router does with the credentials.

**Merge order.** Stacked on #398 (MAG-3685), which edits the same file and test file; merge that first. Until then this PR's diff shows that commit too.

Files: `ecosystem/cache/redisstore/credentials.go`, its test, and one paragraph in `docs/RESP-CACHE.md` under "Credential rotation".

## Tests

`TestRefreshReportsSourceFailureOncePerOutage` captures the log sink and checks, in order: a readable source logs nothing; 21 failed refreshes produce exactly one warning and no push; recovery with unchanged contents produces one "readable again" line and no push; a second outage warns again; a rotation that lands together with the recovery is both reported and applied.

The ticket's suggested binary-level test in the automation repo's `binary_startup` harness (delete the file mid-run, count warnings against the health-loop control) remains a good end-to-end check there.

## Out of scope

The ticket asks whether the "credential source currently unreadable" state should be visible somewhere other than the log, for example on `/debug/cache-state`. Not done here; it is a small follow-up on the same state flag if wanted.

Jira ticket: MAG-3690

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable; one log message reworded, one added)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3690)

## Reviewers Checklist

I have...


AnnaR-prog:
## Review: `/code-review high` on `7914fe7`

Scope: this PR's own commit (`3776d5d...7914fe7`, i.e. without #398's diff that the "Files changed" tab still shows), the whole `StreamingProvider` (`Subscribe`, `Refresh`, `watchCredentials`), `Store.Close`, and the RESP cache's health loop and per-operation error path in `protocol/performance/resp_cache.go`. Each finding was re-verified against a worktree at this head; the verdict says how far.

**Stack note:** stacked on #398 (base `main` on GitHub); #407 is stacked on this one.

### 1. Medium — "keeps the credentials it already holds" is true for live connections only; a connection opened during the outage fails, and nothing reports that at all — **CONFIRMED**

`ecosystem/cache/redisstore/credentials.go:133`, `docs/RESP-CACHE.md:184`

`Subscribe` re-reads the file for every connection go-redis opens, and returns the read error as the connection's init error. Probe at this head (miniredis with `RequireAuth`, `PoolSize: 4`, file removed after the first Ping): the pooled connection keeps serving (`Ping` nil, a `BLPOP` on it times out normally), and three of four concurrent commands that need a new connection fail with `open …/cred: no such file or directory`. So during an outage every pool growth, every reconnect after a network blip, and every idle-timeout replacement is a failed operation — and this PR's one warning is the only line the operator gets: the health probe pings on a pooled connection, so `/debug/cache-state` stays `reachable` and the transition log never fires, and per-operation failures go to `recordOpFailure` metrics only (`resp_cache.go:280`), never to the log.

The provider already tracks `lastRaw`; holding the last good `basicCredentials` beside it and having `Subscribe` return them when the read fails (through the same `refreshFailing` flag, so it still counts as one outage) makes the docs paragraph true for every connection, which is what "keeps the credentials it already holds" reads as. A `Subscribe` that fails before any successful read has nothing to serve and should keep erroring — that is the startup fail-fast's case anyway.

### 2. Low — `captureLog` swaps the process-global logger, the shape #400 just replaced after `-race` caught it — and this PR makes the watcher a goroutine that logs — **CONFIRMED**

`ecosystem/cache/redisstore/credentials_test.go:278`, `ecosystem/cache/redisstore/store.go:363`

#400's `14a4f03` ("capture logs through one synchronized sink instead of swapping the global logger") describes exactly this helper: a goroutine that logs for the life of its owner races the swap and restore of `zerologlog.Logger` and the unsynchronized buffer. Before this PR the credential watcher never logged from its own goroutine; after it, `watchCredentials → Refresh → noteRefreshFailure` does, and `Store.Close` closes `stopWatcher` without waiting for a tick in flight. The package passes `-race` today (full package once, the two watcher tests ×3) only because no current test overlaps a live watcher with a capture — a structural guard, not a verified one. Landing #400's `log_capture_test.go` shape here (a `TestMain` sink, capture attaches a buffer under a lock) is the smaller change; making `Close` join the watcher is the other. Once both chains merge, two `captureLog` helpers with different guarantees exist in two packages.

### Nits

- A change of cause mid-outage (mount gone → permission denied) is silent, since the flag is a bool; keying it on the error string is a one-line change if that distinction is worth a line.
- The recovery line could carry the outage length (`since` captured on the first failure) — the health loop's model does not, so consistent as is.

### Verified and not flagged

- 21 failed refreshes → one warning; recovery → one Info and no push; second outage → new warning; rotation landing with recovery pushed `u:pw2` — the test passes at this head and under `-race` ×3.
- No push during an outage: `Refresh` returns before `lastRaw` or the listeners are touched, and the state flag is read and written under `p.mu`.
- `refreshFailing` starts false, so a source that is unreadable from the first tick is reported on that tick (not swallowed as "already failing").

avitenzer:
Fixed in bd7cb37.

1. A connection opened during an outage now gets the last credentials any read returned. `Subscribe` and `Refresh` read through one `readSource`; the provider remembers the last set a read returned, and a `Subscribe` whose read fails hands that set out, registers the connection as usual so the rotation that ends the outage reaches it, and counts the failure towards the same outage report, so it is one warning whichever noticed first. Before any read has succeeded the error stands, which is `New`'s fail-fast at startup, as you suggested. The last set read is kept apart from the push baseline (`lastRaw`) on purpose: a connection that subscribes between a rotation landing on disk and the watcher's next tick reads the new set first, and the baseline has to stay on the old one for that tick to push to the connections that still have it; the test pins that ordering. The docs paragraph now says the credentials are kept on every connection, live and opened during the outage, and that a rotation landing with the recovery reaches both.

2. The swap-style `captureLog` is gone. `log_capture_test.go` installs one sink for the redisstore test binary in `TestMain` (the package had no other `TestMain`, internal or external; every test file is `package redisstore`), a capture attaches a buffer under a lock, records outside a capture go to stderr, and a nested capture fails the test that tries. I took your other option as well, since it was cheap and cannot deadlock: `Close` now waits for the watcher goroutine to return before releasing the clients. A tick is one file read plus, per live connection, a go-redis `OnNext` that only marks the connection for a background re-auth (`MarkForReAuth` takes a lock and writes a map entry), so the wait is bounded.

Nits: the outage report is keyed on the error's text, so a change of cause mid-outage (the file gone, then the path a directory) is one more line; the built-in sources fail with stable text, and the field comment says a custom source with varying text would report each variation. The recovery line stays without the outage length, as you said.

Verified: from a clean checkout of bd7cb37, `go build ./...`, `go vet` and `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/...` pass; the watcher, rotation and new tests pass `go test -race -count=3` (25 s). `TestSubscribeDuringSourceOutageHandsOutLastGoodCredentials` (the provider: newest set handed out, six subscriptions plus the watcher report one outage, the push that ends it reaches the connection that fell back, the error stands and is reported once before any successful read) and `TestConnectionsOpenedDuringSourceOutageAuthenticate` (go-redis against miniredis with the file removed: four blocking commands force three new dials, all four run, one warning) fail with the test file checked out against 4f250f9 and pass here. `TestCloseWaitsForTheCredentialWatcher` covers the join. golangci-lint reports the same four findings at this head as at 7914fe7, none in changed lines.

Not done: go-redis's dedicated `Conn()` could not be the instrument for the connection-level test, since at v9.22.0 it carries no streaming credentials manager and panics on init with a streaming provider; the test forces pool growth with concurrent blocking commands instead, and says so.

avitenzer:
Rebased onto #398's 30a1a14; tip is d7e0b56. The synchronized log sink now lives in #398, whose tests count the lines the credential reader writes, so this branch's commit that introduced it collapsed into nothing in the rebase and is gone; the once-per-outage report, the `Close` join and the last-good fallback carry through unchanged.

One behaviour changed underneath. A credential file that is empty once trimmed is a failed read since #398, so mid-rotation (a secret mount being rewritten) it is one warning per outage here, a connection opened meanwhile is handed the last credentials read rather than failed on the empty file, and the rotation that lands with the recovery is pushed to it too. `TestEmptyCredentialFileIsOneOutage` (d7e0b56) pins that: five refreshes and a subscribe against an empty file, one line naming the file, nothing pushed, then one recovery line and the push to both connections.

Verified at d7e0b56 from a clean checkout: `go build ./...`, `go vet`, `golangci-lint run --config .golangci.yml` (no findings beyond `main`'s) and `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/...` pass; the credential, watcher, rotation and outage tests pass `go test -race -count=3`.

AnnaR-prog (approved):

avitenzer:
Rebased onto #398's 0b07d9e after #396 merged to `main` (3cf9bcf); no content change of its own, clean replay. Tip is 4e69507; re-verified from a clean checkout.

avitenzer:
Rebased onto `main` after #398 merged there (6ce9cda); the branch no longer carries #398's commits, so the conflicts GitHub reported on `docs/RESP-CACHE.md` and `credentials_test.go` are gone. No content change of its own. Tip is 02184d2, re-verified from a clean checkout (build, lint against `main`'s baseline, `go test -race` over the cache, performance, router and relay packages).

PR #400, opened 2026-09-18
fix(redisstore): refuse a master-name whose topology line was forgotten

avitenzer:
# Description

**Bug (MAG-3671).** A `resp-cache` block with sentinel addresses and a `master-name` but no `topology: sentinel` line started as standalone, dialled the first sentinel address as a plain data node, and failed every cache operation with nothing naming the cause. `Validate()` already refused the opposite mistake and the sibling one (sentinel credentials under another topology); this half was missing. The ticket's follow-up run found the sharper part: the startup line printed the raw topology field, so an omitted topology showed as a blank and the trap configuration was byte for byte indistinguishable from one with no `master-name` at all.

**Fix.**
- `Validate()` refuses a `master-name` under any topology other than sentinel. The message names `master-name` and both ways out, per the ticket's caution that a refusal which does not name it would pass a "router refuses" test while leaving the operator as lost as before. Refuse rather than warn, as the reporter recommends: a `master-name` is read by nothing except the sentinel client, so there is no deployment where the operator wanted what they typed.
- The startup line prints the resolved topology (`EffectiveTopology()`, standalone when the field is empty) and the `master-name` beside it.

**Merge order.** Stacked on #396 (MAG-3683) because both add cases to the same validation function and its test matrix; merge that first. Until then this PR's diff shows that commit too.

Files: `ecosystem/cache/redisstore/config.go`, `protocol/performance/resp_config.go`, their tests, one docs row.

## Tests

- `TestConfigValidateMatrix`: `master-name` with topology omitted, and `master-name` on cluster, both asserting the message names `master-name`.
- `TestLoadRespCacheConfigRefusesMasterNameWithoutSentinelTopology`: the trap YAML through the real loader, with the documented sentinel block as the control that still loads.
- `TestSelectBackendLogsTheResolvedTopology`: captures the log sink and asserts the startup line says `topology=standalone` for a block that omits the field.

The automation test in MAG-3657 was written to hold under either remedy (refuse or warn) by asserting the two configurations are distinguishable; with this change they differ at startup, so its expected-failure marker should turn into a pass once this merges.

Jira ticket: MAG-3671

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable: a previously-silent misconfiguration now fails startup)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3671)

## Reviewers Checklist

I have...

---

Reviewed head [`f3f6dee70368`](https://github.com/Magma-Devs/smart-router/commit/f3f6dee70368398300bc3bbbdb14f5efbd6d1509). Findings below are from code inspection and local Go tests.

### [P2] Synchronize log capture with the health-loop lifetime

In [`captureLog`](https://github.com/Magma-Devs/smart-router/blob/f3f6dee70368398300bc3bbbdb14f5efbd6d1509/protocol/performance/backend_select_test.go#L26-L33), the process-global logger is replaced/restored while the cache created inside the callback has a background health probe. On this branch, `RespCache.Close()` does not wait for that probe to finish. The probe can therefore read the logger during its restoration, and the unsynchronized `bytes.Buffer` can also be accessed by competing goroutines.

Reproduced on this head with:

```bash
go test -race ./protocol/performance -run '^TestSelectBackendLogsTheResolvedTopology$' -count=1
```

The run failed with data races between `captureLog.func1` at line 29 and `RespCache.healthLoop → logUnavailable`. Please use synchronized capture and ensure the probe has exited before reading the captured buffer or replacing/restoring the global logger. The shutdown wait in #404 addresses part of the lifetime problem, but it is not in this PR's dependency chain.

### Resolution (2026-09-20)

Fixed in 14a4f03.

`captureLog` no longer swaps the global logger per capture. `TestMain` installs one synchronized sink for the test binary before any test starts a goroutine, and a capture only attaches a buffer to that sink under a lock; records written outside a capture go to stderr as before. There is nothing left to swap or restore while a probe runs, and every write and read of the captured buffer takes the same lock.

Verified: `go test -race ./protocol/performance -run '^TestSelectBackendLogsTheResolvedTopology$' -count=3` passes, and the whole package passes under `-race` on the chain tip (#402, 3cbe251). The probe-lifetime wait (`Close` cancelling the in-flight probe and waiting for the loop) lands with #404 on its own chain; with a single sink it is no longer needed for the capture to be race-free.


avitenzer:

Reviewed head [`f3f6dee70368`](https://github.com/Magma-Devs/smart-router/commit/f3f6dee70368398300bc3bbbdb14f5efbd6d1509). Findings below are from code inspection and local Go tests.

### [P2] Synchronize log capture with the health-loop lifetime

In [`captureLog`](https://github.com/Magma-Devs/smart-router/blob/f3f6dee70368398300bc3bbbdb14f5efbd6d1509/protocol/performance/backend_select_test.go#L26-L33), the process-global logger is replaced/restored while the cache created inside the callback has a background health probe. On this branch, `RespCache.Close()` does not wait for that probe to finish. The probe can therefore read the logger during its restoration, and the unsynchronized `bytes.Buffer` can also be accessed by competing goroutines.

Reproduced on this head with:

```bash
go test -race ./protocol/performance -run '^TestSelectBackendLogsTheResolvedTopology$' -count=1
```

The run failed with data races between `captureLog.func1` at line 29 and `RespCache.healthLoop → logUnavailable`. Please use synchronized capture and ensure the probe has exited before reading the captured buffer or replacing/restoring the global logger. The shutdown wait in #404 addresses part of the lifetime problem, but it is not in this PR's dependency chain.

avitenzer:
Fixed in 14a4f03.

`captureLog` no longer swaps the global logger per capture. `TestMain` installs one synchronized sink for the test binary before any test starts a goroutine, and a capture only attaches a buffer to that sink under a lock; records written outside a capture go to stderr as before. There is nothing left to swap or restore while a probe runs, and every write and read of the captured buffer takes the same lock.

Verified: `go test -race ./protocol/performance -run '^TestSelectBackendLogsTheResolvedTopology$' -count=3` passes, and the whole package passes under `-race` on the chain tip (#402, 3cbe251). The probe-lifetime wait (`Close` cancelling the in-flight probe and waiting for the loop) lands with #404 on its own chain; with a single sink it is no longer needed for the capture to be race-free.

AnnaR-prog:
## Review: `/code-review high` on `14a4f03`

Scope: this PR's two commits (`dcd32c4...14a4f03`, without #396's diff the "Files changed" tab still shows), `Config.Validate` as a whole, the flag overlay in `LoadRespCacheConfig`, the startup line, and what `/debug/cache-state` reports for the RESP tier. Re-verified against a worktree at this head.

**Stack note:** stacked on #396 (base `main` on GitHub); #401 → #402 are stacked on this one.

The rule and the log fix are right and small; nothing here blocks. The tests pass at this head.

### 1. Low — the ticket's "indistinguishable" finding is fixed at the startup line only; `/debug/cache-state` still cannot say which client was built — **CONFIRMED**

`ecosystem/cache/redisstore/store.go:212`, `protocol/performance/resp_cache.go:212`

`storeEndpoints` carries `Addresses`, `ReadAddresses`, `KeyPrefix`; neither the resolved topology nor `master-name` reaches `DebugCacheState`, so a running router cannot be asked whether it is talking to a sentinel quorum or dialling its first address as a data node — the startup line scrolls away, the endpoint does not. #402 adds `TLSInsecure` to the same struct for the same reason ("distinguishable from one running without"); `Topology`/`MasterName` belong in that change or this one.

### Nits

- `EffectiveTopology()` is `topology()` with an exported name (13 internal call sites); one function, exported, is less to keep in step than two bodies that must stay equal.
- `captureLog` says "captures do not nest" but nothing enforces it — a nested call silently drops the outer buffer's records. `require.Nil(t, logSink.capture)` before attaching makes the contract fail loudly. (#399, on the other chain, adds the swap-style `captureLog` in `redisstore` that this commit replaced here — flagged there.)
- The startup line now prints `master-name=` empty under standalone/cluster, which is what it should say; fine, just noting it is intentional.

### Verified and not flagged

- The rule fires for an omitted topology and for `cluster`; `sentinel` + `master-name` still passes; the earlier "sentinel topology requires master-name" check runs first, so the two messages are mutually exclusive.
- The flag overlay runs before `Validate`, so `--resp-cache-topology standalone` against a YAML sentinel block is refused as dangling, which is the right reading.
- `TestMain` is the only one in the package (internal and external test packages share the binary; the internal one has none).

avitenzer:
Fixed in 26f4851, on a branch rebased onto #396's 7051815. This PR is one fix commit again (bbc497d): the synchronized log sink from 14a4f03 now lives in #396, whose own warning test needed it first, so it is inherited here rather than carried.

1. `storeEndpoints` carries the resolved topology and the master set name, and the tier's `address` on `/debug/cache-state` renders them: `topology=` always (the resolved one, so an omitted line reads as the standalone it became) and `master=` under sentinel, between the read split and the keyspace (bbd8bf2). The `NewWithClient` seams see no config and print neither rather than guess. `TestConfiguredEndpointsNameTheTopology` pins the standalone, sentinel and seam renderings, `TestSelectBackendLogsTheResolvedTopology` asserts the debug state beside the startup line, and the docs' JSON example and the list under it say what the field carries.

Nits: `EffectiveTopology()` is the package's one accessor now; the unexported twin and its thirteen call sites are gone. The nesting contract of `captureLog` fails the test that tries, in #396's copy of the helper. The empty `master-name=` under standalone and cluster is intentional, as you read it.

One thing your run did not reach, because it lives in `protocol/rpcsmartrouter`: `TestRealCommand_FlagsOutrankYAML` overlaid a YAML sentinel block, `master-name` included, with `--resp-cache-topology cluster` and asserted the `master-name` survived. Under this branch's rule that merged configuration is dangling and refused, which is the reading you called right, so the test was red on f3f6dee and 14a4f03 (green on `main`) with nothing saying so, since CI does not run `go test`. 26f4851 overlays keys every topology reads (`key-prefix`, `pool-size`) and still proves the two flags outrank the block, and pins the old shape as `TestRealCommand_FlaggedTopologyLeavesYAMLMasterNameDangling`, refused through the real command.

Verified at 26f4851: `go build ./...`, `go vet`, `golangci-lint run --config .golangci.yml` (no findings beyond `main`'s), and `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/... ./protocol/rpcsmartrouter/... ./types/relay/...` pass; the endpoint and startup-line tests pass 3× under `-race`.

avitenzer:
Rebased onto #396's 67d0e5c after #396 moved there (main advanced to 970ab87); no content change. Tip is 0079640.

AnnaR-prog (approved):

avitenzer:
Rebased onto `main` after #396 merged there (3cf9bcf); the branch no longer carries #396's commits. No content change, tree identical to 0079640. Tip is 46e6554.

avitenzer:
Rebased onto `main` after #397 and #398 merged there. One conflict, in `config.go`: #398 changed `buildClient`'s signature (it now takes the credentials `New` resolves once) on the lines where this branch renames the topology accessor. Resolved by keeping the new signature and the rename; the `topology()` call #398 added in `resolveClientCredentials` is renamed with the rest, since this branch removes the unexported accessor. No behaviour change. Tip is 95d986e, re-verified from a clean checkout (build, lint against `main`'s baseline, `go test -race` over the cache, performance, router and relay packages).

PR #401, opened 2026-09-18
fix(redisstore): refuse a standalone address list longer than one

avitenzer:
# Description

**Bug (MAG-3672).** `resp-cache.addresses` is a list because sentinel and cluster take one. Standalone dials exactly one (`standaloneOptions` takes the first element) and dropped the rest in silence, while the startup line echoed every configured address back, so a spare written for redundancy read as confirmed and did nothing.

**Fix.** `Validate()` refuses more than one address under standalone, naming the addresses that would have been ignored and the two ways out (a discovering topology, or a single address). The same rule covers `read-addresses`, which the standalone builder truncates the same way. Of the ticket's three remedies this takes refusal: a warning would leave the operator with a deployment they misunderstand until it fails, and using the extras as alternatives is a feature that should be decided on, not fallen into.

The check deliberately runs after the `master-name` rule from #400: the MAG-3671 trap (several sentinel addresses, a `master-name`, no topology line) trips both, and the `master-name` message is the precise diagnosis of that mistake.

**Merge order.** Stacked on #400 (MAG-3671), which is stacked on #396 (MAG-3683); all three add cases to the same validation function and test matrix. Merge in that order. Until then this PR's diff shows those commits too.

Files: `ecosystem/cache/redisstore/config.go`, its test, `protocol/performance/resp_config_test.go`, one docs row.

## Tests

- `TestConfigValidateMatrix`: two addresses under explicit standalone, three under the defaulted topology, two `read-addresses` under standalone, each asserting the dropped address is named; sentinel and cluster lists as controls that still validate.
- `TestLoadRespCacheConfigRefusesExtraAddressesUnderStandalone`: the YAML list and the comma-separated `--resp-cache-addresses` flag both reach the rule; the same list under cluster loads.

The automation test in MAG-3658 pins current behaviour (only the first address receives anything, and the startup line echoes all of them) so that it goes red when this changes. With this PR the router refuses at startup instead, so that test needs updating to assert the refusal.

Jira ticket: MAG-3672

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable: a previously-truncated configuration now fails startup)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3672)

## Reviewers Checklist

I have...

---

Reviewed head [`aa75895b7f92`](https://github.com/Magma-Devs/smart-router/commit/aa75895b7f9282c1bdca84eff707fce82a634dcc). Findings below are from code inspection and local Go tests.

### [P2] Carry forward the log-capture fix from #400

This stacked head still includes the unsafe [`captureLog` helper](https://github.com/Magma-Devs/smart-router/blob/aa75895b7f9282c1bdca84eff707fce82a634dcc/protocol/performance/backend_select_test.go#L26-L33) and `TestSelectBackendLogsTheResolvedTopology` from #400. The health goroutine can outlive `Close()` and race restoration of the global logger. The focused `go test -race ./protocol/performance -run '^TestSelectBackendLogsTheResolvedTopology$' -count=1` run reproduced this on the introducing #400 head.

Please include the fix described in [#400's review findings](https://github.com/Magma-Devs/smart-router/pull/400#review-findings-2026-09-18) when updating this stack, then run that race check here. This is an inherited merge prerequisite; I found no additional defect in this PR's standalone address-count validation.

### Resolution (2026-09-20)

Carried forward. This branch is rebased onto #400's sink fix (14a4f03); its head is now b24e51a.

Verified here: `go test -race ./protocol/performance -run '^TestSelectBackendLogsTheResolvedTopology$' -count=3` passes, and the whole package passes under `-race` on the chain tip (#402, 3cbe251). No change to this PR's own address-count validation.


avitenzer:

Reviewed head [`aa75895b7f92`](https://github.com/Magma-Devs/smart-router/commit/aa75895b7f9282c1bdca84eff707fce82a634dcc). Findings below are from code inspection and local Go tests.

### [P2] Carry forward the log-capture fix from #400

This stacked head still includes the unsafe [`captureLog` helper](https://github.com/Magma-Devs/smart-router/blob/aa75895b7f9282c1bdca84eff707fce82a634dcc/protocol/performance/backend_select_test.go#L26-L33) and `TestSelectBackendLogsTheResolvedTopology` from #400. The health goroutine can outlive `Close()` and race restoration of the global logger. The focused `go test -race ./protocol/performance -run '^TestSelectBackendLogsTheResolvedTopology$' -count=1` run reproduced this on the introducing #400 head.

Please include the fix described in [#400's review findings](https://github.com/Magma-Devs/smart-router/pull/400#review-findings-2026-09-18) when updating this stack, then run that race check here. This is an inherited merge prerequisite; I found no additional defect in this PR's standalone address-count validation.

avitenzer:
Carried forward. This branch is rebased onto #400's sink fix (14a4f03); its head is now b24e51a.

Verified here: `go test -race ./protocol/performance -run '^TestSelectBackendLogsTheResolvedTopology$' -count=3` passes, and the whole package passes under `-race` on the chain tip (#402, 3cbe251). No change to this PR's own address-count validation.

AnnaR-prog:
## Review: `/code-review high` on `b24e51a`

Scope: this PR's commit (`14a4f03...b24e51a`, without the #396/#400 diff the "Files changed" tab still shows), `Config.Validate` end to end, the flag overlay, `standaloneOptions`, and the in-repo configs and docs examples. Re-verified against a worktree at this head; tests pass there.

**Stack note:** stacked on #400 → #396 (base `main` on GitHub); #402 is stacked on this one.

The rule is right, the ordering after the `master-name` rule is the correct call, and the message names exactly what would have been dropped. One thing to add, nothing to change.

### 1. Low — a two-address standalone block *worked* before this PR and is refused after it, with no release signal — **CONFIRMED**

`ecosystem/cache/redisstore/config.go:183`

Unlike #400's trap (a `master-name` without its topology never served a request), a standalone block with a spare address served every request from its first element on v1.5.x. After this PR that deployment fails startup on upgrade — the right outcome, and the message tells the operator what to do — but the subject has no `!` and the body no `BREAKING CHANGE:` footer, which are the only signals `scripts/changelog-bump.sh` reads (`:143`, `:167`), so the release notes file it under Bug fixes with no ⚠ block. Same point as on #396; the two should carry the footer consistently, since they land as one release. The PR body's own last paragraph (MAG-3658 pins the old behaviour and must be flipped) is the evidence that behaviour changed.

### Verified and not flagged

- The rule sits after the `master-name` rule, so the MAG-3671 shape (several sentinel addresses + `master-name`, no topology line) gets the `master-name` diagnosis, not this one.
- Both the YAML list and a comma-separated `--resp-cache-addresses` reach the rule (the overlay replaces `Addresses` before `Validate`); the same list under `cluster`/`sentinel` still loads.
- No in-repo lane or docs example lists more than one address under standalone; every multi-address example is sentinel or cluster.
- `read-addresses` under standalone is truncated the same way by the builder, so refusing it under the same rule is consistent, and the docs caveat under "Multi-region reads" (one reader endpoint) already matches.

avitenzer:
Fixed in bdd8203, rebased onto #400's 26f4851. Message-only change to this branch's one commit: the subject is `fix(redisstore)!:` and the body carries the `BREAKING CHANGE:` footer naming the shape that served on v1.5.x and fails now (more than one entry under `addresses` or `read-addresses` on a standalone block) and the two ways out, so `scripts/changelog-bump.sh` files it under the ⚠ block beside #396's. The PR body's checklist line says the same.

The rebase adapted one line: the rule now calls `EffectiveTopology()`, since #400 collapsed the two accessors into that one. Nothing else in the diff moved.

Verified at bdd8203: `go build ./...`, `go vet`, `golangci-lint run --config .golangci.yml` (no findings beyond `main`'s), and `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/... ./protocol/rpcsmartrouter/... ./types/relay/...` pass.

avitenzer:
Rebased onto #400's 0079640 after #396 moved to 67d0e5c (main advanced to 970ab87); no content change. Tip is fbae165.

AnnaR-prog (approved):

avitenzer:
Rebased onto #400's 46e6554 after #396 merged to `main` (3cf9bcf); no content change, tree identical to fbae165. Tip is 1b6062c.

avitenzer:
Rebased onto #400's 95d986e after #397 and #398 merged to `main`; no content change of its own. Tip is 051963a, re-verified from a clean checkout.

PR #402, opened 2026-09-18
fix(redisstore): say out loud when certificate verification is off

avitenzer:
# Description

**Bug (MAG-3684).** `resp-cache.tls.insecure-skip-verify` was accepted in silence: no startup line named it and the debug endpoint's configuration report did not carry it, so a production deployment running without certificate verification looked exactly like one running with it. It is the kind of setting that is set once for a development environment and then travels.

**Fix.** Three surfaces now name it:
- a warning at startup, once, saying the connection is encrypted but the backend's identity is unchecked;
- the "resp-cache backend configured" line, which already reports whether TLS is on, carries `tls-insecure-skip-verify` beside it;
- the configured endpoints that `GET /debug/cache-state` renders as the tier's `address` gain `tls=insecure-skip-verify`, which answers the ticket's open question about whether the debug endpoint can report it without a new field.

**Merge order.** Stacked on #401 (MAG-3672) → #400 (MAG-3671) → #396 (MAG-3683): this PR edits the startup line that #400 changed, and the same docs table. Merge in that order. Until then this PR's diff shows those commits too.

Files: `protocol/performance/resp_config.go`, `ecosystem/cache/redisstore/store.go`, their tests, one docs row.

## Tests

- `TestSelectBackendWarnsOnInsecureSkipVerify`: captures the log sink through the real backend selection and asserts the warning, the configured-line attribute and the debug-state marker; the verifying configuration is the control that none of the three appear for.
- `TestConfiguredEndpointsNameInsecureTLS`: the rendered endpoint string with and without the setting.

The ticket suggests the automation repo's `binary_startup` harness, which already waits for named startup lines, as the end-to-end check; the warning text to wait for is `insecure-skip-verify is set`.

Jira ticket: MAG-3684

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable: one warning and one attribute added, one debug-state string extended)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3684)

## Reviewers Checklist

I have...

---

Reviewed head [`7e17b44f0e13`](https://github.com/Magma-Devs/smart-router/commit/7e17b44f0e1389f078995d14b915d60507940777). Findings below are from code inspection and local Go tests.

### [P2] Stop background probes before switching the captured logger

The new [warning test](https://github.com/Magma-Devs/smart-router/blob/7e17b44f0e1389f078995d14b915d60507940777/protocol/performance/backend_select_test.go#L183-L195) starts two TLS-configured caches against a plaintext miniredis server. Their background probes can fail and log while `captureLog` replaces/restores the global logger or reads its unsynchronized buffer. The first cache also remains alive when the second capture begins.

Reproduced independently on this head with:

```bash
go test -race ./protocol/performance -run '^TestSelectBackendWarnsOnInsecureSkipVerify$' -count=1
```

The test failed with data races involving `captureLog.func1` and `RespCache.healthLoop`. Please use synchronized capture and wait for each cache's probe goroutine to exit before switching/restoring the logger. This shares the helper/lifetime issue reported on #400; the new test needs to pass the race check after that fix is carried into this stack.

### Resolution (2026-09-20)

Fixed in 3cbe251, on top of the #400 sink fix (14a4f03) carried through #401.

The warning test now closes the first cache before the second capture starts, so each capture sees only its own cache's startup; the synchronized sink from #400 makes any overlap harmless to the race detector anyway.

Verified: `go test -race ./protocol/performance -run '^TestSelectBackendWarnsOnInsecureSkipVerify$' -count=3` passes, and the whole package passes under `-race` on this head.


avitenzer:

Reviewed head [`7e17b44f0e13`](https://github.com/Magma-Devs/smart-router/commit/7e17b44f0e1389f078995d14b915d60507940777). Findings below are from code inspection and local Go tests.

### [P2] Stop background probes before switching the captured logger

The new [warning test](https://github.com/Magma-Devs/smart-router/blob/7e17b44f0e1389f078995d14b915d60507940777/protocol/performance/backend_select_test.go#L183-L195) starts two TLS-configured caches against a plaintext miniredis server. Their background probes can fail and log while `captureLog` replaces/restores the global logger or reads its unsynchronized buffer. The first cache also remains alive when the second capture begins.

Reproduced independently on this head with:

```bash
go test -race ./protocol/performance -run '^TestSelectBackendWarnsOnInsecureSkipVerify$' -count=1
```

The test failed with data races involving `captureLog.func1` and `RespCache.healthLoop`. Please use synchronized capture and wait for each cache's probe goroutine to exit before switching/restoring the logger. This shares the helper/lifetime issue reported on #400; the new test needs to pass the race check after that fix is carried into this stack.

avitenzer:
Fixed in 3cbe251, on top of the #400 sink fix (14a4f03) carried through #401.

The warning test now closes the first cache before the second capture starts, so each capture sees only its own cache's startup; the synchronized sink from #400 makes any overlap harmless to the race detector anyway.

Verified: `go test -race ./protocol/performance -run '^TestSelectBackendWarnsOnInsecureSkipVerify$' -count=3` passes, and the whole package passes under `-race` on this head.

AnnaR-prog:
## Review: `/code-review high` on `3cbe251`

Scope: this PR's two commits (`b24e51a...3cbe251`, without the #396/#400/#401 diff the "Files changed" tab still shows), the three surfaces it touches, every caller of `redisstore.New`, and what consumes the debug endpoint's `address` string (including the dashboard repo). Re-verified against a worktree at this head; tests pass there.

**Stack note:** stacked on #401 → #400 → #396 (base `main` on GitHub). This closes the `insecure-skip-verify`-only hole noted on #396.

Correct and complete for what it sets out to do; two placement nits, no fix requests.

### Nits

- **The warning lives in the loader, not the store.** `SelectCacheBackend` is the only caller of `redisstore.New` today, so nothing is missed now; but `TLSConfig.build()` is where the opt-in is honoured (`//nolint:gosec`) and `warnIfReadSplitIsDiscoveryScoped` (`store.go`) is the in-package precedent for a construction-time posture warning. Emitting it there keeps the warning with the thing it warns about for any future caller (the cache server, a tool), and the loader's line keeps its attribute either way.
- **The debug marker overloads a display string.** `address` is now `addr [read=…] [prefix=…] [tls=insecure-skip-verify]`. Nothing parses it today (the dashboard repo does not read `/debug/cache-state` at all), so this is a design note rather than a bug: a boolean field on the tier is what a future consumer would want, and the ticket's "without a new field" was a question, not a constraint. If the string stays, the JSON example at `docs/RESP-CACHE.md:239` is the place to show the marker.

### Verified and not flagged

- The three surfaces fire together for the insecure configuration and none of them for the verifying one (`TestSelectBackendWarnsOnInsecureSkipVerify`); the endpoint string test covers both renderings.
- `TLSInsecure` is `Enabled && InsecureSkipVerify`, while the startup attribute prints the raw field; after #396's rule the two cannot differ, since `insecure-skip-verify` without `enabled: true` is refused before either line runs.
- The second commit (close the first cache before the second capture) is exactly the probe-outlives-capture race #400's sink exists for; correct fix for it.
- The docs row is accurate; no in-repo lane sets the flag except the remote demo lanes, which are the intended audience of the warning.

avitenzer:
Fixed in ca406d8, rebased onto #401's bdd8203.

The warning moved into the store (ca406d8): `warnIfTLSSkipsVerification` sits in `redisstore.New` next to the plaintext-credential warning #396 added, which is the same kind of construction-time posture warning and now the same place, so a second caller of `New` gets both; the loader's "backend configured" line keeps its `tls-insecure-skip-verify` attribute, and `TestSelectBackendWarnsOnInsecureSkipVerify` still drives it through `SelectCacheBackend`. The `address` marker stays a string: nothing parses the field today, and the JSON example in the docs now shows `tls=insecure-skip-verify` on the example address, with the note that its presence is the finding. The exact-string assertion in `TestConfiguredEndpointsNameInsecureTLS` gained the `topology=standalone` that #400 now renders on the same string.

Verified at ca406d8: `go build ./...`, `go vet`, `golangci-lint run --config .golangci.yml` (no findings beyond `main`'s), and `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/... ./protocol/rpcsmartrouter/... ./types/relay/...` pass; the two warning tests and the resolved-topology test pass 3× under `-race`.

avitenzer:
Rebased onto #401's fbae165 after #396 moved to 67d0e5c (main advanced to 970ab87); no content change. Tip is 378284b.

AnnaR-prog (approved):

avitenzer:
Rebased onto #401's 1b6062c after #396 merged to `main` (3cf9bcf); no content change, tree identical to 378284b. Tip is 639f1fd.

avitenzer:
Rebased onto #401's 051963a after #397 and #398 merged to `main`; no content change of its own. Tip is d103759, re-verified from a clean checkout.

PR #403, opened 2026-09-18
fix(redisstore): purge the read endpoint too when reads are split

avitenzer:
# Description

**Bug (MAG-3673).** With `read-addresses` pointing reads at a separate store, `/debug/reset-all` scanned and unlinked through the write client only. A read store the write endpoint never feeds kept every entry and accumulated across resets (the ticket measured 1, 2, 3 keys surviving three rounds), so stale answers kept being served after the one remedy an operator reaches for. `Ping`, `PoolStats` and `Close` already touch both clients; `Purge` was the one operation that did not, which is the ticket's own argument that this was an oversight.

**Fix.** `Store.Purge` scans and unlinks under the prefix on the read endpoint as well, per master under cluster like the write side.
- A read endpoint that is a read-only **replica** of the write endpoint answers `READONLY` to the unlink. That is left to replication, which the write-side purge reaches it through; it is noted at debug level and the purge succeeds. This is why the write-only behaviour may have looked deliberate, and it is now handled rather than assumed.
- Any other failure on the read side is a failed purge, named as such: a reset that silently did half the job is the defect.

**Contract flip, stated so it is reviewed.** `TestReadWriteRouting` pinned "purge is a write-side operation: the read endpoint is out of scope" as intended behaviour. This PR flips that assertion.

Branched from `main`; it touches a different region of the store than the other open RESP PRs and merges cleanly against them.

Files: `ecosystem/cache/redisstore/store.go`, its test, a comment in `protocol/performance/resp_cache.go`, one docs paragraph under "Multi-region reads".

## Tests

- `TestPurgeEmptiesASeparateReadStore`: the ticket's measurement replayed against two miniredis stores, three rounds, with its two controls (the read store holds the entry immediately before each purge; a key outside the prefix survives on both stores).
- `TestPurgeToleratesAReadOnlyReplica`: a client hook answering `READONLY` to every unlink stands in for a replica; the purge succeeds.
- `TestPurgeReportsAnUnreachableReadStore`: the read store is closed; the purge fails naming the read side.
- `TestReadWriteRouting`: the flipped assertion.

The automation test for this ticket (`test_a_reset_empties_the_read_store_as_well_as_the_write_store`, strict xfail naming MAG-3673) should pass once this merges, so its marker needs removing then.

Jira ticket: MAG-3673

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable: a reset now reaches the read endpoint it was documented to reach)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3673)

## Reviewers Checklist

I have...


AnnaR-prog:
## Review: `/code-review high` on `61b7c2b`

Scope: the full diff `main...61b7c2b`, `Purge`/`purgeClient`/`purgeByScan`, the three client shapes the read side can take, and the docs section the paragraph joins. Beyond the PR's miniredis tests, the replica path was run against a real primary/replica pair (two `valkey/valkey:7.2` containers, `--replicaof`). Verdicts say how far each claim was checked.

Branch is standalone (not stacked). The fix is right for both shapes the read endpoint can have, and the real-replica run confirms the READONLY handling: `UNLINK` on the replica answers `READONLY You can't write against a read only replica.`, `redis.HasErrorPrefix` matches it, `Purge` returns nil, and every `sr:*` key is gone from the replica once replication catches up.

### 1. Low — in the documented shape (a managed reader endpoint) the read-side purge is a full keyspace `SCAN` of the replica for no effect, on every reset — **CONFIRMED**

`ecosystem/cache/redisstore/store.go:769`

Measured on the replica during one `Purge` (3,000 keys under the prefix, 20,000 foreign keys, i.e. a shared backend): `cmdstat_scan calls=40` — the whole 23,001-key keyspace was walked — and `rejected_calls` did not move, because the write-side unlinks had already replicated before the read-side scan began, so the scan found nothing to queue and the READONLY branch was never reached. On a lagging replica the branch fires, after the same scan. `SCAN` walks the whole keyspace regardless of `MATCH`, so the cost is O(keys on the shared reader) per reset: ~20,000 round trips for a 10M-key reader at `scanBatchSize` 512, on the read path. `/debug/reset-all` is an operator action, so this is cost, not a defect — but it is paid in the shape the docs call "the shape this feature is for", and nothing tells the operator. Cheap to avoid: one `ROLE` (or `INFO replication`) on the read client first, skip the scan when it reports a replica of the write endpoint and say so at debug level, keeping the READONLY branch as the fallback for anything `ROLE` does not settle. Alternatively keep it and state the cost in the new docs paragraph.

### Nits

- `docs/RESP-CACHE.md:146` still says "writes (including cache population and flush) to `addresses`", two paragraphs above the new "Resets reach both endpoints"; the older sentence now describes the pre-PR behaviour.
- `TestPurgeToleratesAReadOnlyReplica` stands in a hook for the replica; the real error text is `READONLY You can't write against a read only replica.` — worth using verbatim in the hook so the prefix match is pinned against the wire form.

### Verified and not flagged

- Separate-store shape: three rounds against two independent stores empty the read store each time, with the prefix-isolation control (the PR's test, re-run here).
- Batch semantics: `pipe.Exec` returns the first error of the batch, so a batch of READONLY replies is one READONLY, and an empty batch (nothing matched) is not an error.
- Partial failure: the write side is purged before the read side is attempted, so a failed read-side purge leaves the write endpoint clean and a retry is idempotent.
- Cluster and sentinel read clients resolve to writable masters, so READONLY can only come from the standalone reader shape, which is where the branch sits.

avitenzer:
Fixed in 3eb7c7b.

1. Before the read-side scan the store now asks the read endpoint what it is: `ROLE` first, `INFO replication` where `ROLE` is not answered, and one that reports itself a replica is not scanned at all. The master it names, and whether that is the configured write address, are logged at debug level; the comparison is textual, and a replica names its master by IP where an operator may have written a hostname, so a mismatch there is a hint rather than a finding. A read endpoint that answers neither is scanned as before, with `READONLY` still tolerated at the unlink, and a cluster read client is never asked, since it resolves to writable masters and is scanned per master. Of the two options offered I took the role query rather than stating the cost. miniredis answers neither command, so the replica stand-in in the tests now answers `ROLE` and `INFO` as configured and counts the `SCAN`s that reach it: `TestPurgeDoesNotScanAReaderThatReportsItselfAReplica` covers the `ROLE` answer, the `INFO` fallback, and a reader that calls itself a master (scanned and emptied, the MAG-3673 shape); `TestPurgeToleratesAReadOnlyReplica` now also asserts that a reader which could say nothing was scanned; `TestReplicaOfParsesRoleAndInfoReplies` pins both parsers.

Nits. The paragraph under "Multi-region reads" now sends writes (cache population) to `addresses` and points at the reset paragraph, which describes the role query. The hook already carried the verbatim `READONLY You can't write against a read only replica.`; the constant now says it is the wire form, so nobody shortens it.

Verified: `go test -race -count=3 ./ecosystem/cache/redisstore -run 'Purge|ReplicaOf|ReadWriteRouting'` passes; `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/... ./protocol/rpcsmartrouter/...` passes at this head; `go build ./...`, `go vet` and `golangci-lint run` on those packages add nothing over the branch's baseline.

AnnaR-prog (approved):

avitenzer:
Rebased onto `main` after #397 and #398 merged there. One conflict, in `store.go`'s import block: #397 moved the key-prefix validation into `core` and dropped `regexp` here, and this branch adds `net` beside it. Resolved by keeping `net` alone. No behaviour change. Tip is 7cae551, re-verified from a clean checkout (build, lint against `main`'s baseline, `go test -race` over the cache, performance, router and relay packages).

PR #404, opened 2026-09-18
fix(performance): name which half of a split cache failed its probe

avitenzer:
# Description

**Bug (MAG-3674).** With reads and writes split, the health probe reported the first failure it met into one unlabelled gauge and one unlabelled counter, so taking either half down gave an identical reading. The router knew which endpoint failed (its debug route named the address) and did not publish it where an alert reads.

**Fix.**
- `Store.Probe` pings each endpoint on its own and reports each by role (`write` | `read`); `Ping` keeps its first-failure contract on top of it.
- Two new labelled series beside the existing pair: `smartrouter_resp_cache_endpoint_connected{role}` and `smartrouter_resp_cache_endpoint_connection_errors_total{role}`. A read outage never counts against the write endpoint, and vice versa.
- The debug state's `detail` and the transition log name the failing endpoint by role and configured address.

**Design call, stated so it can be challenged.** The ticket left open whether a label or separate series is wanted. This adds labelled series and leaves the existing unlabelled pair unchanged as the whole-cache verdict, so every current alert on `smartrouter_resp_cache_connected == 0` keeps its meaning. Relabelling the existing gauge would have turned one series into two under every existing dashboard.

**A second fix that fell out of the test.** `RespCache.Close` returned while a probe could still be running. That probe then overwrote the "closed" health that `Close` had just published (the exact thing the comment in `Close` says it guards against), and its counter increments landed on whichever cache came next in a test binary sharing the registry. `Close` now cancels the probe context and waits for the loop to exit.

**Merge order.** Stacked on #403 (MAG-3673): both touch the store and share the two-store test fixture. Merge that first. Until then this PR's diff shows that commit too.

Files: `ecosystem/cache/redisstore/store.go`, `protocol/performance/resp_cache.go`, `protocol/performance/resp_metrics.go`, their tests, `docs/METRICS.md`, `docs/RESP-CACHE.md`.

## Tests

- `TestRespCacheHealthNamesTheFailingEndpoint`: two stores; the read half is taken down, then recovery is confirmed before the write half is taken down (the ticket's own trap about the snapshot). For each outage the failing half's gauge flips and its counter rises, the healthy half's do not, the whole-cache gauge still reads down, and the debug detail names the half by address. Run three times in a row to check for flakiness.
- `TestProbeReportsEachEndpointSeparately`: the store-level contract, including the single-endpoint case.

The automation test for this ticket (`test_the_metrics_differ_depending_on_which_half_of_the_cache_is_down`, strict xfail naming MAG-3674, MAG-3665) compares metric names and labels between the two outages. With this PR the labelled series differ by `role`, so it should pass once this merges and its marker needs removing.

Jira ticket: MAG-3674

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable: two series added, none changed)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3674)

## Reviewers Checklist

I have...

---

Reviewed head [`84de552af61a`](https://github.com/Magma-Devs/smart-router/commit/84de552af61ade9f962b6cdf754aafe270d12ce5). Findings below are from code inspection and local Go tests.

### [P2] Give each endpoint a probe that can run before the deadline

[`Store.Probe`](https://github.com/Magma-Devs/smart-router/blob/84de552af61ade9f962b6cdf754aafe270d12ce5/ecosystem/cache/redisstore/store.go#L371-L382) runs the write PING first and then the read PING with the same context. If the writer stalls until the health loop's three-second deadline, the reader receives an already-expired context. A healthy reader is then reported down and its error counter increments too, defeating the endpoint-specific diagnosis this PR adds.

A focused local regression held the write PING until an 80 ms probe deadline while the reader answered PING successfully both before and after the probe. The returned read result was nevertheless `context deadline exceeded`.

Please start both probes concurrently within the overall deadline, or provide independent bounded budgets. Add a regression for a blackholed writer plus a healthy reader; the current closed-port test fails quickly and does not exercise deadline exhaustion.
Docker confirmation (Valkey 7.2): reproduced the same failure without a mock or injected Redis hook. With separate writer/reader containers, `CLIENT PAUSE 750 ALL` on only the writer and an 80 ms probe context returned a writer I/O timeout and a reader `context deadline exceeded`; the reader answered a separate PING successfully immediately afterward. The split-endpoint purge positive control passed and preserved keys under a foreign prefix.

### Resolution (2026-09-20)

Fixed in 4ad47ad.

`Store.Probe` now pings every endpoint on its own goroutine within the caller's one deadline, so a stalled writer cannot spend the budget a healthy reader needed: each result is that endpoint's own verdict, and `Ping` still folds them to the first failure.

Regression: `TestProbeDoesNotBlameAHealthyReaderForAStalledWriter` runs a writer that accepts connections and never answers next to a healthy miniredis reader, under an 80 ms probe context. The writer is reported down and the reader is not. It fails on the sequential probe (the reader's verdict was `context deadline exceeded`) and passes now; `go test ./ecosystem/cache/redisstore ./protocol/performance` pass. #406 is rebased onto this head.


avitenzer:

Reviewed head [`84de552af61a`](https://github.com/Magma-Devs/smart-router/commit/84de552af61ade9f962b6cdf754aafe270d12ce5). Findings below are from code inspection and local Go tests.

### [P2] Give each endpoint a probe that can run before the deadline

[`Store.Probe`](https://github.com/Magma-Devs/smart-router/blob/84de552af61ade9f962b6cdf754aafe270d12ce5/ecosystem/cache/redisstore/store.go#L371-L382) runs the write PING first and then the read PING with the same context. If the writer stalls until the health loop's three-second deadline, the reader receives an already-expired context. A healthy reader is then reported down and its error counter increments too, defeating the endpoint-specific diagnosis this PR adds.

A focused local regression held the write PING until an 80 ms probe deadline while the reader answered PING successfully both before and after the probe. The returned read result was nevertheless `context deadline exceeded`.

Please start both probes concurrently within the overall deadline, or provide independent bounded budgets. Add a regression for a blackholed writer plus a healthy reader; the current closed-port test fails quickly and does not exercise deadline exhaustion.
Docker confirmation (Valkey 7.2): reproduced the same failure without a mock or injected Redis hook. With separate writer/reader containers, `CLIENT PAUSE 750 ALL` on only the writer and an 80 ms probe context returned a writer I/O timeout and a reader `context deadline exceeded`; the reader answered a separate PING successfully immediately afterward. The split-endpoint purge positive control passed and preserved keys under a foreign prefix.

avitenzer:
Fixed in 4ad47ad.

`Store.Probe` now pings every endpoint on its own goroutine within the caller's one deadline, so a stalled writer cannot spend the budget a healthy reader needed: each result is that endpoint's own verdict, and `Ping` still folds them to the first failure.

Regression: `TestProbeDoesNotBlameAHealthyReaderForAStalledWriter` runs a writer that accepts connections and never answers next to a healthy miniredis reader, under an 80 ms probe context. The writer is reported down and the reader is not. It fails on the sequential probe (the reader's verdict was `context deadline exceeded`) and passes now; `go test ./ecosystem/cache/redisstore ./protocol/performance` pass. #406 is rebased onto this head.

avitenzer:
Follow-up aca2110: the stalled-writer probe test moved from the end of the file to beside the other probe test, so this branch merges cleanly with #402, which appends to the same file end. No behaviour change.

AnnaR-prog:
## Review: `/code-review high` on `aca2110`

Scope: this PR's three commits (`61b7c2b...aca2110`, without #403's diff the "Files changed" tab still shows), `Store.Probe`/`Ping`, the whole health loop and `Close`, the metric registrations, and both docs pages. Re-verified against a worktree at this head; the health tests pass 3× under `-race` (51 s).

**Stack note:** stacked on #403 (base `main` on GitHub); #406 is stacked on this one.

Nothing blocking. The design call (labelled series beside the unlabelled pair, not a relabel) is the right one for existing alerts, the concurrent per-endpoint probe fixes the stalled-writer diagnosis the Codex review raised, and the `Close` fix is correct: cancel, wait on `healthDone`, and both probe sites return without publishing once `loopCtx` is done, so the closed snapshot is the last word.

### Nits

- `logUnavailable` picks the auth wording and the `failure` attribute from the **first** failing result only. Writer network-down plus reader auth-rejected logs `failure=unreachable` with a detail that says the reader rejected credentials — the detail is right, the classification is half right. Classifying per result and promoting auth when any endpoint reports it keeps the attribute truthful for the case this PR exists for (two halves, two faults).
- `endpointConnected{role="read"}` is never initialised on a non-split store, so the series is absent rather than 1; the METRICS row says so, which is enough — noting it because `absent()`-style alerts on the read series cannot distinguish "no split configured" from "never probed".

### Verified and not flagged

- `Ping` keeps its first-failure contract on top of `Probe` (write's error wins, as before), and `Probe` is the only non-test caller of the clients' `Ping`.
- Counter arithmetic is as documented: one whole-cache increment per failed probe, one per failed endpoint; `sum(endpoint) >= whole` by construction.
- `safeProbeDetail(nil)` already returned "no error reported", so the healthy `detail` on `/debug/cache-state` is byte-identical before and after.
- Configured addresses (never credentials) are what the detail carries, consistent with the docs' promise about `password-file`.
- `Probe` writes `results[i].Err` from one goroutine per index and reads after `Wait`; no shared writes.

avitenzer:
Fixed in e2b5054, on a branch rebased onto #403's 3eb7c7b.

Nit 1. `logUnavailable` now classifies the probe from every failing result: an authentication rejection on either half wins, since it needs the operator action the class exists to name, and the detail is unchanged and names both. `TestProbeClassificationPromotesAuthFromEitherEndpoint` pins the case from the review, a writer refused beside a reader that rejected the credentials, in both orders, with two network faults and a probe with no failure as controls, and checks the detail still carries neither the username nor the password.

Nit 2, `endpointConnected{role="read"}` absent on a store without a read split: left as it is, with the METRICS row saying so, as you noted was enough.

Verified: `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/...` passes at this head, and the health and probe tests pass three times in a row under `-race` on #406's head, which sits on this one. `go build ./...`, `go vet` and `golangci-lint run` add nothing over the branch's baseline.

AnnaR-prog (approved):

avitenzer:
Rebased onto #403's 7cae551 after #397 and #398 merged to `main`; no content change of its own. Tip is 74c85e6, re-verified from a clean checkout.

avitenzer:
Rebased onto the current `main`, with one change to the commit layout: the branch merged cleanly as a whole, but replaying it commit by commit stopped at "probe every endpoint within the same deadline", which appended its regression test to the end of `store_test.go` where the tests of #402 and #403 now live; the follow-up commit that moved the test beside its sibling is folded into it (the commit message says so), so the branch is three commits that replay onto `main` without conflict. That is what GitHub's "Update with rebase" and "Rebase and merge" both do, which is why they reported a conflict while the merge preview was clean. The tree is byte-identical to the clean three-way merge of the previous tip. Tip is 72280b2f, re-verified from a clean checkout (build, lint against `main`'s baseline, `go test -race` over the cache, performance, router and relay packages). Please do not use "Update with merge commit" on this stack: a merge commit on the branch blocks "Rebase and merge" afterwards.

PR #405, opened 2026-09-18
fix(redisstore): make the cache lifetimes reachable on a RESP backend

avitenzer:
# Description

**Decision taken (MAG-3631 is a "Decide" ticket).** Treated as a gap to fix, not a limitation to document. The chart's own default, a 1.5 multiplier giving every sidecar deployment 90 minutes on settled answers, vanished silently for every customer who moved their cache to a RESP backend, and the PRD's parity clause covers behaviour. If the product answer is "limitation", this PR is the cost of not having to write that down in three places; say so and it can be closed.

**What was happening.** Both cache paths share one TTL rule (`core.Policy.ForRelayEntry`); they differ only in who builds the policy. The sidecar builds it from flags the chart sets. The RESP path called `core.DefaultPolicy()` and nothing anywhere could pass anything else, so the comment promising a table "unless configured otherwise" described a path that was never built.

**Fix.** The `resp-cache:` block gains an `expiration` section, one key per sidecar expiration flag with the same meaning and the same default:

| Key | Sidecar flag | Default |
| --- | --- | --- |
| `expiration.finalized` | `--expiration` | `1h` |
| `expiration.finalized-multiplier` | `--expiration-multiplier` | `1` (the chart sets `1.5`) |
| `expiration.non-finalized` | `--expiration-non-finalized` | `500ms` |
| `expiration.non-finalized-multiplier` | `--expiration-non-finalized-multiplier` | `1` |
| `expiration.node-errors` | `--expiration-finalized-node-errors` | `250ms` |
| `expiration.blocks-hashes-to-heights` | `--expiration-blocks-hashes-to-heights` | `48h` |

The policy is built from it at the seam the ticket names (`resp_config.go`, the `NewRespCache` call). Unset fields keep the engine's defaults, so a block without the section behaves exactly as before. A negative lifetime or multiplier is refused. A misspelled key is refused by the block's strict decoding from #397, which is the failure shape that made the original gap invisible. `GET /debug/cache-state` already reports the lifetimes in force under `lifetimes`, so the fix is verified the way the defect was read.

**Merge order.** Stacked on #397 (MAG-3521) → #396 (MAG-3683), because the new keys must be declared against #397's strict block decoding and it reworked the same loader. Merges cleanly against the other open RESP chains. Until #397 lands this PR's diff shows those commits too.

Files: `ecosystem/cache/redisstore/config.go`, `protocol/performance/resp_config.go`, their tests, the docs table.

## Tests

- `TestExpirationConfigPolicy`: an empty block is exactly the engine's defaults; the chart's multiplier alone gives 90 minutes; a full block computes duration × multiplier per field the way the sidecar does.
- `TestLoadRespCacheConfigExpirationBlock`: a partial block through the real loader keeps the defaults for what it does not name; a misspelled key and a negative value are refused.
- `TestSelectBackendRespExpirationFromConfig`: through the real backend selection, `/debug/cache-state` reports `finalized_seconds: 5400` with the chart's multiplier and `3600` without the block.

## Not in this PR

- Changing any duration. As the ticket says, this only makes them reachable.
- The three documents the ticket lists as stating or implying identical durations (`documentation/secondary_cache/resp-cache-deck.md`, `documentation/secondary_cache/smart-router-cache-decks-combined.md`, `documentation/context/glossary.md:420`) live in the automation repo and should now say: identical once the `expiration` block carries the chart's values.
- The chart itself: if the RESP path is to keep the 90 minutes by default, the chart needs to render `expiration.finalized-multiplier` into the `resp-cache` block the way it renders `miscellaneous.cache.expiration_multiplier` into the sidecar's flags.

Jira ticket: MAG-3631

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable: additive keys, defaults unchanged)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3631)

## Reviewers Checklist

I have...

---

Reviewed head [`596d52398932`](https://github.com/Magma-Devs/smart-router/commit/596d52398932b31c682831d3230082ca27d4e101). Findings below are from code inspection and local Go tests.

### [P2] Validate the computed TTL before converting it to a duration

The multiplication in [`ExpirationConfig.Policy`](https://github.com/Magma-Devs/smart-router/blob/596d52398932b31c682831d3230082ca27d4e101/ecosystem/cache/redisstore/config.go#L143-L154) can truncate a positive configured lifetime to zero, although `validate()` accepts both inputs. For example:

```yaml
expiration:
  finalized: 1ns
  finalized-multiplier: 0.5
```

This loads successfully and produces `policy.Finalized == 0`. `redisstore.SetEntry` interprets zero as a plain SET with no expiry. A focused regression through the real loader/backend selection wrote a finalized entry with this configuration, observed TTL `0s` in miniredis, and confirmed that the entry still existed after advancing its clock by 24 hours. A setting intended to shorten retention therefore creates a permanent key, also making it ineligible for the recommended `volatile-*` eviction policies.

Please validate the effective duration after multiplication and reject or round up positive values that would become zero; also bound the conversion to a representable finite duration. Cover the resulting stored TTL, not only the policy/debug-state value.
Docker confirmation (Valkey 7.2): the same configuration, passed through the real backend selector, wrote a finalized entry whose Redis `PTTL` was `-1` (no expiration). This confirms that the permanent-key failure is not specific to miniredis. The existing three-node `TestClusterDocker` integration test also passed.

### Resolution (2026-09-20)

Fixed in 71230c0.

The multiplied lifetimes now go through one computation, `effectiveLifetime`, that `validate` checks and `Policy` applies. A product under one nanosecond (which truncated to a zero TTL, and a zero TTL is a plain SET to the store, a key that never expires) or past the range of a duration is refused at startup, with the defaults standing in for an unset base, so `finalized: 1ns` with `finalized-multiplier: 0.5` no longer loads. `Policy` clamps to the nearest representable lifetime rather than yielding a zero if validation was ever skipped. The docs row for the multiplier says so.

Tests: `TestExpirationConfigRejectsALifetimeThatRoundsAway` covers the sub-nanosecond product, the default base with a tiny multiplier, the overflow, a valid halved lifetime applied exactly, and the clamp. `TestExpirationBlockReachesTheStoredTTL` goes through the real loader and backend selection: the 1ns configuration is refused, and a `2s × 0.5` block writes a finalized entry whose TTL in miniredis is one second, not zero. That test fails on the previous code (the 1ns block loaded) and passes now; `go test ./ecosystem/cache/redisstore ./protocol/performance` pass.


avitenzer:

Reviewed head [`596d52398932`](https://github.com/Magma-Devs/smart-router/commit/596d52398932b31c682831d3230082ca27d4e101). Findings below are from code inspection and local Go tests.

### [P2] Validate the computed TTL before converting it to a duration

The multiplication in [`ExpirationConfig.Policy`](https://github.com/Magma-Devs/smart-router/blob/596d52398932b31c682831d3230082ca27d4e101/ecosystem/cache/redisstore/config.go#L143-L154) can truncate a positive configured lifetime to zero, although `validate()` accepts both inputs. For example:

```yaml
expiration:
  finalized: 1ns
  finalized-multiplier: 0.5
```

This loads successfully and produces `policy.Finalized == 0`. `redisstore.SetEntry` interprets zero as a plain SET with no expiry. A focused regression through the real loader/backend selection wrote a finalized entry with this configuration, observed TTL `0s` in miniredis, and confirmed that the entry still existed after advancing its clock by 24 hours. A setting intended to shorten retention therefore creates a permanent key, also making it ineligible for the recommended `volatile-*` eviction policies.

Please validate the effective duration after multiplication and reject or round up positive values that would become zero; also bound the conversion to a representable finite duration. Cover the resulting stored TTL, not only the policy/debug-state value.
Docker confirmation (Valkey 7.2): the same configuration, passed through the real backend selector, wrote a finalized entry whose Redis `PTTL` was `-1` (no expiration). This confirms that the permanent-key failure is not specific to miniredis. The existing three-node `TestClusterDocker` integration test also passed.

avitenzer:
Fixed in 71230c0.

The multiplied lifetimes now go through one computation, `effectiveLifetime`, that `validate` checks and `Policy` applies. A product under one nanosecond (which truncated to a zero TTL, and a zero TTL is a plain SET to the store, a key that never expires) or past the range of a duration is refused at startup, with the defaults standing in for an unset base, so `finalized: 1ns` with `finalized-multiplier: 0.5` no longer loads. `Policy` clamps to the nearest representable lifetime rather than yielding a zero if validation was ever skipped. The docs row for the multiplier says so.

Tests: `TestExpirationConfigRejectsALifetimeThatRoundsAway` covers the sub-nanosecond product, the default base with a tiny multiplier, the overflow, a valid halved lifetime applied exactly, and the clamp. `TestExpirationBlockReachesTheStoredTTL` goes through the real loader and backend selection: the 1ns configuration is refused, and a `2s × 0.5` block writes a finalized entry whose TTL in miniredis is one second, not zero. That test fails on the previous code (the 1ns block loaded) and passes now; `go test ./ecosystem/cache/redisstore ./protocol/performance` pass.

AnnaR-prog:
## Review: `/code-review high` on `71230c0`

Scope: this PR's two commits (`156542e...71230c0`, without the #396/#397 diff the "Files changed" tab still shows): `ExpirationConfig`, `Policy`, `effectiveLifetime`, `validate`, the `NewRespCache` seam, the docs table, and what the store actually writes for each resolved lifetime (miniredis `TTL` after a `SetEntry`). Re-verified against a worktree at this head; the PR's tests pass there.

**Stack note:** stacked on #397 → #396 (base `main` on GitHub). This is the schema that fits #397's strict decoding; #388 (same ticket) is the one that does not, as noted there — the typo case and the `<1ns` case from that review are closed here. One of its findings still applies at this head, below.

### 1. Medium — a unitless value passes validation and yields a 1 ms lifetime; the floor that matters on a RESP backend is one millisecond, not one nanosecond — **CONFIRMED**

`ecosystem/cache/redisstore/config.go:188`, `:212`, `docs/RESP-CACHE.md:87`

mapstructure maps a bare YAML integer straight onto the `int64` behind `time.Duration`, so a unitless value is nanoseconds; viper's duration hook only converts strings. Probe through the real loader and a `SetEntry` against miniredis at this head:

| block | `validate` | resolved | stored TTL |
| --- | --- | --- | --- |
| `finalized: 3600` | passes | `3.6µs` | `1ms` |
| `finalized: 1h`, `finalized-multiplier: 1e-12` | passes | `3ns` | `1ms` |
| `node-errors: 250`, `blocks-hashes-to-heights: 172800` | passes | `250ns`, `172.8µs` | `1ms` each |

The first row is the one an operator will type — the sidecar refuses `--expiration 3600` outright (pflag: `missing unit`), so "same names, same numbers" holds for the names and not for this. go-redis rounds any sub-millisecond `PX` up to 1 ms and prints `specified duration is 3.6µs, but minimal supported value is 1ms - truncating to 1ms` through its own logger to stderr — unstructured, and once per write, which at relay rate is a flood in the process's stderr and nothing in the router's log. The `<1ns` rule from the second commit catches the permanent-key case and none of these. One check on the resolved table — every lifetime at least `time.Millisecond`, the precision a RESP `PX` can express — closes all three rows and subsumes the `<1ns` rule (a `>= 1ms` check makes `< 1` unreachable). If bare integers should be refused rather than rounded, a decode hook on the four duration fields is the other route, and the docs table should say a unit is mandatory either way. `/debug/cache-state` does show `finalized_seconds: 3.6e-06` afterwards, so the recipe in the PR body would reveal it — to whoever runs it.

### Nits

- `validate` ranges over two maps, so which of two negative fields the error names is random between runs; a slice keeps the message stable for anyone matching on it.
- The `finalized-multiplier` docs row says the refusal is for a lifetime "below one nanosecond"; after the check above it should say one millisecond, and it is the natural place for "durations need a unit".

### Verified and not flagged

- Key names, meanings and defaults match the sidecar's flags and the multiplier is applied after the base resolves, as `ecosystem/cache/server.go` does it; the effective non-finalized TTL stays `max(avg/8, floor)` from `core/policy.go`.
- An absent block is byte-for-byte `core.DefaultPolicy()` (test and probe), so a router without the section changes nothing on upgrade.
- A misspelled key inside `expiration:` is refused by #397's strict decoding at this head.
- `Policy()` clamps rather than errors, so a caller that bypassed `Validate` still never stores a zero — the case the second commit exists for.

avitenzer:
Fixed in 080e8cc, on a branch rebased onto #397's 659aa75.

1. Every resolved row of the table is held to `time.Millisecond`, named `minLifetime` with the PX reasoning beside it, and the `<1ns` rule is gone since nothing below the floor gets through (77c1540). One function, `resolve`, computes the table in a fixed row order; `Policy` applies it and `validate` checks it, so the same block always names the same field, and a block that skipped validation is clamped to the floor on every row rather than yielding a zero. Your three rows are refused through the real loader in `TestLoadRespCacheConfigRefusesALifetimeBelowOneMillisecond`, and the same numbers with units load; the unit test covers the same shapes plus the clamp. When the configured value itself is below the floor the message shows what the bare number became and how to write it (`expiration.finalized is 3.6µs, shorter than the 1ms a RESP expiry can express; a bare number is read as nanoseconds (3600 is 3.6µs), so write the value with a unit such as 3600s`); a tiny multiplier on a sound base is refused without that hint, since the base is not the problem there. The end-to-end TTL test now refuses the unitless row rather than the `1ns × 0.5` one. The docs table says a unit is required on every `expiration.*` duration, and the multiplier row says one millisecond.

Not taken: a decode hook refusing bare integers. It would apply to `dial-timeout` and the block's other three durations too, which #397's strict decoding did not sign up for, and the floor already turns the unitless trap into a startup refusal for every value an operator would write.

Nits: the two maps are slices, so the negative-value message is stable; the docs wording is fixed. Also fixed the unchecked type assertion in `TestSelectBackendRespExpirationFromConfig` that golangci-lint flagged on this branch (080e8cc).

Verified: `go build ./...`, `go vet`, `golangci-lint run --config .golangci.yml` (no findings beyond `main`'s), and `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/... ./protocol/rpcsmartrouter/... ./types/relay/...` pass at 080e8cc.

avitenzer:
Rebased onto #397's 5a8ee5b after #396 moved to 67d0e5c (main advanced to 970ab87) and #397 took its review fixes; no content change. Tip is cc4650a.

AnnaR-prog (approved):

AnnaR-prog (approved):

avitenzer:
Rebased onto #397's 8103bec after #396 merged to `main` (3cf9bcf); no content change, tree identical to cc4650a. Tip is 90b90bd.

avitenzer:
Rebased onto `main` after #397 merged there (2570da8); the branch no longer carries #397's commits. No content change of its own. Tip is c23dd81, re-verified from a clean checkout (build, lint against `main`'s baseline, `go test -race` over the cache, performance, router and relay packages).

PR #406, opened 2026-09-18
fix(performance): trip a breaker when the RESP cache is unreachable

avitenzer:
# Description

**Bug (MAG-3676).** With its RESP cache unreachable, the router kept serving correctly and about forty times more slowly for as long as the outage lasted. Every lookup was still issued against the dead backend and paid its 50 ms budget; every asynchronous write paid its 5 s budget while holding a pool slot, so under steady traffic the pool filled with stalled writes and lookups queued behind them (the hundred-to-one budget split the ticket names, with nothing capping the writes in flight). Each operation also opened one more connection to a store that never answered.

**Fix.** A breaker on the RESP tier:
- three consecutive backend failures on the relay path, or one failed health probe, open it;
- while open, lookups and writes return at once with `ErrCacheBreakerOpen`, joined with `core.StoreError` so call sites label the outcome as the backend failure it stands for, and no I/O happens. They are counted on `smartrouter_resp_cache_skipped_total{op}` rather than as failures the backend never saw; `smartrouter_resp_cache_breaker_open` is 1 meanwhile;
- the health probe runs every second while open instead of every ten, and the first successful probe closes it. The relay path nudges the loop to probe at once when it opens the breaker.

`GET /debug/cache-state` now reports the RESP tier as `when_unreachable: "skipped"`, like the gRPC tier. The `attempted` value stays defined for readers of older payloads.

**Merge order.** Stacked on #404 (MAG-3674) → #403 (MAG-3673): #404 rewrote the health loop this builds on. Merge in that order. Until then this PR's diff shows those commits too.

**What this does not fix, said plainly.** The ticket's last comment traces the connection leak to MAG-3728, which Jira shows as Done. Its root cause is unchanged on `main`: a connection whose setup fails is kept by the streaming credentials provider, and go-redis is still v9.22.0 (no newer stable release exists). The breaker bounds the leak to the operations issued before it opens, a mitigation, not the fix. MAG-3728's status looks wrong and is worth re-checking.

Files: `protocol/performance/resp_cache.go`, `resp_metrics.go`, `cache_backend.go`, tests, `docs/RESP-CACHE.md`, `docs/METRICS.md`.

## Tests

- `TestRespCacheBreakerBoundsWorkAgainstABlackholedBackend`, the acceptance criterion: a black-holed listener that accepts, never answers, and counts connections. Three lookups at the relay budget open the breaker; a burst of 200 writes with the 5 s budget returns in well under 2 s, every one with `ErrCacheBreakerOpen`, and opens no connection; a lookup during the outage returns before its 50 ms budget instead of at it (the lookup side of the same mechanism: no writes holding pool slots); then a real store takes over the address and the breaker closes within seconds, after which a lookup is an ordinary miss.
  - Run with the breaker's skip disabled, the same burst takes 5.01 s (the write budget) and the test fails on that assertion, which is the ticket's "fails on the code from before the fix".
- `TestRespCacheBreakerOpensOnFailedProbeAndClosesOnRecovery`: dead at startup, the probe opens it and the first lookup is skipped; the store restarts and the breaker closes at its own one-second cadence, not the hour-long configured interval.
- The two dead-at-startup tests now accept "failed at the backend or skipped by the breaker", since the startup probe races the first operation; both outcomes are counted, on their own series.
- Package run twice for flakiness.

The ticket also asks for a re-measurement on the corrected harness (MAG-3666); this PR changes what that measurement should show, from a forty-fold drop to near-normal serving with `smartrouter_resp_cache_skipped_total` rising.

Jira ticket: MAG-3676

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable: two series added; the RESP tier's `when_unreachable` value changes from `attempted` to `skipped` because its behaviour did)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3676)

## Reviewers Checklist

I have...

---

Reviewed head [`6b914c62f453`](https://github.com/Magma-Devs/smart-router/commit/6b914c62f453a96ab0939e2f547a120b9b137b09). Findings below are from code inspection and local Go tests.

### [P2] Count backend failures during symbolic-block resolution

The new [`noteOperation` call](https://github.com/Magma-Devs/smart-router/blob/6b914c62f453a96ab0939e2f547a120b9b137b09/protocol/performance/resp_cache.go#L460-L465) relies on `Engine.GetRelay` returning `core.StoreError`. When a request carries `SAFE_BLOCK`, `FINALIZED_BLOCK`, or `PENDING_BLOCK`, a failed chain-tip read is instead converted to `NOT_APPLICABLE`; the entry lookup is skipped and `GetEntry` returns a miss with a nil error. `noteOperation` then resets the consecutive-failure count.

Reproduced with three `SAFE_BLOCK` lookups against a listener that accepts connections but never replies, each with a 50 ms context: each lookup took approximately 51 ms, returned nil error, and left the failure count at zero and the breaker closed. The background probe eventually opens it, but traffic continues paying the full lookup budget and starting writes until then; such requests also reset a failure streak from other operations.

Please propagate backend errors from chain-tip resolution into the breaker while preserving genuine missing/stale tips as misses. Add the symbolic-block case to the threshold regression. The router pre-resolves ordinary `latest` requests, so the production case here is the other symbolic block tags.

### [P2] Apply the open breaker to sticky-session store calls

The guards added to [`GetEntry` and `SetEntry`](https://github.com/Magma-Devs/smart-router/blob/6b914c62f453a96ab0939e2f547a120b9b137b09/protocol/performance/resp_cache.go#L451-L477) do not cover `GetStickySession` or `SetStickySessionIfAbsent`; both still call the same RESP store directly. With cross-pod stickiness enabled, a new/unconfirmed session therefore continues making I/O against a blackholed backend after the breaker opens. The session manager allows 300 ms per store call, and repeated new sessions continue consuming connections.

A focused regression explicitly opened the breaker and then called each sticky method with a 100 ms context against a blackhole. Both waited approximately 101 ms and returned `context deadline exceeded`, rather than being skipped immediately.

Please guard both sticky methods and feed their backend failures into the breaker. Preserve their fail-closed contract by returning an error immediately; do not convert an unavailable claim registry into a missing claim.

### Inherited prerequisite from #404

This stack also contains the sequential probe/shared-deadline issue described in [#404's review findings](https://github.com/Magma-Devs/smart-router/pull/404#review-findings-2026-09-18). Carry that fix forward so a stalled writer does not report a healthy reader as down.

### Docker validation follow-up (2026-09-18)

Built and ran this head with the repository's Compose files and bundled RESP-cache example. All four HTTP/WebSocket upstream checks returned `ok: true`; both router health endpoints returned HTTP 200. With Valkey running, repeated `eth_getBalance(..., "latest")` requests hit the cache. After stopping only Valkey, the breaker opened, eight requests still returned HTTP 200, and both get/set skipped counters increased by eight. Restarting Valkey restored `connected=1`, closed the breaker, and resumed cache hits without restarting the router.

These checks validate the normal outage/recovery path; the two edge cases above remain reproducible. The inherited #404 probe issue was also reproduced against two real Valkey containers, as recorded in that PR.

The default non-RESP example served RPC and HTTP health successfully, but its full CLI health check returned `ok: false` because the Tenderly endpoint presented a certificate for an unrelated hostname in this environment. The RESP example's PublicNode/dRPC checks all passed; no TLS verification was bypassed.

### Resolution (2026-09-20)

Fixed in 5600c1a, on a branch rebased onto #404's concurrent probe (4ad47ad), which carries the inherited prerequisite.

**Symbolic-block resolution.** `Engine.chainTip` now tells a tip the store could not read apart from an unknown or stale one: the former makes `GetRelay` return a store error (joined with `core.StoreError`), the latter stays a miss. `RespCache.GetEntry` therefore feeds the failure to the breaker and the failed series. `TestRespCacheBreakerCountsSymbolicBlockLookups` runs three `SAFE_BLOCK` lookups against a listener that never answers: each returns a store error, the breaker opens, and the next lookup is skipped at once. `TestGetRelaySymbolicBlockReportsAFailedTipRead` pins the engine's distinction.

**Sticky-session calls.** `GetStickySession` and `SetStickySessionIfAbsent` are now behind the breaker: while it is open they return an error at once (`ErrCacheBreakerOpen` joined with `core.StoreError`), never a missing claim, and open no connection; their own failures count toward opening it, with the engine marking sticky store failures as store errors so `noteOperation` sees them. They have their own `op` labels (`sticky_get`, `sticky_set`) on the skipped and failed series, documented in METRICS.md. `TestRespCacheBreakerGuardsStickySessionCalls` covers both halves: three failing sticky reads open the breaker, then both calls return in under 50 ms with no new dial.

Both backend tests fail on the previous code and pass now; `go test ./ecosystem/cache/... ./protocol/performance` pass.


avitenzer:

Reviewed head [`6b914c62f453`](https://github.com/Magma-Devs/smart-router/commit/6b914c62f453a96ab0939e2f547a120b9b137b09). Findings below are from code inspection and local Go tests.

### [P2] Count backend failures during symbolic-block resolution

The new [`noteOperation` call](https://github.com/Magma-Devs/smart-router/blob/6b914c62f453a96ab0939e2f547a120b9b137b09/protocol/performance/resp_cache.go#L460-L465) relies on `Engine.GetRelay` returning `core.StoreError`. When a request carries `SAFE_BLOCK`, `FINALIZED_BLOCK`, or `PENDING_BLOCK`, a failed chain-tip read is instead converted to `NOT_APPLICABLE`; the entry lookup is skipped and `GetEntry` returns a miss with a nil error. `noteOperation` then resets the consecutive-failure count.

Reproduced with three `SAFE_BLOCK` lookups against a listener that accepts connections but never replies, each with a 50 ms context: each lookup took approximately 51 ms, returned nil error, and left the failure count at zero and the breaker closed. The background probe eventually opens it, but traffic continues paying the full lookup budget and starting writes until then; such requests also reset a failure streak from other operations.

Please propagate backend errors from chain-tip resolution into the breaker while preserving genuine missing/stale tips as misses. Add the symbolic-block case to the threshold regression. The router pre-resolves ordinary `latest` requests, so the production case here is the other symbolic block tags.

### [P2] Apply the open breaker to sticky-session store calls

The guards added to [`GetEntry` and `SetEntry`](https://github.com/Magma-Devs/smart-router/blob/6b914c62f453a96ab0939e2f547a120b9b137b09/protocol/performance/resp_cache.go#L451-L477) do not cover `GetStickySession` or `SetStickySessionIfAbsent`; both still call the same RESP store directly. With cross-pod stickiness enabled, a new/unconfirmed session therefore continues making I/O against a blackholed backend after the breaker opens. The session manager allows 300 ms per store call, and repeated new sessions continue consuming connections.

A focused regression explicitly opened the breaker and then called each sticky method with a 100 ms context against a blackhole. Both waited approximately 101 ms and returned `context deadline exceeded`, rather than being skipped immediately.

Please guard both sticky methods and feed their backend failures into the breaker. Preserve their fail-closed contract by returning an error immediately; do not convert an unavailable claim registry into a missing claim.

### Inherited prerequisite from #404

This stack also contains the sequential probe/shared-deadline issue described in [#404's review findings](https://github.com/Magma-Devs/smart-router/pull/404#review-findings-2026-09-18). Carry that fix forward so a stalled writer does not report a healthy reader as down.

### Docker validation follow-up (2026-09-18)

Built and ran this head with the repository's Compose files and bundled RESP-cache example. All four HTTP/WebSocket upstream checks returned `ok: true`; both router health endpoints returned HTTP 200. With Valkey running, repeated `eth_getBalance(..., "latest")` requests hit the cache. After stopping only Valkey, the breaker opened, eight requests still returned HTTP 200, and both get/set skipped counters increased by eight. Restarting Valkey restored `connected=1`, closed the breaker, and resumed cache hits without restarting the router.

These checks validate the normal outage/recovery path; the two edge cases above remain reproducible. The inherited #404 probe issue was also reproduced against two real Valkey containers, as recorded in that PR.

The default non-RESP example served RPC and HTTP health successfully, but its full CLI health check returned `ok: false` because the Tenderly endpoint presented a certificate for an unrelated hostname in this environment. The RESP example's PublicNode/dRPC checks all passed; no TLS verification was bypassed.

avitenzer:
Fixed in 5600c1a, on a branch rebased onto #404's concurrent probe (4ad47ad), which carries the inherited prerequisite.

**Symbolic-block resolution.** `Engine.chainTip` now tells a tip the store could not read apart from an unknown or stale one: the former makes `GetRelay` return a store error (joined with `core.StoreError`), the latter stays a miss. `RespCache.GetEntry` therefore feeds the failure to the breaker and the failed series. `TestRespCacheBreakerCountsSymbolicBlockLookups` runs three `SAFE_BLOCK` lookups against a listener that never answers: each returns a store error, the breaker opens, and the next lookup is skipped at once. `TestGetRelaySymbolicBlockReportsAFailedTipRead` pins the engine's distinction.

**Sticky-session calls.** `GetStickySession` and `SetStickySessionIfAbsent` are now behind the breaker: while it is open they return an error at once (`ErrCacheBreakerOpen` joined with `core.StoreError`), never a missing claim, and open no connection; their own failures count toward opening it, with the engine marking sticky store failures as store errors so `noteOperation` sees them. They have their own `op` labels (`sticky_get`, `sticky_set`) on the skipped and failed series, documented in METRICS.md. `TestRespCacheBreakerGuardsStickySessionCalls` covers both halves: three failing sticky reads open the breaker, then both calls return in under 50 ms with no new dial.

Both backend tests fail on the previous code and pass now; `go test ./ecosystem/cache/... ./protocol/performance` pass.

avitenzer:
Follow-up a12e806 (branch rebased onto #404's aca2110): the symbolic-block change is reshaped so its hunks sit clear of the lines the keyspace change (#397) rewrites in the lookup. `chainTip` keeps its signature and answers a sentinel block for a read the store failed, `GetRelay` turns that into the store error once the block is resolved (with the context's own error joined in, so a timeout is classified as one), and the engine tests wrap the fake store instead of editing it. Same behaviour as before; every open branch pair now merges cleanly.

AnnaR-prog:
## Review: `/code-review high` on `a12e806`

Scope: this PR's three commits (`aca2110...a12e806`, without the #403/#404 diff the "Files changed" tab still shows): the breaker in `resp_cache.go`, the engine's tip sentinel, the metric registrations, both docs pages, and two shapes the ticket's black-holed harness cannot produce — a split store with one half down, and a backend that is slow rather than dead. Each finding was re-verified against a worktree at this head with a probe; the verdict says how far.

**Stack note:** stacked on #404 → #403 (base `main` on GitHub). The PR body says MAG-3728's root cause is "unchanged on `main`" — #407 (same author, open in this batch) is that fix; the two should point at each other, and with both merged the breaker also bounds the fd count noted there.

The acceptance shape is right: three failures open it, the burst returns at once, recovery lands on the one-second cadence, and `Close` still has the last word. Two findings are about the shapes beside that one.

### 1. Medium — the breaker is whole-cache while the store is split by role: a writer-only outage skips every lookup on a healthy reader that holds the data — **CONFIRMED**

`protocol/performance/resp_cache.go:186`, `:455`

Probe at this head: two stores, reads split, entry present on both; the write store is closed. Timeline of a lookup at the relay budget every 500 ms:

```
t=+0.5s  breaker=false  hit=true
t=+1.0s  breaker=false  hit=true
t=+1.5s  breaker=false  hit=true
WRN resp-cache backend became unavailable … detail="write endpoint 127.0.0.1:64267: … connection refused"
t=+2.0s  breaker=true   hit=false  err=cache store operation failed
t=+2.5s … +4.0s         same
```

Before this PR every one of those lookups was a hit for the whole outage (the first three rows are the pre-PR behaviour, before go-redis's dial retries let the probe fail). After it, the reader — the endpoint `read-addresses` exists for, healthy, holding the entry — is skipped until the writer returns; the reverse shape (reader down) skips writes to a healthy writer, so the cache stops being populated. #404 already gives the probe a `Role` per result, and the op tells the client: `GetEntry` / `GetStickySession` read `s.read`, `SetEntry` / `SetStickySessionIfAbsent` write `s.write`. Two breakers keyed by role — write-side failures and the write probe open `write`, read-side failures and the read probe open `read`, `breaker_open{role}` — make the outage cost exactly the half that is down. The unsplit store has one client behind both roles and behaves as now.

### 2. Medium — no minimum open time, and the close condition is weaker than the open condition, so a slow-but-alive backend flaps the breaker twice a second — **CONFIRMED**

`protocol/performance/resp_cache.go:190`, `:226`, `:286`

The breaker opens on three lookups that missed the 50 ms relay budget; it closes on one `PING` that answered within `respCachePingTimeout` (3 s), on a pooled connection. Probe: a TCP proxy adding 120 ms in front of miniredis, one lookup every 5 ms for 3 s — `6 opens, 5 closes`, i.e. a cycle every ~530 ms (150 ms to trip, one nudged probe to close), and every cycle writes the Warning at `:236` and the Info at `:264`: **11 transition lines in 3 s**, ~320k a day for a backend that is merely slow, with `breaker_open` and `connected` flapping and `/debug/cache-state` alternating. The ticket's black-holed harness cannot show this (its probe fails too), so the acceptance test is green either way. Two changes, either sufficient: while open, probe with the budget the relay path uses (`common.DefaultCacheTimeout`) so a backend that cannot meet it stays open; and hold the breaker open for at least one `respCacheBreakerProbeInterval` so the nudged probe cannot close it inside the same 150 ms window. Rate-limiting the two transition lines is the fallback if flapping is accepted.

### Nits

- `smartrouter_resp_cache_skipped_total`'s Help text says `op (get|set)`; the series also carries `sticky_get` / `sticky_set` (the METRICS row is right).
- `noteOperation` logs the Warning before `openBreaker`'s CAS, so two goroutines crossing the threshold together can both log it; harmless, but the CAS is the natural place for the line.

### Verified and not flagged

- `chainTip` is read only for symbolic blocks (`originalRequestedBlock < 0`), so the `tipReadFailed` sentinel cannot turn a concrete-block lookup into a store error; and a timeout rides along as `ctx.Err()`, so it is classified as one.
- `errors.Join(core.StoreError, ErrCacheBreakerOpen)` satisfies `errors.Is(err, core.StoreError)`, so call sites label a skip as the backend failure it stands for; `skip` does no I/O and no logging.
- `probeNow` is buffered 1 and signalled only inside the CAS, so at most one nudge per opening; the timer is stopped before the nudged probe and reset after, and Go ≥1.23 timer semantics rule out a stale tick.
- The breaker test and the two dead-at-startup tests pass 3× under `-race` at this head.

avitenzer:
Fixed in 9378d0c, on a branch rebased onto #404's e2b5054 (which sits on #403's 3eb7c7b).

1. The breaker is split by side. Writes, the sticky claim and the write endpoint's probe steer the write breaker; lookups (entries, the chain tip, sticky reads) and the read endpoint's probe steer the read breaker, using the `Role` #404's probe already reports. Without a read split one breaker stands behind both sides and the tier behaves as it did with one. `smartrouter_resp_cache_breaker_open` carries a `role` label; the shared breaker of an unsplit store publishes both roles, so each series answers "is this side being skipped" whatever the topology, and both exist as 0 from construction. `/debug/cache-state` gains `breaker: {write_open, read_open}` on the RESP tier and `null` on a tier without one, since with reads split `reachable` alone cannot say which side. `TestRespCacheBreakerIsSplitByEndpoint` runs your two shapes: with the writer closed and the entry on both stores, lookups keep hitting on the reader while writes are skipped; with the reader closed, writes keep landing while lookups are skipped; the gauges and the debug field read per side throughout.

2. Both changes. A breaker now closes only on a probe that answered within the budget of the tightest side it gates, read from the round trip `Probe` now measures per endpoint, and it holds for at least one `respCacheBreakerProbeInterval` once opened, so the nudged probe cannot close it inside the window that opened it. The budget is per side rather than the literal `common.DefaultCacheTimeout` everywhere: lookups and the shared breaker use the relay's cache budget, a write-only breaker uses the write budget, which exceeds the probe deadline, so it closes on any answered probe as before. Probing every endpoint under a 50 ms deadline would have failed a healthy far-away writer for the slow reader's sake, the shape #404 just stopped blaming. `TestRespCacheBreakerStaysOpenOnABackendSlowerThanTheBudget` puts a proxy delaying every reply by 120 ms in front of miniredis and runs lookups every 5 ms for 3 s: one opening, no closing, then the delay is removed and the breaker closes on the cadence; a 1 ms sampler of the breaker state stands in for counting the log lines, since one line is written per transition. `TestRespCacheBreakerHoldsForAProbeInterval` opens the breaker on a frozen proxy, unfreezes it at once, and shows the breaker still open a quarter interval later and closed on the first probe of the open cadence.

Nits. The `skipped_total` help text lists `sticky_get` and `sticky_set`. The opening line is written by the call that won the CAS, in `noteOperation` after `openBreaker` reports it opened, and never by one that found the breaker open; the halves that open on one probe are reported in one line. Related, the probe's verdict is now published before the breakers move, so anything that observes a flip finds the verdict behind it in the debug state.

Stack note: the PR body now points at #407 for MAG-3728, and with both merged the breaker bounds the fd count during an outage while #407 bounds the leak.

One thing to know when reading the gauges: `smartrouter_resp_cache_connected` can read 1 while `breaker_open{role="read"}` reads 1, the endpoint answers and is too slow to serve lookups. The docs say so; it is the slow-not-dead distinction of MAG-3668, and `breaker_open` is the series that makes it.

Verified: `go test -race -count=3 ./protocol/performance -run 'Breaker|DebugState|Reachability|HealthLoop|NamesTheFailing|Probe'` passes; `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/... ./protocol/rpcsmartrouter/...` passes at this head; `go build ./...`, `go vet` and `golangci-lint run` on those packages add nothing over the branch's baseline. Not run against a real Valkey pair: no Docker daemon on this machine today, so the split-store shapes are two miniredis processes.

AnnaR-prog (approved):

avitenzer:
Rebased onto #404's 74c85e6 after #397 and #398 merged to `main`. The final tree merged clean, but the middle commit (count symbolic-block tip failures) conflicted with the lines #397 rewrites in the engine lookup, which its follow-up commit had already moved clear of; the two are folded into one commit so that every commit on the branch builds against `main`, and the message says so. The branch's diff against its parent is byte-identical to before, `core/` included. Tip is 3206009, re-verified from a clean checkout (build, lint against `main`'s baseline, `go test -race` over the cache, performance, router and relay packages).

avitenzer:
Rebased onto #404's 72280b2f (its commits folded so the stack replays commit by commit onto `main`); no content change of its own, and the tree is byte-identical to the clean three-way merge of the previous tip. Tip is f331fab8, re-verified from a clean checkout (build, lint against `main`'s baseline, `go test -race` over the cache, performance, router and relay packages).

avitenzer:
Rebased onto `main` after #404 merged there (691b0682); the branch no longer carries #404's commits, so the conflicts GitHub reported are gone. Three commits of its own, replaying cleanly onto the current `main`; tree byte-identical to f331fab8, which was verified from a clean checkout this morning. Tip is fa98ea12.

PR #407, opened 2026-09-18
fix(redisstore): stop keeping connections that failed setup alive for the life of the process

avitenzer:
# Description

**Bug (MAG-3728).** During a RESP outage every request dials the store, and go-redis (v9.22.0, the version in `go.mod`) abandons a connection whose HELLO fails without ever calling Close on it: the connection is moved to the closed state and its close path, including the unsubscribe it had just stored on the connection, never runs. That is the upstream half. The router's half: its streaming credentials provider held that connection's listener strongly in a map that only an unsubscribe could shrink, so every abandoned connection stayed reachable, with its two 32 KiB buffers and its socket, until the router was killed for memory. The ticket counted 3,510 of them after a short outage. #406's breaker only mitigates: fewer dials during an outage, but every failed one still leaked.

**Fix.** Two parts, both in `ecosystem/cache/redisstore`:

1. **Subscriptions are held weakly**, anchored only by the unsubscribe closure go-redis stores on the connection itself (`pool.Conn.SetOnClose`, whose single caller is `initConn`). A subscription therefore lives exactly as long as its connection is reachable: a live pooled connection carries its own anchor, so rotation still reaches it, and an abandoned one becomes garbage together with everything it held. The runtime closes the socket when it reclaims the connection, so file descriptors close at the next collection rather than at the failure. The same rule covers the case go-redis documents for a re-initialised connection (a second Subscribe replaces the stored closure): the superseded subscription is forgotten instead of orphaned.
2. **Static credentials no longer ride the provider.** It exists to push a rotated file-backed password to live connections, and static credentials never needed it; they now go straight into the client options, the same username and password the provider handed go-redis before, so only a file-backed password touches this path at all. Sentinel, which resolves the file per connection attempt, builds no provider either. A typed-nil provider never reaches the interface field.

**Merge order.** Stacked on #399 (MAG-3690), which sits on #398 (MAG-3685); all three edit the same file and test file. Merge those first; until then this PR's diff shows their commits too.

Files: `credentials.go`, `config.go`, `store.go`, their tests, and one paragraph in `docs/RESP-CACHE.md` under "Credential rotation".

## Tests

Both new tests fail on the previous provider and pass on this one (checked by swapping the old `credentials.go` back in):

- `TestAbandonedConnectionsAreNotKeptByTheProvider` is the defect in-process, in both outage shapes the ticket measured: a listener that accepts every connection and either closes it at once or holds it open and never answers, a store built on it with a password file, twenty operations, then the provider's subscriber count must return to zero once the collector has run. On the old provider it stays at the number of failed dials in both shapes. That also answers the ticket's open question about the "timeout" kind: a timed-out HELLO takes the same go-redis branch as a reset one.
- `TestSubscriptionLivesAsLongAsItsUnsubscribeClosure` is the rule itself: a subscription whose closure nobody holds is reclaimed, the anchored one survives collections and still receives a rotation, and an explicit unsubscribe still removes it at once.
- `TestLiveRotationSmokeOverMiniredis` now forces two collections before the rotation push, so a live pooled connection is shown to keep its subscription.
- `TestStaticCredentialsSkipTheStreamingProvider` and the sentinel watcher test pin part 2: the options carry the credentials directly, a provider is used when given, and only a non-sentinel file-backed config builds one.

`go vet ./...` and `go test ./ecosystem/cache/redisstore/ ./protocol/performance/` pass locally.

## Out of scope

Running the connection's close path when setup fails belongs upstream in go-redis; this PR makes the router correct regardless of it. The ticket's suggested post-outage memory check in the automation repo remains the end-to-end guard.

Jira ticket: MAG-3728

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable; no configuration or API change)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3728)

## Reviewers Checklist

I have...

---

Reviewed head [`3386b10fe8db`](https://github.com/Magma-Devs/smart-router/commit/3386b10fe8db783919ebc8e176f13a0a3bb4ee55), including its incremental changes over #399 (`7914fe738b0c`).

### [P2] Prune collected subscription records when credentials are unchanged

The production cleanup in [`Refresh`](https://github.com/Magma-Devs/smart-router/blob/3386b10fe8db783919ebc8e176f13a0a3bb4ee55/ecosystem/cache/redisstore/credentials.go#L214-L220) only runs after the unchanged-credentials early return. Collection can release the subscription and connection, but every failed handshake still leaves an ID and weak-reference record in `p.listeners`. For a file-backed password that remains unchanged, normal watcher polls never remove those records, so repeated outages still grow the provider's bookkeeping without a bound.

Reproduced through `New(Config{PasswordFile: ..., CredentialRefreshInterval: 10*time.Millisecond})` against a listener that accepts and immediately closes connections. After three batches of ten failed operations, forced collections, and repeated unchanged-source refreshes, the map contained **10 → 20 → 30 records**, with **zero live subscriptions** after each batch.

The new leak test masks this remainder: [`subscriberCount()`](https://github.com/Magma-Devs/smart-router/blob/3386b10fe8db783919ebc8e176f13a0a3bb4ee55/ecosystem/cache/redisstore/credentials.go#L262-L265) calls the cleanup helper itself, whereas production never calls that accessor. Please prune collected records on a regular path independent of a credential change, and check that the underlying registry remains bounded without invoking a cleanup-performing test helper.

### [P2] Retain subscription anchors in the existing rotation tests

With the new ownership rule, retaining the listener alone does not keep its separate `subscription` allocation alive; only the returned unsubscribe closure does. However, [`TestStreamingProviderPushes`](https://github.com/Magma-Devs/smart-router/blob/3386b10fe8db783919ebc8e176f13a0a3bb4ee55/ecosystem/cache/redisstore/credentials_test.go#L177-L180) discards the second listener's closure, and [`TestRefreshReportsSourceFailureOncePerOutage`](https://github.com/Magma-Devs/smart-router/blob/3386b10fe8db783919ebc8e176f13a0a3bb4ee55/ecosystem/cache/redisstore/credentials_test.go#L306-L309) discards its only closure. Both still assert that those listeners receive subsequent rotations. A collection between subscribing and refreshing now makes those assertions fail. This is consistent with [Go's weak-pointer lifetime rules](https://pkg.go.dev/weak#Pointer).

Reproduced on this head with:

```bash
GOGC=1 go test ./ecosystem/cache/redisstore \
  -run '^(TestStreamingProviderPushes|TestRefreshReportsSourceFailureOncePerOutage)$' \
  -count=100
```

Both tests failed with missing rotation notifications; the same command passed on the #399 parent. Please retain each unsubscribe closure, for example in `t.Cleanup`, and force a collection before the delivery assertions so these tests verify the intended live-subscription behavior instead of depending on GC timing.

### Validation notes

- The unmodified Redis-store package passed with `-race`; `protocol/performance` passed its normal package test run.
- Real-Valkey Docker checks passed for explicit-refresh live rotation, TLS/mTLS, Sentinel failover, and the three-node cluster.
- `TestLiveRotationThroughConfiguredStore` intermittently failed with a re-authentication deadline and its original-connection assertion. The same failure also reproduced on #399 during a ten-run comparison, so it is recorded as an inherited validation issue, not an additional regression attributed to this PR.

### Resolution (2026-09-20)

Fixed in be7dc60.

**Pruning on a regular path.** `Refresh` now prunes collected records under the lock before it decides whether the credentials changed, so the watcher's ordinary unchanged-file tick does the housekeeping; `Subscribe` prunes as well, so the registry is bounded by the live connections plus those not yet collected however rarely the watcher polls. The counting helper still prunes as it counts, but the new `TestProviderRegistryStaysBoundedAcrossOutages` reads the raw registry (`registrySize`, no pruning) through three batches of ten failed handshakes, each followed by a collection and an unchanged-credentials refresh: the registry returns to zero after every batch. On the previous code it stayed at ten after the first batch.

**Anchors in the rotation tests.** `TestStreamingProviderPushes` and `TestRefreshReportsSourceFailureOncePerOutage` now hold every unsubscribe closure for the test's life via `t.Cleanup` and force a collection before the delivery assertions. Your reproduction, `GOGC=1 go test ./ecosystem/cache/redisstore -run '^(TestStreamingProviderPushes|TestRefreshReportsSourceFailureOncePerOutage)$' -count=100`, passes.

`go test ./ecosystem/cache/redisstore ./protocol/performance` pass. The intermittent `TestLiveRotationThroughConfiguredStore` failure you recorded as inherited from #399 is not touched here.


avitenzer:

Reviewed head [`3386b10fe8db`](https://github.com/Magma-Devs/smart-router/commit/3386b10fe8db783919ebc8e176f13a0a3bb4ee55), including its incremental changes over #399 (`7914fe738b0c`).

### [P2] Prune collected subscription records when credentials are unchanged

The production cleanup in [`Refresh`](https://github.com/Magma-Devs/smart-router/blob/3386b10fe8db783919ebc8e176f13a0a3bb4ee55/ecosystem/cache/redisstore/credentials.go#L214-L220) only runs after the unchanged-credentials early return. Collection can release the subscription and connection, but every failed handshake still leaves an ID and weak-reference record in `p.listeners`. For a file-backed password that remains unchanged, normal watcher polls never remove those records, so repeated outages still grow the provider's bookkeeping without a bound.

Reproduced through `New(Config{PasswordFile: ..., CredentialRefreshInterval: 10*time.Millisecond})` against a listener that accepts and immediately closes connections. After three batches of ten failed operations, forced collections, and repeated unchanged-source refreshes, the map contained **10 → 20 → 30 records**, with **zero live subscriptions** after each batch.

The new leak test masks this remainder: [`subscriberCount()`](https://github.com/Magma-Devs/smart-router/blob/3386b10fe8db783919ebc8e176f13a0a3bb4ee55/ecosystem/cache/redisstore/credentials.go#L262-L265) calls the cleanup helper itself, whereas production never calls that accessor. Please prune collected records on a regular path independent of a credential change, and check that the underlying registry remains bounded without invoking a cleanup-performing test helper.

### [P2] Retain subscription anchors in the existing rotation tests

With the new ownership rule, retaining the listener alone does not keep its separate `subscription` allocation alive; only the returned unsubscribe closure does. However, [`TestStreamingProviderPushes`](https://github.com/Magma-Devs/smart-router/blob/3386b10fe8db783919ebc8e176f13a0a3bb4ee55/ecosystem/cache/redisstore/credentials_test.go#L177-L180) discards the second listener's closure, and [`TestRefreshReportsSourceFailureOncePerOutage`](https://github.com/Magma-Devs/smart-router/blob/3386b10fe8db783919ebc8e176f13a0a3bb4ee55/ecosystem/cache/redisstore/credentials_test.go#L306-L309) discards its only closure. Both still assert that those listeners receive subsequent rotations. A collection between subscribing and refreshing now makes those assertions fail. This is consistent with [Go's weak-pointer lifetime rules](https://pkg.go.dev/weak#Pointer).

Reproduced on this head with:

```bash
GOGC=1 go test ./ecosystem/cache/redisstore \
  -run '^(TestStreamingProviderPushes|TestRefreshReportsSourceFailureOncePerOutage)$' \
  -count=100
```

Both tests failed with missing rotation notifications; the same command passed on the #399 parent. Please retain each unsubscribe closure, for example in `t.Cleanup`, and force a collection before the delivery assertions so these tests verify the intended live-subscription behavior instead of depending on GC timing.

### Validation notes

- The unmodified Redis-store package passed with `-race`; `protocol/performance` passed its normal package test run.
- Real-Valkey Docker checks passed for explicit-refresh live rotation, TLS/mTLS, Sentinel failover, and the three-node cluster.
- `TestLiveRotationThroughConfiguredStore` intermittently failed with a re-authentication deadline and its original-connection assertion. The same failure also reproduced on #399 during a ten-run comparison, so it is recorded as an inherited validation issue, not an additional regression attributed to this PR.

avitenzer:
Fixed in be7dc60.

**Pruning on a regular path.** `Refresh` now prunes collected records under the lock before it decides whether the credentials changed, so the watcher's ordinary unchanged-file tick does the housekeeping; `Subscribe` prunes as well, so the registry is bounded by the live connections plus those not yet collected however rarely the watcher polls. The counting helper still prunes as it counts, but the new `TestProviderRegistryStaysBoundedAcrossOutages` reads the raw registry (`registrySize`, no pruning) through three batches of ten failed handshakes, each followed by a collection and an unchanged-credentials refresh: the registry returns to zero after every batch. On the previous code it stayed at ten after the first batch.

**Anchors in the rotation tests.** `TestStreamingProviderPushes` and `TestRefreshReportsSourceFailureOncePerOutage` now hold every unsubscribe closure for the test's life via `t.Cleanup` and force a collection before the delivery assertions. Your reproduction, `GOGC=1 go test ./ecosystem/cache/redisstore -run '^(TestStreamingProviderPushes|TestRefreshReportsSourceFailureOncePerOutage)$' -count=100`, passes.

`go test ./ecosystem/cache/redisstore ./protocol/performance` pass. The intermittent `TestLiveRotationThroughConfiguredStore` failure you recorded as inherited from #399 is not touched here.

AnnaR-prog:
## Review: `/code-review high` on `be7dc60`

Scope: this PR's three commits (`7914fe7...be7dc60`, without the #398/#399 diff the "Files changed" tab still shows), the whole `StreamingProvider`, `New`/`buildClient`/the three option builders, and the go-redis v9.22.0 code the design leans on (`initConn`, `initPooledConn`, `pool.Conn.Close`, `SetOnClose`). Each claim was re-verified against a worktree at this head; the verdict says how far.

**Stack note:** stacked on #399 → #398 (base `main` on GitHub).

The mechanism is as described, and I could not break the design. Read in go-redis v9.22.0: a failed `HELLO` does `cn.GetStateMachine().Transition(pool.StateClosed)` (`redis.go:804`), `initPooledConn` then calls `p.Remove` (`redis.go:535`), which reaches `cn.Close()` — and `Close` returns early on `StateClosed` (`conn.go:1114`) before `onClose` or the socket close ever run. `SetOnClose` has exactly one caller (`redis.go:772`), and `net.netFD` carries a `runtime.SetFinalizer(fd, (*netFD).Close)` (`fd_posix.go:44`), so the "socket closes at collection" claim holds. The four new tests pass 5× under `-race` (75 s), and `TestLiveRotationSmokeOverMiniredis` forcing two collections before the push is the right guard for the contract side (a live connection's subscription must survive a GC).

### 1. Low — file descriptors are now bounded by GC cadence, not by the failure, and the bound is worth stating — **CONFIRMED**

`ecosystem/cache/redisstore/credentials.go:113`, `docs/RESP-CACHE.md:189`

An abandoned connection keeps its socket until the collector reclaims `netFD` and its finalizer runs. A cycle is triggered by heap growth (GOGC), and every failed handshake allocates the same two 32 KiB buffers the ticket measured, so the number of sockets open at once during an outage is roughly `live heap × GOGC/100 ÷ 64 KiB`: for a router with 256 MiB live, about 4,000 — the ticket's 3,510 is not a coincidence, it is one cycle's worth. Under a 1024 `nofile` soft limit (systemd's default for a service) that is exhaustion before the first cycle, and the router's other sockets (provider dials, listener accepts) fail with `EMFILE` for the rest of it. The PR is right that #406's breaker only mitigates the *leak*; the reverse also holds — it is the breaker, not this PR, that bounds the *fd count* during an outage, and with both merged the two together are the fix. Two asks: the docs sentence ("Its file descriptor therefore closes at the next collection") should say what the next collection depends on, and the PR body's "the runtime closes the socket when it reclaims the connection" deserves the same qualifier. If a hard bound is wanted without the breaker, the only in-repo seam that sees every socket is `trackingDialer` (`config.go`); a dialer that closes its `net.Conn` when the conn is reclaimed is the same finalizer, so there is no cheap way around GC here — hence Low, not a fix request.

### Nits

- `Subscribe` walks the whole registry under `mu` on every connection init, which is the hot path of an outage; bounded by the uncollected records (thousands between cycles, microseconds per walk), so fine — but a comment saying that bound is what makes it acceptable would stop the next reader from moving the prune to the watcher only, which the Codex review already had to talk out of once.
- `runtime.KeepAlive(sub)` is the closure's only mention of `sub`; the type comment explains the anchor, but a one-line comment at the call ("captured so the closure is the strong reference — not a no-op here") saves the reader a trip.

### Verified and not flagged

- Static credentials now go through `opts.Username/Password` → `resolveCredentials`, which is what the provider handed go-redis before; the typed-nil guard is present in both option builders, and sentinel's `failoverOptions` still gets `cfg.credentialsSource()` per connection attempt, unchanged.
- The watcher is started only when a provider exists (`store.go:208`), so no nil `Refresh`.
- A re-initialised connection's second `Subscribe` orphans nothing: the overwritten closure drops the old `sub`, which is pruned on the next `Subscribe`/`Refresh`.
- `liveListenersLocked` snapshots strong `listener` references under the lock, so a collection between the snapshot and the push cannot nil a listener mid-loop.
- `weak` needs Go 1.24; `go.mod` is at 1.26.6.

avitenzer:
Fixed in df6991c, on the branch rebased onto #399's new head bd7cb37. The last-good fallback, the synchronized log sink and the `Close` join carry through; under the weak-pointer registry the fallback still applies to a connection whose `Subscribe` re-reads during an outage, and #399's outage tests hold their unsubscribe closures the way the rotation tests here do, so your `GOGC=1 -count=100` reproduction extended with them passes.

1. The docs sentence now says what the next collection depends on: heap growth per `GOGC`, not the failure. Every failed handshake allocates the same two 32 KiB buffers, so the sockets open at once during an outage are roughly `live heap × GOGC/100 ÷ 64 KiB`, about 4,000 at 256 MiB live, one cycle's worth and where the ticket's figure came from; under a low `nofile` limit that is exhaustion before the first cycle, with the router's other sockets failing alongside. It names the request-path breaker (MAG-3676) as what bounds how many connections an outage opens, this change as what bounds how long each is held, and the two together as the fix. The PR body carries the same qualifier and now points at #406, as your stack note asked. No hard bound without the breaker, for the reason you gave: the only seam that sees every socket is the dialer, and closing the conn when it is reclaimed is the same finalizer.

Nits: a comment in `Subscribe` says the registry walk on every connection init is acceptable only because it is bounded by the records not yet collected, and what moving the prune to the watcher alone would do; a comment at `runtime.KeepAlive(sub)` says the mention is the capture that anchors the subscription, not a no-op.

Verified: at df6991c, `go build ./...`, `go vet` and `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/...` pass; the rotation, watcher, abandoned-connection and outage tests pass `go test -race -count=3` (67 s); `GOGC=1 go test ./ecosystem/cache/redisstore -run '^(TestStreamingProviderPushes|TestRefreshReportsSourceFailureOncePerOutage|TestSubscribeDuringSourceOutageHandsOutLastGoodCredentials)$' -count=100` passes. golangci-lint's findings at this head are the four inherited from main plus one unchecked type assertion in this PR's own `TestStaticCredentialsSkipTheStreamingProvider`, the same shape as the one on main two tests above it; not touched here.

avitenzer:
Rebased onto #399's d7e0b56; tip is 7aa83a5. The conflicts were in `New` and the imports only. The provider is still built only for a file-backed credential on a client that subscribes, now inside `resolveClientCredentials`, the one place #398 resolves a store's credentials, so the sentinel client is handed the same source as the fail-fast rather than a third `FileCredentials`; the weak-pointer registry, the prune on every refresh and subscribe, the anchored rotation tests and the docs paragraph carry through unchanged. Underneath, a credential file that is empty once trimmed is one warning per outage and keeps the previous credentials on every connection (#398, #399). The one lint finding this branch carried beyond `main`'s, the unchecked assertion in `TestStaticCredentialsSkipTheStreamingProvider`, is gone (7aa83a5): `require.Same` compares the pointers the values carry, so the assertion added nothing.

Verified at 7aa83a5 from a clean checkout: `go build ./...`, `go vet`, `golangci-lint run --config .golangci.yml` (no findings beyond `main`'s) and `go test -race -count=1 ./ecosystem/cache/... ./protocol/performance/...` pass; the credential, watcher, rotation, outage and abandoned-connection tests pass `go test -race -count=3`. The package has one `TestMain`, in `log_capture_test.go`.

AnnaR-prog (approved):

avitenzer:
Rebased onto #399's 4e69507 after #396 merged to `main` (3cf9bcf); no content change of its own, clean replay. Tip is 5206136; re-verified from a clean checkout.

avitenzer:
Rebased onto #399's 02184d2 after #398 merged to `main` (6ce9cda); no content change of its own. Tip is 4d5649e, re-verified from a clean checkout.

avitenzer:
Rebased onto the current `main`; clean replay, no content change, tree byte-identical to the clean three-way merge of the previous tip. Tip is 3699e58d, re-verified from a clean checkout (build, lint against `main`'s baseline, `go test -race` over the cache, performance, router and relay packages).

PR #408, opened 2026-09-19
fix(cacheformat): return a caller's text request id byte for byte on a cache hit

avitenzer:
# Description

**Bug (MAG-3611).** A reply served from the cache carried a text request id wrapped in a second pair of quotation marks: a caller asking under `abc` got `"abc"` back and could not match the answer to its question. Number ids were unaffected, and a reply from a chain node was always right, so the fault sat on the cache read path alone. Three test-framework workarounds had been stripping the extra quotes for as long as they existed.

**Root cause.** `protocol/chainlib/cacheformat/formatter.go` extracted a text id as its raw JSON text, quotes included, and then restored it with a call that encodes a Go string as a JSON string, adding the second pair. A number id was extracted as an int64 instead, which is why numbers survived. That same int64 conversion also returned a fractional id as `1` and an id beyond int64 as a wrong negative number, found while writing the tests.

**Fix.** The id is carried as the raw JSON text it arrived in, for every type, and written back verbatim with `sjson.SetRawBytes`. Nothing is decoded, so nothing can be re-encoded. A request that carries no id (a notification) restores `null`, as it did before. Batches restore each element's id the same way. The cache key is unchanged: every request still hashes under the fixed id `1`.

Files: `formatter.go` and a new `formatter_test.go`. The package had no test file; MAG-3625, which asked for one, is folded into this ticket.

## Tests

`TestOutputFormatterRestoresTheCallersIDByteForByte` covers a text id, a text id with an escaped quote, a number, a fractional number, a number beyond int64 and `null`, and checks that requests differing only by id key identically. `TestOutputFormatterRestoresEachIDOfABatch` and `TestOutputFormatterRestoresNullForARequestWithoutAnID` cover the batch and no-id shapes.

Run against the previous formatter (checked by swapping the old file back in), the text, escaped-text, fractional and oversized cases fail: `"abc"` comes back as `"\"abc\""`, `1.5` as `1`, the large number as `-4362896299872285998`. The number and null cases pass, which is the control the ticket asked for.

`go build ./...`, `go vet ./protocol/... ./ecosystem/...`, `go test ./protocol/chainlib/...` and `go test ./protocol/rpcsmartrouter/` pass locally.

## Out of scope

The expected-failure marker on the automation repo's `test_request_id_cache_key.py` still names MAG-3472 and needs repointing at MAG-3611 there; the test turns green on its own once this ships.

Jira ticket: MAG-3611

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable; a cached reply now carries the id the caller sent)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3611)

## Reviewers Checklist

I have...


AnnaR-prog:
## Review: `/code-review high` on `fe82e59`

Scope: the full diff `main...fe82e59`, both callers of the formatter (`chain_fetcher.go:931`, the websocket subscription path at `consumer_websocket_manager.go:561`), the two places the output formatter runs on a cached reply (`rpcsmartrouter_server.go:4072`, `secondary_cache.go:163`), and a set of ids beyond the test's. Re-verified against a worktree at this head; the package tests pass there.

Branch is standalone. The fix is the right one — carry the caller's bytes, never decode — and I found nothing to change. Probe at this head, restored byte for byte in every case: a text id with an escaped quote and a non-ASCII rune (`"é\"x"`), an object id (`{"a":[1,2]}`), the empty string, an exponent number (`1e3`), and a batch whose middle element is a notification (restored as `null`, the others as `"a"` and `7`).

### Nit

- The same class survives on a different path: `handleRateLimitReached` (`consumer_websocket_manager.go:173`) parses the id with `strconv.Atoi(result.Raw)`, so a text id (or any non-integer) comes back as `0` in the rate-limit error — the raw-text carry this PR introduces would fix it the same way. Separate path, separate ticket if wanted.

### Verified and not flagged

- The cache key is unchanged: every id still becomes `1` before hashing, and a notification inside a batch still gets `"id":1` added for the key, as before.
- `noExtractedID` as raw `"-1"` produces the same wire bytes the old Go-string `"-1"` did through `SetBytes`, so the "output formatter ran without the input formatter" edge is unchanged.
- `Raw` excludes the surrounding whitespace, so `"id" : "abc" ,` restores as `"id":"abc"`; the request's own spacing is preserved in the keyed form exactly as it was before this PR (a pre-existing property of `sjson.SetBytes`, not touched here).
- The websocket subscription path takes the fix for free: it calls the same `inputFormatter` to remember the id and the same `outputFormatter` to restore it.

avitenzer:
Nothing changes on this PR.

The nit is real and is a different path: `handleRateLimitReached` builds the rate-limit error from `common.JsonRpcRateLimitError`, whose `Id` is an `int`, and parses the request's id with `strconv.Atoi` on `result.Raw`, which still carries its quotes, so every text id comes back as `0` there. Carrying the raw text the way this PR does would fix it, but the error type's integer id field has to become raw JSON first, which is its own wire change on every rate-limit reply, so it goes separately: MAG-3786.

The verified-and-not-flagged points stand as written: the cache key is unchanged, `noExtractedID` produces the same bytes as before, surrounding whitespace is dropped as it always was, and the websocket subscription path takes the fix through the same formatter.

AnnaR-prog (approved):

PR #409, opened 2026-09-19
fix(rpcsmartrouter): never cache an answer the spec says is not reproducible across nodes

avitenzer:
# Description

**Bug (MAG-3461).** Both cache gates read only the method's `stateful` flag, so a call that creates or names state on one chain node passed them. `eth_newFilter` asks one node to open a filter and returns an id only that node knows; the router stored the answer and handed the same id to the next caller as `Cached`, who then polled a filter they did not own on nodes that had never heard of it. `eth_newPendingTransactionFilter`, `eth_uninstallFilter`, `eth_getFilterLogs` (the addition in the ticket's comment) and `eth_accounts` behaved the same way. The spec marks every one of them `deterministic: false`; nothing on the cache path read that flag.

**Fix.** One rule, `cacheExclusionReason`, now decides both the lookup (in `sendRelayToEndpoint`) and the write (in `tryCacheWriteResolved`): a stateful request is excluded as before, and so is any request the spec marks non-deterministic, the spec's own statement that the answer is not the same across nodes. A batch is held to its least reproducible member (`SpecCategory.Combine`). `chainlib.IsDeterministic` joins `GetStateful` as the accessor. `eth_getFilterChanges` and `eth_newBlockFilter`, which escaped by accident until now, are excluded by design.

**What this costs, on purpose.** The flag also covers methods whose answers merely differ across nodes without harm, and those stop being cached too. Per spec file, the non-stateful methods that no longer cache:

- `aptos.json`: `/`, `/-/healthy`, `/spec`, `/tables/{table_handle}/item`, `/transactions`, `/transactions/batch`, `/transactions/encode_submission`, `/transactions/simulate`, `/view`
- `arbitrum.json`: `arbtrace_call`, `arbtrace_callMany`, `arbtrace_filter`, `arbtrace_replayBlockTransactions`, `arbtrace_replayTransaction`, `arbtrace_transaction`
- `btc.json`: `bd_estimatefee`, `bd_marketquote`, `getmempoolancestors`, `getmempooldescendants`, `getmempoolentry`, `getmempoolinfo`, `getrawmempool`, `testmempoolaccept`
- `cosmossdk.json`: `/cosmos/base/tendermint/v1beta1/node_info`, `/cosmos/base/tendermint/v1beta1/syncing`, `/node_info`, `/syncing`, `cosmos.authz.v1beta1.Query/GranteeGrants`, `cosmos.authz.v1beta1.Query/GranterGrants`, `cosmos.base.node.v1beta1.Service/Config`, `cosmos.base.tendermint.v1beta1.Service/GetNodeInfo`, `cosmos.base.tendermint.v1beta1.Service/GetSyncing`, `cosmos.tx.v1beta1.Service/GetTxsEvent`, `cosmos.tx.v1beta1.Service/Simulate`
- `cosmoswasm.json`: `/cosmwasm/wasm/v1/code/{code_id}`, `/cosmwasm/wasm/v1/code/{code_id}/contracts`, `/cosmwasm/wasm/v1/contract/build_address`, `/cosmwasm/wasm/v1/contract/{address}`, `/cosmwasm/wasm/v1/contract/{address}/history`, `/cosmwasm/wasm/v1/contract/{address}/raw/{query_data}`, `/cosmwasm/wasm/v1/contract/{address}/smart/{query_data}`, `/cosmwasm/wasm/v1/contract/{address}/state`, `/cosmwasm/wasm/v1/contracts/creator/{creator_address}`, `cosmwasm.wasm.v1.Query/AllContractState`, `cosmwasm.wasm.v1.Query/BuildAddress`, `cosmwasm.wasm.v1.Query/Code`, `cosmwasm.wasm.v1.Query/ContractHistory`, `cosmwasm.wasm.v1.Query/ContractInfo`, `cosmwasm.wasm.v1.Query/ContractsByCode`, `cosmwasm.wasm.v1.Query/ContractsByCreator`, `cosmwasm.wasm.v1.Query/RawContractState`, `cosmwasm.wasm.v1.Query/SmartContractState`
- `ethereum.json`: `debug_getBadBlocks`, `debug_traceBlockByHash`, `debug_traceTransaction`, `eth_accounts`, `eth_coinbase`, `eth_getCode`, `eth_getCompilers`, `eth_getFilterChanges`, `eth_getFilterLogs`, `eth_getLogs`, `eth_hashrate`, `eth_maxPriorityFeePerGas`, `eth_mining`, `eth_newBlockFilter`, `eth_newFilter`, `eth_newPendingTransactionFilter`, `eth_sendUserOperation`, `eth_subscribe`, `eth_supportedEntryPoints`, `eth_syncing`, `eth_uninstallFilter`, `eth_unsubscribe`, `net_listening`, `net_peerCount`, `trace_block`, `trace_call`, `trace_filter`, `trace_get`, `trace_rawTransaction`, `trace_replayBlockTransactions`, `trace_replayTransaction`, `trace_transaction`, `web3_clientVersion`
- `hyperliquid.json`: `eth_bigBlockGasPrice`, `eth_getCode`, `eth_getLogs`, `eth_getSystemTxsByBlockHash`, `eth_getSystemTxsByBlockNumber`, `eth_maxPriorityFeePerGas`, `eth_syncing`
- `solana.json`: the eighteen `*Subscribe`/`*Unsubscribe` methods
- `tendermint.json`: `abci_info`, `block_results`, `consensus_state`, `status`, `subscribe`, `tx`, `unsubscribe`, `unsubscribe_all`

The ones with real cache value are `eth_getLogs`, `eth_getCode`, the `trace_*` and `debug_*` families, Tendermint `block_results` and `tx`, and the CosmWasm contract queries. Taking the spec at its word was chosen over a router-side list of method names because a list needs maintaining per chain and silently misses the next chain (the ticket names Solana and Aptos cases it could not measure), and because the flag has correctness merit for most of these too: a trace from one client implementation is not interchangeable with another's, and a partially indexed node's `eth_getLogs` answer should not be served to everyone for ninety minutes. If a method on that list should cache, the place to say so is its spec flag. The alternative, a deny-list of names in the router, is a one-function change to `cacheExclusionReason` if the reviewer prefers it.

Two of the tickets after this one are touched: `debug_traceTransaction` from MAG-3462's list and `eth_getLogs` with `toBlock: latest` from MAG-3460's list stop being cached here; those PRs cover the remaining methods.

Files: `rpcsmartrouter_server.go` (the rule and both gates), `chainlib/chain_message_queries.go` (the accessor), `README.md` (one sentence on what is never cached), the new `cache_eligibility_test.go`, and `cache_test_helpers_test.go`, shared byte for byte with #410 and #411 so whichever lands first carries it.

**Independent of the other cache PRs.** Based on `main`; its hunks are clear of #410's and #411's, verified pairwise, and the union of #408 to #412 builds and passes the router and chainlib suites.

## Tests

`TestCacheExclusionReasonFollowsTheSpec` parses real requests with the Ethereum spec: the five affected methods report `non-deterministic`, `eth_sendRawTransaction` reports `stateful`, `eth_getBalance` and `eth_getTransactionReceipt` report nothing, and a batch of a balance and a filter is non-deterministic. `TestTryCacheWriteNeverStoresANonDeterministicAnswer` runs the write path against a real cache server: an `eth_newFilter` answer is never stored, while an `eth_getBalance` control written the same way is stored and found. Watched go red first: with the rule in place but the gates still reading only `stateful`, the filter id landed in the cache.

`go vet` and `go test ./protocol/rpcsmartrouter/ ./protocol/chainlib/` pass locally.

## Out of scope

The lookup gate is exercised through the shared rule rather than by driving `sendRelayToEndpoint`, which needs live sessions. The automation repo's `test_side_effecting_call_cache.py` is the end-to-end check and turns green on its own.

Jira ticket: MAG-3461

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable; no API change, though the set of cached methods narrows as listed above)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3461)

## Reviewers Checklist

I have...


AnnaR-prog:
## Review: `/code-review high` on `7482c62`

Scope: the full diff `main...7482c62`, both cache gates, every other reader of `Category.Deterministic` on `main`, the bundled specs (what the flag actually says per method), and the shared test helper against #410/#411 (byte-identical on all three heads, as stated). Re-verified against a worktree at this head; the two new tests pass there.

Branch is standalone. The mechanism is right — one rule, both gates, batches folded by `Combine` — and the ticket's defect (a filter id handed to a caller who does not own it) is closed. The finding is about the flag chosen to close it, which the PR itself names as the design call to challenge.

### 1. Medium — `deterministic` was authored for a different mechanism and its values do not mean "cacheable"; the cost lands on the highest-value methods and is measurable before merging — **CONFIRMED**

`protocol/rpcsmartrouter/rpcsmartrouter_server.go:3617`

Two facts from `main`:

- The flag's only readers before this PR are the cross-validation paths (`rpcsmartrouter_server.go:4723`, `:4919`): "non-deterministic methods legitimately differ", so their content outliers are not counted. That is its provenance — data-reliability comparison across providers — not cache safety.
- Its values are inconsistent under the cache reading. In `specs/ethereum.json`, `eth_getCode` is `false` while `eth_getStorageAt` and `eth_getBalance` are `true`; the three are the same kind of question (state at a block, archive nodes agree, non-archive nodes error rather than lie). Under this PR `eth_getCode` stops caching and `eth_getStorageAt` keeps caching.

Counted over the bundled specs, the non-stateful methods that stop caching: Ethereum 33 of 78, Hyperliquid 7 of 27, CosmWasm 18 of 24 — every `cosmwasm.wasm.v1.Query/*` contract read, which is deterministic at a height by construction. `eth_getCode`, `eth_getLogs` and the `trace_*`/`debug_*` families are the requests dapps and indexers repeat most and the ones nodes pay most to answer; the PR's own list says so. There is a per-function cache-hit metric (`RecordCacheHitRequest`, `smartrouter_metrics_manager.go:1232`), so the hit-rate cost can be read off a production router today — the share of hits on the methods in the PR's list — and that number belongs in the PR before this ships.

The `eth_getLogs` argument ("a partially indexed node's answer served for ninety minutes") is real, and it is an argument for a flag that says *that*. The defect class this ticket names is "the answer names state that lives on one node" — filters, subscriptions, `eth_accounts`, `eth_coinbase`, mempool views — which is small, enumerable, and the same on every chain that has it. Options, in the order I would take them: a dedicated spec category (`node_bound`, or reuse `hanging_api`'s style) set on that family in the specs, with the router gating on it; until the specs carry it, the deny-list the PR offers as its alternative (one function, the family above); and leave `deterministic` to the cross-validation code that owns it. If the flag is kept as the gate, the spec values need a pass first (`eth_getCode`, the CosmWasm reads) and the release note has to say which methods stop caching, because the hit-rate drop will read as a regression to whoever runs the dashboard.

### Verified and not flagged

- Both gates read the one rule; the lookup gate's ordering (cross-validation first, then exclusion, then force-refresh) is unchanged.
- `SpecCategory.Combine` AND-s `Deterministic` and takes the max `Stateful`, so a batch is held to its least reproducible member, as described.
- `eth_getFilterChanges` / `eth_newBlockFilter` were cached by accident on `main` and are correctly excluded under any of the options above.
- The overlap with #410 (`eth_getLogs` with `toBlock: latest`) and #411 (`debug_traceTransaction`) is stated in both directions; if the gate changes per the above, those two PRs' lists are the ones that keep the behaviour.

avitenzer:
Fixed in 76b17e2.

1. The gate no longer reads the spec's `deterministic` flag; it reads a router-side list of node-bound methods, `nodeBoundMethods` in the new `node_bound_methods.go`, and the flag stays with the cross-validation code that owns it (the `IsDeterministic` accessor is gone with it). Of the three options you ordered I took the second, the deny-list, because the first is a lava-specs change: a `node_bound` category in the style of `hanging_api` is the durable home for this family, and the reply on that repo can carry the list as its starting point, but until the specs carry it the list is the gate here. The list is the family the ticket names, built from the bundled specs rather than Ethereum alone, with a reason per entry in four groups: state a node holds for one caller (the EVM filters, `eth_subscribe`/`eth_unsubscribe`, Solana's eighteen subscription methods, Tendermint's `subscribe`/`unsubscribe`/`unsubscribe_all`), the node's own keys (`eth_accounts`, `eth_coinbase`, `getIdentity`), its mempool (the `txpool_*` views, `eth_pendingTransactions`, BTC's six mempool queries, Tendermint's `unconfirmed_txs`/`num_unconfirmed_txs`, and `eth_sendUserOperation`, a submission the spec leaves unflagged), and its own status (`eth_syncing`, `net_peerCount`, `web3_clientVersion` and the rest of that family on EVM, `getHealth`/`getVersion`/`getClusterNodes` on Solana, `status`/`health`/`net_info`/`abci_info`/`consensus_state` on Tendermint, the node-info, syncing, status and config endpoints on Cosmos SDK, `/`, `/-/healthy` and `/spec` on Aptos). It sits where the flag was, in `cacheExclusionReason`, at both gates; a batch's API name is its members' names joined by `chainlib.SEP`, so a batch is still held to its least cacheable member. Your two examples of the flag's inconsistency are now in the file's comment as the reason it is not the gate, and I found a third while building the list: Tendermint's `status` is false while `net_info` is true, and Solana's `getIdentity` and `getHealth` are true.

On the cost: with the list, nothing that was cacheable on `main` and has a reusable answer stops caching. `eth_getCode`, `eth_getStorageAt`, `eth_getLogs`, the `trace_*` and `debug_*` families, Tendermint's `block_results` and `tx` and every CosmWasm read cache exactly as on `main`, so the hit-rate number from production is moot for this PR; the only methods that stop caching are the node-bound ones, and `TestNodeBoundMethodsInTheBundledSpecs` pins exactly which declared methods that is, per spec file, so the blast radius is a diff to read rather than a count to trust. Some of those were `deterministic: true` in the spec (`getHealth`, `getIdentity`, `eth_getWork`, Tendermint's `net_info` and `health`), which is the same inconsistency in the other direction. The overlaps you noted flip back: `debug_traceTransaction` and `eth_getLogs` with `toBlock: latest` keep caching as on `main`, and #411 and #410 now say what happens to each rather than pointing here.

The PR body's design-call paragraph and cost list are replaced accordingly; the title stays.

Verified: from a detached checkout of 76b17e2, `gofmt -l` is clean, `go build ./...`, `go vet` and `golangci-lint run --config .golangci.yml` over `./protocol/rpcsmartrouter/... ./protocol/chainlib/... ./ecosystem/cache/...` add nothing over `origin/main`'s findings, and `go test -race -count=1` over the same three trees passes (rpcsmartrouter 116 s); `TestCacheExclusionReasonRefusesNodeBoundMethods`, `TestTryCacheWriteNeverStoresANodeBoundAnswer` and `TestNodeBoundMethodsInTheBundledSpecs` pass `-race -count=3`. `cache_test_helpers_test.go` is untouched and stays byte-identical to #410's and #411's.

Not done: the spec-level category, which belongs in lava-specs; and the spec's own `stateful` gaps this turned up (`eth_sendUserOperation` and Aptos' `POST /transactions` are submissions the spec does not mark stateful; the user-operation one is on the list, the Aptos path is not, since the same name serves the GET), which are spec fixes too.

AnnaR-prog (approved):

AnnaR-prog (approved):

PR #410, opened 2026-09-19
fix(rpcsmartrouter): file a latest-tagged answer under the key its lookup computes

avitenzer:
# Description

**Bug (MAG-3460).** For a request that names no block, the router resolved "latest" twice and got two different numbers. The lookup used the tracked chain tip; the write used `Reply.LatestBlock`, the block the answer turned out to be about. For a transaction receipt that is the transaction's own block, almost never the newest one, so the entry was filed under a key no lookup ever computed, and every repeat of `eth_getTransactionReceipt` went to a provider while the cache reported a healthy hit rate. The ticket's six measured cases all reduce to one rule: equal numbers are served, different numbers are not.

**Root cause.** Commit `969c076` correctly repointed the lookup at the guarded chain tip, retiring the per-user seen block a lying provider could capture. The write path kept its older preference for the reply's own block, and the comment saying the two must match stopped being true.

**Fix.** One method, `latestCacheBlock`, resolves the block for both ends from the parse-time tip stamped in `RelayPrivateData.SeenBlock`, the single producer every downstream reader is documented to draw from, with the gated tip as the fallback for a request parsed before any tip was known. Neither end reads the reply. A write that can resolve no tip at all still skips, as before. The secondary-cache backfill, which already carries the exact block that hit, is unchanged.

What this means for an entry: a "latest" answer is filed under the tip the request was parsed at, found by every request parsed at that tip, and naturally missed once the tip moves, which is what "latest" means; the short store's lifetime bounds staleness as before. A node one block behind the tip now has its `eth_blockNumber` answer served under the tip for that window, where before that entry was never found at all.

**Independent of the other cache PRs.** Based on `main`; `latestCacheBlock` lives in its own file and the two edited regions are clear of #409's and #411's hunks, verified pairwise. The union of #408 to #412 builds and passes the router and chainlib suites. MAG-3462 (#411) moves receipts and the other by-hash answers to a constant key that survives tip advance; this PR fixes the parity for every "latest" request, and its tests hold with or without #411.

Files: `rpcsmartrouter_server.go` (the lookup and the write), the new `latest_cache_block.go` (the shared resolution), `cache_key_parity_test.go`, and `cache_test_helpers_test.go`, shared byte for byte with #409 and #411.

## Tests

`TestLatestCacheBlockComesFromTheParseTimeTip` pins the resolution: the parse-time tip first, the gated tip when none was stamped, nothing when neither exists. `TestLatestTaggedWriteLandsOnTheLookupKey` runs the write against a real cache server in two shapes where the answer's own block differs from the tip: the head as reported by a node one block behind (`eth_blockNumber`) and the newest block as reported by a node one block ahead (`eth_getBlockByNumber`). Each entry must be found under the lookup's key and not under the answer's block. `TestReceiptIsFiledWhereItsLookupLooks` is the ticket's own case, a receipt of a transaction at 19999000 with the tip at 20000000: it must not sit under the transaction's block, and must sit where a lookup for it looks, which is the parse-time tip here and the identity key once #411 lands. Watched go red first, with the lookup already on the shared resolution and the write still keying on the reply: every case failed to find its entry.

`go vet` and `go test ./protocol/rpcsmartrouter/` pass locally.

## Out of scope

The ticket asks whether REST, Tendermint and gRPC share the mismatch. The fix is transport-agnostic, since both ends read the same relay data, but only the JSON-RPC path was measured and is tested here. The ticket's third question, whether an empty `eth_getLogs` result should be cached at the tip, is moot after #409: `eth_getLogs` is non-deterministic in the spec and is no longer cached.

Jira ticket: MAG-3460

---

## Author Checklist

I have...

* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3460)

## Reviewers Checklist

I have...


AnnaR-prog:
## Review: `/code-review high` on `aa1d806`

Scope: the full diff `main...aa1d806`, the producer of `RelayPrivateData.SeenBlock` on the request path (`rpcsmartrouter_server.go:1057`), the engine's hit validation (`core/engine.go:241`), the write's `finalized`/`seenBlock` handling around the edited hunk, and the shared test helper against #409/#411 (byte-identical). Re-verified against a worktree at this head; the three new tests pass there.

Branch is standalone. The diagnosis is right and the fix does what it says: one resolution, both ends, the reply never consulted. The anti-lie property of `969c076` is preserved — `SeenBlock` is stamped from `getLatestBlock()` at parse time (`:1057`), so preferring it over "the tip now" is a snapshot of the same guarded value, not a return to the per-user block. One consequence the PR names in passing deserves a number and, I think, a one-line guard.

### 1. Medium — parity is restored by filing a lagging node's answer under the tip, and nothing downstream catches that — **CONFIRMED**

`protocol/rpcsmartrouter/rpcsmartrouter_server.go:3727`, `ecosystem/cache/core/engine.go:241`

Before this PR a node one block behind had its `eth_blockNumber` / `eth_getBlockByNumber("latest")` answer filed under `N-1` — a key no lookup at tip `N` computed, which was the bug for receipts and, by accident, a freshness guard for head-shaped answers: the stale answer was never served. After it the same answer is filed under `N` with `SeenBlock = N` (the parse-time tip), and the engine's hit rule is `stored.SeenBlock < min(req.SeenBlock, req.RequestedBlock)` → `N < N` is false, so every caller parsed at tip `N` is served block `N-1` as "latest" for the non-finalized lifetime (`max(avgBlockTime/8, 500ms)`, ~1.5 s on Ethereum) while the router's own tip says `N`. The PR body says this; what it does not say is how often the first answer at a new tip comes from a lagging node. The optimizer does pick them — the MAG-1748 work measured stale-provider picks at 35% before and 29% after its fix on a router against the provider simulator, which has lagging providers by construction; production is lower, not zero — so this is not a corner.

The cheap guard: at the write, for a `LATEST`-tagged request, skip when `relayResult.Reply.LatestBlock > 0 && Reply.LatestBlock < requestedBlockForCache` — the answer is about a block older than the key claims. For head-shaped methods that is exactly "the node is behind". For a receipt or a block fetched by hash it is also true (their `Reply.LatestBlock` is the object's block), which puts those back to `main`'s behaviour until #411 moves them to a constant key — which is that PR's whole point, and it is in this batch. With #411 merged, the only `LATEST` writes left are head-shaped, and the guard then reads as the freshness rule it is. If the trade is preferred as is, the docs' "a lagging replica can never serve data older than what the client has already seen" needs the qualifier that the *router* may have seen one block more.

### Verified and not flagged

- The secondary-cache backfill path still carries `resolvedBlock` and is untouched; the `seenBlock` lift that follows the hunk is unchanged.
- `latestCacheBlock` returns 0 only when neither the parse-time tip nor the gated tip exists; the write skips and the lookup asks for block 0, as before.
- Receipt case: filed under the parse-time tip and found there (test); with #411 the key becomes the identity key, and the test says so.
- The `finalized` decision is still computed from `Reply.LatestBlock` and the local tip, so a `LATEST` answer stays non-finalized and short-lived; the staleness above is bounded by that lifetime.

avitenzer:
Fixed in d2ad607.

1. The guard is in, at the write, in the branch that resolves a latest-tagged key: a `Reply.LatestBlock` above zero and below the block the entry would be keyed under skips the write, since the answer is about an older block than the key claims. For `eth_blockNumber` and `eth_getBlockByNumber("latest")` that is the node being behind, and your timeline no longer happens: a node at the tip and a node one block ahead are filed under the lookup's key, a node one block behind is filed nowhere. For a receipt or a block by hash, whose `Reply.LatestBlock` is the object's own block, it skips too, which is what `main` amounted to for every caller (filed under the object's block, never found), until #411 keys those under a constant block above this branch; with it merged the only latest writes left are head-shaped and the guard reads as the freshness rule it is. One harvest reports neither a tip nor an object's block, and I looked at it because #409 no longer takes `eth_getLogs` out of the cache: that harvest yields the first log's block, so a latest-tagged log query with results skips here, which is again `main`'s effective behaviour, while an empty result harvests nothing and is filed under the tip for the short lifetime, also as on `main`. The comment at the guard and the PR body say so; a harvest that reported the range's upper bound would let those cache and is a separate change. The docs sentence about a lagging replica stays as written: with the guard, an entry under N is about N or newer, so the router's tip cannot be ahead of what it serves.

`TestLatestTaggedWriteLandsOnTheLookupKey` now runs four shapes against a real cache server, the two lagging ones asserting the entry is under neither the tip nor the answer's block, with a same-tip control entry written after each case so the async write path is shown to have drained before the absence is judged; `TestReceiptIsNeverFiledUnderABlockItDoesNotDescribe` is the ticket's receipt, never under its transaction's block and never under the tip it lags, with #411 named as where it lives.

Verified: from a detached checkout of d2ad607, `gofmt -l` is clean, `go build ./...`, `go vet` and `golangci-lint run --config .golangci.yml` over `./protocol/rpcsmartrouter/... ./protocol/chainlib/... ./ecosystem/cache/...` add nothing over `origin/main`'s findings, and `go test -race -count=1` over the same three trees passes (rpcsmartrouter 117 s, which covers the secondary-cache backfill tests the guard sits beside); `TestLatestCacheBlockComesFromTheParseTimeTip`, `TestLatestTaggedWriteLandsOnTheLookupKey` and `TestReceiptIsNeverFiledUnderABlockItDoesNotDescribe` pass `-race -count=3`. `cache_test_helpers_test.go` is untouched and stays byte-identical to #409's and #411's.

AnnaR-prog (approved):

PR #411, opened 2026-09-19
fix(rpcsmartrouter): keep an answer fetched by hash for the finalized lifetime once its block is final

avitenzer:
# Description

**Bug (MAG-3462).** A transaction fetched by its hash, its receipt, a block fetched by its hash, the transaction count of such a block: none of these can come back different once the object's block is final, yet the router kept them for the short store's second and a half and fetched the same unchanging answer again on every poll. The spec's block parsing resolves these methods to "latest" because the request names a hash and no block, and a "latest" entry is keyed under the tip and can never be marked finalized.

**Fix.** The ticket left the choice between a spec change and a router rule open. The spec route was rejected: the one method that already carries a `BLOCK_HASH` parser with this block parsing, `trace_transaction`, resolves to "not applicable" today, which switches the cache off for it entirely, and receipts given the same parser would follow. The router rule instead:

1. **Identity keying.** An EVM by-hash request, "latest" by default value on one of nine listed methods (`eth_getTransactionByHash`, `eth_getRawTransactionByHash`, `eth_getTransactionReceipt`, `eth_getBlockByHash`, `eth_getBlockTransactionCountByHash`, `eth_getTransactionByBlockHashAndIndex`, `eth_getRawTransactionByBlockHashAndIndex`, `eth_getUncleByBlockHashAndIndex`, `eth_getUncleCountByBlockHash`), is keyed under one constant block on both ends, so its entry survives tip advance. The list is by name, like the EVM harvest beside it and for the same reason. The method check is load-bearing: `eth_blockNumber` and `eth_gasPrice` also parse to "latest" by default and stay tip-keyed.
2. **Settling by the answer's own block.** The store is decided from the block the answer is about, re-read from the response rather than taken from `Reply.LatestBlock`: a reply with no block is stamped downstream with the endpoint's observed tip, and that value must never settle an answer that can still change. Past the finalization distance the entry goes to the long store; otherwise to the short one, as today. An answer that carries no block, a pending transaction's `null` or a count, is never settled, because the finalization predicate would take a zero for the genesis block.
3. **Harvest.** `eth_getTransactionByHash`, `eth_getTransactionByBlockHashAndIndex` and `eth_getUncleByBlockHashAndIndex` now carry their block like a receipt or a block does. Of the three methods the ticket measured: transaction by hash settles; block transaction count by hash has no block in its answer and stays in the short store, on a stable key now; `debug_traceTransaction` is non-deterministic in the spec and, since #409, is not cached at all.

**Independent of the other cache PRs.** Based on `main`; the rule lives in its own file and its three hunks in the write and lookup sit between #410's, verified pairwise. The union of #408 to #412 builds and passes the router and chainlib suites. With #410 as well, a "latest" answer that is not by-hash keys under the parse-time tip and a by-hash one under the identity key; without it, the by-hash rule stands on its own.

Files: the new `by_hash_cache.go` (the rule), `rpcsmartrouter_server.go` (the lookup and the write), `direct_rpc_relay.go` (the harvest), the new `cache_by_hash_test.go`, and `cache_test_helpers_test.go`, shared byte for byte with #409 and #410.

## Tests

`TestIdentityKeyedIsTheEVMByHashFamily` pins the rule on real parsed requests, including the three that must stay tip-keyed and a batch. `TestByHashAnswerSettlesOnceItsBlockIsFinal` runs the write against a real cache server with a tracked tip and the shortest short-store lifetime: six by-hash answers all land under the identity key, where a lookup carrying the tip as its seen block finds them (the cache server's seen-block floor is not tripped) and not under the tip; then the short store empties and only the three whose block is eight or more blocks behind the tip survive, while a transaction in the newest block, an unknown transaction's `null` and a count do not. Watched go red first: with the lookup on the identity key and the write still on the tip, no entry was found.

`go vet` and `go test ./protocol/rpcsmartrouter/` pass locally.

## Out of scope

Chains other than EVM: a Cosmos transaction by hash already carries a `BLOCK_HASH` parser in its spec and the REST path harvests no block, so it stays as it is. The block-hash-to-height mapping the cache server can store (`BlocksHashesToHeights`) is not used; nothing writes it today, and identity keying needs no height.

Jira ticket: MAG-3462

---

## Author Checklist

I have...

* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3462)

## Reviewers Checklist

I have...


AnnaR-prog:
## Review: `/code-review high` on `76cb10c`

Scope: the full diff `main...76cb10c`, the engine's key derivation and hit validation for a block-0 key (`core/engine.go`), where the extended harvest flows (`tipBlockFromRelay`, the site-B `latestBlock`, `OnSessionDone`'s sync block), the Ethereum spec's finalization distance, and the shared test helper against #409/#410 (byte-identical). Re-verified against a worktree at this head; both new tests pass there.

Branch is standalone. The rule is sound and the choice of router rule over spec change is argued correctly (`trace_transaction` shows what the spec route does). I found nothing that needs changing; the notes below are what I checked because they could have gone wrong.

### Nits

- `evmByHashMethods` and the `case` lists in `extractBlockHeightFromEVMResponse` are two by-name lists that must agree (the PR extends the second so the first can settle); deriving the harvest cases from the map, or a test that walks the map and asserts each method harvests a block from a representative answer, keeps the next method from landing in one list only.
- `identityKeyBlock = 0` is safe for the reason the comment gives, verified: the engine only consults the chain tip for negative blocks, and the hit rule `stored.SeenBlock < min(req.SeenBlock, 0)` can never be true. Worth one line in the comment that the lookup's `SeenBlock` therefore does nothing for these entries — the finalization decision at the write is the only freshness control.

### Verified and not flagged

- The extended harvest cannot move a tip or a sync score: `tipBlockFromRelay` admits only `eth_blockNumber` / `eth_getBlockByNumber(latest)`, and `OnSessionDone` prefers the per-endpoint tip store, falling back to the harvested block only for an endpoint with no tip record — the same exposure receipts already had on `main`.
- Store choice: `byHashFinalized` takes `answerBlock` from the response, never `Reply.LatestBlock`, so a `null` (pending/unknown) answer and a bare count stay in the short store; an answer about a block ahead of the gated tip stays there too.
- Finalization distance is the spec's `block_distance_for_finalized_data` (8 on ETH1, ~100 s), the same notion by-number data has always settled on; a receipt now inherits it. A reorg deeper than that would serve the old receipt for the finalized lifetime — accepted risk, pre-existing for by-number reads, and rare post-merge.
- Ordering against #410 in both hunks: the by-hash branch takes precedence over the exact-key backfill in the write and over the parse-time tip in the lookup, so with both merged a `LATEST` answer that is not by-hash keys under the parse-time tip and a by-hash one under the identity key, as the PR body states.
- A mixed fleet during rollout (old routers writing receipts under tip keys, new ones reading at 0) only misses; nothing is served wrongly.

avitenzer:
Fixed in 8a1a455.

Nits. `TestEveryByHashMethodHasItsHarvest` walks `evmByHashMethods` with a representative answer per method through `extractBlockHeightFromEVMResponse`: the five that carry a block harvest it, the four that carry none (raw bytes, a count) are pinned as harvesting nothing, a method added to the map without a row fails, and a row for a method that is not in the map fails, so the two by-name lists cannot drift apart in either direction. The comment at `identityKeyBlock` now says the other half of the trade: the seen-block floor is inert for these entries, a lookup's `SeenBlock` does nothing, and the finalization decision at the write is their only freshness control.

One correction that fell out of #409's change of gate: the comment on `evmByHashMethods` said the trace and debug by-hash methods "never reach the cache", and the PR body said `debug_traceTransaction` "is no longer cached at all". With #409 gating on a node-bound list instead of the spec's flag, both cache again as on `main`, tip-keyed in the short store. They are still not identity-keyed, and the comment and body now give the real reason: a trace's shape depends on the client implementation that produced it, so an identity-keyed trace kept for the finalized lifetime would pin one implementation's answer on every caller.

Verified: from a detached checkout of 8a1a455, `gofmt -l` is clean, `go build ./...`, `go vet` and `golangci-lint run --config .golangci.yml` over `./protocol/rpcsmartrouter/... ./protocol/chainlib/... ./ecosystem/cache/...` add nothing over `origin/main`'s findings, and `go test -race -count=1` over the same three trees passes (rpcsmartrouter 118 s); `TestEveryByHashMethodHasItsHarvest`, `TestIdentityKeyedIsTheEVMByHashFamily` and `TestByHashAnswerSettlesOnceItsBlockIsFinal` pass `-race -count=3`. `cache_test_helpers_test.go` is untouched and stays byte-identical to #409's and #410's.

AnnaR-prog (approved):

PR #412, opened 2026-09-19
fix(rpcsmartrouter): judge a secondary cache entry by its contents before serving it

avitenzer:
# Description

**Bug (MAG-3597).** When the router borrows an answer from another zone's cache, it never looked at the answer. Two consequences, and a third the ticket's comment added:

1. The entry's kind came from the two signals the writing zone attached, the `IsNodeError` label and the legacy placeholder. A zone whose build predates the label writes none, so its JSON-RPC error body reached the caller as HTTP 200 with `Lava-Cache-Tier: secondary` and no `lava-identified-node-error` header, against what `docs/SECONDARY-CACHE.md` promises.
2. The bytes went through the id formatter before anyone checked they were a reply. `sjson` turned `<html>502 Bad Gateway</html>` into `{"id":7}` and `{"jsonrpc":"2.0","resu` into `,"id":7}`, both served as successes.
3. An empty entry was served as an empty 200. The router's own envelope check did see it (the reply carried node-error retries), but its verdict never reached the reply.

**Fix.** The entry's contents decide, on the sanitized copy, in two steps:

- **Before anything is reshaped**, an entry that is not a reply at all (no bytes, a web page, a truncated envelope, a bare value) is dropped as a secondary *error*, like a copy failure, and the request proceeds to a provider. It is deliberately not served as a node error: that would put the retry loop through the same entry again, which is what the comment's `Lava-Retries: 5` shows. The formatter itself now refuses to restore an id into anything that is not a JSON object or a batch of them, so no reply can be manufactured on any path.
- **After the placeholder substitution**, `CheckResponseError`, the verdict a live answer gets, classifies a well-formed error body as the node error it is: served with the header, and never backfilled, through the populator's existing node-error check. A labelled entry is unaffected.

Both steps read the envelope on JSON-RPC and Tendermint. REST and gRPC entries carry no envelope and stay judged by their recorded status, as the ticket's coverage note anticipates. The edge the ticket flags in `jsonRPCMessage.go` (an error object without a message reads as a success) is locked by its own test and is unchanged here.

**Independent of the other cache PRs.** Based on `main`; the shape rule lives in its own file and the formatter change is one guard line clear of #408's hunks, verified pairwise. The union of #408 to #412 builds and passes the router and chainlib suites.

Files: `secondary_cache.go` (both checks), the new `cacheformat/reply.go` (the exported shape rule) and one guard line in `cacheformat/formatter.go`, their tests (`reply_test.go`, `secondary_cache_content_test.go`), and the promise in `docs/SECONDARY-CACHE.md`.

## Tests

The existing secondary harness is REST-only and cannot exercise an envelope check, so `secondary_cache_content_test.go` builds the same scene on Ethereum JSON-RPC with the real spec. `TestSecondaryEntryKindComesFromItsContents` stages an error body with neither signal: served, `IsNodeError` set, the caller's id restored, nothing backfilled into a real primary; its success control is served as a success. `TestSecondaryEntryThatIsNotAReplyIsNotServed` stages the ticket's web page and truncated envelope, the comment's empty entry, and a bare value: none is served, each is recorded as a secondary error, and nothing reaches the primary. `TestOutputFormatterLeavesANonReplyAlone` pins the formatter guard. Watched go red first: with the checks absent, the error body was served as a success and all four non-replies were served.

`go vet` and `go test ./protocol/rpcsmartrouter/ ./protocol/chainlib/cacheformat/` pass locally.

## Out of scope

The ticket's first open question, whether a served entry is copied into the router's own cache: the tests here show an error entry is not, which the populator's node-error check explains. REST and gRPC content inspection, and the version gap across the fleet, are as the ticket left them. MAG-3573 (the Helm chart cannot enable the tier) stays blocked on this PR, as linked.

Jira ticket: MAG-3597

---

## Author Checklist

I have...

* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3597)

## Reviewers Checklist

I have...

---

## Codex review — 2026-09-19

Reviewed head [`db8b28e`](https://github.com/Magma-Devs/smart-router/commit/db8b28e9884b291faee65253f3ea662a74e1a926).

### [P2] Reject RPC envelopes with no payload before accepting the cache hit

Location: [`protocol/chainlib/cacheformat/reply.go:20–22`](https://github.com/Magma-Devs/smart-router/blob/db8b28e9884b291faee65253f3ea662a74e1a926/protocol/chainlib/cacheformat/reply.go#L20-L22), and the corresponding batch-element check at lines 30–33.

The new guard accepts every JSON object, including `{}` and `{"jsonrpc":"2.0","id":1}`, although neither contains an RPC response payload. This also admits the `{"id":7}` entries that an older zone could already have produced through the HTML-rewriting bug described above. The later `CheckResponseError` only marks these as node errors: `trySecondaryCacheLookup` still returns `served=true` and records a hit, so the request cannot fall through to an upstream on that attempt and retries can keep selecting the same unusable entry.

I reproduced this against this head with a temporary Go test overlay, reusing `runSecondaryEthLookup` and the real primary cache server. For an `eth_getBalance` request at `0x64`, both missing-payload examples produced `served=true, SecondaryOutcome=hit, IsNodeError=true`. The empty object was rewritten to `{"id":7}`.

The batch case is worse: for two balance requests, this cached body:

```json
[{"jsonrpc":"2.0","id":1,"result":"0x64"},{"jsonrpc":"2.0","id":2}]
```

produced `served=true, SecondaryOutcome=hit, IsNodeError=false`, and the malformed batch was **backfilled into the primary cache**. The default batch error classifier lets the successful sibling mask the missing payload.

Please reject an object with neither a `result` field (allow `null`) nor a non-null `error` payload, checking every batch element before formatting or hit accounting. Route these entries through the existing secondary-error/fallback branch, while keeping valid RPC error responses on the served-node-error path. Add regression cases for the two single objects and the mixed batch; each should return `served=false`, record a secondary error, and leave the primary empty. This is distinct from the explicitly deferred empty-error-message behavior.

Validation: the existing formatter, RPC response-classifier, and secondary-cache tests passed. All three additional missing-payload fallback assertions failed; the mixed-batch test also observed the malformed entry through the real primary cache's `GetRelay`. Repository source files were not changed.

### Resolution (2026-09-20)

Fixed in 5125bce.

`IsRestorableJSONRPCReply` now requires a payload: a `result` member (null included, since an unknown transaction answers null) or a non-null `error`. A batch is judged element by element, so one good sibling no longer vouches for an element that answers nothing. Everything that fails the rule takes the existing secondary-error branch before formatting or hit accounting: `served=false`, `SecondaryOutcome=error`, nothing backfilled, and the request proceeds to an upstream. A well-formed error object still passes and is served as a node error, as before.

Regressions: `TestSecondaryEntryThatIsNotAReplyIsNotServed` gains `{}` and `{"jsonrpc":"2.0","id":1}`; `TestSecondaryBatchWithAnElementThatAnswersNothingIsNotServed` stages your mixed batch against two balance requests and checks the real primary stays empty; the formatter rule test covers the same shapes plus `{"id":7}` and a null error. All three fail on the previous rule and pass now; `go test ./protocol/chainlib/cacheformat ./protocol/rpcsmartrouter` pass. The docs line in SECONDARY-CACHE.md names the new refusal.


avitenzer:

## Codex review — 2026-09-19

Reviewed head [`db8b28e`](https://github.com/Magma-Devs/smart-router/commit/db8b28e9884b291faee65253f3ea662a74e1a926).

### [P2] Reject RPC envelopes with no payload before accepting the cache hit

Location: [`protocol/chainlib/cacheformat/reply.go:20–22`](https://github.com/Magma-Devs/smart-router/blob/db8b28e9884b291faee65253f3ea662a74e1a926/protocol/chainlib/cacheformat/reply.go#L20-L22), and the corresponding batch-element check at lines 30–33.

The new guard accepts every JSON object, including `{}` and `{"jsonrpc":"2.0","id":1}`, although neither contains an RPC response payload. This also admits the `{"id":7}` entries that an older zone could already have produced through the HTML-rewriting bug described above. The later `CheckResponseError` only marks these as node errors: `trySecondaryCacheLookup` still returns `served=true` and records a hit, so the request cannot fall through to an upstream on that attempt and retries can keep selecting the same unusable entry.

I reproduced this against this head with a temporary Go test overlay, reusing `runSecondaryEthLookup` and the real primary cache server. For an `eth_getBalance` request at `0x64`, both missing-payload examples produced `served=true, SecondaryOutcome=hit, IsNodeError=true`. The empty object was rewritten to `{"id":7}`.

The batch case is worse: for two balance requests, this cached body:

```json
[{"jsonrpc":"2.0","id":1,"result":"0x64"},{"jsonrpc":"2.0","id":2}]
```

produced `served=true, SecondaryOutcome=hit, IsNodeError=false`, and the malformed batch was **backfilled into the primary cache**. The default batch error classifier lets the successful sibling mask the missing payload.

Please reject an object with neither a `result` field (allow `null`) nor a non-null `error` payload, checking every batch element before formatting or hit accounting. Route these entries through the existing secondary-error/fallback branch, while keeping valid RPC error responses on the served-node-error path. Add regression cases for the two single objects and the mixed batch; each should return `served=false`, record a secondary error, and leave the primary empty. This is distinct from the explicitly deferred empty-error-message behavior.

Validation: the existing formatter, RPC response-classifier, and secondary-cache tests passed. All three additional missing-payload fallback assertions failed; the mixed-batch test also observed the malformed entry through the real primary cache's `GetRelay`. Repository source files were not changed.

avitenzer:
Fixed in 5125bce.

`IsRestorableJSONRPCReply` now requires a payload: a `result` member (null included, since an unknown transaction answers null) or a non-null `error`. A batch is judged element by element, so one good sibling no longer vouches for an element that answers nothing. Everything that fails the rule takes the existing secondary-error branch before formatting or hit accounting: `served=false`, `SecondaryOutcome=error`, nothing backfilled, and the request proceeds to an upstream. A well-formed error object still passes and is served as a node error, as before.

Regressions: `TestSecondaryEntryThatIsNotAReplyIsNotServed` gains `{}` and `{"jsonrpc":"2.0","id":1}`; `TestSecondaryBatchWithAnElementThatAnswersNothingIsNotServed` stages your mixed batch against two balance requests and checks the real primary stays empty; the formatter rule test covers the same shapes plus `{"id":7}` and a null error. All three fail on the previous rule and pass now; `go test ./protocol/chainlib/cacheformat ./protocol/rpcsmartrouter` pass. The docs line in SECONDARY-CACHE.md names the new refusal.

AnnaR-prog:
## Review: `/code-review high` on `5125bce`

Scope: the full diff `main...5125bce`, every caller of the JSON-RPC output formatter (the two cache hit paths and the three websocket sites in `consumer_websocket_manager.go`), the primary hit path beside the secondary one, and the shape rule against replies the tests do not stage. Re-verified against a worktree at this head; the PR's tests pass there.

Branch is standalone. The two-step judgement on the secondary tier is right, the "drop as an error, never as a node error" reasoning is the correct reading of the `Lava-Retries: 5` comment, and the shape rule holds for what I threw at it. One thing the guard line changes that the PR does not mention.

### 1. Medium — the formatter guard changes the websocket wire: every subscription notification loses the `id` the router has been injecting — **CONFIRMED**

`protocol/chainlib/cacheformat/formatter.go:75`, `protocol/chainlib/consumer_websocket_manager.go:610`

`outputFormatter` runs on every streamed subscription message, not only on cache hits (`:610`, inside the `subscriptionMsgsChan` loop). A notification has no top-level `result`/`error` (`{"jsonrpc":"2.0","method":"eth_subscription","params":{"subscription":…,"result":…}}`), so `IsRestorableJSONRPCReply` is false and the guard returns it untouched. Probe, same input through both formatters:

```
main:    {"jsonrpc":"2.0","method":"eth_subscription","params":{…},"id":"\"sub-1\""}
5125bce: {"jsonrpc":"2.0","method":"eth_subscription","params":{…}}
```

JSON-RPC 2.0 says a notification carries no `id`, so the new output is the correct one — but it is a change on every push to every websocket subscriber, made by a PR about the secondary cache, with no test on the websocket path and no mention in the body or the docs. A client that keyed pushes on that injected id (non-standard, but it has been on the wire since the formatter existed, with #408's double-quote bug on top) stops matching. Either say so — release note plus a websocket-path test that pins the new shape — or scope the guard to the cache paths (`usableCachedReply` already gates the secondary; the primary path could call the same rule) and leave the formatter's behaviour to a deliberate websocket change. I would take the first: the old behaviour was wrong, and this is the moment to own it.

### Low — the primary hit path judges the label, not the contents

`protocol/rpcsmartrouter/rpcsmartrouter_server.go:4076`

The primary path runs `resolveCachedEntryKind` but not the `CheckResponseError` step this PR adds to the secondary. The scenario is the same one, one hop closer: a primary shared by a fleet mid-rollout (an older build wrote an unlabelled error body) serves it as a success with no header until every writer is upgraded. The secondary's step is four lines and reads the same envelope; applying it to both keeps "which cache answered makes no difference" true for the primary too.

### Verified and not flagged

- The shape rule: `result: null` counts as a payload (an unknown transaction), `error: null` does not, a batch is judged per element, and `json.Valid` runs before `gjson` so a truncated envelope cannot be read as partial.
- An unusable entry is counted as a secondary error, never served, never backfilled (`hit = false` before the populator runs), and the request proceeds to a provider — the retry loop does not see it again.
- REST and gRPC entries keep their status-based verdict, as the docs now say; an HTML body under a recorded 200 on REST is still served, which the ticket's coverage note anticipates.
- Textually clear of #408's hunks as stated; semantically independent too (raw-id restoration and the shape guard compose).

avitenzer:
Fixed in 14edb0b. The root commit is amended to 94c87ea (message only): the subject carries `!` and the body the `BREAKING CHANGE:` footer, since that is the commit that introduced the formatter guard; 5125bce is replayed as 5661123, and two commits sit on top.

1. Owned, as you suggested. The footer names the change: streamed notifications (no top-level `result` or `error`, the `eth_subscription` pushes) no longer carry the id the router injected from the subscribe request, a JSON-RPC 2.0 notification has none, and a client that matched pushes on it keys on `params.subscription` instead; replies still get the caller's id back. The loop's formatter construction is one named function now, `subscriptionReplyFormatter`, so `TestSubscriptionReplyFormatterLeavesNotificationsWithoutAnId` (8cecf77) drives the exact formatter the loop runs on every frame: the subscribe reply, an acknowledgement and an error reply get the id back, and a notification is pushed byte for byte, no `id`, `params.subscription` untouched. It fails against the formatter as it was before the guard, checked by overlaying `main`'s `formatter.go`. A Tendermint event carries a `result` and keeps getting the caller's id restored, as before. There is no websocket page under `docs/` to add a sentence to; the PR body carries it.

Low. Both tiers run the same judgement now (14edb0b). `classifyCachedEntry` is the label-then-contents step, one helper the secondary calls where its inline block was and the primary calls where it read the label alone; and `primaryLookupVerdict` applies the shape rule the secondary already applied (`usableCachedReply`) to the primary: a hit whose bytes are not a reply is no hit, recorded as an error rather than a miss, and the request proceeds as on a miss, to the secondary if there is one and to a provider, whose answer replaces the entry through the ordinary write. The primary lookup lives inline in `sendRelayToEndpoint`, which no test drives end to end, so `TestPrimaryEntryIsJudgedByItsContents` stages the entries in a real cache server, reads them back through the router's own client and puts the reply through the two steps that path runs: an unlabelled error body is served as a node error with the id restored, a success as a success, and a web page or an envelope with no payload is no hit and an error outcome. `TestPrimaryLookupVerdictOnNonHits` pins miss, timeout and the REST exemption.

Verified: from a detached checkout of 14edb0b, `go build ./...`, `go vet` and `golangci-lint run --config .golangci.yml` over `./protocol/chainlib/... ./protocol/rpcsmartrouter/...` (no findings beyond `main`'s eleven) and `go test -race -count=1 ./protocol/chainlib/... ./protocol/rpcsmartrouter/...` pass; the new tests and the secondary content tests pass `-race -count=3`. Independence from #408 re-checked: its head fe82e59 merged into this branch builds and passes the chainlib and router suites.

Not done: nothing from the review. The primary path's own wiring, the closure returning the verdict and the `!unusable` on the serving branch, is exercised through the helpers rather than through `sendRelayToEndpoint`, for the reason above.

AnnaR-prog (approved):

PR #413, opened 2026-09-20
test(redisstore): assert what rotation guarantees under go-redis's re-auth race: in place or replaced, never a failed operation

avitenzer:
# Description

**Flake (MAG-3769).** `TestLiveRotationThroughConfiguredStore` failed about one run in three, and its hand-built sibling `TestLiveRotationAgainstRealValkey` about one in five, always the same way: the connection recorded before the rotation never showed up authenticated as the new user, and after the pool timeout it was gone from the server, a fresh connection under the new credentials having replaced it. Operations never failed.

**Root cause: go-redis v9.22.0, still on master.** A pooled connection is re-authenticated by a background worker that must first see the connection go idle. The pool marks the connection idle on its Put hot path (`Conn.Release`, `internal/pool/conn.go`) with a bare compare-and-swap that never notifies waiters, so a worker already parked in `AwaitAndTransition` (`internal/pool/conn_state.go`) is never woken; in a narrower ordering the release lands between the worker's failed fast path and its enqueue, and the wake-up is lost the same way. Captured on a failing run: the worker sits in that method's select (`conn_state.go:267`, spawned from `pool_hook.go:159`). Every checkout rejects the marked connection meanwhile, so it serves nothing; when the worker's wait expires after the pool timeout (`read-timeout` + 1 s when a read timeout is set, six seconds with the client's default of five; thirty seconds when `read-timeout` is negative) the client closes it and dials a fresh connection with the new credentials.

Measured against Valkey 8.1: 4 of 20 rotations took that path with a single push of the credentials, and a second push of the same credentials does not rescue the worker (also 4 of 20), because later workers queue behind the stuck one. `v9.23.0-beta.1` behaves the same (5 of 20). Reported upstream as [redis/go-redis#4027](https://github.com/redis/go-redis/issues/4027) with a self-contained reproduction, an instrumented trace of a stalled rotation, a regression test and a verified two-hunk fix (`Release` notifies waiters, and `AwaitAndTransition` re-checks once after enqueueing): 0 of 100 rotations stall with it applied, against 3 to 8 of 20 without; the fix is opened upstream as [redis/go-redis#4028](https://github.com/redis/go-redis/pull/4028).

**Fix here.** The router cannot reach into the pool, so the guarantee it can honestly test and document is the one the client gives: after a rotation, every connection established before it is either re-authenticated in place (same id, new user) or replaced (closed by the client, a fresh one under the new credentials serving), with no failed operation and no reconnect storm, and, while operations keep flowing, within the pool timeout. Both real-Valkey tests assert exactly that: `rotationSettled` checks retirement of the established ids and logs which path each run took; `servedAs` then proves the identity (`ACL WHOAMI` through the store's own client answers the new user, and `CLIENT LIST` shows no connection under the previous user and at least one under the new); a `dialCounter` hook on each client bounds dials during the rotation to one and requires convergence to dial-free serving afterwards. The configured-store test also fails at once on any operation error. A negative control kills the established connection without rotating: retirement alone reads as a settled rotation and the operation succeeds, while the identity proof rejects it. The hand-built test uses a pool of two so the replacement has a slot: a stalled connection keeps its slot until the pool timeout, and with the pool at its size limit every operation in that window dials and drops a replacement, which the docs say as well. `docs/RESP-CACHE.md` states both caveats in the rotation section: re-authentication starts only when traffic reaches a marked connection, so an idle connection keeps the previous identity meanwhile, and the stall, whose pool timeout bounds only the scheduled worker's wait. The operational advice follows: verify completion with `CLIENT LIST`, or force it with `ACL DELUSER` / `CLIENT KILL USER`, keeping the previous credential valid until then; a timed wait proves nothing.

Files: `ecosystem/cache/redisstore/valkey_rotation_test.go` and `docs/RESP-CACHE.md`. No production code changes.

## Tests

Thirty runs of both tests against Valkey 8.1 in Docker: zero failures. Of the sixty rotations, 52 re-authenticated in place and 8 were replaced, the split the diagnosis predicts; the previous assertion failed on exactly the replaced runs. The ordinary package run, which skips the Valkey-backed tests, passes.

## Out of scope

The upstream defect itself, now filed as redis/go-redis#4027; until it lands, in-place re-authentication is best effort and the documented bound is what holds.

Jira ticket: MAG-3769

---

## Author Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change (not applicable; tests and documentation only)
* [x] provided a link to the relevant issue or specification (https://magmadevs.atlassian.net/browse/MAG-3769)

## Reviewers Checklist

I have...

### Resolution (2026-09-20)

Codex review, three P2 findings, fixed in 02cde65:

- **Replacement identity**: `servedAs` (ACL WHOAMI through the store's client; CLIENT LIST shows no connection under the previous user and one under the new) runs after `rotationSettled` in both live tests. Negative control `TestLiveRotationProofsRejectReplacementUnderPreviousUser`: a server-side kill without a rotation passes `rotationSettled` and fails `servedAs`.
- **Reconnect storm**: the configured store's client gets a `dialCounter` hook; at most one dial during the rotation and convergence to dial-free serving replace the live-connection count, which cannot see dropped connections.
- **Idle connections and the pool timeout**: `docs/RESP-CACHE.md` separates the traffic-driven start of re-authentication from the stall; the pool timeout bounds only the scheduled worker's wait, and completion is verified with `CLIENT LIST` or forced with `ACL DELUSER` / `CLIENT KILL USER`.


avitenzer:
## Review findings

Reviewed [`c12f017`](https://github.com/Magma-Devs/smart-router/commit/c12f0173a829fd8451d2601ca1b68f5dc6591fe2), including the documentation/comment updates pushed during this review. The executable test code is unchanged from the `6852a1e` version used for the runs below.

### [P2] Verify that a replacement actually uses the new ACL user

In [`rotationSettled`](https://github.com/Magma-Devs/smart-router/blob/c12f0173a829fd8451d2601ca1b68f5dc6591fe2/ecosystem/cache/redisstore/valkey_rotation_test.go#L254-L265), disappearance of an original connection is sufficient for success. The subsequent `op()` does not establish the replacement's identity: both ACL users retain `+@all`, so the operation succeeds under either user. The configured-store count assertion also permits **zero** connections under the new user.

Negative control against Valkey 8.1: retain the old credentials, establish a connection, close its server-side ID, then issue another store operation. The new helper returned `settled=true, replaced=1`, the operation succeeded, and the count bound passed, while `ACL WHOAMI` still returned the **old user** and no new-user connections existed. Thus an unrelated reconnect can make failed credential propagation satisfy the new acceptance checks.

Keep the original-ID retirement check, but also prove which user serves the post-rotation operation—for example, with [`ACL WHOAMI`](https://valkey.io/commands/acl-whoami/) on the actual tested connection, or a probe only the new ACL user can execute. Add the old-credentials replacement as a negative control.

### [P2] Count cumulative connections when checking for a reconnect storm

The new [configured-store assertion](https://github.com/Magma-Devs/smart-router/blob/c12f0173a829fd8451d2601ca1b68f5dc6591fe2/ecosystem/cache/redisstore/valkey_rotation_test.go#L201-L202) counts only connections still alive at the final snapshot. Connections opened and immediately discarded disappear from that count, so it cannot detect the repeated dial/drop behavior this PR describes.

A separate negative control forced twenty reconnects while operations continued successfully. The instrumented dialer recorded **21 dials**, but `CLIENT LIST` contained **one** live new-user connection; the proposed bound of two still passed.

Please attach a counting `DialHook` to the configured store's client before the initial operation, or measure another cumulative connection counter on the isolated server. Assert on connections opened during rotation and subsequent convergence, as the hand-built test already does with its dial counter.

### [P2] Qualify the documented timeout for connections that remain idle

The [new rotation guidance](https://github.com/Magma-Devs/smart-router/blob/c12f0173a829fd8451d2601ca1b68f5dc6591fe2/docs/RESP-CACHE.md#L178-L180) bounds the old connection's identity by the pool timeout. That timeout starts only once the re-auth worker is scheduled. In go-redis v9.22, `MarkForReAuth` only records the pending work; `OnPut` starts the worker. A connection already sitting idle when credentials are pushed may not reach that hook until a later pool checkout/return.

Reproduced with the same streaming provider and default client timeouts: establish one connection, change the credential file, call `Refresh()`, and leave the data client idle while inspecting it through the separate admin client. After **6.5 seconds**, with `PoolTimeout=6s`, the original connection was still authenticated as the old user. The router's credential polling and background health-probe cadences can add further delay.

Please describe the pool timeout as a bound on the scheduled worker's wait, not on total retirement time after a credential change. Document how to verify retirement or explicitly disconnect the previous ACL user's sessions; a fixed pool-timeout sleep alone does not prove rotation has completed.

### Validation

- The ordinary Redis-store package tests passed.
- Both unchanged live-rotation tests passed **10 runs each** against Docker Valkey 8.1: 17 rotations re-authenticated in place and 3 replaced the connection.
- Temporary Go test overlays reproduced all three cases above. The first two deliberately inject invalid behavior to check the assertions; the third observes an idle client without fault injection.
- No repository code or computer configuration was changed.

avitenzer:
Fixed in 02cde65. All three findings are addressed; the live lane (both rotation tests plus the new negative control) passed 10 consecutive runs against Valkey 8.1, taking both the in-place and the replaced path.

**[P2] Replacement must run as the new ACL user.** New `servedAs` proof, applied in both live tests after `rotationSettled`: `ACL WHOAMI` through the store's own client must answer the new user, and `CLIENT LIST` must show no connection left under the previous user and at least one under the new one. `rotationSettled` is now documented as a retirement check only. Negative control `TestLiveRotationProofsRejectReplacementUnderPreviousUser` reproduces your scenario (credentials unchanged, `CLIENT KILL ID` on the established connection, an operation reconnects): it asserts that `rotationSettled` reads it as settled with one replacement, that the operation succeeds, and that `servedAs` rejects it naming the previous user.

**[P2] Cumulative dials for the storm check.** The configured store's client now gets a counting hook through `AddHook` (`dialCounter`, shared with the hand-built test, which used a `Dialer` option before), attached before the initial operation. The assertions are on dials during the rotation (at most one beyond those before it) and on convergence to dial-free serving, the same two checks the hand-built test makes; the live-count assertion is gone. The negative control also drops the connection three more times and shows the live count staying within the pool size while the cumulative count reaches five.

**[P2] Pool timeout qualified for idle connections.** `docs/RESP-CACHE.md` now states the two caveats separately: re-authentication is traffic-driven (the push marks a connection; the worker, and with it the pool-timeout clock, starts when that connection is next checked out and returned; an idle connection keeps the previous identity for as long as it stays idle), and the stall, whose pool timeout bounds the scheduled worker's wait only. Completion is to be verified with `CLIENT LIST` (`user=`) or forced with `ACL DELUSER` / `CLIENT KILL USER`, with the previous credential kept valid until then; a timed wait is stated to prove nothing. The `rotationSettleTimeout` comment says the tests' bound holds because they keep operations flowing.

